PR #1, opened 2017-12-11
initial spec

duglin:
Signed-off-by: Austen Collins <austen@serverless.com>

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1,0 +1,442 @@
+# OpenEventing - Verson 0.1
+
+OpenEventing is a vendor-neutral specification for event data.
+
+## Overview
+Events are everywhere.  However, event publishers tend to describe events
+differently.
+
+The lack of a common way of describing events means developers must constantly
+re-learn how to receive events.  This also limits the potential for libraries,
+tooling and infrastructure to aide the delivery of event data across
+environments, like SDKs, event routers or tracing systems.  The portability
+and productivity we can achieve from event data is hindered overall.
+
+Enter OpenEventing, a specification for describing event data in a common way.
+OpenEventing seeks to ease event declaration and delivery across services,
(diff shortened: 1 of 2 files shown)


spec.md
```diff
@@ -0,0 +1,442 @@
...
+* Description: ID of the event.  Can be specified by the producer.  The
+  semantics of this string are explicitly undefined to ease the implementation
+  of producers.  Enables deduplication.
+* Examples:
+  * A database commit ID
+* Constraints:
+  * Required
+  * Unique per producer
+
+### correlation-id
```
ac360: @duglin Could we move this to the `Context Attributes Backlog`.  This was suggested in the Google doc by a contributor, but wasn't agreed to by anyone else.  We need to discuss and review new properties before adding them in.

spec.md
```diff
@@ -0,0 +1,442 @@
...
+  event. Enables traceability through systems.
+* Examples:
+  * Event A generated with unique correlation-id. Event handler receives event
+    and generates a new event in response, Event B. Event B copies the
+    correlation-id of Event A so that the chain of events can be properly
+    correlated.
+* Constraints:
+  * Optional
+
+### causation-id
```
ac360: @duglin Could we move this to the `Context Attributes Backlog`.  This was suggested in the Google doc by a contributor, but wasn't agreed to by anyone else.  We need to discuss and review new properties before adding them in.

duglin: No idea why this comment didn't go away when I moved the entry, but I did move it as you can see.

duglin:
@ac360 can you add a comment saying that you agree with the edit and are "signing off" on it - in essence signing a DCO, just so we transfer right to the group?

ac360:
These edits look good to me.

PR #2, opened 2017-12-14
Add initial pass at our governance model

duglin:

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -41,13 +41,15 @@ merged:
 
 If a vote is taken during a WG meeting, the follow rules will be followed:
 
+* There is only 1 vote per participating company, or nonaffiliated individual.
+* Each participating company can assign a primary and secondary representative.
+* A participating company, or nonaffiliated individual, attains voting rights
+  by having any of their assigned representative(s) attend 3 out of the last
+  4 meetings. They obtain voting rights after the 3rd meeting, not during.
 * Only WG members with voting rights will be allowed to vote.
-* Only one WG member per company can vote.
 * A vote passes if more than 50% of the votes cast approve the motion.
 * Only "yes" or "no" votes count, "abstain" votes do not count towards the
   total.
-* WG members attain voting rights by attending 3 out of the last 4 WG
-  meetings.  They obtain rights after the 3rd meeting, not during.
 * Meeting attendence will be formally tracked
   [here](https://docs.google.com/spreadsheets/d/1bw5s9sC2ggYyAiGJHEk7xm-q2KG6jyrfBy69ifkdmt0/edit#gid=0).
   Members must acknowledge their presence verbally, meaning, adding yourself
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -1,0 +1,55 @@
+# Governance
+
+This document describes the governance process under which the Serverless
+Working Group (WG) will manage this repository.
+
+## Working Group Meetings
+
+In order to provide equitable rights to all Working Group members,
+the following process will be followed:
+
+* Official WG meetings will be announced at least a week in advance.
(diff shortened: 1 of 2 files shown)


caniszczyk:
LGTM, but I'd also add a MAINTAINERS file that lists the actual maintainers with voting rights

duglin:
The problem with keeping a MAINTAINERS file up to date is that it might change ever week based on attendance - which will be kind of annoying. I plan on having the spreadsheet automagically calculate who has rights (see https://docs.google.com/spreadsheets/d/1bw5s9sC2ggYyAiGJHEk7xm-q2KG6jyrfBy69ifkdmt0/edit#gid=0 for my working prototype), so the info will be easily available. I guess we could create a MAINTAINERS.md file that just say "go look at the [link] spreadsheet".

carimura:
I could write an Fn func that grabs names and updates MAINTAINERS.md. you know, dogfooding working group projects. :)

GOVERNANCE.md
```diff
@@ -0,0 +1,55 @@
...
+  Note, administrivia PRs that do not materially modify WG output documents
+  may be processed by WG admins as needed.
+* Resolving PRs ("merging" or "closing with no action") will be done as a
+  result of a motion made during a WG meeting, and approved.
+* Reopening a PR can be done if new information is made available, and a
+  motion to do so is approved.
+* Any motion that does not have "unanimous consent" will result in a formal
+  vote. See [Voting](#voting).
+
+## PRs
```
schatekar: Do we need to raise an issue that the PR addresses first?

duglin: I know some groups like to do that, but I think its kind of silly overhead to require an issue. I'm personally ok with someone opening a PR and explaining why they think the change is needed directly in the PR. I think having it all in one spot anyway  ;-)

duglin:
On last week's call @markpeek agreed to suggest some wording around how to deal with multiple people from the same company.

GOVERNANCE.md
```diff
@@ -0,0 +1,55 @@
...
+merged:
+
+* The author of the PR indicates that it is ready for review by asking for it
+  to be discussed in an upcoming meeting - eg. by adding it to the agenda
+  document.
+* All comments have been addressed.
+* PRs that have objections/concerns will be discussed off-line by interested
+  parties. A resolution, updated PR, will be expected from those talks.
+
+## Voting
```
markpeek: Here is a proposed rewrite of the Voting section to take into account when  primary members may not be able to attend.

## Voting

If a vote is taken during a WG meeting, the follow rules will be followed:

* Only WG members with voting rights will be allowed to vote.
* Only one WG member per company is designated primary and can vote.
* A vote passes if more than 50% of the votes cast approve the motion.
* Only "yes" or "no" votes count, "abstain" votes do not count towards the
  total.
* WG members attain voting rights by attending 3 out of the last 4 WG
  meetings.  They obtain rights after the 3rd meeting, not during.
* A primary may designate a secondary company attendee to attend and still
  be counted for attendance voting rights. This allows for times when the
  primary may not be able to attend.
* For purposes of a vote, a secondary can vote when the primary is not
  in attendance.
* Meeting attendence will be formally tracked [here](https://docs.google.com/spreadsheets/d/1bw5s9sC2ggYyAiGJHEk7xm-q2KG6jyrfBy69ifkdmt0/edit#gid=0).
  Members must acknowledge their presence verbally, meaning, adding yourself
  to the "Attendees" section of the Agenda document is not sufficient.
  Secondary members should indicate their role at this time.

duglin: i've updated the PR with the above + some slight tweaks based on a chat Mark and I had.  So removing this comment to avoid confusion as to the latest version of the proposal.

duglin:
Voting section has been updated -please review for today's call.

duglin:
Per today's call this was approved by unanimous consent

PR #3, opened 2018-01-07
add ref to RFC2119 and href checker

duglin:
Any good spec needs to use RFC2119 keywords, so this starts us down that path
by added a ref to it. Doesn't introduce any uses of the keywords yet.

Added an href checker tool so that we can make sure we don't have any
bad hrefs in our docs.

Also added a travis hook so we run the verification script as our build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,3 @@
+language: generic
+script:
+- make
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,6 @@
+all: verify
+
+verify:
+	@echo Running href checker:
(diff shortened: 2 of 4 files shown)


caniszczyk:
LGTM

duglin:
Just a reminder for people, per the proposed governance process, we will be approving PRs during our calls.  We will not be using a traditional "LGTM" process yet. However, people should still LGTM (or "NOT LGTM" w/comments) in PRs to express their (dis)agreement with the PRs. This will help us know which have general consensus and can be quickly approved during our calls.

PR #11, opened 2018-01-18
Add DCO

caniszczyk:
Signed-off-by: Chris Aniszczyk <caniszczyk@gmail.com>

### DIFF
--- a/DCO
+++ b/DCO
@@ -1,0 +1,38 @@
+Developer Certificate of Origin
+Version 1.1
+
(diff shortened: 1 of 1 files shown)


duglin:
@caniszczyk the CONTRIBUTING.md has a copy of this embedded - do we need both? Should that point to this instead?

caniszczyk:
@duglin missed it in the CONTRIBUTING.md, we're all set then

duglin:
ah ok thanks!

PR #15, opened 2018-01-22
First pass at using RFC 2119 keywords

duglin:
Also:
- removed some editorial comments that aren't really appropriate for a spec
- added a tool to verify we upper-case our RFC2119 keywords

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1,6 +1,7 @@
-# CloudEvents - Verson 0.1
+# CloudEvents - Version 0.1
 
-CloudEvents is a vendor-neutral specification for event data.
+CloudEvents is a vendor-neutral specification for defining the format
+of event data.
 
 ## Table of Contents
 - [Overview](#overview)
@@ -12,38 +13,21 @@ CloudEvents is a vendor-neutral specification for event data.
 - [Reference](#reference)
 
 ## Overview
-Events are everywhere.  However, event publishers tend to describe events
+Events are everywhere. However, event publishers tend to describe events
 differently.
 
-The lack of a common way of describing events means developers must constantly
-re-learn how to receive events.  This also limits the potential for libraries,
+The lack of a common way of describing events means developers are constantly
+re-learning how to receive events. This also limits the potential for libraries,
 tooling and infrastructure to aide the delivery of event data across
-environments, like SDKs, event routers or tracing systems.  The portability
-and productivity we can achieve from event data is hindered overall.
+environments, like SDKs, event routers or tracing systems. The portability
+and productivity that can be achieved from event data is hindered overall.
 
 Enter CloudEvents, a specification for describing event data in a common way.
 CloudEvents seeks to ease event declaration and delivery across services,
 platforms and beyond.
 
-CloudEvents is a new effort and it's still under active development.  However,
-its working group has received a surprising amount of industry interest,
-ranging from major cloud providers to popular SaaS companies.  Our end goal is
-to offer this specification to the Cloud Native Computing Foundation.
-
 ## Status
-A coalition of industry stakeholders have expressed interest in collaborating
-on this specification (though this does not constitute an endorsement on their
-behalf) and have begun iterating on early drafts.
-
-Version 0.4 of CloudEvents represents the end result of a working session
-between engineers at Amazon, Google, and Microsoft.  Many items are left to be
-discussed, but this specification does represent the beginning of a level of
-consensus between these major industry stakeholders, which is exciting.  Now,
-we are opening up this specification for public feedback and conducting all
-working sessions through the Serverless Working Group within the CNCF.
-Overall, in this early stage, almost all stakeholders have requested the
-initial scope of the specification be kept small to ease adoption and
-implementation.  At this time we are focused on the following scope:
+At this time the specification is focused on the following scope:
 
 * Agree upon a set of event metadata attributes (“context”) that:
   * Offer a basic description of the event and the data it carries.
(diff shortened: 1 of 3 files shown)


carimura:
Overall LGTM. Agree with formatting, etc.

duglin:
@ultrasaurus not sure what you mean by the "formatting in the preamble".  I removed a chunk of text that I didn't think was appropriate for a spec, but the rest of it should be just removing of words that are reserved for RFC2119 usage (or fixing a few minor typos).

tools/verify-phrases.sh
```diff
@@ -0,0 +1,180 @@
+#!/bin/bash
```
duglin: This script will check to make sure all uses of the RFC2119 keywords are upper-cased. This will help make sure that we don't include them by mistake.

ultrasaurus: @duglin there are a lot of changes from top of file to ### namespace -- pulling those into a separate PR will make it easier for reviewer to focus on the things that could potentially change the meaning of the spec

duglin: yes but a lot of them are related to removing rfc2119 keywords in places that we didn't mean to use them.

spec.md
```diff
@@ -103,133 +84,162 @@ occurrence that tools and developers can rely upon to better handle the event.
...
+that contains both context and data).
 
 
 ### namespace
 * Type: String
 * Description: Identifier that uniquely identifies the organization publishing
   the event.
+* Constraints:
+  * OPTIONAL
+  * If present, MUST be a non-empty string
```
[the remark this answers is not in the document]
duglin: I was trying to keep things consistent with the current doc and save the harder discussions of changing the semantics for follow-on PRs. But let's see what other people think about making this a required field (I need to check too...)

spec.md
```diff
@@ -103,133 +84,162 @@ occurrence that tools and developers can rely upon to better handle the event.
...
 * Notes:
   * It is up for discussion whether this is applicable to the whole event or
-    to the data payload alone.  Currently, event-type-version covers the data,
+    to the data payload alone. Currently, event-type-version covers the data,
     while open-events-version covers the context.
 
 ### open-events-version
 * Type: String
```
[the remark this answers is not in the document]
duglin: Seems reasonable.  Any objection to this one being REQUIRED?  @ac360 ?

spec.md
```diff
@@ -103,133 +84,162 @@ occurrence that tools and developers can rely upon to better handle the event.
...
+* Description: The event payload. The payload depends on the event-type,
   schema-url and event-type-version.
+* Constraints:
+  * OPTIONAL
 * Notes:
   * TBD where the producer specifies format/encoding of the data elsewhere in
     this schema.
+  * TBD is it allowed to be empty if present?
 
 ## Context Attributes Backlog
```
ultrasaurus: Note: in today's meeting we decided to move these into bugs, so you might want to remove from this PR to avoid merge conflict.

tools/verify-phrases.sh
ultrasaurus: True, though just thought the reviews might be easier in two parts that are logically separate, even though they are related.  Just a suggestion.

spec.md
```diff
@@ -1,6 +1,7 @@
 # CloudEvents - Verson 0.1
 
-CloudEvents is a vendor-neutral specification for event data.
```
rabbah: note typo above Verson -> Version.

duglin: Even though I didn't touch that line as part of this PR, I fixed it. thanks!

spec.md
```diff
@@ -103,133 +84,162 @@ occurrence that tools and developers can rely upon to better handle the event.
...
 * Description: Timestamp of when the event happened.
+* Constraints:
+  * OPTIONAL
+  * If present, MUST adhere to the format specified in
+    [RFC 3339](https://tools.ietf.org/html/rfc3339)
 
 ### schema-url
-* Type: String
+* Type: URI per [RFC 3986](https://tools.ietf.org/html/rfc3986)
 * Description: A link to the schema which can be optionally specified by the
```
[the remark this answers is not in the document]
duglin: good catch - fixed

spec.md
duglin: yep - done!

spec.md
```diff
@@ -103,133 +84,162 @@ occurrence that tools and developers can rely upon to better handle the event.
...
 * Type: String
 * Description: The version of the CloudEvents specification which the event
-  uses.  This enables the interpretation of the context.
+  uses. This enables the interpretation of the context.
+* Constraints:
+  * OPTIONAL
+  * If present, MUST be a non-empty string
 
 ### resource
 * Type: Object
```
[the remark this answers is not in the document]
duglin: No strong opinion on this one yet so I'll make it REQUIRED. We can always change our minds later if needed.

spec.md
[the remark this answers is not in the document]
duglin: Done

spec.md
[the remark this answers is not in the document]
duglin: made it REQUIRED for now so we can try to resolve this PR on the next call. We can always revert it later if we want.

duglin:
I addressed all comments.

One thing I was wondering about is how someone will deal with our REQUIRED properties. For example, what if the producer just doesn't have a "namespace" to associate with the event? Does this mean that they can ever be CloudEvents compliant? Or must they just generate a dummy value? I'm not making a formal suggestion yet, but I do kind of wonder if we should word things to be more like "this property is OPTIONAL, but if you do include this data then it MUST use this property name".  Then we still get interop w/o forcing people to inject dummy values that don't really provide any value.  As I said, no formal proposal, but I'd like to hear from event producers to know how they feel about our REQUIRED properties.

duglin:
The travis failure is due to some incorrectly cased keywords in the backlog section that will be removed in #39. So #39 needs to be merged first.

duglin:
Approved on Feb 1st call

PR #39, opened 2018-01-27
Remove Notes/Attribute-backlog from spec now that we have issues for them

duglin:
During the last call we agreed to open up issues for the "Notes" in the spec
for the backlog of context properties. Now that those action items are done
we can remove those bits of text from the spec.

Note, some of the backlog context attributes might be added back in as
"agreed to" attibutes, but we'll decide that while resolving the corresponding
issue.


sslavic:
LGTM

duglin:
Approved on Feb 1st call.

PR #44, opened 2018-01-30
Update spec.md

lfourie:
Signed-off-by: lfourie <louis.fourie@huawei.com>

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -167,13 +167,13 @@ and data).
 * Description: The version of the CloudEvents specification which the event
   uses.  This enables the interpretation of the context.
 
-### resource
+### source
 * Type: Object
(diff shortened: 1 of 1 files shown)


duglin:
@lfourie what is your git client's "user.name" value? I think it needs to match what's listed in the DCO of the commit.

lfourie:
Signed-off-by: lfourie louis.fourie@huawei.com

duglin:
So, in this case I think it needs to be "Louis Fourie"

duglin:
@lfourie you can fix an existing commit by using: git commit --amend -s
that will allow you to edit the commit message and the -s will add your signature if not already there.
No need to open new PRs. If you have more than one commit, you may find it easier to just squash them down to one.

PR #49, opened 2018-01-31
Issue 6 - Definition of Event in the terminology section is confusing

### LINKED ISSUE #6: Definition of Event in the terminology section is confusing
[Definition of Event](https://github.com/openeventing/spec/blob/master/spec.md#event) introduces terms `notification` and `message` without clearly defining their meaning in the context of an event-driven system.  There may be some implicit assumptions/understanding in there that be explicitly written. 

Assigned to: @schatekar

schatekar:
Reworded the description of the term event to remove ambiguity around words message and notification

Fixes #6 

Signed-off-by: Suhas Chatekar <suhas.chatekar@gmail.com>

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -71,11 +71,11 @@ system. This is most typically when that system receives an external signal
 
 #### Event
 Data representing an occurrence, a change in state, that something happened
-(or did not happen), usually used for notification. Events include context
-and data. Each occurrence might be uniquely identified with data in the event.
-Events are considered as facts that have no given destination, whereas
-messages contain intent and tend to transport data from a source to a given
-destination.
+(or did not happen).  Events include context and data.  Each occurrence MAY be
+uniquely identified with data in the event. Events ought not to be confused 
+with messages which are used to transport or distribute data without assumptions 
+regarding its semantic. Events are considered to be facts that have no given 
+destination. Events are used to notify other systems that something has happened.
 
 #### Context
 A set of consistent metadata attributes included with the event about the


duglin:
@schatekar you'll need to sign all of your commits. It might be easiest if you just squash this down to one and sign that one.

schatekar:
I have messed it up even more. It's easier to do it all over again.

schatekar:
Other PR here  https://github.com/cloudevents/spec/pull/50

Still trying to figure out how to fix the failing DCO check. This PR can be rejected.

duglin:
Do you want me to try to fix this one or the other PR?

schatekar:
Both have the same content so you can fix any one.

duglin:
ok DCO is happy now

duglin:
rebase needed

schatekar:
I will need some more time to fix this. I had deleted the original branch after recurring DCO failures which messed up commit history. I need to see how can I get that branch back now.

duglin:
@schatekar ping me thru slack if you need some help

schatekar:
@duglin I think its all in order now. Thanks for the help.

duglin:
LGTM

spec.md
```diff
@@ -71,11 +71,11 @@ system. This is most typically when that system receives an external signal
...
 #### Event
 Data representing an occurrence, a change in state, that something happened
-(or did not happen), usually used for notification. Events include context
-and data. Each occurrence might be uniquely identified with data in the event.
-Events are considered as facts that have no given destination, whereas
-messages contain intent and tend to transport data from a source to a given
-destination.
+(or did not happen).  Events include context and data.  Each occurrence MAY be
+uniquely identified with data in the event. Events ought not be confused with
+messages which are used to carry intent and data to a given destination. Events
```
deissnerk: It's a good idea to explain the difference between messages and events, but I think messages are usually considered as containers to transport or distribute data between systems without specifying the semantic of the data.
I would see them in a layer below events. Usually on the same layer, events are distinguished from commands. Commands carry an intent about something that is supposed to happen on the target. Events inform about something that happened at the source. Both can be transported as messages.

schatekar: That is exactly what was going on in my head when I wrote this. But I just wanted to disambiguate what was already there. I thought if I added a new terminology i.e. `command`, not only would it be very specific to an architectural guideline, it may cause further confusion.

deissnerk: I get your point. That's the problem with terms like messages and events. They have quite some history in software engineering. I mentioned commands to have some distinction on the same layer, but I understand if you don't want to introduce yet another term.
I just think that messages will play an important role in this spec, as protocols like AMQP are mentioned as "transport protocols" for events. 
What about: 
```
Events ought not to be confused with messages which are used to transport or distribute data without assumptions regarding its semantic.
```

schatekar: That looks good. I have changed my commit to reflect this.

duglin:
still LGTM

deissnerk:
LGTM

ac360:
LGTM

duglin:
Approved on 2/8 call

Issue #37, opened 2018-01-26
System Architecture diagram / document explaining the context for events


ultrasaurus:
We would like to write up some info so that folks can arrive at this repo and understand what kind of event system we're talking about.

Starting docs:
- [presentation](https://onedrive.live.com/view.aspx?resid=123CCD2A7AB10107!744659&ithint=file%2cpptx&app=PowerPoint&authkey=!AIUYJmTPLT_D9Os) -- slide 23 (below) shows the difference between general messaging, and "Eventing" which is what CloudEvents is focused on.  Events are statements of fact.
![image](https://user-images.githubusercontent.com/41906/36404501-e05bbfa6-159f-11e8-87a6-7495b5ff0d18.png)
- https://github.com/cncf/wg-serverless/tree/master/proposals/cloudevents
- [Cloud-Native Event Mapping (CNEM)](https://docs.google.com/document/d/1dFfnWnbRDTvGwpwaBsvXRlw0ZyL3xYl7iXhYqbAPJjk/edit#heading=h.a30wdiuh8lnh)
- [Common Event Use Cases](https://docs.google.com/document/d/1E9J3yPIoqtjVKdDe7f3p7RxJml57Dc64T4b3o445cYk/edit)

- [OpenWhisk](https://openwhisk.apache.org/) has a nice illustration of the kind of system we're talking about CloudEvents:
![image](https://user-images.githubusercontent.com/41906/36404396-3acb2608-159f-11e8-9dda-6f4b2593f649.png)

Assigned to: @ultrasaurus @yaronha

duglin:
@ultrasaurus you ok with owning this one?

duglin:
Perhaps including the current "Reference" section of the spec (where we show what some providers do today).

ultrasaurus:
@duglin I think you can assign multiple owners.  Yaron volunteered to help, but I don't know his github!

duglin:
Took a sec I managed to hunt him down @yaronha :-)

duglin:
@ultrasaurus I like what you have so far, and I like that it appears you used a lot of the same text we have in the spec itself. That consistency will help newbies.

[...]

PR #53, opened 2018-02-01
starter doc explaining key concepts from the spec

ultrasaurus:
For reviewers (and later implementers) to understand the specification, we need to provide 
some context for the kinds of systems that use events and the problems we're solving.

See https://github.com/cloudevents/spec/issues/37 

This PR defines a few terms with an overly simplified diagram.  Once we agree on these terms,
then it will be easier to add the other components that will need to be present in a typical system.


deissnerk:
@ultrasaurus @clemensv Having our recent user story discussion in mind, I wonder if we could derive some architectural patterns of systems dealing with events. I don't mean complete systems. I observed that during discussions there are some reoccuring examples like the API gateway and the web hook, or mapping events to a topic structure on a broker. Maybe there are also patterns specific to interoperability like event gateway that do more than simple forwarding. If the term "pattern" causes some people to roll their eyes, we could also just call them examples. I just think that in our discussions it might help to simply refer to examples/patterns from the list.

duglin:
if #283 is adopted then I think moving this info into the primer.md would be good

PR #63, opened 2018-02-02
Add log-level to context attributes

### LINKED ISSUE #27: log-level context attribute
This is a proposed metadata attribute (from Context Attribute Backlog):

log-level
    Type: String
    Description: The level of logging that the publisher may want to log this specific event (e.g. debug)

vbmade2000:
Add log-level to context attributes from Context Attribute Backlog

Fixes #27

Signed-off-by: Malhar Vora <mlvora.2010@gmail.com>

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -231,6 +231,12 @@ that contains both context and data).
 * Constraints:
   * OPTIONAL
 
+### log-level
+* Type: String
+* Description: The severity level of logging. The intent of this attribute is to be considered as a mechanism to tell the receiving infrastructure what the logging level of the infrastructure SHOULD be.
+* Constraints:
+  * OPTIONAL
+  * If present, MUST be one of debug, info, warning, error, fatal   
 
 ## Additional Topics & Questions
 


vbmade2000:
travis says following.

> 
> Running the RFC2119 keyword checker:
> > spec.md
> line 226: 'may' should be 'MAY'
> make: *** [verify] Error 1

It doesn't make sense to replace "may" with "MAY" in following line.
`The level of logging that the publisher may want to log this specific event (e.g. debug)`

duglin:
@vbmade2000 can you follow the pattern of the other attributes, in particular add a "Constraints" section?  Then you can reword the description w/o duplicating the optionality - for example:

```
* Type: String
* Description:  The level of logging of the event (e.g. `debug`).
 * Constraints:
   * OPTIONAL
```

Do we want to have some well-defined valued (debug, info, ...) that people can interop on? Would we want to allow for extensions?

vbmade2000:
@duglin 

> can you follow the pattern of the other attributes, in particular add a "Constraints" section? 

I am sorry, I should have checked pattern. 

> Do we want to have some well-defined valued (debug, info, ...) that people can interop on?

Yes I think it would be good to have well defined levels like debug, warning, error etc

> Would we want to allow for extensions?

Not sure about it.

duglin:
I agree on having well defined values. Perhaps you could add that to this PR too:
This specification defines the following values: debug, info, ....

We may, at some point, need to also define what these values mean,and when they should be used. You could try to do that in this PR or in a follow-on if you want.

vbmade2000:
@duglin Sounds cool.

schatekar:
Should log level be part of this specification? Consider the following two examples

1. An angular app publishes an event which is transported over HTTP. Angular App wants the event to be logged so sets the log-level field to a value. What does that mean? Who should log the event? Where should the event be logged? 
2. The event in the above example goes via API Gateway which has it's logging set to the lowest level by default, maybe because it's running in production mode. Having the log-level field in the cloud event set to debug would override that setting and log the whole message? 

Also, consider the other aspects e.g. event contains sensitive information and should never be logged unless the system logging the event know what it is doing. 

I believe it would be an anti-pattern to have an event tell the system that it needs to be logged. The system, based on other factors, should decide whether it wants to log the event or not. 

Having said that, it could be useful to have a way for the publisher to state that the event should never be logged (probably it contains sensitive information). But this is possible only when every intermediate system that the event passes through understands cloud-events and hence respects the semantics defined by it.

duglin:
I was wondering if log-level should be part of `data`, and therefore not in scope for us, since not all events are logging events.  Or, if people need it outside of the data so they can do processing on it w/o looking at the `data` attribute, one thing we could consider is producing a set of "profiles" for categories of events. For example, perhaps we could define a "logging event profile" that defines some well-known properties (like `log-level`) that are extensions to the base specification. This way we have the core spec + well-defined extensions for interop in a particular domain.

schatekar:
+1 for profiles. 

OpenId Connect has taken this approach where there is core spec and then there are domain specific extensions that build on top of core spec. This has two advantages

1. Core spec remains domain neutral and focussed
2. Gives an opportunity for the spec to grow across multiple domains while core team is busy fine tuning the core

For this to work effectively, the core specification should be close to its final stage. That guarantees that the core will not change significantly while people are building domain specific profiles.

spec.md
```diff
@@ -221,6 +221,11 @@ that contains both context and data).
 * Constraints:
   * OPTIONAL
 
+### log-level
+* Type: String
+* Description:  The level of logging that the publisher may want to log this
```
duglin: need to remove the 'may'.
Perhaps: The level of logging associated with this specific event (e.g. debug)

Can you also add a "Constraints" section? Probably with OPTIONAL

duglin:
@vbmade2000 what is the status of this PR?

vbmade2000:
@duglin I am sorry for the delay. I have added log levels. Set of log-levels is a subset of syslog protocol log levels. I felt few levels unnecessary so removed them.

barkerd427:
If the intention of this is to manipulate the logging level of a consumer, then I would expect this to become a vector for an attack or completely ignored. If a malicious actor can elevate the logging level of a victim system to debug, then they would likely collapse most systems under the load generated in that mode. There could be a control in the consumer, but it seems non-intuitive to have an incoming message manipulate your logging level. If this is meant to force additional logging based on a single message, then I would expect that logic to exist in the consumer and be based on a combination of other attributes.

yaronha:
To sum up, the log-level is a way to specify the verbosity level for the event processing, not related to the specific even type, e.g. a developer issues an event and want to track why it failed or had certain result, he can turn on debug level and get more details in the serverless function logging report. 

Given that it is not widely adopted and may create some security issues it is recommended to add this through the extensions mechanism.

markpeek:
Aligned with some of the discussion in the 2018-03-15 meeting, this could be classified as an extension rather than formalizing as optional in the spec. I'd like to see this tagged as a possible extension for further discussion after the minimum fields have been agreed/approved in the main spec.

duglin:
@vbmade2000 to clarify - we need to know what the intent of this attribute is. Is it:
1 - a mechanism to tell the receiving infrastructure what the logging level of the infrastructure should be, or
2 - the logging level of the sending infrastructure sending the event, or
3 - this IS a logging message/event and this log-level attribute is tells us what logging level value the sending infrastructure was set to (almost like 2)

SeanFeldman:
The systems our customers build we consider `log-level` a meta-data that would typically be provided and consumed by customers systems. As such, it would be under `extensions`.

duglin:
@vbmade2000 are you still interested in this one?

vbmade2000:
@duglin Yes I am. The intent of this attribute is to be considered as a mechanism to tell the receiving infrastructure what the logging level of the infrastructure should be. I am not eperienced in this area but after going through replies I think I would agree with @yaronha and @markpeek that this attibute can be considered as an extension.

duglin:
ok, maybe we should consider it for this: https://github.com/cloudevents/spec/pull/126  if adopted.

duglin:
@vbmade2000 perhaps you could modify the text to make the intent clear since there was some confusion?

vbmade2000:
@duglin Sure.

duglin:
Any more comments on this one?

duglin:
At 6/15 f2f we agreed to close this in favor of tracing

PR #64, opened 2018-02-03
added content-type attribute

### LINKED ISSUE #23: content-type context attribute
This is a proposed metadata attribute (from Context Attribute Backlog):

content-type
    Type: String
    Description: The data encoding scheme (e.g. application/json). The data content type.
    Constraints:
        Required if the event contains a body.

### LINKED ISSUE #59: content-type should be explicit
We need the `content-type` attribute to be explicit (not part ofthe extensions), much like the schema is, without it there is no way to know how to de-serialize the `data`  

Assigned to: @yaronha

yaronha:
Closes #59 
Closes #23

Signed-off-by: Yaron Haviv  yaronh@iguazio.com

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -200,6 +200,15 @@ that contains both context and data).
   * If present, MUST adhere to the format specified in
     [RFC 3986](https://tools.ietf.org/html/rfc3986)
 
+### content-type
+* Type: String per [RFC 2046](https://tools.ietf.org/html/rfc2046)
+* Description: Describe the data encoding format 
+* Constraints:
+  * OPTIONAL
+  * If present, MUST adhere to the format specified in
+    [RFC 2046](https://tools.ietf.org/html/rfc2046)
+* For Media Type examples see [IANA Media Types](http://www.iana.org/assignments/media-types/media-types.xhtml)
+
 ### extensions
 * Type: Map <String, Object>
 * Description: This is for additional metadata and this does not have a
@@ -212,12 +221,12 @@ that contains both context and data).
   * If present, MUST contain at least one entry
 * Examples:
   * authorization data
-  * content type
 
 ### data
 * Type: Arbitrary payload
 * Description: The event payload. The payload depends on the event-type,
-  schema-url and event-type-version.
+  schema-url and event-type-version, the pyload is encoded into a media format 
+  which is specified by the content-type attribute (e.g. application/json).
 * Constraints:
   * OPTIONAL
 


yaronha:
added 
Signed-off-by: Yaron Haviv yaronh@iguazio.com

duglin:
@yaronha it needs to be part of the commit.  Try:
git commit -s --amend --no-edit
git push BRANCH-NAME -f

duglin:
make sure you have your git client user.name and user.email config properties set first. See the git notes from the call for more info: https://docs.google.com/presentation/d/1tWYPiwmGrzf_38e4qbPc4EhVQRdVoLwQktbUIHto2G8/edit#slide=id.g306755fb49_0_7

spec.md
```diff
@@ -200,6 +200,15 @@ that contains both context and data).
...
     [RFC 3986](https://tools.ietf.org/html/rfc3986)
 
+### content-type
+* Type: String per [RFC 2046](https://tools.ietf.org/html/rfc2046)
+* Description: Describe the data encoding format 
+* Constraints:
+  * OPTIONAL
+  * If present, MUST adhere to the format specified in
+    [RFC 2046](https://tools.ietf.org/html/rfc2046)
+  * For Media Type examples see [IANA Media Types](http://www.iana.org/assignments/media-types/media-types.xhtml)
```
duglin: Minor tweak: remove the space before the "*" so the media type example text is not under "Constraints".

spec.md
```diff
@@ -212,12 +221,12 @@ that contains both context and data).
   * If present, MUST contain at least one entry
 * Examples:
   * authorization data
-  * content type
 
 ### data
 * Type: Arbitrary payload
 * Description: The event payload. The payload depends on the event-type,
-  schema-url and event-type-version.
+  schema-url and event-type-version, the pyload is encoded into a media format
```
duglin: s/pyload/payload/

duglin: you may want to make this more normative, something like:
If the `content-type` attribute is specified then the `data`
MUST be encoded per the rules defined by the specified format.

spec.md
```diff
@@ -200,6 +200,15 @@ that contains both context and data).
   * If present, MUST adhere to the format specified in
     [RFC 3986](https://tools.ietf.org/html/rfc3986)
 
+### content-type
+* Type: String per [RFC 2046](https://tools.ietf.org/html/rfc2046)
+* Description: Describe the data encoding format 
+* Constraints:
+  * OPTIONAL
+  * If present, MUST adhere to the format specified in
```
schatekar: Is there a default if this attribute is not present?

yaronha: its not different than schema-url, if you dont have it you need to figure out how to interpret the data/body yourself

spec.md
```diff
@@ -212,12 +221,12 @@ that contains both context and data).
...
 * Examples:
   * authorization data
-  * content type
 
 ### data
 * Type: Arbitrary payload
 * Description: The event payload. The payload depends on the event-type,
-  schema-url and event-type-version.
+  schema-url and event-type-version. If `content-type` attribute is specified 
+  then the `data` MUST be enconded accordingly to the specified `content-type`.
```
duglin: s/enconded/encoded/

duglin:
minor typo, but otherwise LGTM

duglin:
@yaronha see #23 - it talks about how, perhaps, content-type should be REQUIRED if there's a `data` property. You have it as OPTIONAL.  Any comment on that?

yaronha:
@duglin fixed the typo, i see content-type is not mandatory in HTTP, so suggest we keep it optional

duglin:
@yaronha thanks for addressing the other comment.
still LGTM

ac360:
LGTM.  This was also in our original scope, since this is an essential property.

ultrasaurus:
doesn't event-type cover this?

djrosanova:
I don't think this property makes sense at this scope. In order to understand the envelope (where this property lives) you already needed to know the encoding before hand, this would allow you to change encoding in the middle of an event for one particular field. This seems odd to me and could cause problems. E.g. the event is JSON, but the data field is something else like XML or protobuf or CSV, it could easily contain non-well formed JSON which would break the entire object / envelope.

maplebed:
Given that `data` is described as "Arbitrary payload", it is implied that the envelope (the `Event`) should not make assertions about how the data is encoded. In order to ease the job of the recipient of the Event, adding this field allows the sender to include information about how the data is encoded.  This is distinct from how the `Event` is encoded (which would belong in a header in whatever transport layer is used). 

It makes sense to me to include and provide an inline alternative to using the `schema-url` field.

Concerns about the content of the `data` field breaking the encoding of the surrounding envelope must be dealt with no matter what, given that data is already defined as "arbitrary". Regardless of encoding, it will be able to include characters that the envelope encoding might need to escape. That must be dealt with separately, and is independent of this PR.

LGTM.

schatekar:
Before we decide, whether we need `Content-Type` attribute or not, where does it sit, how does it get handled etc. I would like to lay out my understanding of events get represented in different places. I may be wrong (in part of fully) so please correct me. 

This specification, like HTTP or JWT spec, defines an implementation-neutral structure for a particular kind of information i.e. events, to be (1) represented in a software system (2) transmitted from one software system to another. I don't think that the core of this specification directly deals with both of those aspects as both are implementation specific representations of the specification. We need two different representations of an event in order to be able to use the CloudEvents meaningfully

#### 1. In-memory representation

This is realised when a software instance is handling the event. There are two ways an event could end up in having an in-memory representation

__[A] When a software system creates an instance of a CloudEvent__

Any software system that wants to publish a CloudEvent, will first need to create one. It will most likely depend on a language/framework specific implementation of the CloudEvents spec in the form of an OSS library. This library will use the primitives of that specific language/framework for an in-memory representation of the event e.g. Java may use a class, JavaScript may use a JSON object etc. These libraries may offer multiple ways for a developer to define how their arbitrary data payload looks. There may also be a need to encrypt or/and sign `data`. 

__[B] When a software system receives an instance of CloudEvent__

There is a bit of a chicken-and-egg situation with this. Difficult to talk about this without we have talked about __On-wire representation__, so feel free to read the next section before this. 

A consumer, when it receives a CloudEvent, it needs to be able to deserialise the whole event i.e. both `context` and `data`.  For the deserialisation to work, it needs to know 

1. How is the event encoded?
2. How is the event structured? 

The event will be encoded by the transport for its own purpose. The spec can rely on that for the encoding of `context` at least. However, `data` can be encoded in the same way or differently depending on a number of different factors. 

The structure of the event is pretty straightforward.  The structure of `context` is always fixed, defined by this spec. The structure of `data` is dynamic and defined using `schema-url`. 

Once deserialised, the in-memory representation of the event, in this case, will be no different from the case [A]


#### 2. On-wire representation

This is realised when the producer of the event wants to transmit the event to the target/intermediate system. This representation will depend on 

1. What type of transport is being used for transmitting the event? Some transports will have a limited set of encoding options available. 
2. What is accepted by the target endpoint that the event is being transmitted to

Again, the library that the developers use for in-memory representation of the event, can provide with options to transmit the events using different transports. There may be a requirement to transfer some of the attributes from `context` over into the transport-specific mechanics e.g. headers, envelopes etc. 
	
#### Questions we need to answer

1. Does this spec need to define how a language/framework specific implementation represents the event at runtime i.e. in-memory representation? 
2. Does the spec need to define context attributes to support encrypting/signing of data?
3. Does the spec (optionally/mandatorily) define that `data` can be encoded separately from the whole event while in transit?
4. If the answer to 3 is affirmative, what does the spec need to include in order to prevent one encoding from corrupting the other encoding?
5. Does the core spec need to constraint how the transport encodes the event?
6. Does the core spec need to define which context attributes CAN/MAY/MUST be copied over to transport specific mechanics e.g. headers, envelopes? Or is this rather left to the understanding between the publisher and the consumer of the events?

Where I say "core spec" I am indicating that the specific item has a place in some spec but may be not in this one.

duglin:
While I agree that we need to come to an agreement around what it is we're actually producing (on-the-wire representation of events, in-memory, an abstraction....) I think we should talk about that outside of this issue. 

To me `content-type` states the high-level format of the `data` field. It will tell you whether its plain text, xml, json or something else. Whether or not the `data` field, or any other field including the entire event, was transmitted over the wire in a different format is not relevant. I would prefer if we assumed, for the moment, that the current fields are defining the abstract set of fields that make up an event - nothing more.

We can then, in follow-on PRs (or even work streams) decide how these fields are manifested on the wire, or presented to the applications (functions) using them.

schatekar:
@duglin I agree with your comment. 

However, in the call, we had agreed that all the suggestions/counter-views be presented here on this PR. If we want to merge/close this PR, what's the best way to capture those suggestions/ideas?

duglin:
@schatekar I think we could just agree, out of the spec, that for now we're just talking about the abstract definition of an event, or if people want some in the spec itself we could add something like:
```
```
as part of this PR.

And then, either way, accept the PR.

schatekar:
@duglin LGTM with the adoption of your suggestion above.

duglin:
@yaronha what do you think?

yaronha:
@schatekar @duglin im not sure what does `in-mem representation` mean, its simpler when talking about Python or JS, but for C++ or Go (typed lang) we cannot assume the framework de-serialize the `data` for the consumer app, the application may want to gain control of that (e.g. get the Data as Blob, get individual fields, get an iterator, or get it fully de-serialized). 

i guess having `content-type` makes our spec more complete/usable, we can start with very simple APIs such as `GetHeader(extension-name) -> string` + `GetData() -> Blob`, and later on add flexibility (e.g. `GetDataStruct() -> Language-struct` and/or `GetField(data-field/path) -> Object`, BTW the content may be unstructured, e.g. contain a JPEG pic sent from a remote camera.

In the broader spectrum IMO we cant stay at the high level and need a reference implementation of event consumer API for major languages as the minimum, and preferably a definition how such events traverse the wire in an interoperable way, going back to my original CloudEvent scope diagram

Scope

schatekar:
@yaronha  `in-memory representation` is the term I picked up from our last meeting. I think it refers to how an event gets represented in a system at runtime. For example, a statically typed language like Java may use a class with templates to represent an event. 

I think your example of JPEG sent from a remote camera makes it clearer now. The consumer can work out from the `event-type` attribute that they are expecting an image blob but they would not know whether it is a PNG/JPEG or a proprietary format unless there is another field that specifies this.

deissnerk:
So we may have libraries that provide transport of events over some L7 protocol like HTTP or AMQP. The content-type is, what event providers and consumers put into and retrieve from those libraries. Correct?
LGTM

oritnm:
LGTM

djrosanova:
I think this field really only makes sense in terms of non-schematizable inline data like JPG or Zip. As such it should be mutually exclusive to schema-url, which I would think could tell you if the schema itself is JSON or XML. It wouldn't make sense to have a JGP and a schema-url.

yaronha:
@duglin your suggestion LGTM

duglin:
Add https://github.com/cloudevents/spec/pull/84 to address my comment.
Perhaps we could merge this PR and then next week merge #84 and that might cover
people's concerns.

sslavic:
We have specific use-case like https://github.com/cloudevents/spec/blob/master/use-cases.md#increasing-portability-of-functions-as-a-service
Is it possible to address such use case with abstract definitions? Or should we come up with a practically applicable spec and supporting resources which are not subject to interpretation.

duglin:
Approved on the 2/15 call

duglin:
@djrosanova please open a new PR if you think we need to add text about when those 2 fields can be used (or when only one can be used).

Issue #58, opened 2018-02-01
Revisit the name "source" in Events


inlined:
On February 1st 2018, we tentatively changed "resource" to "source". We agreed that "resource" is too confusing, but also that "source" may not be the best name. After clarify the meaning of this term we can reaffirm or change the name.

Assigned to: @inlined

schatekar:
@duglin the idea of profiles can simplify this issue greatly. Irrespective of the domain/use case, there will always be an `origin` or `source` indicating who the publisher of the event is. 

Then a FaaS profile can add relevant extension fields to capture the information about various intermediate systems that the event passes through before it hits the function.

inlined:
This seems to be quite settled.

PR #68, opened 2018-02-07
Clarify the meaning of "source" as where the event occurs.

inlined:
Related to #58, this is a follow-up to last week's meeting to clarify the meaning of source (fka resource).

The change is minimal and builds on the vocabulary of "occurrence" to be clearer that this is the first piece of software where the event originates, not the latest software to forward the event in a federated system. I've also added real examples from Google Firestore in case our interpretation helps others.

I'll add another PR to suggest an additional source-service.

Signed-off-by: inlined <inlined@google.com>


spec.md
```diff
@@ -148,30 +148,30 @@ that contains both context and data).
...
   * s3
+  * "google.firestore.document"
 
 ### source-id
 * Type: String
-* Description: ID of the event source.
+* Description: A unique identifier of the object on which an event occurred. The source-id is of kind source-type.
 * Constraints:
   * REQUIRED
   * MUST be a non-empty string
```
duglin: I think we should add a statement in this Constraints section about the uniqueness aspect - is it globally unique or just unique within the scope of the source system?

spec.md
```diff
@@ -148,30 +148,30 @@ that contains both context and data).
...
 * Type: String
-* Description: Type of the event source. Providers define list of event
-  sources.
+* Description: The kind of object on which an an event occurred. Providers define list of source-types.
 * Constraints:
   * REQUIRED
   * MUST be a non-empty string
 * Examples:
   * s3
+  * "google.firestore.document"
```
duglin: Not sure which way we want to go (quotes vs no no-quotes) but since the rest of the spec (as of now) doesn't have quotes on these examples can you remove them? If you think we should have quotes on all then let's open up a separate issue/PR and get them all in one shot.

spec.md
```diff
@@ -148,30 +148,30 @@ that contains both context and data).
 
 ### source
 * Type: Object
-* Description: This describes the software instance that emits the event at
-  runtime (i.e. the producer). It contains sub-properties (listed below)
+* Description: This describes the software where an event occurred. It contains sub-properties (listed below)
```
duglin: Can you wrap still at 80 cols to align with the rest of the spec?

spec.md
schatekar: `object` may be ambiguous? e.g. s3 is not an object but a service.

inlined: Honestly I struggled here a bit and decided to toss it back for discussion. If we add `source-server`, then it's very easy to say that `source-server` is a physical or logical address and `source-id` is a unique resource hosted by `source-server`.

As it is, I can't actually say that this is a globally-unique value. 10.x.x.x IP addresses are totally valid but may not be _globally_ unique.

spec.md
```diff
@@ -148,30 +148,33 @@ that contains both context and data).
...
   * MUST be a non-empty string
 * Examples:
   * s3
+  * google.firestore.document
 
 ### source-id
 * Type: String
-* Description: ID of the event source.
+* Description: A unique identifier of the object on which an event occurred. The
+  source-id is of kind source-type.
```
duglin: I'm not sure what you mean by it "is of kind source-type". Does this mean it can't just be a UUID?

inlined: Maybe the Google examples could help. In a google.storage.object.finalize event (when an object is ready to serve), the object is of type "google.storage.object" so source-ids must always be a path to an object, not a bucket, pubsub topic, etc.

Do you have suggestions to better describe this?

duglin: That seems implementation specific since to some sources an ID could just be a uuid.
Would this wording work for you:
- - -
An identifier of the object on which the event occurred. The format
of this ID will be source specific.
- - -
And under Constraints, add:
- - -
MUST be unique within the scope of the `source`.
- - -
Then each source implementation can define the shape/syntax of this ID/string w/o forcing one model on everyone.

duglin:
On a related question... does it make sense to have the spec define multiple fields for `source`? It seems to me that if we split things into multiple fields then we have to get agreement on how many different ways people may want to split the info (eg. type, id, service....) - and that might be a problem since not all systems will have all of those bits of data. Additionally, it seems like the definition of those fields would most likely be source-specific and only meaningful to someone who understands the source environment.  What I'm wondering is whether it would make more sense to just have one field (as a string) and tell people to encode all of the various bits into one field - e.g. `source-type:source-id:source-service`.  Then each impl can decide how to encode their data w/o needing to get buy-in from all WG participants on all possible fields and ways to encode the data. We don't really need interop on this because its going to be source-defined anyway - so in order to use this data (aside from just storing it for audit purposes) you'd need to understand it anyway and you'd then know how to split up the single string.

Just a thought.

duglin:
ping @inlined

deissnerk:
I think that at least a separate type field would help a processor to easily determine, if it can interpret the rest of the source information. I actually also like the idea of the proposed three fields.
@inlined Thanks for the example!

djrosanova:
Changing the name to source doesn't seem to have solved problems here, looks like it make more. If source was a resource, or a Universal Resource Identifier, there wouldn't be a question of source ID would there?

deissnerk:
Resource is a very generic term. When I read the draft spec first, I did not understand the intention. Source to me made it much clearer. IMO the name and the format of the attribute need to be distinguished. We could call it Source and require it to be a URL, although I prefer the current proposal.

djrosanova:
I just don't think we'd need a source-id if we all knew what source was. Naming it source instead of resource doesn't change that. Ignoring 30 years of HTTP and the web just seems.... interesting.

inlined:
@djrosanova I'm not 100% sure what you mean by "if we all knew what the source was."

The original proposal shape for Source was the Istio "Resource" type, which is based on, but not strictly, a URI + additional metadata.
The service and id field (known as "name" in Istio) are generally the URI. In Google APIs they tend to be the URI without the version number attached to it. Additionally the "type" field adds semantic meaning, and Istio has an "attributes" field for labels on the resource.

spec.md
inlined: "That seems implementation specific since to some sources an ID could just be a uuid." In some systems, a uuid would work, but I would say if and only if that UUID is used to serve data. I would have assumed that a PUT to a RESTful service at ${source.service}/${source.id} would generate an event with those respective fields.

spec.md
```diff
@@ -148,30 +148,33 @@ that contains both context and data).
...
-* Description: Type of the event source. Providers define list of event
-  sources.
+* Description: The kind of object on which an event occurred. Providers
+  define list of source-types.
 * Constraints:
   * REQUIRED
   * MUST be a non-empty string
 * Examples:
   * s3
+  * google.firestore.document
```
ultrasaurus: `s3` and `google.firestore.document` seem to be conflicting examples -- can you revise the example list or add text to clarify if this is intentional?

inlined:
@duglin I've taken your suggestions verbatim to keep the conversation moving, though I think we'll need tomorrow's set of presentations to get everyone on the same page. I think there are three different discussions that are getting entangled:

1. **What do we call it?** I think there's generally consensus that "source" is better than resource because many things could be called a resource (e.g. the event source or the action).
2. **Which system does it describe?** The point of this pull request was to suggest/clarify that the "source" is the _original_ software. If an event broker/relay is used to deliver the event, it does not become the new source.
3. **What is its format?** Contrary to my rejected proposal to split `source-id` into a tuple of service + id, I'm not generally against URIs as a SHOULD/MUST restriction once it became clear that the committee allows relative URIs. This does add some odd complications for those handling the event. A generic action doesn't actually know the meaning of the URI other than it is unique; the semantic parsing of `source-id` depends on the `namespace` or `event-type` fields (or the tuple of the two?).
I'm honestly still wrapping my head around the implications, especially for open source software. What if I want to receive events from my open source database deployment? The database clearly has software to decide the relative portion of the URI for each event, though I as the deployer should possibly get to influence what the reported host portion of `source-id` is. Should the database have hard-coded that the host is the local IP address? On which interface (if serving 0.0.0.0)? Can I instead tell the database to report its authority as the K8S service selector associated with the replicaset?

(Meta) I wonder if we could use a better markdown layout for defining object properties. I'm not sure it's clear that `source-id` is just the section header for the `id` subproperty of `source`. I'm confused by the confusion between source & source-id.

spec.md
```diff
@@ -148,30 +148,34 @@ that contains both context and data).
...
-* Description: Type of the event source. Providers define list of event
-  sources.
+* Description: The kind of object on which an event occurred. Providers
+  define list of source-types.
 * Constraints:
   * REQUIRED
   * MUST be a non-empty string
 * Examples:
-  * s3
+  * aws:s3:object
```
inlined: ALERT AWS USERS:
I previously avoided touching the AWS samples because I'm a Google employee and don't want to give the impression that I'm dictating what Amazon should do. @ultrasaurus pointed out that leaving the sample untouched made the examples incongruous so I made one up. I'd love an Amazon employee or pro to help me here. My thought process:

1. Amazon seems to be relatively inconsistent between using an "aws:" prefix (e.g. ARNs) or not (e.g. event types). I chose the former on a whim.
2. Amazon clearly prefers colon separators to period separators.
3. Since all S3 event types are on objects, the source-type here was an object, not an "s3" (bucket).
4. I couldn't find a policy "type" in Amazon for an S3 object; Amazon IAM uses <verb><noun> nomenclature like "s3:PutObject", so I made up a type string using the noun ("object") that their API uses.

duglin:
@inlined given today's call, can we close this?

inlined:
Yup. I think we've made a lot of progress since this PR was needed.

PR #69, opened 2018-02-07
Propose a new "service" sub-field of "source"

inlined:
(Note: this is a compound review; it may be more useful to skip the first commit when reviewing)

At Google we have found it useful to separate the service name from the rest of the resource string:

* When registering an event trigger, the service name is a key into an address where we register our hooks. This service name can be a key into a private API hosted on another server address or on another protocol (e.g. gRPC)
* The service name can tersely be the only difference in trigger or event objects between:
  * Production, integration, and development tiers of a service.
  * Common APIs hosted across multiple services by a cloud provider (e.g. IAM access control).
  * Enterprise versions of software such as IBM OpenWhisk, GitHub, or HipChat.
  * Open-source services running in local infrastructure.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -170,8 +170,21 @@ that contains both context and data).
 * Constraints:
   * REQUIRED
   * MUST be a non-empty string
+  * MUST be a unique resource relative to source-service
 * Examples:
-  * my.s3.bucket
+  * mybucket.s3.amazonaws.com
+  * /projects/myProject/databases/(default)/documents/users/aNewUser
+
+### source-service
+* Type: String
+* Description: The physical (e.g. IP address) or logical (e.g. DNS address) of
+  the source software.
+* Constraints:
+  * REQUIRED
+  * MUST be a non-empty string
+* Examples:
+  * s3.amazonaws.com
+  * firestore.googleapis.com
 
 ### event-id
 * Type: String


spec.md
```diff
@@ -148,30 +148,39 @@ that contains both context and data).
 
 ### source
 * Type: Object
-* Description: This describes the software instance that emits the event at
-  runtime (i.e. the producer). It contains sub-properties (listed below)
+* Description: This describes the software where an event occurred. It contains sub-properties (listed below)
```
duglin: Can you wrap still at 80 cols to align with the rest of the spec?

spec.md
```diff
@@ -148,30 +148,39 @@ that contains both context and data).
...
 * Type: String
-* Description: Type of the event source. Providers define list of event
-  sources.
+* Description: The kind of object on which an an event occurred. Providers define list of source-types.
 * Constraints:
   * REQUIRED
   * MUST be a non-empty string
 * Examples:
   * s3
+  * "google.firestore.document"
```
duglin: Not sure which way we want to go (quotes vs no no-quotes) but since the rest of the spec (as of now) doesn't have quotes on these examples can you remove them?  If you think we should have quotes on all then let's open up a separate issue/PR and get them all in one shot.

spec.md
```diff
@@ -148,30 +148,39 @@ that contains both context and data).
...
 
 ### source-id
 * Type: String
-* Description: ID of the event source.
+* Description: A unique identifier of the object on which an event occurred. The source-id is of kind source-type.
 * Constraints:
   * REQUIRED
   * MUST be a non-empty string
```
duglin: I think we should add a statement in this Constraints section about the uniqueness aspect - is it globally unique or just unique within the scope of the source system?

spec.md
```diff
@@ -148,30 +148,39 @@ that contains both context and data).
...
+* Description: This describes the software where an event occurred. It contains sub-properties (listed below)
 * Constraints:
   * REQUIRED
   * MUST contain at least one non-empty sub-property.
 
 ### source-type
 * Type: String
-* Description: Type of the event source. Providers define list of event
-  sources.
+* Description: The kind of object on which an an event occurred. Providers define list of source-types.
```
sslavic: ".. an an.."

duglin:
I moved my comments over to the other PR #68

spec.md
```diff
@@ -148,30 +148,39 @@ that contains both context and data).
...
+  * "/projects/myProject/databases/(default)/documents/users/aNewUser"
+
+### source-service
+* Type: String
+* Description: The physical (e.g. IP address) or logical (e.g. DNS address) of the source software.
+* Constraints:
+  * REQUIRED
+  * MUST be a non-empty string
+* Examples:
+  * "firestore.googleapis.com"
```
duglin: remove quotes for now.

spec.md
```diff
@@ -148,30 +148,39 @@ that contains both context and data).
...
 * Constraints:
   * REQUIRED
   * MUST be a non-empty string
 * Examples:
   * my.s3.bucket
+  * "/projects/myProject/databases/(default)/documents/users/aNewUser"
+
+### source-service
+* Type: String
+* Description: The physical (e.g. IP address) or logical (e.g. DNS address) of the source software.
```
duglin: please wrap at 80 cols

spec.md
duglin: @inlined you may want to rebase this PR so it just has the new field and not the other PR, otherwise you may continue to get comments that are best addressed in #68

spec.md
sslavic: What would be `source-service` for s3 use case? If s3 is `source-service`, what should be `source-type`? "bucket"?

spec.md
inlined: My GitHub foo may be lacking. I made this a second PR because the source-id lines will otherwise conflict. That might be better though and I can handle the conflict when both are approved.

spec.md
inlined: Will do

spec.md
inlined: (@sslavic fixed in #68 thanks)

inlined:
Rebased to isolate the change and addressed feedback.

spec.md
```diff
@@ -170,8 +170,21 @@ that contains both context and data).
...
+  * MUST be a unique resource relative to source-service
 * Examples:
-  * my.s3.bucket
+  * mybucket.s3.amazonaws.com
+  * /projects/myProject/databases/(default)/documents/users/aNewUser
+
+### source-service
+* Type: String
+* Description: The physical (e.g. IP address) or logical (e.g. DNS address) of
+  the source software.
```
duglin: In an environment where software components can move easily, how would someone use this field later on? In particular if its the IP.

Is the source-id not something that can be mapped to the source-service from within the source's environment?

I'm not asking to push back - I'm just trying to understand because it feels like this might be a dup of source-id to some extent.

inlined: Were I using this in my own software I would strongly prefer something other than IP (e.g. DNS). I didn't want to require DNS as a prerequisite to an events system.

deissnerk: Could the source service also be an instance id according to the [Open Service Broker API](https://www.openservicebrokerapi.org/)?

duglin: yea, I would prefer a description that allowed more flexibility w.r.t. what the "source" considers useful to help identify where the event came from.  At least, that's what I view this (and other source-*) fields being used for. Is that consistent with others?

duglin: Also, see this comment: https://github.com/cloudevents/spec/pull/68#issuecomment-365525662

inlined: @deissnerk Absolutely. my intuition would be that the source-service would be the domain where the Service Broker were hosted and the source-id would be "/service_instances/:instance_id". Using a full resource path rather than just the ID would help us distinguish from an event, say, coming from the catalog when a new service was made available.

duglin:
Agreed to CWNA (close with no action) on the 2/15 call.

There was general consensus that people would prefer a single value (e.g. URI) to identify the "source" rather than to try to split it out into separate fields.  If people want to reduce the number of source-related fields we currently have then please open PRs with concrete proposals to do so.

spec.md
deissnerk: @inlined Yes, that would be nice, but is "/service_instances/:instance_id" really a source-id here? For the source-id you have to refer to some object inside the service like a queue or a table. "/service_instances/:instance_id/:object_id" would be a confusing URI, as it would look like a URL.

deissnerk:
> There was general consensus that people would prefer a single value (e.g. URI) to identify the "source" 

@duglin This PR was just about the source-service, right? So I hope the source-type is still there.

duglin:
@deissnerk it wasn't totally clear to me how many of the fields people wanted to consolidate. But, until someone opens a PR to remove it, yes source-type is still there.  :-)

PR #77, opened 2018-02-08
removing trailing whitespace from the end of lines

maplebed:
My editor keeps automatically stripping whitespace from the ends of lines. I don't think the linter for this repo cares one way or the other, but it seems unfortunate for 3 lines to end with a space while the rest don't. 

Normally I would just include this in some more relevant actual PR but in the interests of not confusing things, making it its own change. 

Signed-off-by: Ben Hartshorne <ben@honeycomb.io>


duglin:
my OCD thanks you :-)
LGTM

duglin:
ok I'm going to take @markpeek's thumbs up as an implicit LGTM on this controversial PR and merge it :-)

PR #80, opened 2018-02-13
Add description for jwt_claims extension

inlined:
Adds a new section to the spec to list known extensions. This will help us reuse common ideas.

Adds a new extension for JWT claims.

For background, the Google event structure has a currently-undocumented feature where the Realtime Database can forward information about the user who triggered the event. The `firebase-functions` library uses this, though we would like to document it and expose it to our developers. We are proposing a name & semantics to the committee in case we can agree on a format together before release.

### DIFF
--- a/extensions.md
+++ b/extensions.md
@@ -1,0 +1,59 @@
+# Registered Extensions
+
+The `extensions` field of a Cloud Event holds additional metadata in an event.
+This field may be both used for experimental features not yet ready for
+standardization or for fields that are domain-specific and not suitable for
+standardization.
+
+To encourage common design patterns where they make sense, an event source MAY
+choose to use a registered extension. When using an extension with a registered
+name, an event source SHOULD behave consistently with the description.
+
+## jwtClaims
+* Type: Object
+* Description: If the Event Source supports JWT authentication, this extension
+  allows the Source to provide the identity of the user who triggered the
+  occurrence of the Event. The jwt_claims extension is unsigned and decoded
+  claims portion of a JWT. This allows event systems to forward identity
+  information without sufficient information to authorize further requests.
+  An empty object MAY be used to indicate that the Event was triggered by an
+  anonymous user. For more information about JWTs, see 
+  [jwt.io/introduction](https://jwt.io/introduction).
+* Constraints:
+  * OPTIONAL
+  * If empty, MUST indicate that the Event was triggered by an anonymous user.
+  * The 
+  [registered claim names](https://tools.ietf.org/html/rfc7519#section-4.1)
+   of JWTs (e.g. 'iss', 'sub', 'aud', and 'exp') are reserved and MUST 
+   have the same meaning as RFC3519 when present.
+* Examples:
+
+```json
+// Actual value of the user signed in with Firebase Auth in the project
+// "inlined-junkdrawer". The user has signed in using Google as an
+// identity provider and their email address "inlined@google.com".
+{
+  "iss": "https://securetoken.google.com/inlined-junkdrawer",
+  "name": "Thomas Bouldin",
+  "picture": "https://lh4.googleusercontent.com/-_LyEJyXZklU/AAAAAAAAAAI/AAAAAAAAAW4/hiJ6WHKP8AI/photo.jpg",
+  "aud": "inlined-junkdrawer",
+  "auth_time": 1519171715,
+  "user_id": "TSNVi5u1tyRhSCbhtIXbk9Go9IK2",
+  "sub": "TSNVi5u1tyRhSCbhtIXbk9Go9IK2",
+  "iat": 1519171715,
+  "exp": 1519175315,
+  "email": "inlined@google.com",
+  "email_verified": true,
+  "firebase": {
+    "identities": {
+      "google.com": [
+        "113635245217169028038"
+      ],
+      "email": [
+        "inlined@google.com"
+      ]
+    },
+    "sign_in_provider": "google.com"
+  }
+}
+```
--- a/spec.md
+++ b/spec.md
@@ -242,7 +242,7 @@ they already suit the Event Source's needs.
   * If empty, MUST indicate that the Event was triggered by an anonymous user.
   * The 
   [registered claim names](https://tools.ietf.org/html/rfc7519#section-4.1)
-   of JWTs (e.g. 'iss', 'sub', 'aud', and 'exp') are reserved and must
+   of JWTs (e.g. 'iss', 'sub', 'aud', and 'exp') are reserved and MUST 
    have the same meaning as RFC3519 when present.
 
 ## Additional Topics & Questions
--- a/spec.md
+++ b/spec.md
@@ -222,6 +222,29 @@ that contains both context and data).
   * OPTIONAL
 
 
+## Known extensions
+The following extensions are in active use. To better support promotion of
+extensions to full-features, Event Sources SHOULD use known extensions when
+they already suit the Event Source's needs.
+
+### jwt_claims
+* Type: Object
+* Description: If the Event Source supports JWT authentication, this extension
+  allows the Source to provide information about which user triggered the
+  occurrence for the Event. The jwt_claims extension is unsigned claims portion
+  of a JWT. This allows event systems to forward authentication information
+  without fear of leaking end-user authorization. An empty object MAY be used
+  to indicate that the Event was triggered by an anonymous user. For more
+  information about JWTs, see 
+  [jwt.io/introduction](https://jwt.io/introduction).
+* Constraints:
+  * OPTIONAL
+  * If empty, MUST indicate that the Event was triggered by an anonymous user.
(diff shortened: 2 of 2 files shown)


spec.md
```diff
@@ -222,6 +222,29 @@ that contains both context and data).
...
+extensions to full-features, Event Sources SHOULD use known extensions when
+they already suit the Event Source's needs.
+
+### jwt_claims
+* Type: Object
+* Description: If the Event Source supports JWT authentication, this extension
+  allows the Source to provide information about which user triggered the
+  occurrence for the Event. The jwt_claims extension is unsigned claims portion
+  of a JWT. This allows event systems to forward authentication information
+  without fear of leaking end-user authorization. An empty object MAY be used
```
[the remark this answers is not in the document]
inlined: Yes, this is part of the reason why I had suggested passing the actual claims Object instead of the signed JWT token. It mitigates part of the risk to be PII disclosure rather than also allowing unauthorized impersonation of the user. Do you have wording you'd suggest?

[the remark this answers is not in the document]
schatekar: If the purpose of this field is to transmit information about the user who triggered the event then include only the `sub` claim is enough? There is nothing PII in `sub` claim and no unauthorised user can impersonate using just the `sub` claim. 

However, if the purpose of including this information is to actually authenticate the sender of the event, then you would need to include the full token. How is this any more secure than `HTTP Bearer Token` scheme? 

Going back to the wording question - my suggestion would be to remove any statement about how the spec helps stop leaking of sensitive information.

spec.md
```diff
@@ -222,6 +222,29 @@ that contains both context and data).
   * OPTIONAL
 
 
+## Known extensions
```
duglin: I'm not sure what it means to have a "known extension" in a spec. Is it a 2nd class field in some way? Is it less normative than the others?

If the intent is to define a set of fields that we think are good ideas but we're not yet prepared to make them part of the spec, I think it might be more appropriate to put them into a different doc so that its clear that they are experimental.  WDYT?

clemensv:
As a matter of principle and considering privacy implications in particular, it is rather problematic to leak the contents of a token to any party other than issuer, bearer, or designated audience. The fact that you're stripping the signature is a giveaway that you're now disclosing the token to a party that isn't supposed to have it (since it doesn't hold the key). Instead of funneling the token out, you should simply break the contained claims out into regular custom event properties. If an implementation insists on leaking the token contents, that's a private implementation decision, but I/we don't think it ought to be in the standard spec.

duglin:
@inlined there are some comments - did you want to address them?

inlined:
TL;DR (everyone): Thanks for the feedback. I'm going to move this into a separate document and use the word "identity" not "authentication" for what is forwarded. I think this makes the intent much more clear that the feature is about understanding _who_ the user was, not the action of (re)authenticating as that user in subsequent API calls.

@clemensv 
> As a matter of principle and considering privacy implications in particular, it is rather problematic to leak the contents of a token to any party other than issuer, bearer, or designated audience

I agree with your principle if not your conclusion. To use the STRIDE terms, there are two risks to worry about here: Spoofing and Information disclosure. Because the signature is stripped, we disallow spoofing unless the Action already had spoofing permission. This also has the benefit of preventing edge cases where the token expires before the Cloud Function can execute. Information disclosure is critically important, as are other fields like `data`. I've assumed that this whole eventing system rests upon a well-defined and secure way for a developer to register an even trigger and for that registration to be subject to authorization checks. I've also assumed that TLS and other protocols can be used to secure data in transit between the source and the action. If either of these are non-obvious, then we should add these discussions to our roadmap.

FWIW, this is an attempt to be transparent and document a feature already widely in use. The Firebase Realtime Database forwards end-user token information in its events, and the [`firebase-functions`](https://www.npmjs.com/package/firebase-functions) SDK uses this to let developers opt into downscoping their use of the database in Cloud Functions (the [databaseAuthVariableOverride](https://firebase.google.com/docs/reference/admin/node/admin.app.AppOptions#databaseAuthVariableOverride) feature tells the connection that reads/writes must be approved by _both_ software and user validation). It's absolutely not a security boundary, but multiple developers have told me that it's saved them from accidentally deleting their database due to software bugs.

Multiple devs have discovered the (currently) undocumented feature and depend on it to do something with respect to the end-user in their Cloud Functions when it is unreasonable to include the user ID in the document they write. I want to help those developers out by documenting our behavior and I want to help other platform providers by documenting it as a known use of the extensions field.

spec.md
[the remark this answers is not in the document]
inlined: There's a lot of good stuff densely packed in this comment. I'll try to inline my response:

> There is nothing PII in sub claim

A username is commonly considered PII, for example when the username is an email address or a common alias. Our architectures will be subjected to the same scrutiny internally regardless of whether we add more claims or not.

> and no unauthorised user can impersonate using just the sub claim.

No unauthorized actor can _impersonate_ the user unless they have the issuer's private key either. I thought about using just the sub claim. I backed out for two reasons:
1. When designing a platform, I often use the litmus test of "could this be used to itself build a platform?" All of these claims would be needed by a platform developer; their customers would each have their own issuer and that issuer would mint tokens for the platform developer's customers' customers.
2. These seem necessary for serverless/event-driven development to not regress from more traditional architectures (see next inlined comment)

> However, if the purpose of including this information is to actually authenticate the sender of the event, then you would need to include the full token. How is this any more secure than HTTP Bearer Token scheme?

This is a complex/emerging topic; some background might help. Our oldest best practices consider a single identity. In many systems (esp. with FaaS and events), there tend to be two kinds of identities:

1. Software identity (in Google, a "service account") that the developer manually configures and controls with access control like IAM. This identity system is what, for example, authorizes Google Cloud Storage to publish its events to the developer's Pub/Sub topic. 
2. End user identities that customers of a product use.

In traditional monolith architecture where the developer builds an API server for every client <-> server communication, this is easily accessed. The rise of BaaS means that a platform provider, not the developer, now handles the original request and is the authenticated identity that calls developer code (e.g. FaaS).

This information is not enough to further act _as_ the end user (though if the Action that receives the event has an "act as" permission it may be able to exchange the claims for a new token), but it is enough to act _on behalf_ of the user.

> my suggestion would be to remove any statement about how the spec helps stop leaking of sensitive information.

I completely agree. I'll remove this since it added confusion/contention.

spec.md
inlined: I'll move this into a second doc. I think there is value in allowing parties to publicly document their extensions so that others can decide to follow suit and use the same meaning. Many standards, such as JWTs, iana media types, etc. do this.

I'm not 100% sure whether I'd ever push this to be a top-level field for a few reasons:
1. Not all services implement JWTs
2. Requiring a service to forward its JWT claims severely raises the bar since it requires more plumbing (and may not be acceptable to many)
3. Being an extension means there's a proto-safe way to distinguish between exists-but-empty (forwarding that a request was unauthenticated) and does not exist (the service does not forward JWT tokens or the request was authenticated a different way)

duglin: +1
I like having a well known spot for the community to work together on extensions that might not ever become part of the spec, but still provides some level of interop value by just being well-defined some place.

duglin: To be clear, I'm not commenting on whether or not this particular extension should be included - I'm more focused on creating the structure (secondary doc) in which we can place extensions in general.

clemensv:
Information leakage from tokens may be done in practice, but that doesn't make it good. A token is specifically issued to a relying party (RP, audience) and signed for consumption by that RP.  If you are choosing to break the information out of that token and flow it downstream, that's a choice your application (the audience) can make, but at that point it is application-level information that has been intentionally separated from the JWT context. That data no longer represents "jwt_claims". It's information that the token audience specifically chose to extract from the token to pass on. 

An STS may choose to include a user's complete list of group memberships in a JWT on the suspicion that the RP may want to make an authorization decision based on that list. Handing that list to any other party is inappropriate. Also, consider that with OAuth2, for instance, the user gives consent for specific information disclosure on a service by service basis. In the age of GDPR, I would be rather careful with where that information, which makes up the JWT claims, is going.

extensions.md
```diff
@@ -0,0 +1,59 @@
+# Registered Extensions
+
+The `extensions` field of a Cloud Event holds additional metadata in an event.
+This field may be both used for experimental features not yet ready for
+standardization or for fields that are domain-specific and not suitable for
+standardization.
+
+To encourage common design patterns where they make sense, an event source MAY
+choose to use a registered extension. When using an extension with a registered
+name, an event source SHOULD behave consistently with the description.
```
duglin: s/SHOULD/MUST/ otherwise we don't get interop and lose the value.  In fact, you may want to add the boilerplate RFC2119 text to this doc too.

duglin: @inlined ^^^

inlined:
Sorry for the delay. We've been discussing this internally and will take long enough that it makes sense to close the PR for now. Will re-open the discussion after more research.

Issue #19, opened 2018-01-25
Remove "references" section


duglin:
Before we ship a formal release I think we should remove the "References" section that shows what existing providers do for their events. This is very useful today as we work through the initial design, but once we get agreement on our stuff we no longer need to see what other people are doing since it doesn't align with what we're recommending.

Assigned to: @notque

ultrasaurus:
I think this would be good material to keep around as background of what drove some of the design decisions.  Maybe we could create a sub-folder...  I expect / hope we'll also be documenting what we learn about real-world use cases and infrastrucutre that developers need to work with when implementing a system that supports events.  Not sure if these belong together, but could provide context for people understanding the spec.

duglin:
@ultrasaurus you have an AI (AI: Sarah to open an issue for a doc to explain the spec, ecosystem, how it all fits together, etc… work with Yaron), I wonder if that doc would be a good spot for this info, if we keep it around. I just don't think it belongs in our spec.

duglin:
Closing since this is now done

PR #89, opened 2018-02-20
move references out of the specification into supporting material

ultrasaurus:
addresses https://github.com/cloudevents/spec/issues/19

- named directory `/about` since that is the name of the directory in work-in-progress PR https://github.com/cloudevents/spec/pull/53
- proposed `/landscape` sub-directory, since I thought it would be nice to refactor this later into an area with a page for each product/software that implements this type of a system -- currently is for historical context, but in the future could contain references to systems that implement the spec

### DIFF
--- a/about/landscape/README.md
+++ b/about/landscape/README.md
@@ -1,0 +1,129 @@
+Prior to the CloudEvents specification, many systems implemented similar models for events that allowed late-binding of services to actions.
+
+The folowing references were evaluated in the formation of the first draft
+of the specification. (We welcome pull requests with more context or
+additional references)
+
+### Microsoft - Event Grid
+```
+{
+    "topic":"/subscriptions/{subscription-id}",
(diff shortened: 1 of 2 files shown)


about/landscape/README.md
```diff
@@ -0,0 +1,129 @@
+Prior to the CloudEvents specification, many systems implemented similar models for events that allowed late-binding of services to actions.
```
duglin: can you wrap this at 80 cols?

duglin:
Do we need a 2-level dir structure just for one file? I'd prefer to not create dirs until we have more than one file that lives in the dir.

Also, we'll need some kind of pointer to that file (e.g. from top README?), otherwise its just dangling and no one will find it.

duglin:
travis error looks real:
```
Running href checker:
> ./about/landscape/README.md
> ./CONTRIBUTING.md
> ./GOVERNANCE.md
> ./README.md
> ./roadmap.md
> ./spec.md
./spec.md: Can't find section '#reference' in ./spec.md
```

duglin:
@ultrasaurus can you fix the bad hrefs?

duglin:
oh and a rebase is needed

duglin:
I believe this one can be closed - all of the edits appear to already be done via some other PRs.

Issue #83, opened 2018-02-15
Definition: Occurrence


clemensv:
https://github.com/cloudevents/spec/blob/master/spec.md#occurrence 

The definition of "occurrence" is unnecessarily narrowing the scope of what an occurrence may be, especially by calling out specific technologies/protocols. 

Suggestion:

An "occurrence" is the capture of a statement of fact during the operation of a software system. This may occur because of a signal raised by the system or a signal being observed by the system, because of a state change, because of a timer elapsing, or any other noteworthy activity.

Assigned to: @clemensv

duglin:
@clemensv can you open a PR with your proposed change?

clemensv:
Yes. FWIW, I think it's useful to have issues for PRs.

duglin:
PR #90 is the PR for this one, its merged, so closing this issue.

Issue #85, opened 2018-02-15
Definition: Event


clemensv:
https://github.com/cloudevents/spec/blob/master/spec.md#event

The specification should either drop "occurrence" or nor restate a partial definition in "event".  

Suggestion:

An "event" is a data record expressing an occurrence and its context or a set of occurrences and their shared context. Events are routed from the emitting [re]source to interested parties for the purpose of notifying the about the [re]source occurrence. The routing may be performed based on information contained in the event, but an event will generally not identify a specific routing destination.

Assigned to: @clemensv

duglin:
@clemensv wanna open a PR?

clemensv:
Yes, I am logging issues first and then doing PRs

duglin:
@clemensv ok - we're not that formal here, you can go straight to a PR if you want. The extra noise isn't necessary.

ultrasaurus:
By the way, we agreed to use the term "source" for now, and I think dropping the "[re]source" will make the definition more readable.  Introducing the idea that an event should represent "a set of occurrences" is new -- I would think that packaging a group of events together would be a transport-level optimization, though I haven't thought about that deeply.

I do like your wording.  Would love to see a PR that just clarifies the wording, without adding the multiple occurrences for now.

PR #90, opened 2018-02-20
Definition of "Occurrence" #83

clemensv:
https://github.com/cloudevents/spec/issues/83

The definition of "occurrence" is unnecessarily narrowing the scope of what an occurrence may be, especially by calling out specific technologies/protocols.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -67,7 +67,9 @@ This specification defines the following terms:
 An "occurrence" is the capture of a statement of fact during the operation of 
 a software system. This might occur because of a signal raised by the system or 
 a signal being observed by the system, because of a state change, because of 
-a timer elapsing, or any other noteworthy activity.
+a timer elapsing, or any other noteworthy activity. For example, a device may
+go into an alert state because the battery is low, or a virtual machine is
+about to perform a scheduled reboot.
 
 #### Event
 Data representing an occurrence, a change in state, that something happened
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -64,10 +64,10 @@ be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
 This specification defines the following terms:
 
 #### Occurrence
-When something happens (or doesn’t happen) and is detected by a software
-system. This is most typically when that system receives an external signal
-(e.g. HTTP or RPC), though could also be through observing a changing value
-(e.g. an IoT sensor or period of inactivity).
+An "occurrence" is the capture of a statement of fact during the operation of 
+a software system. This may occur because of a signal raised by the system or
```
duglin: s/may/might/

duglin:
LGTM - after RFC issue is fixed

duglin:
still LGTM

inlined:
See comment in #85

duglin:
@clemensv rfc issue

duglin:
s/may/might/

duglin:
@clemensv correct me if I'm wrong but the diff between what is in there now vs your original proposal is just the last sentence:
- - -
For example, a device might go into an alert state because the battery is low, or a virtual machine is about to perform a scheduled reboot.
- - -
Which I don't consider to be a substantial change and so this should still be ok to consider for today's call.

Issue #86, opened 2018-02-15
Definition: Context


clemensv:
https://github.com/cloudevents/spec/blob/master/spec.md#context 

Suggestion for clarification:

A set of metadata attributes included with the event that describe circumstances of the occurrence that are not specific to the occurrence. Tools and application code use context information to identify the relationship of events to aspects of the system or to other events. Context attributes may describe the event source, the originating system, the emitting process, and other circumstantial information.

Assigned to: @clemensv

PR #92, opened 2018-02-20
Definition "Context" #86

clemensv:
Closes: https://github.com/cloudevents/spec/issues/86

Clarification that context information describes the circumstances of the event and not the occurrence per-se.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -80,8 +80,8 @@ destination. Events are used to notify other systems that something has happened
 #### Context
 A set of metadata attributes included with the event that describe circumstances
 of the occurrence which are not specific to the occurrence. Tools and application
-code may use context information to identify the relationship of events to aspects of
-the system or to other events. Context attributes may describe the event source,
+code can use context information to identify the relationship of events to aspects of
+the system or to other events. Context attributes describe the event source,
 the originating system, the emitting process, and other circumstantial information.
 
 #### Message
--- a/spec.md
+++ b/spec.md
@@ -78,11 +78,12 @@ regarding its semantic. Events are considered to be facts that have no given
 destination. Events are used to notify other systems that something has happened.
 
 #### Context
-A set of metadata attributes included with the event that describe circumstances
-of the occurrence which are not specific to the occurrence. Tools and application
-code can use context information to identify the relationship of events to aspects of
-the system or to other events. Context attributes describe the event source,
-the originating system, the emitting process, and other circumstantial information.
+As described in the Event definition, an Event contains two parts, the data
+representing the Occurrence and additional metadata that provides other
+circumstantial information about the Occurrence (e.g. information about the
+originating system). This additional metadata is referred to as Context data.
+Tools and application code can use this information to identify the
+relationship of Events to aspects of the system or to other Events.
 
 #### Message
 Events are transported from a source to a destination via messages.
--- a/spec.md
+++ b/spec.md
@@ -78,10 +78,11 @@ regarding its semantic. Events are considered to be facts that have no given
 destination. Events are used to notify other systems that something has happened.
 
 #### Context
-A set of consistent metadata attributes included with the event about the
-occurrence that tools and developers can rely upon to better handle the event.
-These attributes describe the event and the structure of its data, include
-information about the originating system, and more.
+A set of metadata attributes included with the event that describe circumstances
+of the occurrence which are not specific to the occurrence. Tools and application
+code may use context information to identify the relationship of events to aspects of
+the system or to other events. Context attributes may describe the event source,
+the originating system, the emitting process, and other circumstantial information.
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -78,10 +78,11 @@ regarding its semantic. Events are considered to be facts that have no given
 destination. Events are used to notify other systems that something has happened.
 
 #### Context
-A set of consistent metadata attributes included with the event about the
-occurrence that tools and developers can rely upon to better handle the event.
-These attributes describe the event and the structure of its data, include
-information about the originating system, and more.
+A set of metadata attributes included with the event that describe circumstances
+of the occurrence which are not specific to the occurrence. Tools and application
+code may use context information to identify the relationship of events to aspects of
```
duglin: s/may/can/

spec.md
```diff
@@ -78,10 +78,11 @@ regarding its semantic. Events are considered to be facts that have no given
...
+the system or to other events. Context attributes may describe the event source,
```
duglin: s/may/can/
I think.  I don't think this is meant to be normative, but if I'm wrong then we can make it a "MAY" instead.

duglin:
LGTM once RFC/travis issues are resolved.

duglin:
LGTM

maplebed:
I'm not sure I understand how context can both be "not specific to the occurrence" at the same time that it can be used "...to identify the relationship of events ... to other events".

Could you say more about what you're going for here?

clemensv:
The context is everything not specific to the event. That includes anything that's intentionally common across two or more events for the specific purpose of detecting them as being related.

markito:
I like how this talks about the context helping you co-relate events based on the circumstances but what about execution environment properties ?  Are they already implied there ?

Things like function version, resource constraints (time/memory) and identity would also be part of the context I'm assuming.

deissnerk:
LGTM
@markito The question came up today in the kubeless technical call, too. The attributes you mention are part of the function context. As an event may be associated to multiple functions, the "event context" would then look different in each function.

spec.md
```diff
@@ -78,10 +78,11 @@ regarding its semantic. Events are considered to be facts that have no given
 destination. Events are used to notify other systems that something has happened.
 
 #### Context
-A set of consistent metadata attributes included with the event about the
-occurrence that tools and developers can rely upon to better handle the event.
-These attributes describe the event and the structure of its data, include
-information about the originating system, and more.
+A set of metadata attributes included with the event that describe circumstances
+of the occurrence which are not specific to the occurrence. Tools and application
```
ultrasaurus: The ID is definitely specific to the occurrence at least on some systems... I find this confusing.

deissnerk: The ID is specific, but the circumstance of having an event ID is not.

maplebed: I'm still not getting it either. What does "the circumstance of having an ID" mean? Just that the concept of an `event-id` field is not specific? But having an `event-id` field contain a value that is unique to this event is somehow not specific to the occurrence? My understanding of the `event-id` field is that it *does* uniquely identify this occurrence.

cathyhongzhang: The original description seems to be more clear to me.  in addition to "... tools and developers can rely upon to better handle the event" , we may want to say "... tools, developers, serverless platforms can reply upon to better handle the event"

duglin: @clemensv ^^^

clemensv: I condensed "tools and developers can rely upon to better handle" to "code can use"

clemensv: The event ID is context data. It's externally attached to the occurrence data for the purpose of embedding the occurrence data in an event. The event has uniqueness requirements for the ID. The best the occurrence data might give you is a timestamp of its local clock. 

It'd be great if anyone has a better term than "specific" to express the following two concepts:
* not part of the core data (occurrence information)
* describes the circumstances of the event being emitted (system scope, event type, etc)

duglin: I wonder if this slight rewording would address some of the concerns:
- - -
As described in the **Event** definition, an Event contains two parts, the data representing the Occurrence and additional metadata that provides other circumstantial information about the Occurrence (e.g. information about the originating system). This additional metadata is referred to as **Context** data. Tools and application code can use this information to identify the relationship of Events to aspects of the system or to other Events.
- - -

duglin: Since @clemensv is out, @djrosanova what do you think of the wording I've proposed?

@maplebed @deissnerk @cathyhongzhang @ultrasaurus could you guys live with it?

cathyhongzhang: Doug, LGTM. Thanks.

clemensv:
I snapped to the definition that Doug provided above.

duglin:
LGTM

duglin:
Approved on 4/26 call

Issue #94, opened 2018-02-21
Disambiguation/clarification of source/source-id/source-type


clemensv:
The "source" property is, irrespective of being interpreted as a URI or URL, already an identifier for the source itself. The "source-id" therefore appears superfluous. 

However, the present schema does not provide any further context qualification for the event relative to the source, so I am proposing to repurpose the source-id to provide that context qualification. In common messaging APIs and protocols (including the emails stack), the qualification property is called "subject".

Furthermore, with "source" already being a structured identifier (URI) and the event being sufficiently qualified by namespace/event-type and the subject, I question the practical value of the "source-type" property and propose for it to be removed. 

Assigned to: @clemensv

ultrasaurus:
I don't really understand the problem you are trying to solve.  I'm having trouble following this series of recommendations, since I'm hearing a critique without a clear statement of how the set of attributes are used together and for what use cases.

(This was an issue with the spec before @clemensv arrived, and I really appreciate the focus on detail -- I just want to make sure everyone is aligned on the big picture)

yaronha:
@clemensv @ultrasaurus i think we only need a single source field which calls out the entity or a proxy which pushed the cloud event, additional source info can be found in the schema. e.g. if there is an S3/Blob update event the source need to be the s3 service, and the source S3 object which pushed the event needs to be in the message body along with other "sources" (users, location, ..). we must keep CloudEvents simple, metadata only describes the "envelop" not the Message itself. dont want us to become yet another Soap/XML/..

ultrasaurus:
please see note in issue #82 -- I have the same concern with all of the separate `source` issues

duglin:
@clemensv do you think we can close this one now?

duglin:
tagging as "GoingToClose" - so speak before EOD tomorrow if you think we should keep this open

duglin:
Closing per WG agreement

PR #95, opened 2018-02-21
Renaming/broadening "source" to "topic", consolidating source-id/source-type into "subject"

clemensv:
See issue https://github.com/cloudevents/spec/issues/94

Restated "source-id" as "subject", which now more clearly serving to qualify the event in relation to the source, which is missing as an explicit concept. I also dropped the "source-type" in this proposal as we already have namespace/event-type as event qualifier and namespace/source/subject as context qualifiers.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -161,9 +161,11 @@ that contains both context and data).
   system might identify the CRM system as the "source", might further qualify the
   event as 'new-customer-added' in the "event-type" relative to its "namespace",
   and then further qualify the subject of the event (the new record) with the content
-  of this field. The subject is a free-form string defined by the publisher.
+  of this property. The subject is a free-form string defined by the publisher. The
+  property is optional, because the subject might already be self-evident from 
+  the "source" context.
 * Constraints:
-  * OPTIONAL. The subject may be self-evident from the "source" context.
+  * OPTIONAL 
 * Examples:
   * File system event, a newly created file being the subject, with a partial URI 
     relative to the source URI: /data/app/alice.txt
--- a/spec.md
+++ b/spec.md
@@ -161,11 +161,9 @@ that contains both context and data).
   system might identify the CRM system as the "source", might further qualify the
   event as 'new-customer-added' in the "event-type" relative to its "namespace",
   and then further qualify the subject of the event (the new record) with the content
-  of this property. The subject is a free-form string defined by the publisher. The
-  property is optional, because the subject might already be self-evident from 
-  the "source" context.
+  of this property. The subject is a free-form string defined by the publisher. 
 * Constraints:
-  * OPTIONAL 
+  * OPTIONAL. The subject might already be self-evident from the "source" context. 
 * Examples:
   * File system event, a newly created file being the subject, with a partial URI 
     relative to the source URI: /data/app/alice.txt
--- a/spec.md
+++ b/spec.md
@@ -154,24 +154,21 @@ that contains both context and data).
   * REQUIRED
   * MUST contain at least one non-empty sub-property.
 
-### source-type
+### subject
 * Type: String
-* Description: Type of the event source. Providers define list of event
-  sources.
+* Description: Identifies the subject of the event in relation to the source.
+  For example, a notification about a new customer record being added to a CRM
+  system might identify the CRM system as the "source", might further qualify the
+  event as 'new-customer-added' in the "event-type" relative to its "namespace",
+  and then further qualify the subject of the event (the new record) with the content
+  of this field. The subject is a free-form string defined by the publisher.
 * Constraints:
-  * REQUIRED
-  * MUST be a non-empty string
+  * OPTIONAL. The subject may be self-evident from the "source" context.
 * Examples:
-  * s3
-
-### source-id
-* Type: String
-* Description: ID of the event source.
-* Constraints:
-  * REQUIRED
-  * MUST be a non-empty string
-* Examples:
-  * my.s3.bucket
+  * File system event, a newly created file being the subject, with a partial URI 
+    relative to the source URI: /data/app/alice.txt
+  * Customer relationship management (CRM) solution event, new customer record being 
+    the subject, with fully qualified URI: https://example.com/crm/c/2662773
 
 ### event-id
 * Type: String
--- a/spec.md
+++ b/spec.md
@@ -146,29 +146,48 @@ that contains both context and data).
   * REQUIRED
   * MUST be a non-empty string
 
-### source
-* Type: Object
-* Description: This describes the software instance that emits the event at
-  runtime (i.e. the producer). It contains sub-properties (listed below)
+### topic
+* Type: String
+* Description: This expression is a classification of the event that 
+  enables groupwise publishing of and subscription to events of the 
+  described class. The classification might reflect the identity of 
+  the party raising the event, and/or the context of where and/or why 
+  the event was raised, and/or a structured classification of the 
+  information contained in the event. The value SHOULD be a URI. 
+  The value MAY be used to map directly to the "topic" concept found
+  in many middleware products. The value MAY also be a interpreted as 
+  a URL. The topic property value's scope is set by the "namespace" 
+  property. 
 * Constraints:
   * REQUIRED
-  * MUST contain at least one non-empty sub-property.
+* Examples:
+  * Resource path, relative URI reference: 
+    /tenant/group/type/myresource
+  * Machine component, relative URI reference:
+    /robot/drives/3/temperature
+  * US City, relative URI reference:
+    /USA/Alaska/Juneau
+  * Virtual machine alerting, URI with custom scheme: 
+    alerts://mymachine.example.com/
+  * Web service notification triggers, URI/URL network resolvable: 
+    https://myservice.example.com, https://myservice.example.com/mytrigger   
 
 ### subject
 * Type: String
-* Description: Identifies the subject of the event in relation to the source.
-  For example, a notification about a new customer record being added to a CRM
-  system might identify the CRM system as the "source", might further qualify the
-  event as 'new-customer-added' in the "event-type" relative to its "namespace",
-  and then further qualify the subject of the event (the new record) with the content
-  of this property. The subject is a free-form string defined by the publisher. 
+* Description: Identifies the subject of the event in relation to the topic.
+  For example, a notification about a new customer record being added to 
+  a CRM system might identify a CRM system's tenant as the ´topic´, might 
+  further qualify the ´event-type´ as 'new-customer-added', and then further 
+  qualify the subject of the event (the new record) with the content of this 
+  property. The subject is a free-form string defined by the publisher. 
 * Constraints:
-  * OPTIONAL. The subject might already be self-evident from the "source" context. 
+  OPTIONAL. The subject might already be self-evident from the "source" context. 
 * Examples:
-  * File system event, a newly created file being the subject, with a partial URI 
-    relative to the source URI: /data/app/alice.txt
-  * Customer relationship management (CRM) solution event, new customer record being 
-    the subject, with fully qualified URI: https://example.com/crm/c/2662773
+  * File system event, with `topic` reflecting the emitting storage container,
+    and the newly created file-name being the subject, a partial URI 
+    relative to the source URI: `/data/app/alice.txt`.
+  * Customer relationship management (CRM) solution event, new customer record 
+    being the subject, with fully qualified URI: https://example.com/crm/c/2662773
 
 ### event-id
 * Type: String


spec.md
```diff
@@ -154,24 +154,21 @@ that contains both context and data).
...
-* Examples:
-  * s3
-
-### source-id
-* Type: String
-* Description: ID of the event source.
-* Constraints:
-  * REQUIRED
-  * MUST be a non-empty string
+  * OPTIONAL. The subject may be self-evident from the "source" context.
```
duglin: I'd prefer to put this explanatory text in the Description field.
But either way:  s/may/might/

clemensv: I'll move it.

spec.md
```diff
@@ -161,9 +161,11 @@ that contains both context and data).
   system might identify the CRM system as the "source", might further qualify the
   event as 'new-customer-added' in the "event-type" relative to its "namespace",
   and then further qualify the subject of the event (the new record) with the content
-  of this field. The subject is a free-form string defined by the publisher.
+  of this property. The subject is a free-form string defined by the publisher. The
+  property is optional, because the subject might already be self-evident from
```
duglin: Perhaps: The property might not be present, because the subject...
?

clemensv: It's a bit weird to tap dance around OPTIONAL here. Shall I uppercase or move it back?

duglin: I think the bigger concern, at least to me, is saying its OPTIONAL twice - once here and once below. Duplicating normative statements isn't good, its opening the possibility of someone updating one of them but not the other - then we're inconsistent.

clemensv: If we have a constraints section, we should also allow rationalization of the constraint in that section, IMO. I'm moving it back.

duglin: I don't necessary object to that - I'm just trying to avoid two sections that give lots of explanatory text and people not knowing what text goes where - so they end up duplicating thoughts.

duglin: Actually, on a side note, this is partially why I tend to prefer to not have a "constraints" section because it forces this kind of split. I'd prefer just a set of sentences with the RFC keywords in the appropriate spots and they can live next to the explanatory text.  But I didn't want to drastically change the original format of the spec when we first pulled it in.  But if people are open to the change I can take a pass at a PR.

duglin: ping @ac360 - any thoughts on this?

deissnerk: Providing descriptions without mentioning the constraints will in many cases be hard. @duglin I agree with you that a split of the text is hard to achieve. On the other hand I see some value in a restrictions section that only lists the restrictions without providing additional text. An implementor of the spec who just wants to quickly check the restrictions of an attribute, might benefit from that. Using the same normative statements in the description would be even better than paraphrasing them, because it would help maintaining consistency between description and constaints.

maplebed:
I'm sorry, I don't really understand what's going on here.  It looks like this diff
* removes `source-type`
* removes `source-id`
* adds `subject`
Could those be separate diffs, or at least separate the removal of `source-type` and `source-id` from the addition of `subject`?

Lastly, I don't understand what sort of data would go in this field, or what the it means to be called `subject`. Is this trying to evoke images of the Subject line in an email? 

I'm having trouble forming any opinion about this PR given my inability to figure out what's going on.

clemensv:
I rename source-id.

Here's the scenario:

The emitter of the event, the "source" is some object that you subscribe on. Let's say that's a storage service container/bucket. You are interested when a new object is created in that container/bucket.

```
namespace: azure.com
source: /subscriptions/{id}/providers/Microsoft.Storage/storageAccounts/myStorageAccount
eventType: Microsoft.Storage.BlobCreated
```
To do any generic pub/sub filtering on that event with pattern matching or startsWith/endsWith in generic infra that doesn't want to reason about the details of the event, you now still need a hint here what that event is actually about, i.e. which blob was created here. Thus:

```
subject: /blobServices/default/containers/testcontainer/blobs/testfile.txt 
```

duglin:
@clemensv your usecase makes me wonder whether the source (or "entity that generated the event") should be the bucket or the newly created resource? Based on how I look at it I can see an argument for either one. My concern is that we now have, sort of, two "sources": 1) the entity to which you subscribed to for the event stream, and 2) the entity that did something that caused the event, which may or may not be the same as no. 1 - but I suspect in some (many?) cases will be. Which means we then have to have a very clear set of rules so people know when to use one vs the other since they might appear to overlap.

Not trying to cram too much into the 'source', but couldn't this information be included within the "source" value? Perhaps as the tail part of the URL path or query parameter?  You can still do your regex filtering that way I believe.

clemensv:
You will always have a a notion of a thing that emits events and matters that are going on inside of that thing that the events are about. Same as with email. Or github issues. I am the source, identified by my SMTP URI, and the subject line is what I want to tell you today. 

Since we say the "source" SHOULD be a URI, but we're not mandating it (and I do think there are good reasons not to), having a separate descriptor for the subject makes sense, and there's plenty of prior art to look at in messaging for that split.

clemensv:
I'm amending this proposal by renaming "source" to "topic" and proving a broader definition that allows for both a concrete concept of originating source and also a more abstract notion of classification that isn't bound to a concrete origin as a primary consideration. Origin/provenance considerations can be enormously complex in some verticals and need to be resolved by processing of the event payload. 

The rationale for this proposal of consolidating the route/dispatch metadata to `topic`, `subject`, and `eventType`, including links to prior art and current practice across a range of platforms and infrastructures is laid out in detail in this issue: https://github.com/cloudevents/spec/issues/112

spec.md
```diff
@@ -146,32 +146,48 @@ that contains both context and data).
...
   * REQUIRED
-  * MUST be a non-empty string
 * Examples:
-  * s3
-
-### source-id
+  * Resource path, relative URI reference: 
+    /tenant/group/type/myresource
+  * Machine component, relative URI reference:
+    /robot/drives/3/temperature
```
shalka: Request: add clarification for URI segments, like /namespaces/{namespaceId}/buckets/{bucketId}/objects or /robots/{robotId}/drives/{driveId}/temperature

markfisher:
The following was motivated specifically by some of the discussion on the WG call today...

To what extent is this spec prescribing behavior for the middleware? It was my understanding that the primary goal was to describe a common format for events without making assumptions about the way those attributes would be interpreted, since event brokers don't all behave the same way, even when they use common terminology.

For example, while there's common usage of "topics" across brokers, not all brokers support hierarchical and wildcard subscriptions. Some only support subscribing at the granularity of the literal topic name. In the latter case, it can be a common pattern to create derived streams from a more general upstream topic.

The limitations of naming topics on a specific broker won't always be flexible enough to support the range of syntax used in these examples, such as URI (Kafka for example). So it could lead to confusion between the logical and "physical" topic. Also, it's not typically a requirement to include the topic name in the event itself. A consumer subscribes to the topic on the broker, often that information is provided via declarative configuration within the context of the consuming application. Consumers might subscribe to multiple topics, and producers might even send the same event to multiple topics.

deissnerk:
@markfisher I agree that the term _topic_ is misleading. In my opinion it belongs to the domain of messaging and not to events. I was actually happy with the term _source_. Unfortunately I wasn't able to attend the calls this week. So I don't know what concerns people have regarding _source_. What about _origin_?
With the terms topic and subject I don't see any attribute left that is specific for an event instead of a message. Probably this is what Vaughn Vernon meant when he addressed the group on the mailing list a few weeks ago. Maybe it is a good preparation for tomorrow's call to think of reasons why the spec is called CloudEvents and not CloudMessages.

ultrasaurus:
From my perspective, the whole purpose of this spec is to separate source/producer and destination/action/consumer from transport/middleware.  From my perspective, this proposal completely changes what we're doing here.

notque:
Source is clear to me, Topic is not clear, and perhaps even misleading. Not sure I understand the thought process.

clemensv:
@notque @ultrasaurus @deissnerk @markfisher 

"Topic" is a concept that is broader than and a full superset of "source". It often reflects the originating context, but also works in cases where any party other than the originator needs to publish an event on behalf of the same context. We even have a fresh example of that in the discussion of the [usage scenarios](https://github.com/cloudevents/spec/pull/117#discussion_r176705151)

The abstract topic concept also happens to map straight to numerous existing middleware platforms and also maps right onto RESTful resource graphs, i.e. onto HTTP.  

From the mailing list, I'm also putting the following on the record:

[Topics and Subjects Presentation Doc](https://1drv.ms/p/s!AgcBsXoqzTwSrc4vf0at9rLY4JHErQ)
[Topics and Subjects Presentation Screencast](https://1drv.ms/v/s!AgcBsXoqzTwSrc4xtiQ25VGGln6zWw)

Add supporting evidence for how broadly the topic concept is supported across the industry and with existing standards and platforms that many users of this specification will most certainly want to leverage, consider this incomplete list of examples:

HTTP WebHook specs:
* [IETF RFC8030 “Generic Event Delivery Using HTTP Push”] (https://tools.ietf.org/html/rfc8030#section-5.4) 
* [Shopify WebHook Topics](https://help.shopify.com/api/reference/webhook)
* [Facebook Graph API WebHook Topics](https://developers.facebook.com/docs/workplace/integrations/custom-integrations/webhooks)
* [Twitch.TV WebHook Topics](https://dev.twitch.tv/docs/api/webhooks-reference)

Messaging Standards:
* [MQTT v5](http://docs.oasis-open.org/mqtt/mqtt/v5.0/cs01/mqtt-v5.0-cs01.html#_Toc502667330)

IoT Platforms:
* [AWS IoT Topics](https://docs.aws.amazon.com/iot/latest/developerguide/topics.html)
* [IBM Watson IoT Topics](https://console.bluemix.net/docs/services/IoT/applications/mqtt.html#mqtt)
* [Azure IoT Hub Topics](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support)

Cloud Event Routers (Push)
* [AWS SNS Topics](https://docs.aws.amazon.com/sns/latest/dg/welcome.html)
* [Azure Event Grid Topics](https://docs.microsoft.com/en-us/azure/event-grid/concepts)

Cloud Event Ingestion Platforms
* [Apache Kafka Topics, IBM Message Hub, Oracle Event Hub](https://kafka.apache.org/documentation/#gettingStarted)
* [Azure Event Hubs; "hub"](https://docs.microsoft.com/en-us/azure/event-hubs/event-hubs-features)

Cloud and On-Prem PubSub brokers:
[IBM MQ Topics](https://www.ibm.com/support/knowledgecenter/en/SSFKSJ_9.0.0/com.ibm.mq.pro.doc/q005050_.htm)
[IBM MQ Topic Streams](https://www.ibm.com/support/knowledgecenter/en/SSFKSJ_9.0.0/com.ibm.mq.pro.doc/q005080_.htm)
[RabbitMQ Exchanges (and Topics)](https://www.rabbitmq.com/tutorials/tutorial-five-python.html)    
[Apache ActiveMQ Address](https://activemq.apache.org/artemis/docs/2.0.0/address-model.html)
[Solace Topics](https://docs.solace.com/Features/Topic-Support-and-Syntax.htm)
[Azure Service Bus Topics](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-java-how-to-use-topics-subscriptions)
[Google Cloud PubSub](https://cloud.google.com/pubsub/docs/admin)
[Google FCM](https://firebase.google.com/docs/cloud-messaging/js/topic-messaging)

It's impossible to argue that the "topic" concept and terminology choice isn't grounded in an existing industry consensus. For the few deviations from the word "Topic" in the list above, there are actually good reasons. RabbitMQ uses "exchange" and ActiveMQ uses "address" for their top-level concept to avoid a name clash with the JMS topic that they also both implement on top of their base construct. Event Hubs doesn't use "Topic" because we have the same name already taken by Service Bus as a top level concept. "Streams" in MQ disambiguates from native topics.

markfisher:
> It's impossible to argue that the "topic" concept and terminology choice isn't grounded in an existing industry consensus

I don't think anyone is arguing that. In fact, my comments were motivated by the fact that topics _are_ ubiquitous but unable to map one-to-one in a consistent way to either the syntax or semantics that you are suggesting. So it would inevitably lead to situations where the "topic" header value does not match the actual topic name where the event has been published. It's not common to include the topic name in messages in the first place. I'd even consider it an anti-pattern, since it's possible to have a many-to-many relationship between events and topics.

clemensv:
@ultrasaurus - regarding [this comment](https://github.com/cloudevents/spec/pull/95#issuecomment-375330745)

We have a [shared definition of the goals](https://github.com/cloudevents/spec/blob/master/spec.md#design-goals).

In those goals, we specifically call out that "... producer and consumer can be developed and deployed independently. A producer can generate events before a consumer is listening, and a consumer can express an interest in an event or class of events that is not yet being produced."

That level of decoupling is typically achieved by introducing "middleware" intermediaries between producer and consumer. The minimal intermediary typically used for this is the dispatch logic of an HTTP request framework that maps paths to method invocations.

NodeJS "Express", ASP.NET, and Apache Struts all fit the middleware definition of the usage scenario PR. 

Their "topic" is the [request target](https://tools.ietf.org/html/rfc7230#section-5.3) as defined in HTTP; the request URI is generally a reference to a graph, most simply into the file system.

The abstract nature of "topic" indeed strongly echoes that of [a resource in REST](https://www.ics.uci.edu/~fielding/pubs/dissertation/rest_arch_style.htm#sec_5_2_1_1), and middleware is a more generalized notion than caches, proxies and gateways that Fielding enumerates in his seminal dissertation.


Please explain how my proposal concretely contradicts the existing shared definition of goals.

clemensv:
Closing in support of #123

Issue #82, opened 2018-02-15
"source" name/explanation


clemensv:
"This describes the software instance that emits the event at runtime (i.e. the producer). It contains sub-properties (listed below)"

The description using a "i.e." seems indicative of either the property name being too ambiguous to stand, or the description needing work. 

"Software instance" is also often an incorrect term to describe a resource/source/producer, because a resource/source/producer is often a logical concept that may run multiple instances and even be made up of multiple separate implementations that all yield the same events from the same virtual context. 

Assigned to: @clemensv

ultrasaurus:
I've been working on [cloud events concepts](https://docs.google.com/presentation/d/1xippFyX9fXqlTWUh3waZcanVMHF12z0AbZ8EK3fk3ys/edit#slide=id.g3244de5a2d_0_0).  After last weeks discussion of what we mean by `source` I've attempted to illustrate how the different attributes might be used (currently slide 11, including image below for easy reference):

![image](https://user-images.githubusercontent.com/41906/36771466-7b7d024a-1c05-11e8-9f96-0c054f0e0c71.png)

yaronha:
@ultrasaurus @clemensv @duglin  there seem to be an overlap (same prefix) between source-type & event-type, and between namespace & source-id, IMO Cloud events "envelop" doesn't need the level of details in source-id since this is event class specific and belongs in the event body/data.  

I'm still not sure what the "namespace" stands for (whats the diff between acme.com & org.samba), and how all of this relates the the Schema field which may overlap with the event-type.

We need to keep it simple have one field for source RN and one field for Schema, all the rest of the details need to go into the event body, the source in the "envelop" needs to serve for auditing/logging and routing purposes, not for understanding the content of the event.

duglin:
@clemensv given the current text in the spec for `source` is this issue resolved?

duglin:
@clemensv can we close this one?

duglin:
Agreed to close on the 5/24 call

PR #96, opened 2018-02-21
Proposed clarification of "source" #82

clemensv:
Last week's discussion and issue https://github.com/cloudevents/spec/issues/82 
Supporting: https://github.com/cloudevents/spec/issues/94 

This is a proposed clarification of "source", also as a follow-up to last week's debate about the structure of the "source" property (with Matt and myself advocating for a URI). 

As discussed last week, I will support this tomorrow with a brief scenario description of event routing from within industrial automation and other contexts with segmented networking, where considering the "source" value as being a network-resolvable address from the event consumer perspective is impractical, and using hierarchical identifiers has yet proven sufficient.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -149,18 +149,22 @@ that contains both context and data).
 ### source
 * Type: String
 * Description: This property identifies the source of the event as defined
-  by the publishing organization. The value SHOULD be a URI. The source property
-  value MUST uniquely identify the source within the scope set by the "namespace"
-  property. While the value MAY be a interpreted as a URL, the purpose of
-  it to logically identify the source and not to establish the location of the 
-  event source on the network.
+  by the publishing organization. The value SHOULD be a URI. The source 
+  property value MUST uniquely identify the source within the scope set by the
+  "namespace" property. While the value MAY be a interpreted as a URL, the 
+  purpose of it to logically identify the source and not to establish the 
+  location of the event source on the network.
 * Constraints:
   * REQUIRED
 * Examples:
-  * Resource path, relative URI reference: /tenant/group/type/myresource
-  * Machine component, relative URI reference: /robot/drives/3/temperature
-  * Virtual machine alerting, URI with custom scheme: alerts://mymachine.example.com/
-  * Web service notification triggers, URI/URL network resolvable: https://myservice.example.com
+  * Resource path, relative URI reference: 
+    /tenant/group/type/myresource
+  * Machine component, relative URI reference:
+    /robot/drives/3/temperature
+  * Virtual machine alerting, URI with custom scheme: 
+    alerts://mymachine.example.com/
+  * Web service notification triggers, URI/URL network resolvable: 
+    https://myservice.example.com
 
 ### source-type
 * Type: String
--- a/spec.md
+++ b/spec.md
@@ -147,12 +147,20 @@ that contains both context and data).
   * MUST be a non-empty string
 
 ### source
-* Type: Object
-* Description: This describes the software instance that emits the event at
-  runtime (i.e. the producer). It contains sub-properties (listed below)
+* Type: String
+* Description: This property identifies the source of the event as defined
+  by the publishing organization. The value SHOULD be a URI. The source property
+  value MUST uniquely identify the source within the scope set by the "namespace"
+  property. While the value MAY be a interpreted as a URL, the purpose of
+  it to logically identify the source and not to establish the location of the 
+  event source on the network.
 * Constraints:
   * REQUIRED
-  * MUST contain at least one non-empty sub-property.
+* Examples:
+  * Resource path, relative URI reference: /tenant/group/type/myresource
+  * Machine component, relative URI reference: /robot/drives/3/temperature
+  * Virtual machine alerting, URI with custom scheme: alerts://mymachine.example.com/
+  * Web service notification triggers, URI/URL network resolvable: https://myservice.example.com
 
 ### source-type
 * Type: String


duglin:
Can you wrap things at 80 cols?
Aside from that - LGTM

spec.md
```diff
@@ -147,12 +147,20 @@ that contains both context and data).
...
-* Type: Object
-* Description: This describes the software instance that emits the event at
-  runtime (i.e. the producer). It contains sub-properties (listed below)
+* Type: String
+* Description: This property identifies the source of the event as defined
+  by the publishing organization. The value SHOULD be a URI. The source property
+  value MUST uniquely identify the source within the scope set by the "namespace"
+  property. While the value MAY be a interpreted as a URL, the purpose of
+  it to logically identify the source and not to establish the location of the 
+  event source on the network.
```
ultrasaurus: This certainly clarifies the description more; however, it raises as many questions as it answers.  How do we expect that the source will be used?  We want to consider use cases for tools, providers, and enabling an action to call back to the source, when appropriate.

Given these open questions, it is hard to evaluate whether the option suggested will be effective.

spec.md
```diff
@@ -147,12 +147,20 @@ that contains both context and data).
   * MUST be a non-empty string
 
 ### source
-* Type: Object
-* Description: This describes the software instance that emits the event at
-  runtime (i.e. the producer). It contains sub-properties (listed below)
+* Type: String
+* Description: This property identifies the source of the event as defined
+  by the publishing organization. The value SHOULD be a URI. The source property
```
inlined: Who is "publishing organization"? Is this the `namespace` field? If so, we should reuse terms. If not, it sounds like the definition is circular: the publisher (identified by `source`) determines the format and meaning of `source`.

A litmus test I frequently use is an on-prem install of open source software. Which field is apache.org and which field is the domain/IP address where the software is hosted?

spec.md
```diff
@@ -147,12 +147,20 @@ that contains both context and data).
...
+  value MUST uniquely identify the source within the scope set by the "namespace"
+  property. While the value MAY be a interpreted as a URL, the purpose of
+  it to logically identify the source and not to establish the location of the 
+  event source on the network.
 * Constraints:
   * REQUIRED
-  * MUST contain at least one non-empty sub-property.
+* Examples:
```
inlined: Let me know if I'm getting ahead of myself: I'm curious in practical use for the OSS case whether this will be easy to provide. Oftentimes the service will have a private IP but will want to share a load-balanced name. Do we expect this is just something that software vendors will need to expose as a config file when their software is an event source?

clemensv: Source is URI not a URL. It's an identifier. Many legitimate event sources can't be identified in public network endpoint terms, at all. URI is a universal and rather successful structured identifier concept. What is "the OSS use-case"?

spec.md
clemensv: "Publishing organization" is whoever writes the event. It's the namespace owner and we should add that to the definitions. "namespace" per-se is a qualifier that represents the publishing organization's chosen scope.

spec.md
clemensv: The source identifies who emits the event, scoped by the namespace. The source can be any string that makes sense for the publishing organization and their subscribers (in the extreme it could hold a JSON object), but it's suggested that it SHOULD be a URI. I will lay out in the call that the source property information by itself cannot be sufficient to determine a callback path to the source in non-trivial scenarios. The statement "MAY be a interpreted as a URL" explicitly enables interpretation of the identifier as a callback location if an implementation wants that.

spec.md
```diff
@@ -147,12 +147,24 @@ that contains both context and data).
...
 ### source
-* Type: Object
-* Description: This describes the software instance that emits the event at
-  runtime (i.e. the producer). It contains sub-properties (listed below)
+* Type: String
+* Description: This property identifies the source of the event as defined
+  by the publishing organization. The value SHOULD be a URI. The source 
+  property value MUST uniquely identify the source within the scope set by the
+  "namespace" property. While the value MAY be a interpreted as a URL, the 
+  purpose of it to logically identify the source and not to establish the
```
markito: "**is** to logically..." ?

duglin: @clemensv ^^

yaronha:
@clemensv @duglin IMO the source should be the reporting entity or one that proxies for it, it should contain the namespace, vendor, etc. Why do we need 4 fields to identify the source, just like in email or URL we have a single string that specify the domain and entity. any other "source" (type, ..) can be delegated to the message body it self and is not required for the event transportation.

duglin:
@yaronha I tend to agree and I was hoping that these discussions would lead us in that direction.

Right now we have "source", "source-type" and "source-id". It seems to me that for any given source the definition of these values might vary quite a bit - which means in order for the receiver of an event to really understand and make use of them they would have to share the source's understanding of those fields. Therefore, I don't see why (if they have that knowledge) they couldn't extract all 3 values from a single value (string, URL, URI... whatever). And then we can just let the source decide what data to include for identification purposes, and how to encode it into that value.

If the receiver doesn't understand all of that, then they probably are going to treat these fields as opaque strings and blindly process them w/o any analysis (perhaps just stick them in an audit log). In which case, why force them to pass around multiple values when a single value will do?

If, in the end, there is consensus that we do want a set of well-defined source fields, then I claim we can do that by mandating the encoding of those fields into a single value so if a receiver does want to split them up they know exactly how to break it up. But as an incremental step we could decide to just have a single value (defined by the source) and then open an issue to decide if we want to mandate an encoding scheme for a set of fields to be included in that value.

spec.md
ultrasaurus: I don't think it makes any sense to modify source by itself.  As documented, it is an "object" that has properties `type` and `id` (listed in the spec as `source-type` and `source-id`).  I agree that, as written, the spec is confusing, but I believe we need to address how to specify the source holistically.

ultrasaurus:
FYI -- added more context in issue #82

spec.md
clemensv: I've read the table you put into the associated issue, and I can't say that it makes sense to me. The application receiving an event needs to know who sent it. A URI appears to be sufficient for that purpose. A composite identifier that has an explicit notion of an "id" and "type" is a complication.

yaronha: I agree, i dont see why we need a hierarchy (namespace, id, type), a single string value should do for source (identifying the event reporting entity/resource, not the modified S3 object names..). event specific source data should be written into the event structure itself.

clemensv:
Closing this PR in favor of expanding PR #95

PR #98, opened 2018-02-21
clarify namespace to explicitly declare the convention used in the examples

maplebed:
The examples of values for `namespace` both use the convention of a domain name in reverse order to identify an organization. This convention is the same as to https://tools.ietf.org/html/rfc7595 when suggesting how to namespace URI schemes.  Borrowing language from that RFC to clarify how to declare `namespace` in this spec.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -110,7 +110,10 @@ that contains both context and data).
 ### namespace
 * Type: String
 * Description: Identifier that uniquely identifies the organization publishing
-  the event.
+  the event. It is important to prevent collisions and to make it possible to
+  identify the owner of a namespace. To accomplish these two goals,
+  organizations SHOULD use a prefix based on their domain name, expressed in
+  reverse order.
 * Constraints:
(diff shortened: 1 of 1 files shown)


duglin:
Just curious, why reverse?  Do we need interop around the shape of the string?

If we do keep the SHOULD in there, I would recommend removing the example that doesn't adhere to it.

clemensv:
It's sufficient for the namespace to be a unique identifier, IMO. For it to be a URI would be helpful. Reverse order suggests some expectation around being able to interpret that string in some particular way. For the "namespace" identifier to merely be different from all other namespace identifiers makes it serve its purpose.

maplebed:
whoops... I don't know how I read both examples as being reversed - clearly one is forward and the other reverse.  Going to close this PR. If we're not expecting that `namespace` is always going to be a domain name, it would be lovely to have some example in there that's not.

Issue #73, opened 2018-02-08
design goals for CloudEvents V.1


ultrasaurus:
Roadmap says: Agree on design goals for CloudEvents V.1. for January

I don't think we've actually accomplished this.  

@ac360 got together last week and brainstormed what might be an interim goal that would be useful in aligning a core subset of meta data:

* draft specification that WG members agreed *could* work
* at least 3 sample events that conform to the specification

Assigned to: @ultrasaurus

duglin:
@ultrasaurus does #88 address this or is an additional PR needed?

ultrasaurus:
The above interim goals are addressed in PR#88, but we still lack design goals, see #101

ultrasaurus:
yes https://github.com/cloudevents/spec/pull/101 resolves this issue. 

I've opened https://github.com/cloudevents/spec/issues/121 for the separate roadmap about the scope of the spec.

PR #101, opened 2018-02-23
Draft design goals for CloudEvents specification

ultrasaurus:
addresses #73


about/design-goals.md
```diff
@@ -0,0 +1,29 @@
+# CloudEvents Specification Design Goals
+
+CloudEvents are typically used in a distributed system to allow for services to be loosely coupled during development, then then configured with tight-coupling for specific applications.
```
markpeek: Typo "then then"?

about/design-goals.md
```diff
@@ -0,0 +1,29 @@
...
+
+The goal of the CloudEvents specification is to allow for different cloud providers to create interoperable event systems, that allow for third-party tools and services to provide:
```
markpeek: This seems a bit too specific to cloud providers. Perhaps change the emphasis onto event systems rather than cloud providers?
```The goal of the CloudEvents specification is to define interoperability of event systems that allow cloud providers, third-party tools and services to provide:```

File: about/design-goals.md [hunk 3]
```diff
@@ -0,0 +1,29 @@
...
+- new event sources
+- new event consumers
+- observability, such as monitoring, debugging and visualization,
+- developer tools and libraries
+
+Key properties of the system:
```
markpeek: My first read of this made me think we needed to define these terms. I then read further and saw you mention the WG needs to define this vocabulary. :-) Not sure how or whether to rectify.

about/design-goals.md
```diff
@@ -0,0 +1,29 @@
...
+- Source & action may be developed and deployed independently
+- Trigger configuration allows for actions to be invoked based on a condition (e.g. for a specific event type generated by a source, optionally filtered by specific attribute values, such as a Function which is invoked when an image is uploaded to specific S3 or Google Cloud Storage “bucket” path
+- Events are uniquely identified so that developers can write code that is idempotent
```
markpeek: Perhaps it is just me but I'm not sure I understand the distinction being made by this statement. Is this about event uniqueness from a namespace or source? Or having a common definition to allow the consumer to rely on a specific format?

duglin:
Can you wrap at 80 cold?

about/design-goals.md
duglin: perhaps "rather than"  ?

about/design-goals.md
duglin: Maybe its just me, but I suspect many people will never look at this doc - which is a shame because I think its very useful to help set the context for what the spec will talk about. Add to that, Mark's comment about needing to define the terms used in the following points, and my desire not to duplicate info, I'm wondering what people think about changing this PR so that it adds all of this info into the spec itself. This would serve a couple of purposes: 1) it will increase the changes of people reading it, 2) avoid duplication of very important concepts, 3) it'll make the spec more self-sufficient by beefing-up our intro (o where ever) section.

about/design-goals.md
```diff
@@ -0,0 +1,29 @@
...
+- Common meta-data attributes allow for observability via tools that are developed without knowledge of the specific implementation of an event system
```
duglin: I think this would also include: w/o knowledge of the specifics of the event itself - meaning the 'data'/

about/design-goals.md
ultrasaurus: I was attempting to describe how late-binding of trigger + action allows for tight-binding when an application uses events. I believe that this is exactly what allows for observability.  Open to suggestions on how to clarify that!

about/design-goals.md
ultrasaurus: If other folks agree, I could split this PR into two parts:
1. the goals/non-goals => spec
2. WG actions => README with link to Goals section

about/design-goals.md
ultrasaurus: Now that you mention it, both!  I was just thinking about the former -- will update language to include both points

about/design-goals.md
```diff
@@ -0,0 +1,52 @@
...
+uploaded to specific S3 or Google Cloud Storage “bucket” path)
+- Events are uniquely identified to facilitate developer velocity and tooling:
+  - a unique idenfier allows for idempotency for reliable implementation
+  - structured event definitions allow the consumer to rely on a specific format
+- Common metadata attributes allow for observability via tools whithout
+knowledge of the specific implementation of an event system and also can provide
+value without detailed handling of the specifics of the unique event data
+- CloudEvents may be transported from source to action with different protocols.
+
+In order to achieve the above goals and desired system properties, the
```
duglin: It feels like this paragraph should come after line 12 since that's where you introduce our goal of interop, so it seems like listing the things we need to do to achieve that interop right after it makes sense.

about/design-goals.md
```diff
@@ -0,0 +1,52 @@
...
+Serverless WG must agree on a common vocabulary with precise definitions
+describing the key elements of the system including:
+- [Source](../spec.md#source)
+- Action <use definition from starter doc>
+- Trigger: the condition that causes an event to be transmitted from source to
+action
+- Rule: the association of trigger + action
+
+In order to achieve the interoperability goals, the Serverless WG must describe:
```
duglin: I would merge this paragraph with the previous one and just have one list of "goals" rather than split it into two. I would also avoid defining the terms here since we're going to do that later on (if we merge this into the spec). I would just state the goal itself - meaning, something like: Define the common vocabulary of an eventing system

about/design-goals.md
```diff
@@ -0,0 +1,52 @@
...
+interoperability can be achieved independently, then shared configuration and/or
+API definition can be addressed as follow-on work
+- Authorization model(s) that control access and execution permissions
+
+
+## Non-Goals of this spec
+The following non-goals may be follow-on work, yet would not be included in the
+1.0 specification.
+- Function build and invocation process
+- Language-specific runtime APIs
```
clemensv: If we are achieving semantic and wire-level interoperability across systems, we have already achieved a lot. 

An API definition, or an API language binding, or a dispatch model for how to map event data to code would seem to define how a user application interacts with events on either the source side or the subscriber side, and that's even another layer further out into the app from "Trigger" and "Action" as laid out above. 

In the messaging space, we begrudgingly need to carry the "shared API" JMS legacy around with us because of the installed base and are doing mapping work in OASIS AMQP BINDMAP, but you could few of us excited about building another shared API standard. JMS severely constrains access to features, it imposes a particular notion of topology, it imposes a particular notion of send/receive patterns, it imposes a particular notion of filters, and so on. The innovation now all happens on different client stacks, on in some cases, like Apache Qpid Proton or AMQP for Net, those client stacks either work across products with their own API, or are the basis for product SDKs that wrap them. That all exists interoperably without the OASIS AMQP TC having imposed an API shape on those projects.

about/design-goals.md
```diff
@@ -0,0 +1,52 @@
...
+
+In order to achieve the above goals and desired system properties, the
+Serverless WG must agree on a common vocabulary with precise definitions
+describing the key elements of the system including:
+- [Source](../spec.md#source)
+- Action <use definition from starter doc>
+- Trigger: the condition that causes an event to be transmitted from source to
+action
+- Rule: the association of trigger + action
+
```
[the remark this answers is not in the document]
duglin: I have a simpler rule... if we use a term (as a CloudEvent term and not in a traditional English usage) then we need to define it. So, if in the body of the spec we talk about 'actions' or 'triggers' as CloudEvent things then we need to define them. If we don't use those words then we don't need to define them. Then the question becomes whether we want to use those terms or not - and that's up to the PR that tries to add them.  Right now our only use of the words "action" and "trigger" are more traditional English to me and not formal terms.

about/design-goals.md
```diff
@@ -0,0 +1,52 @@
+# CloudEvents Specification Design Goals
+
+CloudEvents are typically used in a distributed system to allow for services to
+be loosely coupled during development, then configured with tight-coupling for
+specific applications.
```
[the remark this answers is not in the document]
duglin: Would it address your concern if this just said:
- - - 
CloudEvents are typically used in distributed systems to allow for services to be loosely
coupled.
- - -

about/design-goals.md
duglin: @clemensv are you looking for a change or is this just a commentary?

about/design-goals.md
```diff
@@ -0,0 +1,52 @@
...
+- One or more common architectures that are in active use today or planned to be
+built by WG members
+- How events are transported from source to action via at least one protocol
+- Whether a shared trigger specification is needed, or if runtime
+interoperability can be achieved independently, then shared configuration and/or
+API definition can be addressed as follow-on work
+- Authorization model(s) that control access and execution permissions
+
+
+## Non-Goals of this spec
```
inlined: I would also add a non-goal that we don't need to standardize on a single identity/access control system. (E.g. we're not requiring a single spec like OpenID or inventing a pan-cloud version of IAM). This seems to follow naturally from CNCF's "No kingmakers" rule.

about/design-goals.md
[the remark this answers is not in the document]
duglin: @clemensv ^^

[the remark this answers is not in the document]
clemensv: Yes.

about/design-goals.md
clemensv: I am suggesting for qualification above to be dropped and clearly state that we're not defining APIs.

ultrasaurus: @clemensv I'm inclined to agree, yet others have advocated for APIs as a project of the Serverless WG.  In any case, I think I can re-word so as to simply clarify it is outside of our current scope

about/design-goals.md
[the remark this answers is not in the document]
ultrasaurus: In the conversations over the past couple of months, I have found that there are many different kinds of events.  In the industry an event can mean:
1. any async message
2. time-serious data
3. what we're talking about here

There are many messaging systems (#1) that allow services to be loosely coupled which are far more general than what I believe we're discussing here.  

To address concern by @clemensv, I suggest:

CloudEvents are typically used in a distributed system to allow for services to
be loosely coupled during development, then may be configured with
tight-coupling for specific applications or dynamically configured at runtime.

Will update PR

about/design-goals.md
ultrasaurus: agreed. will add.

README.md
```diff
@@ -21,6 +21,25 @@ ranging from major cloud providers to popular SaaS companies.  Our end goal is
...
+The CNCF Serverless WG is working to formalize the [specification](spec.md)
+based on [design goals](about/design-goals.md) which focus on interoperability
+between systems which generate and respond to events.
+
+In order to achieve these goals, the Serverless WG must describe:
+- Common attributes of an *event* that facilitate interoperability
+- One or more common architectures that are in active use today or planned to be
+built by WG members
+- How events are transported from source to action via at least one protocol
+- Authorization model(s) that control access and execution permissions
```
duglin: This one feels out of scope for this spec - at least for now since it would into impl details of how to process the events, right?

README.md
```diff
@@ -21,6 +21,25 @@ ranging from major cloud providers to popular SaaS companies.  Our end goal is
...
+
+### Non-Goals of this specification
+The following non-goals will not be part of the specification.
```
duglin: s/./:/

about/design-goals.md
```diff
@@ -0,0 +1,57 @@
...
+be loosely coupled during development, then may be configured with
+tight-coupling for specific applications or dynamically configured at runtime.
+
+The goal of the CloudEvents specification is to define interoperability of event
+systems that allow cloud providers, third-party tools and services to provide:
+- new event sources
+- new event consumers
+- observability, such as monitoring, debugging and visualization,
+- developer tools and libraries
+
```
deissnerk: I think it is also an important design goal to define the event attributes in a way they can be added as extension to existing protocol headers like it is done for HTTP. So it's not about introducing a full-fledged event layer.

duglin:
@ultrasaurus I view this PR has having two aspects to it:
1 - outlining what the spec if for, what its trying do, the problem space, etc... and I think that should go into the spec in some kind of "overview" or "intro" section. That will help set the context for readers of the spec. Putting that into a side-doc someplace else in the repo will just be ignored - and this info is too important (IMO).
2 - process related topics - things like "the wg must do ...." and I think that info should go either into the README (like in the new "WG process" section you propose) or into the "Status" section of the spec - which is already kind of our todo list. Once those items are completed we don't need to list them any place and they can be deleted.

Net: I like the concepts but I think the placement of the text should be tweaked so that its incorporated into our existing docs rather than in another one that I suspect will be hidden from most causal readers.  This will also avoid the problem of potentially duplicating definition in two docs.

about/design-goals.md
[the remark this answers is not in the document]
ac360: @duglin +1 Good rule.  Our company maintains a glossary.  It greatly helps alignment overall.

about/design-goals.md
ac360: +1

about/design-goals.md
rachelmyers: I think splitting this up into the spec and goals is a great idea. Let me know if you'd like me to help with that.

ultrasaurus:
I had moved WG process notes  into README in prior commit -- just moved the design goals into the spec

README.md
ultrasaurus: I don't see how systems can interop without addressing authorization in some way

duglin: True, but that will come into play if we're talking about APIs. If we're limited (for now) to just "shape of an event" I don't think auth comes into play. Right?

spec.md
```diff
@@ -26,6 +27,61 @@ Enter CloudEvents, a specification for describing event data in a common way.
...
+- new event consumers
+- observability, such as monitoring, debugging and visualization,
+- developer tools and libraries
+
+## Key properties of the system:
+- Source & action may be developed and deployed independently
+- Trigger configuration allows for actions to be invoked based on a condition
+(e.g. for a specific event type generated by a source, optionally filtered by
+specific attribute values, such as a Function which is invoked when an image is
+uploaded to specific S3 or Google Cloud Storage “bucket” path)
```
duglin: let's just say: ... when an image is uploaded to an object store.
and leave off product specific names.

spec.md
```diff
@@ -26,6 +27,61 @@ Enter CloudEvents, a specification for describing event data in a common way.
...
+- Action: An action consumes an event, producing a behavior or effect which was
+triggered by a specific *occurrence* from a specific *source*.  While the
+details of how events are delivered to an action is outside of the scope of the
+specification, the purpose of generating an *event* is typcially to allow other
+systems to easily react to changes in a source that they do not control. The
+*source* and action are often built by different developers.
+
+Examples:
+- the *source* is a managed service and the *action* is custom
+code in a serverless Function (such as AWS Lambda or Google Cloud Functions).
```
duglin: let's leave the product names out - just remove the parens stuff - I don't think anything is lost by doing so.

README.md
```diff
@@ -21,6 +21,25 @@ ranging from major cloud providers to popular SaaS companies.  Our end goal is
...
+The CNCF Serverless WG is working to formalize the [specification](spec.md)
+based on [design goals](spec.md#design-goals) which focus on interoperability
+between systems which generate and respond to events.
+
+In order to achieve these goals, the Serverless WG must describe:
+- Common attributes of an *event* that facilitate interoperability
+- One or more common architectures that are in active use today or planned to be
+built by WG members
+- How events are transported from source to action via at least one protocol
+- Authorization model(s) that control access and execution permissions
```
duglin: I think its important to talk about having implementations of the spec prior to going to v1.0.
No code, no verification, no spec :-)

spec.md
```diff
@@ -26,6 +27,61 @@ Enter CloudEvents, a specification for describing event data in a common way.
...
+The goal of the CloudEvents specification is to define interoperability of event
+systems that allow cloud providers, third-party tools and services to provide:
+- new event sources
+- new event consumers
+- observability, such as monitoring, debugging and visualization,
+- developer tools and libraries
+
+## Key properties of the system:
+- Source & action may be developed and deployed independently
+- Trigger configuration allows for actions to be invoked based on a condition
```
clemensv: I agree with the general goal of event sources and event handlers to be developed and deployed independently, but it's not clear to me how that goal can be specifically tied to this specification. You can easily sabotage this goal for yourself by expressing the event payload in a proprietary fashion that causes hard coupling between event source and handler, and there's little this spec can do to help you out of that. 

The core elements of the current specification do not define or mandate a "source"/"trigger"/"action" in the sense of resources that can be "developed and deployed". Those are very reasonable naming choices for concepts of a particular implementation that is using the specification, but so are "exchange"/"binding"/"queue" (RabbitMQ), "address"/"binding"/"queue" (ActiveMQ), "topic"/"selector"/"subscriber" (JMS), "topic"/"rule"/"subscriber" (Azure Service Bus), "topic"/"filter"/"subscriber" (Azure Event Grid), etc. 

* as an event publisher I want to make the event available in a way that it can be distributed to an interested audience and the audience can get at a particularly interesting subset with minimal own effort (i.e. delegate filtering to generic middleware)
* as an event consumer I want to have enough fidelity in the event metadata so that I can tell that generic middleware to pick out, and deliver a subset of events to me, and for there to be enough fidelity in the metadata for local dispatch to a handler.

I prefer for the spec not to pick favorites for what code concept might produce an event, what code concept might filter/dispatch an event, and what code concept might handle the event. We can certainly talk about concrete examples like sensors or cash registers and their respective contexts on one hand and about web hooks as event handlers on the other, but I prefer for us to stay neutral regarding the architecture model and nomenclature for how those events flow from A to B in a middleware usage scenario of this spec, precisely because there's already no shortage of those models.

(more context https://github.com/cloudevents/spec/issues/112 )

spec.md
```diff
@@ -26,6 +27,61 @@ Enter CloudEvents, a specification for describing event data in a common way.
...
+uploaded to specific S3 or Google Cloud Storage “bucket” path)
+- Events are uniquely identified to facilitate developer velocity and tooling:
+  - a unique idenfier allows for idempotency for reliable implementation
+  - structured event definitions allow the consumer to rely on a specific format
+- Common metadata attributes allow for observability via tools whithout
+knowledge of the specific implementation of an event system and also can provide
+value without detailed handling of the specifics of the unique event data
+- CloudEvents may be transported from source to action with different protocols.
+
+In order to achieve the above goals and desired system properties, the
```
clemensv: Since we're inlining the goals section into the spec, it's now odd for the section to argue for specific elements of the spec itself rather than discussing the joint interoperability objective.

spec.md
```diff
@@ -26,6 +27,61 @@ Enter CloudEvents, a specification for describing event data in a common way.
...
+
+Examples:
+- the *source* is a managed service and the *action* is custom
+code in a serverless Function (such as AWS Lambda or Google Cloud Functions).
+- the *source* is an IoT device and the *event* is transmitted through an API
+Gateway to the *action*, a service provided by a different developer.
+
+The events system provides a way for the developer to specify which *event*
+should be delivered to which *action*. The specific mechanism is outside the
+scope of the specification, yet the following definitions faciltate an
```
rabbah: s/faciltate/facilitate

spec.md
```diff
@@ -26,6 +27,61 @@ Enter CloudEvents, a specification for describing event data in a common way.
...
+understanding of how CloudEvent are typically used in context:
```
rabbah: s/CloudEvent/CloudEvents

README.md
yaronha: @ultrasaurus not sure if the goal is to define the common attributes of an event (there can be many types of events w/o common structure), rather the attributes that allow us to pass the event from the source to the function reliably, observe its movement, and decipher its content (i.e. what i call the event envelop). The interop at the event level will be achieved by (future) definition of some common event types/schema.

spec.md
rachelmyers: I'll change this to `an image is uploaded to a specific bucket in an object store` to emphasize that this is about the individual instance

spec.md
rachelmyers: That sounds fine; I'll make this change.

README.md
rachelmyers: The [roadmap](https://github.com/cloudevents/spec/blob/master/roadmap.md) calls for creating reference implementations by v0.2, I assume with the expectation that implementations will be updated as we refine the spec. 

I'm resistant to including implementations here because I see this section as being the sticking points around interoperability that the spec should address, while implementations are a demonstration of the spec working.

How strongly do you feel that the implementations should be included here?

README.md
```diff
@@ -21,6 +21,25 @@ ranging from major cloud providers to popular SaaS companies.  Our end goal is
...
+
+### Non-Goals of this specification
+The following non-goals will not be part of the specification:
+- Function build and invocation process
+- Language-specific runtime APIs
+- Selecting a single identity/access control system
+
```
csantanapr: I like that the "Non-Goals" are included it helps having a clear what is the scope of the spec.

README.md
duglin: re: `Authorization model(s)...`  can someone elaborate on this one? access and permissions for what?

spec.md
```diff
@@ -26,6 +27,61 @@ Enter CloudEvents, a specification for describing event data in a common way.
 CloudEvents seeks to ease event declaration and delivery across services,
 platforms and beyond.
 
+# Design Goals
+
+CloudEvents are typically used in a distributed system to allow for services to
+be loosely coupled during development, then may be configured with
```
duglin: s/may/can/

spec.md
```diff
@@ -26,6 +27,61 @@ Enter CloudEvents, a specification for describing event data in a common way.
...
+(e.g. for a specific event type generated by a source, optionally filtered by
+specific attribute values, such as a Function which is invoked when an image is
+uploaded to a specific bucket in an object store).
+- Events are uniquely identified to facilitate developer velocity and tooling:
+  - a unique identifier allows for idempotency for reliable implementation
+  - structured event definitions allow the consumer to rely on a specific format
+- Common metadata attributes allow for observability via tools without
+knowledge of the specific implementation of an event system and also can provide
+value without detailed handling of the specifics of the unique event data
+- CloudEvents may be transported from source to action with different protocols.
```
duglin: s/may/can/

duglin:
All commits need to be signed to make the DCO checker happy

spec.md
```diff
@@ -26,6 +27,61 @@ Enter CloudEvents, a specification for describing event data in a common way.
...
+# Design Goals
+
+CloudEvents are typically used in a distributed system to allow for services to
+be loosely coupled during development, then may be configured with
+tight-coupling for specific applications or dynamically configured at runtime.
+
+The goal of the CloudEvents specification is to define interoperability of event
+systems that allow cloud providers, third-party tools and services to provide:
+- new event sources
+- new event consumers
```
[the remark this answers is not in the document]
ultrasaurus: Imagine there is an SMS service that can receive specific kinds of common event types and knows how to transform them into natural language.  "Serverless" is a bit broader than FaaS.

spec.md
```diff
@@ -26,6 +27,61 @@ Enter CloudEvents, a specification for describing event data in a common way.
...
+- observability, such as monitoring, debugging and visualization,
+- developer tools and libraries
```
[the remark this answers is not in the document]
duglin: I don't think the items in this list are the things the spec (or even the wg) will do, rather this are some of the things that people should be able to do *with* the spec, when completed.

README.md
yaronha: @duglin LGTM

ultrasaurus:
FYI -- this was discussed at today's Serverless WG meeting -- please add comments / suggestions in [Google Doc](https://docs.google.com/document/d/1kadXOdcvZ1ex2kIuXY1Yoy_OEVZawyYN8jLImM9Myog/edit)

README.md
rachelmyers: @duglin, An example use case for handling Authorization is that an Event Consumer receives an Event, and will take an action of, say, unlocking the door. First it wants to confirm that this user has permission to unlock this door.

duglin: @rachelmyers yup - I agree with that kind of action/check being done - what's not clear to me is whether that's part of what our spec needs to define. That feels out of scope w.r.t. "defining an event" - and gets more info the processing model of the event.

spec.md
```diff
@@ -26,6 +27,29 @@ Enter CloudEvents, a specification for describing event data in a common way.
...
+to express an interest in an event or class of events that is not yet being
+produced.
+
+To this end, the specification will include common metadata attributes of an
+event that facilitate interoperability, where the event does not contain any
+details about the consumer or transport that might be use to send the event.
+
+## Non-Goals
+The following will not be part of the specification:
+Function build and invocation process
```
duglin: missing bullet markers ("-") for each

ultrasaurus:
FYI -- this is a step forward with high level description of goals, based on discussions this week.

The group that met live, felt there still needed to be additional work done via user stories or use cases.  Thinking through next steps, I think we'll need a sequence like this:

1. PR to add the list of Personas, to the README, that will be referenced in future work.  Extracted from Personas being developed down [the doc](https://docs.google.com/document/d/1kadXOdcvZ1ex2kIuXY1Yoy_OEVZawyYN8jLImM9Myog/edit)
2. PR to create a user-stories.md file to track/store our guiding user-stories
3. Discussion prioritizing the user stories
4. Validate that metadata attributes enable the user stories

There was some side discussion of Personas/User Stories vs Use Cases, maybe we need both... probably merits discussion in WG meeting.

ultrasaurus:
FYI -- that was a rebase, which had a merge conflict.  My editor is set to remove whitespace at the end of lines, which is minor cosmetic change.

spec.md
```diff
@@ -35,8 +35,8 @@ can be connected to create new applications.
 
 The goal of the CloudEvents specification is to define interoperability of event
 systems that allow services to produce or consume events, where the producer and
-consumer can be developed and deployed independently.  A producer must be able
-to generate events before a consumer is listening, and a consumer must be able
+consumer can be developed and deployed independently.  A producer MUST be able
+to generate events before a consumer is listening, and a consumer MUST be able
```
duglin: I'd switch these to "needs to" because this isn't technically part of the normative section of the spec. These don't place any formal requirements on people implementing the spec.

spec.md
```diff
@@ -26,6 +27,29 @@ Enter CloudEvents, a specification for describing event data in a common way.
...
+# Design Goals
+
+CloudEvents are typically used in a distributed system to allow for services to
+be loosely coupled during development, deployed independently, and later
+can be connected to create new applications.
+
+The goal of the CloudEvents specification is to define interoperability of event
+systems that allow services to produce or consume events, where the producer and
+consumer can be developed and deployed independently.  A producer MUST be able
+to generate events before a consumer is listening, and a consumer MUST be able
```
duglin: I'd switch these to "needs to" because this isn't technically part of the normative section of the spec. These don't place any formal requirements on people implementing the spec.

duglin:
LGTM for text changes and to the proposed next steps: https://github.com/cloudevents/spec/pull/101#issuecomment-373027752

ac360:
@ultrasaurus @duglin LGTM.  The original **Scope** section is still in the spec.  Now that we have the new Road Map and the Design Goals, we should take this section out (in a separate PR).

duglin:
@ac360 yes, once this PR is merged I was going to open one to remove, or radically update, the "Status" section

duglin:
Approved on the 3/15 call

PR #114, opened 2018-03-14
Collect supporting docs in `about` directory`

rachelmyers:
Since we're creating supporting content for the spec that provides context, guidance, and intermediate decisions, and we've talked at various points about creating a glossary, personas or user stories, and documentation of prior art, I want to make a space for more content that's _about_ the spec but not part of the spec itself. As a bonus, it collects reading material for people who join the group. 

I've only moved `references.md` and `use-cases.md` into a new `about` directory. If people like this change we could consider what other content could be pulled into `about` as well. If `about` is the wrong word, let me know; I'm happy to change it. My other thought was a `docs` directory, but I thought we should to reserve for more user-facing documentation about the spec.

What are people's thoughts?

P.S. I've worked on some projects with git commit [emoji conventions](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#styleguides). Since we don't do that, I'll leave them here in the PR: 
:art: => changes to formatting or organization
:memo: => updating the docs


ultrasaurus:
This looks great!  I think it will be super helpful to collect these docs in a sub-directory.

duglin:
Approved on the 3/22/2018 call.

Issue #112, opened 2018-03-05
Topics/Subject Prior Art and Example


clemensv:
We (Microsoft) believe that the technical core artifact that the WG has developed to date, the property definitions, are quite close to where the WG needs to be for an initial "stable" draft that teams can start to do some bi- or multilateral interop work on. In other organizations, there's a differentiation between a
working draft that changes day-to-day and a committee draft that has been promoted to a status where implementation makes sense. That doesn't mean at all that a subsequent draft wouldn't have breaking changes; it simply means that a quorum of WG members are willing to try out the status quo.

Allow me to substantiate this in the following, with several examples and pointers to prior art, and also some principles to apply to goals.

The goal of this WG could be a holistic definition and shared base infrastructure implementation for creation, routing, dispatch, and handling of "events". We do not think such a goal would be realistic.

At Microsoft alone, we have several dozen of such infrastructures, each with its own native notion of events, and most existing with good reason since they aim either at a particular audience or address certain architectural needs. This doesn't look different in other larger firms.

What this WG can achieve, and in large part already has achieved, is to establish consensus on a sufficient set of well-defined [criteria](#Which-Criteria-Do-We-Need) that allow generic infrastructure to route and dispatch events to event handlers, and to enable operational monitoring of such infrastructure to track and display the events in a uniform fashion.

**The primary purpose of achieving such a consensus is interoperability.** The specification may also be useful for consistently wiring up the various parts of a single vendor's stack internally, but we're ultimately not pooling resources for everyone to build their own stuff better in isolation.

Interoperability has many facets, so let me call out the most fundamental interoperability goal that I believe most of the group is chasing:

With apologies to everyone whose products I'm not mentioning: An event raised in the Oracle Cloud should be routable through Azure Event Grid, with the subscriber being an IBM Cloud Function. An event raised in the Google Cloud should be routable through AWS SNS be subscriber being someone's custom web-hook handler. And with all that, the code in that web-hook handler shouldn't be looking like [Yaron's example (slide 2)](https://github.com/cloudevents/spec/blob/master/share/2018-02-22%20Yaron%20CloudEvents.pdf)

Once we jointly declare victory on that goal, we can proceed to see whether there is more common ground for achieving consensus on specific categories of events, such as time-series telemetry.

## Which criteria do we *really* need?

Which routing and dispatch criteria do we need? Whether it's headers, or properties, or fields, or metadata, an event will always carry information that answers some basic questions about the event.

[...]

PR #117, opened 2018-03-19
Usage Scenarios

clemensv:
Collapsed the 12 scenarios from the email to 4 and added clarifying discussion

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -63,12 +63,11 @@ concurrently, including being both a producer and a consumer of events.
    application's capabilities with event-driven extensions.
 
    Events are typically produced related to a context or a producer-chosen
-   base classification. A temperature sensor in a room might be
+   classification. For example, a temperature sensor in a room might be
    context-qualified by mount position, room, floor, and building. A sports
    result might be classified by league and team.
 
-   The producer application might reside on server hosts or on a device of
-   any kind.
+   The producer application could run anywhere, such as on a server or a device.
 
    The produced events might be rendered and emitted directly by the producer
    or by an intermediary; as example for the latter, consider event data
@@ -77,16 +76,19 @@ concurrently, including being both a producer and a consumer of events.
    specification will be rendered by a network gateway on behalf of the
    producer.
 
-2) Applications consume events originating from devices and/or people
-   and/or applications/-modules/-instances for display, archival, analytics,
-   workflow processing, or other handling.
+   For example, a weather station transmits a 12 byte event payload indicating
+   weather conditions once every 5 minutes over LoRaWAN. A LoRaWAN gateway
+   might then be used to publish the event to an Internet destination and
+   in the Cloud Events format. The weather station is the producer and
+   the intermediary gateway plays a middleware role (see 3).
 
-   Platform tools and applications consume events for the purpose of
-   monitoring the condition and/or providing transparency into the operation
-   of a business solution and its foundational building blocks.
+2) Applications consume events for the purposes such as display, archival,
+   analytics, workflow processing, monitoring the condition and/or providing
+   transparency into the operation of a business solution and its foundational
+   building blocks.
 
-   The consumer application might reside on server hosts or on a device
-   of any kind.
+   The consumer application could run anywhere, such as on a server or a
+   device.
 
    A consuming application will typically be interested in:
    - distinguishing events such that the exact same event is not
@@ -97,11 +99,17 @@ concurrently, including being both a producer and a consumer of events.
      originating context and/or relative to a wall-clock.
    - understanding the context-related detail information carried
      in the event.
+   - correlating event instances from multiple event producers and send
+     them to the same consumer context.
 
    In some cases, the consuming application might be interested in:
    - obtaining further details about the event's subject from the
      originating context, like obtaining detail information about a
      changed object that requires privileged access authorization.
+     For example, a HR solution might only publish very limited
+     information in events for privacy reasons, and any event consumer
+     needing more data will have to obtain details related to the event
+     from the HR system under their own authorization context.
    - interact with the event's subject at the originating context,
      for instance reading a storage blob after having been informed
      that this blob has just been created.
@@ -130,10 +138,17 @@ concurrently, including being both a producer and a consumer of events.
      diagnostics purposes.
 
    To satisfy these needs, middleware will be interested in:
-   - A metadata discriminator usable for classification of events so that
-     consumers can express interest in one or multiple such classes.
+   - A metadata discriminator usable for classification or
+     contextualization of events so that consumers can express interest
+     in one or multiple such classes or contexts.
+     For instance, a consumer might be interested in all events related
+     to a specific directory inside a file storage account.
    - A metadata discriminator that allows distinguishing the subject of
-     a particular event of that class.
+     a particular event of that class or context.
+     For instance, a consumer might want to filter out all events related
+     to new files ending with ".jpg" (the file name being the "new file"
+     event's subject) for the context describing specific directory
+     inside a file storage account that it has registered interest on.
    - An indicator for the encoding of the event and its data.
    - An indicator for the structural layout (schema) for the event and
      its data.
@@ -152,7 +167,8 @@ concurrently, including being both a producer and a consumer of events.
    on a particular subject.
 
    Frameworks are most interested in semantic metadata commonality
-   across the platforms they abstract.
+   across the platforms they abstract, so that similar activities can
+   be handled uniformly.
 
    For a sports application, a developer using the framework might be
    interested in all events from today's game (subject) of a team in a
--- a/spec.md
+++ b/spec.md
@@ -48,6 +48,144 @@ The following will not be part of the specification:
 * Language-specific runtime APIs
 * Selecting a single identity/access control system
 
+## Usage Scenarios
+
+The list below enumerates key usage scenarios and developer perspectives
+that have been considered for the development of this specification.
+
+In these scenarios, we keep the roles of event producer and event consumer
+distinct. A single application context can always take on multiple roles
+concurrently, including being both a producer and a consumer of events.
+
+1) Applications produce events for consumption by other parties, for instance
+   for providing consumers with insights about end-user activities, state
+   changes or environment observations, or for allowing complementing the
+   application's capabilities with event-driven extensions.
+
+   Events are typically produced related to a context or a producer-chosen
+   base classification. A temperature sensor in a room might be
+   context-qualified by mount position, room, floor, and building. A sports
+   result may be classified by league and team.
+
+   The producer application might reside on server hosts or on a device of 
+   any kind.
+
+   The produced events might be rendered and emitted directly by the producer
+   or by an intermediary; as example for the latter, consider event data
+   transmitted by a device over payload-size-constrained networks such as
+   LoRaWAN or ModBus, and where events compliant to this
+   specification will be rendered by a network gateway on behalf of the
+   producer.
+
+2) Applications consume events originating from devices and/or people
+   and/or applications/-modules/-instances for display, archival, analytics,
+   workflow processing, or other handling.
+
+   Platform tools and applications consume events for the purpose of
+   monitoring the condition and/or providing transparency into the operation
+   of a business solution and its foundational building blocks.
+
+   The consumer application might reside on server hosts or on a device
+   of any kind.
+
+   A consuming application will typically be interested in:
+   - distinguishing events such that the exact same event is not
+     processed twice.
+   - identifying and selecting the origin context or the
+     producer-assigned classification.
+   - identifying the temporal order of the events relative to the
+     originating context and/or relative to a wall-clock.
+   - understanding the context-related detail information carried
+     in the event.
+
+   In some cases, the consuming application might be interested in:
+   - obtaining further details about the event's subject from the
+     originating context, like obtaining detail information about a
+     changed object that requires privileged access authorization.
+   - interact with the event's subject at the originating context,
+     for instance reading a storage blob after having been informed
+     that this blob has just been created.
+
+   Consumer interests motivate requirements for which information
+   producers shall include in an event.
+
+3) Middleware routes events from producers to consumers, or onwards
+   to other middleware. Applications producing events might delegate
+   certain tasks arising from their consumers' requirements to
+   middleware:
+
+   - Management of many concurrent interested consumers for one of
+     multiple classes or originating contexts of events
+   - Processing of filter conditions over a class or originating context
+     of events on behalf of consumers.
+   - Transcoding, like encoding in MsgPack after decoding from JSON
+   - Transformation that changes the event's structure, like mapping from
+     a proprietary format to Cloud Events, while preserving the
+     identity and semantic integrity of the event.
+   - Instant "push-style" delivery to interested consumers.
+   - Storing events for eventual delivery, either for pick-up initiated
+     by the consumer ("pull"), or initiated by the middleware ("push")
+     after a delay.
+   - Observing event content or event flow for monitoring or
+     diagnostics purposes.
+
+   To satisfy these needs, middleware will be interested in:
+   - A metadata discriminator usable for classification of events so that
+     consumers can express interest in one or multiple such classes.
+   - A metadata discriminator that allows distinguishing the subject of
+     a particular event of that class.
+   - An indicator for the encoding of the event and its data.
+   - An indicator for the structural layout (schema) for the event and
+     its data.
+
+   Whether its events are available for consumption via a middleware is
+   a delegation choice of the producer.
+
+4) Frameworks and other abstractions make interactions with event platform
+   infrastructure simpler, and often provide common API surface areas
+   for multiple event platform infrastructures.
+
+   Frameworks are often used for turning events into an object graph,
+   and to dispatch the event to some specific handling user-code or
+   user-rule that permits the consuming application to react to
+   a particular kind of occurrence in the originating context and
+   on a particular subject.
+
+   Frameworks are most interested in semantic metadata commonality
+   across the platforms they abstract.
+
+   For a sports application, a developer using the framework may be
+   interested in all events from today's game (subject) of a team in a
+   league (topic of interest), but wanting to handle reports
+   of "goal" differently than reports of "substitution".
+   For this, the framework will need a suitable metadata discriminator
+   that frees it from having to understand the event details.
+
+## Status
+At this time the specification is focused on the following scope:
+
+* Agree upon a set of event metadata attributes (“context”) that:
+  * Offer a basic description of the event and the data it carries.
+  * Are currently implemented and semantically similar across multiple
+    platforms.
+  * Can be delivered separately from the event data in the transport headers
+    (e.g. HTTP, AMQP, Kafka) or together with the data in a serialized fashion
+    (e.g. JSON, protobuf, Avro).
+  * Include a description of the transport/protocol and encoding, with an
+    initial focus on HTTP.
+  * Can be extended to support experimental or uncommon features, while being
+    clearly indicated as an extension (e.g. extensions use a common prefix).
+  * Allow for evolution of both the payload and CloudEvents definition (e.g.
+    versioning).
+  * Can be embedded at different stages along the route of the event by
+    middleware (e.g. a router can add transport or auth information).
+* Establish a backlog of prospective event metadata attributes (“context”)
+  for potential inclusion in the future.
+* Include use-case examples to help users understand the value of CloudEvents,
+  with an initial focus on HTTP and Functions-as-a-Service/Serverless computing.
+* Determine process and overall governance of the specification.
+* Discuss additional architecture components that complement this specification.
+
 ## Notations and Terminology
 
 ### Notational Conventions
--- a/spec.md
+++ b/spec.md
@@ -52,6 +52,12 @@ The following will not be part of the specification:
 
 The list below enumerates key usage scenarios and developer perspectives
 that have been considered for the development of this specification.
+These usage scenarios are by no means exhaustive, and the specification
+does not aim to be prescriptive about usage.
+
+These scenarios are not normative; anyone is free to create a system that
+mixes these scenarios. These cases establish a common vocabulary of event
+producer, consumer, middleware, and framework.
 
 In these scenarios, we keep the roles of event producer and event consumer
 distinct. A single application context can always take on multiple roles
@@ -76,11 +82,12 @@ concurrently, including being both a producer and a consumer of events.
    specification will be rendered by a network gateway on behalf of the
    producer.
 
-   For example, a weather station transmits a 12 byte event payload indicating
-   weather conditions once every 5 minutes over LoRaWAN. A LoRaWAN gateway
-   might then be used to publish the event to an Internet destination and
-   in the Cloud Events format. The weather station is the producer and
-   the intermediary gateway plays a middleware role (see 3).
+   For example, a weather station transmits a 12-byte, proprietary event
+   payload indicating weather conditions once every 5 minutes over LoRaWAN. A
+   LoRaWAN gateway is then used to publish the event to an Internet destination
+   in the Cloud Events format. The LoRaWAN gateway is the event producer,
+   publishing on behalf of the weather station, and will set event metadata
+   appropriately to reflect the source of the event.
 
 2) Applications consume events for the purposes such as display, archival,
    analytics, workflow processing, monitoring the condition and/or providing
@@ -155,6 +162,11 @@ concurrently, including being both a producer and a consumer of events.
 
    Whether its events are available for consumption via a middleware is
    a delegation choice of the producer.
+
+   In practice, middleware can take on role of a producer when it changes
+   the semantic meaning of an event, a consumer when it takes action based
+   on an event, or middleware when it routes events without making semantic
+   changes.
 
 4) Frameworks and other abstractions make interactions with event platform
    infrastructure simpler, and often provide common API surface areas
--- a/spec.md
+++ b/spec.md
@@ -65,9 +65,9 @@ concurrently, including being both a producer and a consumer of events.
    Events are typically produced related to a context or a producer-chosen
    base classification. A temperature sensor in a room might be
    context-qualified by mount position, room, floor, and building. A sports
-   result may be classified by league and team.
+   result might be classified by league and team.
 
-   The producer application might reside on server hosts or on a device of 
+   The producer application might reside on server hosts or on a device of
    any kind.
 
    The produced events might be rendered and emitted directly by the producer
@@ -107,7 +107,7 @@ concurrently, including being both a producer and a consumer of events.
      that this blob has just been created.
 
    Consumer interests motivate requirements for which information
-   producers shall include in an event.
+   producers ought to include an event.
 
 3) Middleware routes events from producers to consumers, or onwards
    to other middleware. Applications producing events might delegate
@@ -154,7 +154,7 @@ concurrently, including being both a producer and a consumer of events.
    Frameworks are most interested in semantic metadata commonality
    across the platforms they abstract.
 
-   For a sports application, a developer using the framework may be
+   For a sports application, a developer using the framework might be
    interested in all events from today's game (subject) of a team in a
    league (topic of interest), but wanting to handle reports
    of "goal" differently than reports of "substitution".


spec.md
```diff
@@ -49,6 +49,119 @@ The following will not be part of the specification:
...
+
+1) Applications produce events for consumption by other parties, for instance
+   for providing consumers with insights about end-user activities, state
+   changes or environment observations, or for allowing complementing the
+   application's capabilities with event-driven extensions.
+
+   Events are typically produced related to a context or a producer-chosen
+   base classification. A temperature sensor in a room might be
+   context-qualified by mount position, room, floor, and building. A sports
+   result might be classified by league and team.
```
ultrasaurus: Suggested wording to clarify:

Events are typically produced related to a context or a producer-chosen classification.  For example: a temperature sensor in a room might be context-qualified by room, floor, and building.  A sports result might be classified by league and team.

+   context-qualified by mount position, room, floor, and building. 

-- removing "base" might help

spec.md
```diff
@@ -49,6 +49,119 @@ The following will not be part of the specification:
...
+
+   The producer application might reside on server hosts or on a device of
+   any kind.
```
ultrasaurus: The producer application could run anywhere, such as on a server or a device.

spec.md
```diff
@@ -49,6 +49,119 @@ The following will not be part of the specification:
...
+
+   The produced events might be rendered and emitted directly by the producer
+   or by an intermediary; as example for the latter, consider event data
+   transmitted by a device over payload-size-constrained networks such as
+   LoRaWAN or ModBus, and where events compliant to this
+   specification will be rendered by a network gateway on behalf of the
+   producer.
```
ultrasaurus: Note: the rendering of the event by an intermediary does not change the semantic meaning of the event, specifically that the producer remains the same, even if the event were not originally produced in the CloudEvent format.

spec.md
```diff
@@ -49,6 +49,119 @@ The following will not be part of the specification:
...
+
+2) Applications consume events originating from devices and/or people
+   and/or applications/-modules/-instances for display, archival, analytics,
+   workflow processing, or other handling.
+
+   Platform tools and applications consume events for the purpose of
+   monitoring the condition and/or providing transparency into the operation
+   of a business solution and its foundational building blocks.
```
ultrasaurus: graft onto above paragraphs...

display, archival, analytics, workflow processing, monitoring the condition and/or providing transparency into the operation of a business solution and its foundational building blocks.
 
Proposed merging of two paragraphs:

Applications consume events for the purposes such as display, archival, analytics, workflow processing, monitoring the condition and/or providing transparency into the operation of a business solution and its foundational building blocks.

spec.md
```diff
@@ -49,6 +49,119 @@ The following will not be part of the specification:
...
+
+   The consumer application might reside on server hosts or on a device
```
ultrasaurus: see above

spec.md
```diff
@@ -49,6 +49,119 @@ The following will not be part of the specification:
...
+   of any kind.
+
+   A consuming application will typically be interested in:
```
ultrasaurus: numbered list to help discussion

spec.md
```diff
@@ -49,6 +49,119 @@ The following will not be part of the specification:
...
+   - distinguishing events such that the exact same event is not
+     processed twice.
+   - identifying and selecting the origin context or the
+     producer-assigned classification.
+   - identifying the temporal order of the events relative to the
+     originating context and/or relative to a wall-clock.
```
ultrasaurus: relative to the originating context (which could be clock-time or other sequence number)

spec.md
```diff
@@ -49,6 +49,119 @@ The following will not be part of the specification:
...
+   - understanding the context-related detail information carried
+     in the event.
+
+   In some cases, the consuming application might be interested in:
+   - obtaining further details about the event's subject from the
+     originating context, like obtaining detail information about a
+     changed object that requires privileged access authorization.
+   - interact with the event's subject at the originating context,
+     for instance reading a storage blob after having been informed
+     that this blob has just been created.
```
ultrasaurus: For example: an HR producer application may choose to use a customer ID in the event (to limit PII exposure), because different consumers may have different access permissions, and may need additional authorization, obtain additional detail about the customer that may not be present in the event or act on the customer record.

spec.md
cathyhongzhang: Sara: If we decide that the producer will still be the the original entity that emits the event, then we need to spell this out clearly. Let's take the following two examples to make sure we are on the same page. 

1. In an IoT example, a motion detection event is originally emitted by a motion detector and then sent to the cloud platform via a HTTP API GW, the producer will be the motion detector, not the API GW. Could you confirm this is what you propose? 
2. In the same IoT example, if the motion detection event triggers a video/image to be saved into a storage which  in turn triggers a motion video/image storage event to the cloud platform, does this mean the semantic meaning of the event has  been changed? Is the producer the storage entity or is it still the motion detector? I think we need to clearly define what constitutes "change of the semantic meaning of the event"

ultrasaurus: @cathyhongzhang agree with your interpretation.   If we added the following text, do you think that would be clear?

For example, an IoT devices capture an image and provides that as the payload of its motion detection event:
1. The devices transmits a stream of events using a proprietary protocol, there exists an API Gateway that transforms these into CloudEvents, the IoT device is the producer, even though the API Gateway created the data in the CloudEvent format because the semantic meaning of the event remains unchanged
2. The devices transmits a stream of events using a proprietary protocol and an intermediary saves the image in an ftp server, which then emits a object.created event.  The ftp server is the producer of the object.created event (not the motion detection event).

spec.md
cathyhongzhang: The consuming application is also interested in 
-  correlating event instances from multiple event producers and send them to the same workflow instance.

spec.md
```diff
@@ -49,6 +49,119 @@ The following will not be part of the specification:
...
+
+   Consumer interests motivate requirements for which information
+   producers ought to include an event.
+
+3) Middleware routes events from producers to consumers, or onwards
+   to other middleware. Applications producing events might delegate
```
rachelmyers: Suggested change: Middleware routes events from producers to consumers, or on to other middleware, without changing the event.

deissnerk: > without changing the event.

Does that include the context attributes or just the payload?

spec.md
```diff
@@ -49,6 +49,119 @@ The following will not be part of the specification:
...
+   middleware:
+
+   - Management of many concurrent interested consumers for one of
+     multiple classes or originating contexts of events
+   - Processing of filter conditions over a class or originating context
+     of events on behalf of consumers.
+   - Transcoding, like encoding in MsgPack after decoding from JSON
+   - Transformation that changes the event's structure, like mapping from
+     a proprietary format to Cloud Events, while preserving the
+     identity and semantic integrity of the event.
```
ultrasaurus: This seems to overlap with rendering in #1 above

spec.md
```diff
@@ -49,6 +49,119 @@ The following will not be part of the specification:
...
+     by the consumer ("pull"), or initiated by the middleware ("push")
+     after a delay.
+   - Observing event content or event flow for monitoring or
+     diagnostics purposes.
+
+   To satisfy these needs, middleware will be interested in:
+   - A metadata discriminator usable for classification of events so that
+     consumers can express interest in one or multiple such classes.
+   - A metadata discriminator that allows distinguishing the subject of
+     a particular event of that class.
```
ultrasaurus: what does "subject" mean in this context?

spec.md
```diff
@@ -49,6 +49,119 @@ The following will not be part of the specification:
...
+   - An indicator for the encoding of the event and its data.
```
ultrasaurus: Is a concern of the event or of the transport from producer to middleware and from middleware to consumer?

spec.md
```diff
@@ -49,6 +49,119 @@ The following will not be part of the specification:
...
+   - An indicator for the structural layout (schema) for the event and
+     its data.
+
+   Whether its events are available for consumption via a middleware is
+   a delegation choice of the producer.
```
ultrasaurus: If middleware consumes the event, isn't it a consumer then?

spec.md
```diff
@@ -49,6 +49,119 @@ The following will not be part of the specification:
...
+
+4) Frameworks and other abstractions make interactions with event platform
+   infrastructure simpler, and often provide common API surface areas
+   for multiple event platform infrastructures.
+
+   Frameworks are often used for turning events into an object graph,
+   and to dispatch the event to some specific handling user-code or
+   user-rule that permits the consuming application to react to
+   a particular kind of occurrence in the originating context and
+   on a particular subject.
```
ultrasaurus: nit: user => application developer

spec.md
```diff
@@ -49,6 +49,119 @@ The following will not be part of the specification:
...
+
+   Frameworks are most interested in semantic metadata commonality
+   across the platforms they abstract.
```
ultrasaurus: Would be good to include what it is about the metadata commonality that is interesting to a framework

spec.md
```diff
@@ -49,6 +49,119 @@ The following will not be part of the specification:
...
 * Selecting a single identity/access control system
 
+## Usage Scenarios
+
+The list below enumerates key usage scenarios and developer perspectives
+that have been considered for the development of this specification.
+
+In these scenarios, we keep the roles of event producer and event consumer
+distinct. A single application context can always take on multiple roles
+concurrently, including being both a producer and a consumer of events.
```
ultrasaurus: In this context, we use the term "application" to be inclusive of custom applications or platforms that host applications.

spec.md
ultrasaurus: we could change the point two bullets above ("identifying and selecting the origin context or the producer-assigned classification") to say:

"identifying and selecting the origin context or the producer-assigned classification, potentially correlating multiple events"

duglin: Agreed to add this. Cathy is not asking for a new attribute at this time, she just wants to make sure the spec doesn't preclude the usecase

spec.md
```diff
@@ -49,6 +49,119 @@ The following will not be part of the specification:
...
+
+3) Middleware routes events from producers to consumers, or onwards
+   to other middleware. Applications producing events might delegate
+   certain tasks arising from their consumers' requirements to
+   middleware:
+
+   - Management of many concurrent interested consumers for one of
+     multiple classes or originating contexts of events
+   - Processing of filter conditions over a class or originating context
+     of events on behalf of consumers.
```
ultrasaurus: From today's conversation, I think adding this clarification to the above point could address question raised by @cathyhongzhang : Filtering does not change the content of the event, rather selects particular events to forward to a consumer.

spec.md
ultrasaurus: in today's calls @clemensv clarified that this might be a specific file that is being acted on, which led to the event being generated.  I think this needs just a little more explanation.

spec.md
ultrasaurus: Propose additional text to clarify (based on today's discussion):
In the list below, a specific platform or application could take on multiple roles.

spec.md
```diff
@@ -49,6 +49,119 @@ The following will not be part of the specification:
...
+   - Instant "push-style" delivery to interested consumers.
+   - Storing events for eventual delivery, either for pick-up initiated
+     by the consumer ("pull"), or initiated by the middleware ("push")
+     after a delay.
+   - Observing event content or event flow for monitoring or
+     diagnostics purposes.
+
+   To satisfy these needs, middleware will be interested in:
+   - A metadata discriminator usable for classification of events so that
+     consumers can express interest in one or multiple such classes.
```
ultrasaurus: Is the intent here to suggest that there be a single discriminator for this purpose?  I think it might move us forward more quickly to not be overly specific at this point.  Maybe just change line 133 to say:

To satisfy these needs, middleware will be interested in common metadata that allows:
- classification of events so that  consumers can express interest in one or multiple such classes.
- distinguishing the subject of a particular event of that class.

etc.

spec.md
```diff
@@ -49,6 +49,119 @@ The following will not be part of the specification:
...
+     identity and semantic integrity of the event.
+   - Instant "push-style" delivery to interested consumers.
+   - Storing events for eventual delivery, either for pick-up initiated
+     by the consumer ("pull"), or initiated by the middleware ("push")
+     after a delay.
+   - Observing event content or event flow for monitoring or
+     diagnostics purposes.
+
+   To satisfy these needs, middleware will be interested in:
+   - A metadata discriminator usable for classification of events so that
```
lfourie: Looks like you have several levels of discriminator for classification and filtering. This first one seems to map to the source-type, eg, s3 storage, second one maps to source-id, eg bucket within s3 storage.

spec.md
lfourie: As discussed in the meeting this appears to map to namespace, eg, the event encoding and data is to be interpreted within the context of a specific namespace value, eg, azure, aws, oracle

spec.md
```diff
@@ -49,6 +49,119 @@ The following will not be part of the specification:
...
+     for instance reading a storage blob after having been informed
+     that this blob has just been created.
+
+   Consumer interests motivate requirements for which information
+   producers ought to include an event.
+
+3) Middleware routes events from producers to consumers, or onwards
+   to other middleware. Applications producing events might delegate
+   certain tasks arising from their consumers' requirements to
+   middleware:
```
lfourie: Is the event emitted by the middleware identical to the received event apart from possible event _payload_ transformation? If not there is a need to be able to correlate events received by middleware with the events emitted by middleware.

spec.md
deissnerk: @ultrasaurus Good point. To me _consuming an event_ means that some action is triggered by it and that it is not forwarded. Of course some new event can be emitted as part of the action. The text above mentions consumption **via** a middleware. In that case the middleware provides the  means to consume the event but is not the consumer.

spec.md
deissnerk: @inlined I like the idea of more examples, too. There is still #53 open. A starter doc could contain exactly the desired examples. I wrote a comment there a few days ago regarding patterns. This gateway that renders events as CloudEvents could be one of them. Mapping the context attributes to a topic hierarchy to propagate events using a message broker could be another one.

spec.md
cathyhongzhang: I think when middleware routes the event, it should include both the original context attributes and payload since the consumer has interest in the context attributes too. For example, some context attribute is needed to identify/extract a field in the payload.

spec.md
cathyhongzhang: Agree with Sara. I think we need to define/explain what is meant by the subject. We may also need to define/explain classification of events. Classification can be based on different event source type (eg. storage, message notification etc.) or based on different cloud event vendor. What is this classification?

spec.md
cathyhongzhang: Agree with both comments here. For example, a serverless platform will consume the event and trigger a function execution and pass filtered event data to the function. Is the serverless platform the middleware or the consumer or both?

spec.md
cathyhongzhang: Yes, I would like to makes sure "correlating event instances from multiple event producers" is a key interest point of the consuming application and the event data (metadata+payload) should carry "correlation-token" information. As to how to identify this correlation-token in the event data, there could be multiple ways of doing it depending on where/how the event producer put this info.  For example, this correlation-token could be part of the origin context or could be a unique string in the event payload or could be an explicitly defined metadata field of the event.

spec.md
deissnerk: I was referring to _without changing_. Do we exclude that there might be context attributes that can be changed, removed or added by the middleware?

spec.md
clemensv: the second block would have to be 5. and 6. and that seems to throw some markdown tooling off.

spec.md
clemensv: "Without changing" would be contradicted by the details of the section, because transcoding and transforming is doubtlessly a change. The transform description has a clause on it that semantic changes are not allowed, and that includes omissions. 

On @deissnerk's point: I'd like to rename the "extensions" section we currently have in the spec to "annotations" and clarify that this section MAY be edited by intermediaries but MUST be forwarded. That would align with the notion that the AMQP TC arrived at, where the "bare message" which holds the content and core properties is immutable, but can be annotated in transit.

spec.md
clemensv: I don't see the overlap/conflict

spec.md
clemensv: No. Even if we were to decide that Cloud Event events were always self-contained JSON, the event data itself might quite well still be a base64 encoded binary holding Protobuf encoded data and the middleware may want to transcode that into Avro. That's not a transport concern.

spec.md
clemensv: @deissnerk @ultrasaurus **via** is indeed the keyword here, as @deissnerk explains

spec.md
clemensv: I'm telling an app to specifically show the MSFT, GOOG, and IBM stock ticks for me, it's arguable that the subscription rule has been user supplied. This section is about frameworks that deal with dispatching events and therefore I chose this to be a bit broader than just application developers.

spec.md
clemensv: "so that similar activities can be handled uniformly."

spec.md
clemensv: We chose to eliminate "platform" related to events in this discussion. The only place where I keep "platform" is in the Frameworks section where it refers to "event platform infrastructures" that frameworks aim to provide commonality over. That should be sufficient for framework folks to know what this means.

spec.md
clemensv: None of the activities change the event semantically, and the one activity that touches the event structure is specifically stating that semantics may not be changed. A clarification for filtering specifically seems superfluous to me.

spec.md
```diff
@@ -49,6 +49,119 @@ The following will not be part of the specification:
...
+
+   The produced events might be rendered and emitted directly by the producer
+   or by an intermediary; as example for the latter, consider event data
+   transmitted by a device over payload-size-constrained networks such as
+   LoRaWAN or ModBus, and where events compliant to this
+   specification will be rendered by a network gateway on behalf of the
+   producer.
+
+2) Applications consume events originating from devices and/or people
+   and/or applications/-modules/-instances for display, archival, analytics,
```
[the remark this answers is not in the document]
clemensv: I'm just using "applications, application-modules, and application-instances" (what the above expands to in common usage) to catch most concepts people may have in mind for where the event may come from. To some "application" is a deployed instance on a device, to some it's any deployment anywhere, some cares about the specific piece of an app that emits and event and I'm just trying make that all the same here. 

Being nitty in return, "OSS" refers to a collaboration model when building software ("OSS" doesn't define a qualitative difference between Apache Kafka and Azure Event Hubs once you have deployable bits) and "SaaS" refers to a delivery model for a product category, so neither of them fit as examples in the respective places, IMO.

spec.md
```diff
@@ -49,6 +49,119 @@ The following will not be part of the specification:
...
+   - Observing event content or event flow for monitoring or
+     diagnostics purposes.
+
+   To satisfy these needs, middleware will be interested in:
+   - A metadata discriminator usable for classification of events so that
+     consumers can express interest in one or multiple such classes.
+   - A metadata discriminator that allows distinguishing the subject of
+     a particular event of that class.
+   - An indicator for the encoding of the event and its data.
+   - An indicator for the structural layout (schema) for the event and
```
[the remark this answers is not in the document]
clemensv: We're effectively defining a one-way flow here, so the middleware only gets to operate on what it gets.

spec.md
clemensv: this isn't meant to jump to conclusions but to call out what the middleware needs.

spec.md
clemensv: the first maps to "topic" and the second to "subject" in https://github.com/cloudevents/spec/pull/95 as further elaborated on in https://github.com/cloudevents/spec/issues/112

spec.md
clemensv: No. The encoding maps to the specification's "content-type" field.

spec.md
clemensv: It's assumed that the middleware doesn't make semantic changes and as I explain [here](https://github.com/cloudevents/spec/pull/117#discussion_r176042677) I would be in favor declaring the event metadata immutable except for an annotation section.

spec.md
duglin: I agree that the text in this PR is very helpful, and on one of our calls this week I asked whether this should live in the spec or in a companion doc because its non-normative text. For now, and to keep us moving forward, I think putting this into the spec is fine and we can do follow-on PRs to extract it if we think it would be better to live in another doc (one in which we could really elaborate on things with more examples, and not bloat the spec).

To me the biggest goal of this PR is to get alignment on our thinking first , which I think we're really close on - we can leave word-for-word wordsmith-ing and document(s) organization for later.

duglin:
I believe that @clemensv has addressed all of the comments on this PR.

I'd like to ask that people review this PR for conceptual correctness in prep for tomorrow's call. Is this more right than wrong? Does it move us forward in the right direction? If so, then please LGTM it. That does not mean we can not make editorial tweaks and additions in follow-on PRs. The goal here is to make incremental progress, not achieve perfection on the first try.

LGTM

markpeek:
Agreed, the content look really good. LGTM to commit and we could tweak with subsequent PR's.

spec.md
ac360: > "Without changing" would be contradicted by the details of the section, because transcoding and transforming is doubtlessly a change. The transform description has a clause on it that semantic changes are not allowed, and that includes omissions.

+1 

> On @deissnerk's point: I'd like to rename the "extensions" section we currently have in the spec to "annotations" and clarify that this section MAY be edited by intermediaries but MUST be forwarded. That would align with the notion that the AMQP TC arrived at, where the "bare message" which holds the content and core properties is immutable, but can be annotated in transit.

@clemensv Some brief context: We chose `extensions` because it is term OpenAPI uses to describe custom attributes to its specification and we felt having some similarities between CloudEvents and OpenAPI would be helpful.  That said, I like your suggestion of having that section be editable.  It means that section will take on a new purpose, in addition to merely being a place for custom attributes.  I believe having a section containing editable attributes may solve a handful of problems.

ac360:
LGTM.  Nice work @clemensv, @ultrasaurus, @duglin, @cathyhongzhang, @rachelmyers and others! There are some helpful suggestions in the comments.  However, the PR in its current state offers enough clarity that I'm in favor of merging it now to help improve our focus, and optimizing it later.

spec.md
alexdebrie: Is it true that middleware should *never* change the content of an event? A few examples I could think of:

- Scrubbing PII from an event before forwarding to downstream consumers;
- Enriching an event, such as hydrating an object on the event. Imagine a producer includes a `productId` in the event payload, then the middleware hydrates the `product` object based on that id to include product name, price, inventory, etc.

In both of these situations, I would still consider the original producer the `source` of the event rather than the middleware, as suggested by [this thread](https://github.com/cloudevents/spec/pull/117/files#r175473795).

spec.md
ultrasaurus: no conflict, just would be more readable if it was only in one place.  I would move the bit about "rendering" from producer section to here  (since transcoding seems like what you were describing as "rendering" to me, yet the language above speaks to a somewhat different use case)

spec.md
ultrasaurus: My point is that the producer determines how it will communicate about its events.  There seemed to be agreement when we discussed this.  I would like the text here to make clear that an operator can connect producer and middleware, not that a producer somehow specifies it's middleware -- I wouldn't prevent that in the specification, but this language seems to imply that we always encode middleware configuration in the producer, which I don't think is your intent.

clemensv: The "producer" is a piece of software and as such it is written, configured, and deployed. The operator may be changing configuration. I think those are IT ground rules that need no elaboration.

spec.md
lfourie: @alexdebrie middleware may modify (transcode/transform) the content/payload of the event as mentioned in the paragraphs below. The question is whether middleware can modify the common event metadata attributes?

alexdebrie: @lfourie Yea, I'm wondering if it's ever appropriate for middleware to modify the event payload before passing along. Example:

**Removing personally identifiable information (PII) from payload**

Perhaps my Users service emits a `user.created` event which includes the social security number of the newly created user. I want to configure my middleware to scrub that attribute before sending to downstream consumers.

From:

```json
{
  "firstName": "Alex",
  "lastName": "DeBrie",
  "dateOfBirth": "01-01-1980",
  "ssn": "555-55-5555",
}
```

to: 

```json
{
  "firstName": "Alex",
  "lastName": "DeBrie",
  "dateOfBirth": "01-01-1980",
}
```

In my mind, this is still a `user.created` event that's been emitted from my Users service rather than an entirely new event emitted by the middleware as suggested by @ultrasaurus's [object.created example](https://github.com/cloudevents/spec/pull/117#discussion_r175625483).

clemensv: @alexdebrie  IMO, if you're removing (or adding) anything, including PII, you're making a semantic change. A scrubbed event is no longer the original but rather originates from whoever does the scrubbing (or enriching); it is a derivative. A very basic test might be whether you can compute the same integrity hash over the payload values (not metadata or bits on wire) before and after transcoding or transform.

ESPECIALLY if you have PII involved, you'll want to have a clear break between the event data before and after the scrub, because you will want to have absolutely no doubt in any audit trails about what kind of data you're dealing with when the GDPR police shows up. 

All that does not at all mean that such an intermediary weren't allowed to publish the derived event as to relate to the *same context*. Your event doesn't originate from the "users" service, but it still relates to exact same user context. 

The event will have a different id (to delineate it from the unscrubbed original) and if you want to include provenance details in the event detail data, you might have a trail there that gives the id of the original event and a hash of that original event. 

This is actually brilliant example for why "source" isn't the right concept in the current spec, because your scenario calls for an original and a derivative event to be posted related to the exact same context by different actors for different segments of a system.

spec.md
```diff
@@ -49,6 +49,135 @@ The following will not be part of the specification:
...
+   The producer application could run anywhere, such as on a server or a device.
+
+   The produced events might be rendered and emitted directly by the producer
+   or by an intermediary; as example for the latter, consider event data
+   transmitted by a device over payload-size-constrained networks such as
+   LoRaWAN or ModBus, and where events compliant to this
+   specification will be rendered by a network gateway on behalf of the
+   producer.
+
+   For example, a weather station transmits a 12 byte event payload indicating
```
shalka: "For example, a weather station transmits a 12 byte event payload indicating weather conditions once every 5 minutes over LoRaWAN. A LoRaWAN gateway might then be used to publish the event to an Internet destination and in the Cloud Events format. The weather station is the producer and the intermediary gateway plays a middleware role (see 3)."

should be

"For example, a weather station transmits a 12-byte, proprietary event payload indicating weather conditions once every 5 minutes over LoRaWAN. A LoRaWAN gateway is then used to publish the event to an Internet destination in the Cloud Events format. The LoRaWAN gateway is the event producer, publishing on behalf of the weather station, and will set event metadata appropriately to reflect the source of the event."

spec.md
```diff
@@ -49,6 +49,135 @@ The following will not be part of the specification:
...
+     to new files ending with ".jpg" (the file name being the "new file"
+     event's subject) for the context describing specific directory
+     inside a file storage account that it has registered interest on.
+   - An indicator for the encoding of the event and its data.
+   - An indicator for the structural layout (schema) for the event and
+     its data.
+
+   Whether its events are available for consumption via a middleware is
+   a delegation choice of the producer.
+
```
rachelmyers: Proposed addition based on conversation during March 23 call: "In practice, Middleware can take on role of a Producer when it changes the semantic meaning of an event, Consumers when it takes action based on an event, or Middleware when it routes events without making semantic changes." cc @duglin

spec.md
```diff
@@ -49,6 +49,135 @@ The following will not be part of the specification:
 * Language-specific runtime APIs
 * Selecting a single identity/access control system
 
+## Usage Scenarios
+
+The list below enumerates key usage scenarios and developer perspectives
+that have been considered for the development of this specification.
```
markfisher: based on March 23 WG call, I propose adding something like:

"These usage scenarios are by no means exhaustive, and the specification does not aim to be prescriptive about usage."

spec.md
```diff
@@ -49,6 +49,135 @@ The following will not be part of the specification:
 * Language-specific runtime APIs
 * Selecting a single identity/access control system
 
+## Usage Scenarios
+
```
rachelmyers: Addition based on March 23 call: ```These scenarios are not normative; anyone is free to create a system that mixes these scenarios. These cases establish a common vocabulary of event Producer, Consumer, Middleware, and Framework.```

duglin:
I added some changes based on today's call. I believe this URL will show the diff: 
https://github.com/cloudevents/spec/compare/d07a1b9fef53d71...clemensv:user-stories

@clemensv looked these over before he vanished on his vacation and approved them.

duglin:
With all of the good work people put into this over the last week (with near daily meetings!), I think it covers the main points people wanted to see. Any additional tweaks can be done via follow-on PRs.

LGTM

markpeek:
@duglin your edits LGTM to consolidate the input from the 2018-02-23 meeting and the above input from @rachelmyers and @markfisher.

My only suggestion (which I could put into a follow-on PR) would be to move the middleware addition to the top. This would be combine:
```In these scenarios, we keep the roles of event producer and event consumer distinct. A single application context can always take on multiple roles concurrently, including being both a producer and a consumer of events.```
and
```In practice, middleware can take on role of a producer when it changes the semantic meaning of an event, a consumer when it takes action based on an event, or middleware when it routes events without making semantic changes.```
and replace at the top with:
```In these scenarios, we keep the roles of event producer, event consumer, and middleware distinct. A single application context can always take on multiple roles concurrently, such as being both a producer and a consumer of events. In the case of middleware, it could take on role of a producer when it changes the semantic meaning of an event, a consumer when it takes action based on an event, or as a more typical middleware when it routes events without making semantic changes.```

duglin:
@markpeek seems like a reasonable follow-on PR to me.

spec.md
cathyhongzhang: Like these additions!

duglin:
Approved on 3/29 call.

PR #118, opened 2018-03-21
add a use case for correlation of multiple events

cathyhongzhang:

### DIFF
--- a/about/use-cases.md
+++ b/about/use-cases.md
@@ -104,3 +104,16 @@ In a lot of cases these messages are binary encoded instead of textual.
 Whether directly from the device or transformed via a gateway, CloudEvents
 would allow for a better description of the origin of the message and the
 format of the data contained within the message.
+
+### Event Correlation
+A serverless application/workflow could be associated with multiple events from 
+different event sources/producers. For example, a burglary detection 
+application/workflow could involve both a motion event and a door/window open event. 
+A serverless platform could receive many instances of each type of events, 
+e.g. it could receive motion events and window open events from different houses. 
+The serverless platform needs to correlate one type of event instance correctly 
+with other types of event instances and map a received event instance to the 
+correct application/workflow instance. CloudEvents will provide a standard way 
+for any event consumer (eg. the serverless platform) to locate the event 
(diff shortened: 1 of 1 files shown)


duglin:
@cathyhongzhang can you rebase?

cathyhongzhang:
How about I do a new PR which might be simpler than rebase?

duglin:
I guess if you really wanted to, but it might be a good learning experience  :-)

If you don't want that much excitement, I can try to do the rebase for you if you want.

inlined:
1. Was it intentional to delete data in `spec.md`? I'm not entirely against this since this aligns closely with my vision that context and data are clearly separate, though that might be a different PR.
2. How would an implementer of your burglary detection system know to correlate a particular window open and motion detection event? Is this likely just that the first event after a period of inactivity starts a new house-wide correlation ID? I'm trying to think about whether this is best implemented in the IoT gateway or the delivery infrastructure.

duglin:
LOL that's funny - I didn't notice the deletion of `data`.  @cathyhongzhang can you fix that?

cathyhongzhang:
@inlined @duglin  I have fixed it and will do the removal of the data from Context Attributes in a separate PR. 

Good question. W.r.t. the burglary detection system, the correlation-token/id could be the house address or any other unique identifier of that house . In order to support a burglary detection system on serverless framework, this correlation-token value must be carried in both the window/door open event and the motion event. Based on the correlation-token, the serverless platform/infrastructure can then correlate the window/door open event instance with the motion event instance that are associated with the same house and send them to the same burglary detection instance for processing. IoT GW could be a good place to add this correlation-token value. For use cases that involve other multiple events, the correlation-token value would be different.  In order to support consistency among event producers, we need to provide a standard way for them to specify the correlation token/id. Hope I answer your question:-)  You may want to take a look at PR #128 .

duglin:
@cathyhongzhang thanks!

cathyhongzhang:
@duglin Np. Yes, let's do it next week. It should be an easy one.

deissnerk:
LGTM

duglin:
Approved on 4/5 call

PR #120, opened 2018-03-23
add video from Clemens

duglin:


sslavic:
LGTM

SeanFeldman:
Question: a link could break. Would it make sense to capture the binary (PPTX) within this repository?

duglin:
@SeanFeldman good point, I forgot that he made the ppt available. I'll grab that and add it to the PR.

duglin:
Done. The deletes are just old files that I renamed to not have spaces in the names.

SeanFeldman:
👍

ultrasaurus:
This PR seems to be deleting `share/2018-02-22 CloudEvents.pdf` -- am I reading that correctly?

duglin:
@ultrasaurus yes - weird, for some reason I thought I added a comment about how those were old files that I forgot to erase when I renamed things to not have spaces in the names. Sorry, the comment must have just been in my head :-)

ultrasaurus:
@duglin I think those were from a presentation from @yaronha before we had an index.  Deleting old content should be in a different PR (and not sure these should be deleted -- would need to go back thru notes to be sure)

duglin:
@ultrasaurus I didn't delete content, I just renamed it and forgot to delete the old named version.

duglin:
Approved on 3/29 call

PR #123, opened 2018-03-23
Source as a URI

duglin:
In an attempt to get to a MVP for the spec, I'm proposing to collapse source-related attributes (source* & namespace) down to a single URI whose format/encoding is event producer defined.

Future PRs can suggest that we split out certain bits of data from this URI (e.g. for performance, optimization or other reasons), but then those discussions will be more focused on just the specific need being addressed, so hopefully will be easier to reach a consensus.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -107,17 +107,6 @@ the event data. The context might also need to be serialized with the event
 data for some use cases (e.g. a JSON implementation might use one JSON object
 that contains both context and data).
 
-### namespace
-* Type: String
-* Description: Identifier that uniquely identifies the organization publishing
-  the event.
-* Constraints:
-  * REQUIRED
-  * MUST be a non-empty string
-* Examples:
-  * kafka.apache.org
-  * com.microsoft.azure
-
 ### event-type
 * Type: String
 * Description: Type of the event `data`. Producers can specify the format of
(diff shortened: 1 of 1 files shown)


yaronha:
@dug LGTM

SeanFeldman:
👍

clemensv:
I just closed #95 in support of this PR. This simplification does align with what we believe we need for addressing our customers' needs. We will have optimization comments subsequent to this PR.

shalka:
Agree with @clemensv .. My only (minor) concern would be that we should type 'source' as a string instead of a URI since we're consolidating namespace into it as well. The format (which is yet to be determined) may or may not take the form of a URI, but we can't make that distinction yet.

duglin:
I thought about that but I recall that when I casually mentioned it as a string on a call a long time ago some folks pushed back, so I picked URI as the starting point.  Would you be ok with discussing a possible re-typing in a follow-on PR?

shalka:
@duglin That's reasonable. I like the consolidation, so long as we agree that a follow-on PR may (or may not) change the format of 'source'.
:shipit:

markpeek:
LGTM

deissnerk:
LGTM
Under the assumption that it is not carved in stone, and that we are free to change this, after we learned from discussions and early implementations.

inlined:
I don't want to let great be the enemy of good. I will ask in the future that we specify some subset of URI to build APIs on top of this (e.g. one that includes an authority) but LGTM.

ultrasaurus:
LGTM

dklyle:
LGTM

duglin:
Approved on 3/29 call.

PR #124, opened 2018-03-24
add json serialization

### LINKED ISSUE #17: Show sample json serialization for each context attribute
I think it would be good to show some sample json serialization for each context attribute so there is no ambiguity around what it would look like. It would also be good to show what a complete "context" object would look like.  These actually might not be "samples" - by that I mean it might be good to normatively say something like "When serialized in JSON, the format of the event MUST adhere to the following...".  We'll see...

duglin:
Closes #17 

I would actually prefer if we just move this into the main spec itself since its so small, but for now I started out with it being in a separate doc with the assumption that we can decide on whether we want to merge it later.

### DIFF
--- a/serialization.md
+++ b/serialization.md
@@ -1,0 +1,48 @@
+# CloudEvents Serialization Profile - Version 0.1
+
+This document specifies how a CloudEvent is to be serialized into certain
+encoding formats. Compliant CloudEvents implementations that support these
+formats MUST adhere to these rules for these formats.
+
+## Table of Contents
+- [Notational Conventions](#notational-conventions)
+- [JSON](#json)
+
+## Notational Conventions
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to
+be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+## JSON
+
+When serialized in [JSON](https://tools.ietf.org/html/rfc7159) a CloudEvent
+MUST adhere to the following format:
+
+```
+{
+  "cloud-events-version": "cloud-events-version value",
+  "event-id": "event-id value",
+  "namespace": "namespace value",
+  "source": {
+    "type": "source-id value",
+    "id": "source-id value"
+  },
+  "event-type": "event-type value",
+  "event-type-version": "event-type value",
+  "schema-url": "schema-url value",
+  "extensions": {
+    ... extensions values ...
+  },
+  "data": ... data value ...
+}
+```
+
+Notes:
+- The use of whitespace is not significant.
+- The order of the properties, at any level, is not significant.
+- CloudEvent producers MAY include additional properties in the JSON
+  but receivers MAY choose to ignore them. However, it is RECOMMENDED that
+  they be placed as children of the `extensions` property. Receivers
+  MUST NOT treat unkonwn additional properties as an error and MUST NOT stop
+  processing of the event as a result of their presence.
--- a/serialization.md
+++ b/serialization.md
@@ -46,3 +46,24 @@ Notes:
   they be placed as children of the `extensions` property. Receivers
   MUST NOT treat unkonwn additional properties as an error and MUST NOT stop
   processing of the event as a result of their presence.
+- Non-mandatory properties are NOT REQUIRED to be included in the JSON, but
+  if they are present then they MUST adhere to the format described.
+
+Example:
+```
+{
+  "cloud-events-version": "0.1",
+  "event-id": "6480da1a-5028-4301-acc3-fbae628207b3",
+  "namespace": "product.example.com",
+  "source": {
+    "type": "repository",
+    "id": "AFN234769347623"
+  },
+  "event-type": "created",
+  "event-type-version": "v1.5",
+  "schema-url": "https://product.example.com/schema/repo-create",
+  "data": {
+    "path": "/JaneDoe/repos/mycode"
+  }
+}
+```
--- a/spec.md
+++ b/spec.md
@@ -26,6 +26,11 @@ Enter CloudEvents, a specification for describing event data in a common way.
 CloudEvents seeks to ease event declaration and delivery across services,
 platforms and beyond.
 
+The [Serialization Profile](serialization.md) specifies how to
+serialize a CloudEvent into certain encoding formats. Compliant CloudEvents
+implementations that support those formats MUST adhere to the encoding rules
+specified in the profile for those formats.
+
 # Design Goals
 
 CloudEvents are typically used in a distributed system to allow for services to


serialization.md
```diff
@@ -0,0 +1,43 @@
...
+
+When serialized in [JSON](https://tools.ietf.org/html/rfc7159) a CloudEvent
+MUST adhere to the following format:
+
+```
+{
+  "cloud-events-version": "cloud-events-version value",
+  "event-id": "event-id value",
+  "namespace": "namespace value",
+  "source": {
```
SeanFeldman: @duglin isn't `source` a URI?

serialization.md
```diff
@@ -0,0 +1,43 @@
...
+  "event-type-version": "event-type value",
+  "schema-url": "schema-url value",
+  "extensions": {
+    ... extensions values ...
+  },
+  "data": ... data value ...
+}
+```
+
+Note:
```
SeanFeldman: nitpicking - sais `Note`, but these are notes.

serialization.md
```diff
@@ -0,0 +1,43 @@
...
+- the use of whitespace is not significant
```
SeanFeldman: Capitalize?

serialization.md
duglin: Not yet :-)

serialization.md
duglin: fixed

serialization.md
duglin: fixed

serialization.md
SeanFeldman: Fair enough. Pending #123 🙂

serialization.md
```diff
@@ -0,0 +1,43 @@
+# CloudEvents Serialization Profile - Version 0.1
+
+This document specified how a CloudEvent is to be serialized into certain
```
markpeek: Should this be applied here? s/specified/specifies/

serialization.md
```diff
@@ -0,0 +1,43 @@
...
+  "event-type-version": "event-type value",
+  "schema-url": "schema-url value",
+  "extensions": {
+    ... extensions values ...
+  },
+  "data": ... data value ...
+}
+```
+
+Notes:
```
markpeek: Should there be a comment about extra fields in the serialized data? Such as:
```
- It is undefined whether extra fields are passed along or removed depending on the usage role and implementation. Those fields are best placed in the extensions map.
```

serialization.md
duglin: fixed the "specified" typo. What did you mean by "should this applied here?" ?

serialization.md
duglin: I expanded on the thought a little - see what you think.

markpeek: Thanks for the update. LGTM.

spec.md
```diff
@@ -26,6 +26,11 @@ Enter CloudEvents, a specification for describing event data in a common way.
 CloudEvents seeks to ease event declaration and delivery across services,
 platforms and beyond.
 
+The [Serialization Profile](serialization.md) specified how to
```
markpeek: This might be better as "specifies". (Or am I being too tense?)

duglin: No, you're right - just a disconnect between the brain and fingers.

serialization.md
```diff
@@ -0,0 +1,48 @@
...
+be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+## JSON
+
+When serialized in [JSON](https://tools.ietf.org/html/rfc7159) a CloudEvent
+MUST adhere to the following format:
+
+```
+{
+  "cloud-events-version": "cloud-events-version value",
```
inlined: I'm curious whether it would be more useful to use valid fake values rather than prose in quotes.

serialization.md
```diff
@@ -0,0 +1,48 @@
...
+    "type": "source-id value",
+    "id": "source-id value"
+  },
+  "event-type": "event-type value",
+  "event-type-version": "event-type value",
+  "schema-url": "schema-url value",
+  "extensions": {
+    ... extensions values ...
+  },
+  "data": ... data value ...
```
inlined: As an FYI: though Google&Firebase started out with a model like this where all context is at the top level with one magic field called "data", we generally considered this to be a mistake. Many languages have better ways to expose context (e.g. Golang's `Context.context`, or JavaScript's variadic parameter system) so we are migrating to a world where data & context are exposed separately/natively our runtimes. Alpha testers have given very positive feedback for this because it allows any existing function(SomeNativeType) to be a CloudFunction<SomeNativeType>.

To match this, the REST JSON is being migrated to a two-field JSON object with `{"context", "data"}`

serialization.md
duglin: I struggled with this because if we put fake values in there then we need something to say "property X from the spec goes into json property 'x'".  While in many cases it should be really really obvious, I felt that a spec should be more explicit about it.  If all we cared about was JSON I actually would have proposed that we have the serialization of each property next to its definition in the spec itself. But I didn't want to assume everyone would be ok with JSON getting special treatment.

serialization.md
duglin: I could live with that. What do other people think?

SeanFeldman: @inlined would `extensions` be under `context` or another sibling at the root level?

duglin: @inlined where would "could-events-version" go? I can see it being useful as a top-level thing since we may choose to rename `data` or `context` in a future version, but I can also see a need for that info to be passed downstream so people know how to interpret the `context` attributes based on our spec's version number.

I do want to explore this option, it feels cleaner - would make for a great follow-on PR

duglin:
Added an example and one line to make it clear that optional fields are still optional, but if present then they MUST adhere to the rules defined.  I think it was implied before but now its explicitly stated.

serialization.md
duglin: @inlined I added an example to try to address your concern. See if that helps.

serialization.md
deissnerk: @duglin I agree with @inlined that context and data should be on the same level as he suggested. 

One more thing:
Do we need rules how to encode data, if the content-type is something binary or text/xml?

duglin: would that be something that the definition of the content-type (e.g. json spec) would need to define?

deissnerk: afaik json spec does not cover this explicitly. Most people seem to base64 encode data.

duglin: yea, I did a quick search too and didn't find a lot. I guess we could offer some recommendations...

inlined: Honestly I don't have the practical experience with `content-type` and `cloud-events-version` that I do with the other fields. Here's my first SWAG:
* In my opinion, `content-type` is/should be meta to the whole spec (why are we dictating the content type within a payload that's already explicitly JSON encoded?). This should be dictated in the _transport_ layer, not the encoding layer.
* `cloud-events-version` looks cleanest at the top level in JSON though if it's going to show up in the programming model it might be easier to put it in `context`.
* I have the strong opinion (weakly held) that `extensions` should be in `context`. The developer is likely to depend on these extensions and they should be easily accessible.

ultrasaurus:
This will have to be modified to be consistent with changed accepted today

serialization.md
deissnerk: @inlined My understanding is that this PR is about how to package an arbitrary cloudevent into json. I was just wondering what happens if the content-type (the context attribute) is text/xml or image/png. The result is supposed to be a valid json object. So data would need to be encoded somehow. I don't see how this can be delegated to another layer. For events that use json for their data section, there is of course no need for encoding.

duglin:
Approved on 3/29 call

PR #126, opened 2018-03-26
First pass at an 'extension attributes' doc

duglin:
There is a strong desire by the WG to keep the list of attributes we include in the specification down to the minimum necessary to achieve our goals. However, there have been some attributes proposed that seem to be popular in some usecases, but just not enough usecases to cross the threshold of being included in the spec as formal attributes. We've talked about possibly having a spot to put those attributes so they can be used as extensions - then at least there's the possibility of some amount of interop around them even if they have no official standing from our WG.

To that end, this PR proposes a starting doc to host some extensions that people can use.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,4 +5,4 @@ verify:
 	@# Use "-x" if you want to skip exernal links
 	@tools/verify-links.sh -v .
 	@echo Running the RFC2119 keyword checker:
-	@tools/verify-phrases.sh -v spec.md
+	@tools/verify-phrases.sh -v spec.md extensions.md
--- a/extensions.md
+++ b/extensions.md
@@ -1,0 +1,23 @@
+# CloudEvents Extension Attributes
+
+The [CloudEvents specification](spec.md) defines a set of metadata attributes
+than can be used when tranforming a generic event into a CloudEvent.
+The list of attributes specified in that document represent the minimal set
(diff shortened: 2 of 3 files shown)


alexdebrie:
@duglin I'd love to see an endorsed `authorization` or `identity` extension in the spec.

Use case is that a piece of middleware might accept a CloudEvent only if the publisher is authenticatd in some way, such as an HTTP Authorization header. The middleware might not want to pass the header along but may want to pass some context around the user or publisher that sent the event.

Could be a very useful area to have some standardization.

inlined:
@alexdebrie I had proposed something similar in the past, though I retracted it until I could have some time to finalize what we would use within Google. Would definitely love to chat if you have opinions/desires for what these extensions would contain.

inlined:
This LGTM. I'd like some additional ideas about how an extension should be graduated into a normal property. Ideally an Event Source isn't required to double-publish the data as an extension and normal property. Maybe they just support both but decide which field to use based on the `cloud-events-version` attribute? That would imply breaking changes though.

alexdebrie:
Thanks, @duglin. Sounds like @inlined and I will work together to get an auth-related extension proposal put together. 😄

Also, this PR LGTM.

duglin:
Approved on 3/29 call

PR #16, opened 2018-01-23
WIP: include more async participation

ultrasaurus:
- proposed member list (link goes nowhere, will create separately if others like this idea)
- slight increase in commitment by active members
    - members identify themselves, affiliation & interest
    - members need to comment on a PR in advance of the meeting to be eligible to vote
- also added something to suggest that a minority view could strengthen their point with an implementation

Signed-off-by: Sarah Allen <sarahallen@google.com>

PR #127, opened 2018-03-26
idea for more async process to improve WG meetings

ultrasaurus:
factoring out the key point from https://github.com/cloudevents/spec/pull/16/files 
with the goal of fostering more async communication by:
1. prioritizing PRs with most support by active members (which I hope/expect will lead to more active review and commenting in advance of meetings)
2. require more evidence (via code sample) that a change is required if there are minority voices who disagree with a change

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -3,16 +3,31 @@
 This document describes the governance process under which the Serverless
 Working Group (WG) will manage this repository.
 
-## Working Group Meetings
+## Working Group Process
 
 In order to provide equitable rights to all Working Group members,
 the following process will be followed:
 
 * Official WG meetings will be announced at least a week in advance.
-* Proposed changes to any document will be done via a Pull Request (PR).
-* PRs will be reviewed during official WG meetings.
-* During meetings, priority will be given to PRs that appear to be ready for
-  a vote over those that appear to require discussions.
+* Proposed changes to any document will be done via a Pull Request (PR)
+* PRs will be reviewed in advance of WG meetings
+* Active members will vote on PRs in advance of Working Group meetings to
+  promote them to active consideration
+* Changes to the [spec.md](spec.md) require active member participation:
+  * if there are no negative votes, then any PR with 3 affirmative votes can be
+    raised for consideration
+  * if there are negative votes, then 5 affirmative votes and a code sample
+  illustrating the benefit of the change or drawback of current spec which would
+  be addressed by the change
+* Changes outside of the spec require only a single affirmative vote from
+  another active member for consideration
+* Working Group will discuss PRs under consideration at Working Group meetings
+  in advance of merging any PR (see details below)
+
+
+## Working Group Meetings
+* During meetings, priority will be given to PRs with the most votes from
+  active members
 * PRs should not be merged if they have had substantial changes made within
   two days of the meeting.
   Rebases, typo fixes, etc. do not count as substantial.
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -11,8 +11,8 @@ the following process will be followed:
 * Official WG meetings will be announced at least a week in advance.
 * Proposed changes to any document will be done via a Pull Request (PR)
 * PRs will be reviewed in advance of WG meetings
-* Active members will vote on PRs in advance of Working Group meetings to
-  promote them to active consideration
+* [Active Members](#active-members) will vote on PRs in advance of Working Group
+  meetings to promote them to active consideration
 * Changes to the [spec.md](spec.md) require active member participation:
   * if there are no negative votes, then any PR with 3 affirmative votes can be
     raised for consideration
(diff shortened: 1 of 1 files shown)


GOVERNANCE.md
```diff
@@ -3,16 +3,31 @@
...
 
 In order to provide equitable rights to all Working Group members,
 the following process will be followed:
 
 * Official WG meetings will be announced at least a week in advance.
-* Proposed changes to any document will be done via a Pull Request (PR).
-* PRs will be reviewed during official WG meetings.
-* During meetings, priority will be given to PRs that appear to be ready for
-  a vote over those that appear to require discussions.
+* Proposed changes to any document will be done via a Pull Request (PR)
```
SeanFeldman: via a Pull Request**s** (PR**s**)

SeanFeldman: ~~a~~

duglin:
@ultrasaurus can you elaborate on the problem you're trying to solve?

duglin:
Can we close #16?

inlined:
A lot of times our weekly meetings feel like group screen reading sessions. I like that this incentivizes reviewing PRs earlier and can save more time in the meeting for identifying problems/theorizing solutions that will be proposed in the next meeting. LGTM (if a vote from another employee at the same company counts)

GOVERNANCE.md
```diff
@@ -3,16 +3,31 @@
...
+* PRs will be reviewed in advance of WG meetings
```
duglin: This feels a bit odd to say since we can't really enforce it.

GOVERNANCE.md
```diff
@@ -3,16 +3,31 @@
...
+* Active members will vote on PRs in advance of Working Group meetings to
+  promote them to active consideration
```
duglin: I don't know what "Active members" mean since we don't have levels of membership.

GOVERNANCE.md
```diff
@@ -3,16 +3,31 @@
...
+* Changes to the [spec.md](spec.md) require active member participation:
+  * if there are no negative votes, then any PR with 3 affirmative votes can be
+    raised for consideration
```
duglin: While I suspect this line is there to force people to do something before the meetings, it feels more like it would have a bad negative effect because it means I can't schedule an PR for discussion that I'm pretty sure will be accepted if no one bothers to LGTM it.  I think I (or anyone else who wants to add a topic to the agenda) should have that freedom to understand the dynamics and sentiment of the group w.r.t. certain topics.

As an example, the logging level attribute PR had pretty much no comments yet I could guess that we'd come to a pretty quick agreement on what to do with it based on previous conversations. I think scheduling that one for our quick chat was the right thing to do - but this would have prevented me from doing that.

GOVERNANCE.md
```diff
@@ -3,16 +3,31 @@
...
+  * if there are negative votes, then 5 affirmative votes and a code sample
+  illustrating the benefit of the change or drawback of current spec which would
```
duglin: I'm not sure why we should force a higher bar (ie. code) for controversial PRs than non-controversial ones. My feeling is that seeing running code won't actually change anyone's mind since anyone can code anything. I think people should discuss usecases and requirements to explain why a change is needed - but I think that should be true for all PRs.

GOVERNANCE.md
ultrasaurus: sure we can... right now our governance model is moving things forward by default when there isn't clear support from multiple members of the group

GOVERNANCE.md
ultrasaurus: Yes we do.  We have members with [voting rights](https://docs.google.com/spreadsheets/d/1bw5s9sC2ggYyAiGJHEk7xm-q2KG6jyrfBy69ifkdmt0/edit?pli=1#gid=0)

GOVERNANCE.md
ultrasaurus: The goal is to create momentum for interoperability.  The current undocumented process of scheduling unobjectionable PRs, tends to prioritize optional attributes, which does nothing to moving us toward a 0.1 specification

GOVERNANCE.md
ultrasaurus: This is an attempt to move us toward working code samples that illustrate a point, rather than taking action based on the loudest voices

GOVERNANCE.md
duglin: Well, to be fair,I think that's only because most of our meaty attributes were blocked by several weeks of "what is source?". That didn't leave us with a lot of other things to discuss. Hopefully that will be behind us soon.

ultrasaurus: I believe that "what is source?" discussion would have been accelerated by this process.  We could have focused discussion on proposals where WG members were aligned -- encouraging people to address issues from minority dissenters in the PR in order to turn a "no" vote into a "yes" and focus discussion on proposals that had alignment.  In that case, I think the dissent was all about the lack of clarity in definition and this process would have moved us forward quicker. 

If no PRs are ready, then we could use WG meeting for discussion of issues to create greater alignment and accelerate forward process, rather than using the time to discuss lower priority issues.

ultrasaurus:
I've refactored the "voting" section to create clear "active member" terminology.  I think there should be a very high bar for overruling a minority perspective, and stronger encouragement to have positive discussions that drive alignment in written comments and include clear language in proposed text.  

There may be many factors that influence behavior, yet it may be that our governance model is encouraging people to show up and listen (in order to have voting rights), rather than actively reviewing PRs by sharing their thoughts in writing in advance of a meeting.  Note that a recent WG meeting had 35 attended, whereas, there are just a few people who are very active in proposing and reviewing PRs.

GOVERNANCE.md
leecalcote: @ultrasaurus that's true, however, mandating PRs are reviewed prior to the call can be difficult to enforce in terms of determining how many reviewers will have to have chimed-in on a PR in advance for it to be discussed and in terms of the day that we reach a nominal backlog (only have a couple PRs open to review). 

We can encourage pre-review and optimize our time spent on those first.

ultrasaurus: Right now, PRs which have had little to no discussion are prioritized, which seems to be an unwritten rule which IMHO is not optimal.  Discussion is not always a sign of disagreement.

duglin:
Yup. Since we do not have an approved list of "maintainers" that would allow us to merge PRs outside of the weekly calls we are forced to do it during the calls. As a result I give priority to those that appear to have consensus and I think are easy to resolve (merge or close). But it is a guessing game based on what I see happening in the PRs.

However, once we get past those (which is usually pretty quick), we then have PRs on the agenda that require some discussion - which is exactly what you're asking for us to do. So, net, I do think we're already doing what you want. If the order of the "PRs that need more discussion" in the agenda doesn't have what people consider to be the more important ones first then people should suggest an alternative ordering - my ordering is just a guess and I could be reading the group's sentiments incorrectly.

I will add though, I think the current process tries to force off-line chats which, IMO, is the faster and better way to get agreement. Trying to fit more than one of those discussions into our weekly calls will be a challenge and will most likely end up with me asking people to take it off-line anyway.

I've added this to the f2f agenda - I think its a good topic for that.

(rebase needed)

duglin:
What's the status of this one? Do we still need this?

duglin:
If I don't see any movement on this one by this week's call I'm going to support we close it. It can always be reopened later if needed.

duglin:
Agreed to close on the 11/29 call

Issue #24, opened 2018-01-25
correlation-id context attribute


lfourie:
This is a proposed metadata attribute (from Context Attribute Backlog):

correlation-id
    Type: String
    Description: Correlation ID of event that triggered this event to be created (if any). If not triggered in response to an event, a unique ID. Can be specified by the producer if event not produced in response to an event. Enables traceability through systems.
    Examples:
        Event A generated with unique correlation-id. Event handler receives event and generates a new event in response, Event B. Event B copies the correlation-id of Event A so that the chain of events can be properly correlated.
    Constraints:
        Optional

cathyhongzhang:
I can take ownership of this issue.

duglin:
See: https://github.com/cloudevents/spec/pull/128

rektide:
I feel like this should be closed, since #227 was merged. I agree with [jroper's comments there](https://github.com/cloudevents/spec/pull/227#issuecomment-395612199),

>  I don't think CloudEvents should support (whether it's an extension or part of the top level spec) two different mechanisms for attaching trace metadata, that would just be confusing and defeat the purpose because if one party used one, and another used another, then the tracing would not work.

The way correlation-context works in Distributed Tracing is that it is a property-bag, so I believe a `caused-by=.....` property would allow this work to also cover the case of #25 as well.

duglin:
Based on our previous discussions, I think we can close this one because it appears that there will be many different ways to do correlation and at this time there doesn't seem to be enough momentary around just one mechanism (or property) to justify adding it to the core spec.  We discussed how specific producers can add (via extensions) their own correlation properties to the events though.

I'll add this to the list of possible issues to close for the next call - if you have concerns about this please add a comment.

PR #128, opened 2018-03-27
Add property-bucket to correlate multiple events associated with a serverless application

cathyhongzhang:
A path string pointing to a field in the event data that can be used to correlate all the events associated with the same application instance. Different event types or different applications can have different fields in the event data/message as the correlation token. The correlation token/identifier could sit in any location of the event data. 

Closes: #24

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -212,8 +212,8 @@ that contains both context and data).
 ### property-bucket
 * Type: <Key Value>
 * Description: A list of “key : value” pairs in flat strcuture that specify the properties of the event source.
-  The value could be a string or a JSON object. One example usage of the property-field is to
-  correlate multiple events involved with a serverless application.  
+  The value could be a string or a JSON object. There should be no duplication of field in this property-bucket. 
+  One example usage of the property-field is to correlate multiple events involved with a serverless application.  
 * Constraints:
   * MANDATORY
   * MUST be in the format of <key value> pair. 
--- a/spec.md
+++ b/spec.md
@@ -209,17 +209,23 @@ that contains both context and data).
     [RFC 2046](https://tools.ietf.org/html/rfc2046)
 * For Media Type examples see [IANA Media Types](http://www.iana.org/assignments/media-types/media-types.xhtml)
 
-### correlation-field
-* Type: String
-* Description: A path string pointing to a field in the event data 
-  that can be used to correlate all the events associated with 
-  the same application instance. 
+### property-bucket
+* Type: <Key Value>
+* Description: A list of “key : value” pairs in flat strcuture that specify the properties of the event source.
+  The value could be a string or a JSON object. One example usage of the property-field is to
+  correlate multiple events involved with a serverless application.  
 * Constraints:
-  * OPTIONAL
-  * If present, MUST be a non-empty string
+  * MANDATORY
+  * MUST be in the format of <key value> pair. 
 * Examples:
-  * headers.body.cathy-house-number
-
+  property-bucket
+  { 
+     "building address": "12 Main Street, Copenhagen, Denmark, 123456"
+     "floor": "2"
+     "apartmentId": "345"
+     "sensor-location": "window"
+  }
+  
 ### extensions
 * Type: Map <String, Object>
 * Description: This is for additional metadata and this does not have a
--- a/spec.md
+++ b/spec.md
@@ -209,36 +209,25 @@ that contains both context and data).
     [RFC 2046](https://tools.ietf.org/html/rfc2046)
 * For Media Type examples see [IANA Media Types](http://www.iana.org/assignments/media-types/media-types.xhtml)
 
-### properties
+### application-properties
 * Type: <Key Value>
-* Description: A list of “key : value” pairs in flat structure that specify the properties of the event source.
-  The value could be a string or a JSON object. There should be no duplication of field in this property-bucket. 
-  One example usage of the property-field is to correlate multiple events involved with a serverless application.  
+* Description: A list of “key : value” pairs providing information about the event, which are 
+  used by an event-based application. The “key : value” pairs should be in flat structure.
+  The value should be a string. There should be no duplication of keys.
+  One usage example: an application workflow can use a "key:value" to correlate multiple events 
+  (from different event sources) asscociated with a serverless application workflow.  
 * Constraints:
   * MANDATORY
   * MUST be in the format of <key value> pair. 
 * Examples:
-  properties
+  application-properties
   { 
      "building address": "12 Main Street, Copenhagen, Denmark, 123456"
      "floor": "2"
      "apartmentId": "345"
-     "sensor-location": "window"
+     "sensor-location": "room1.window2"
   }
   
-### extensions
-* Type: Map <String, Object>
-* Description: This is for additional metadata and this does not have a
-  mandated structure. This enables a place for custom fields a producer or
-  middleware might want to include and provides a place to test metadata before
-  adding them to the CloudEvents specification. TBD - Determine a shorter
-  prefix for this (e.g. OpenAPI uses “x-”)
-* Constraints:
-  * OPTIONAL
-  * If present, MUST contain at least one entry
-* Examples:
-  * authorization data
-
 ### data
 * Type: Arbitrary payload
 * Description: The event payload. The payload depends on the event-type,
--- a/spec.md
+++ b/spec.md
@@ -209,16 +209,16 @@ that contains both context and data).
     [RFC 2046](https://tools.ietf.org/html/rfc2046)
 * For Media Type examples see [IANA Media Types](http://www.iana.org/assignments/media-types/media-types.xhtml)
 
-### property-bucket
+### properties
 * Type: <Key Value>
-* Description: A list of “key : value” pairs in flat strcuture that specify the properties of the event source.
+* Description: A list of “key : value” pairs in flat structure that specify the properties of the event source.
   The value could be a string or a JSON object. There should be no duplication of field in this property-bucket. 
   One example usage of the property-field is to correlate multiple events involved with a serverless application.  
 * Constraints:
   * MANDATORY
   * MUST be in the format of <key value> pair. 
 * Examples:
-  property-bucket
+  properties
   { 
      "building address": "12 Main Street, Copenhagen, Denmark, 123456"
      "floor": "2"
--- a/spec.md
+++ b/spec.md
@@ -209,6 +209,17 @@ that contains both context and data).
     [RFC 2046](https://tools.ietf.org/html/rfc2046)
 * For Media Type examples see [IANA Media Types](http://www.iana.org/assignments/media-types/media-types.xhtml)
 
+### correlation-field
+* Type: String
+* Description: A path string pointing to a field in the event data 
+  that can be used to correlate all the events associated with 
+  the same application instance. 
+* Constraints:
+  * OPTIONAL
+  * If present, MUST be a non-empty string
+* Examples:
+  * headers.body.cathy-house-number
+
 ### extensions
 * Type: Map <String, Object>
 * Description: This is for additional metadata and this does not have a


spec.md
```diff
@@ -209,6 +209,17 @@ that contains both context and data).
     [RFC 2046](https://tools.ietf.org/html/rfc2046)
 * For Media Type examples see [IANA Media Types](http://www.iana.org/assignments/media-types/media-types.xhtml)
 
+### correlation-field
```
SeanFeldman: I'm not sure about this one.
It feels it should be an attribute with the value rather than a path in the data. Reading an attribute value for correlation ID is also more practical then reading a path and then fetch the value from the data.

Having this field on the spec that is optional _and_ application/system specific does not appeal to me.

duglin: I would think that you'd want to be able to do the correlation w/o having to look at the event data itself. So I kind of assumed that if we added info like this it would be based on something in the metadata, like the event-id.

cathyhongzhang: Sean: are you suggesting adding an explicit attribute, e.g., correlation-token or correlation-id, to the Event Context/Metadata? If so, I am fine with it. The multi-event correlation ID information can either be fetched by reading a path into the data (in this case the path string needs to be available and read from the event metadata) or reading directly from an explicitly defined correlation-token value from the event metadata. 

Doug, looks like you are thinking about the same. Correct me if I misunderstand your point. 

I will update the PR soon.

SeanFeldman: @cathyhongzhang not quite. Correlation is a system/application concern. Not every single event will have it. As such, mandating an explicit attribute feels off. I'd think an applicaiton / system that required correlation would be a part of event medatata. And 💯% not to point back to the `data` to retrieve correlation value.

cathyhongzhang:
@SeanFeldman First we are not mandating this attribute and that is why it is defined as OPTIONAL, similar to schema-url, event-time, content-type etc. which are OPTIONAL Context Attributes.  For those use cases, e.g. use case that only has a single event, the event does not need to have this correlation token/ID attribute. But there are many use cases (such as IoT use cases) that require correlation of multiple events. To support such serverless use cases, the event context must have an attribute for the event consumer to get the correlation token value to correlate multiple event instances correctly and send them to the same application/processing instance.

SeanFeldman:
Could you articulate why would it be benefitial to pass correlation value using an optional `event` attribute rather than `extensions` @cathyhongzhang?

cathyhongzhang:
@SeanFeldman Could you first articulate why it is good to put schema-url, event-time, content-type as optional event attributes while it is not good to put correlation as event attribute :-) ?

lfourie:
I think that an event attribute that contains a value to correlate/associate several events from different sources is very useful, particularly in IoT use cases, where there are multiple events sources (sensors). I think this warrants inclusion as a cloudevents attribute.

duglin:
Only writing this in this PR because it was at the top of my list, I'm not implying anything about which bucket it belongs, but for each non-REQUIRED attribute the WG will need to decide which of 3 categories it belongs:
1 - in the CloudEvents spec, marked as OPTIONAL
2 - in something like a "CloudEvents Extension Attributes" doc we produce, like proposal https://github.com/cloudevents/spec/pull/126, that lists well-known extensions but have no official standing
3 - defined by a vendor or some other WG - but not this WG

The bar I believe I've heard mentioned so far is that the OPTIONAL ones in the spec should be considered used by a majority of the implementations. How do we measure that? I don't know. But a start would be for people to speak up on this PR, https://github.com/cloudevents/spec/pull/63 and https://github.com/cloudevents/spec/pull/76 to say whether your org will use it.

SeanFeldman:
Discussions around `schema-url`, `event-time`, and `content-type` predates my participation on this WG.
Regardless, here's my thought on those.

✔️ `event-time` - a timestamp of when the original event has occured is logical. As a consumer of an event such as "AWS S3 object created" or "VM provisioned" I'd like to know when that took place.
❌ `schema-url` and `content-type` - the way those are defined in the current spec make little sense to me.

Still would want to hear what do you think the benefits of passing correlation value using an optional event attribute rather than extensions are. Thanks.

cathyhongzhang:
@SeanFeldman My understanding is that Optional Event attributes refer to attributes that are needed in some use cases but not in other use cases.  Use cases that involve multiple events are common and are a key category use case for serverless. One goal of this spec is to facilitate interoperability/consistency. Adding Correlation-token/id into the Event Context Attributes will help with the  consistency/interoperability among different event producers and consumers. Extension is vendor specific feature, which is not the case for correlation of multiple events.

inlined:
So I think the use case is incredibly important, though I worry that explicitly including the "correlation-id" in the event definition (as the example is described) is bordering on including destination information in the Event definition. We agreed that this was an explicit non-goal.

What if we could have free-form attributes (labels) like K8S, AWS, GCP, etc support and an event broker supported "correlation-id"s for subscription/delivery? In my mind, your IoT example event could have the label `{"house": "CathyHouse"}` and an Action subscribe for events with `{"correlation-id": "house"}`. Putting the `correlation-id` directly in the event seems more brittle. To use another IoT example, I might have temperature sensors. Depending on my use case, I might want to correlate the sensors on all boilers  or on all devices in a warehouse.

cathyhongzhang:
@inlined  Thanks for your input! We are on the same page w.r.t most of your points. As mentioned in the commit message, different applications can have different fields in the event data/message as the correlation token/id. I think we need to solve the following issues to support this category of use cases: 
1. How can we ensure that the event producer puts a label which a serverless platform or other event middle ware/consumer can use for correlation?  
2. Suppose we go down the path of event labels, an event could carry multiple event labels. How could the serverless platform know which label to use for correlation? This "correlation-id" must be specified somewhere.

inlined:
I think that different applications will want to correlate on different fields. In fact, the same application might deploy multiple actions that each listen to different event streams correlated by different fields. Thus I think a good event broker/router/whatever it's called would want to support _as part of the subscription_ filtering and ordering constraints by any field in the event, including labels.

This _does_ add the constraint that the IoT action must know the exact label used for "house" if it wants to correlate on houses. This does mean that some part of the system (either the Action or the subscription–a concept currently out of scope) must understand some details of the specific event provider if it's going to provide these kinds of features. I'm not entirely sure that this is unexpected or unavoidable.

duglin:
@cathyhongzhang I added `Closes: #24 ` to the first comment since this PR covers that issue and will then github will close that issue automatically as we resolve this issue. Hope that's ok.

cathyhongzhang:
@duglin Sure.

duglin:
@cathyhongzhang what's the status of this one?

cathyhongzhang:
@duglin Could we discuss this PR in tomorrow's meeting?

Vlaaaaaaad:
Hi,

This was discussed at Kubecon+CloudnativeCon Face-to-Face, where I could not keep my mouth shut and promised to type out the example.

What is desired by adding a correlation id is to be able to see where an event originated from. An event source might not have all the details necessary to complete all the details about its origin. This was discussed and an example was given, but there was some confusion hence this post.

Say there is a motion sensor that is generating an event: "_motion detected_". Now, the motion detector knows details about itself( manufacturer, version, OS, what motion was detected and so on), but has no idea what room it is into, if it's in a house or in an apartment or what the building complex address is. A security system for a building would want to know all those details.

The motion sensor sends the event to an event gateway. The gateway knows about the motion sensor and about the environment it's in. Knowing such information it can add some more data such as the room name. The event may go to other higher level gateways which add even more information to the event. See an example below:

At the **motion detector level** we can create the following:
```json
"correlation":
{
	"properties":
	{
		"name": "MotionSensorCorp Motion Plus Ultra",
		"version": "2",
		"uniqueId": "baevgr8eqdc2ww80f",
	},
	"source_data":
	{}
}
```

The `source_data` field is empty since this is the event being created the first time( there is no previous).

At the **room level** gateway we can further enrich the event before routing it:

```json
"correlation":
{
	"properties":
	{
		"roomId": "21312",
		"roomName": "Living Room"
	},
	"source_data":
	{
		"properties":
		{
			"name": "MotionSensorCorp Motion Plus Ultra",
			"version": "2",
			"uniqueId": "baevgr8eqdc2ww80f",
		},
		"source_data":
		{}
	}
}
```

At an even higher level( **apartment level**) we can further enrich the event with data that night not be available at a previous level:
```json
"correlation":
{
	"properties":
	{
		"apartmentId": "e12e312e12d",
		"apartmentNumber": "10140",
		"floor": "2",
		"address": "12, Main Street, Copenhagen, Denmark, 123456",
		"tenant": ["Jane Doe", "John Doe", "Sparky the dog"]
	},
	"source_data":
	{
		"properties":
		{
			"roomId": "21312",
			"roomName": "Living Room"
		},
		"source_data":
		{
			"properties":
			{
				"name": "MotionSensorCorp Motion Plus Ultra",
				"version": "2",
				"uniqueId": "baevgr8eqdc2ww80f",
			},
			"source_data":
			{}
		}
	}
}
```


And the event is now arriving at the function which executes it.
  
  
  
This was the whole workflow initially envisioned by the developers. But some time passes and there are now multiple buildings and the system needs to evolve. The same event can be enriched further, without any modifications to any lower levels. At the **building level** we would get something like:

```json
"correlation":
{
	"properties":
	{
		"complexName": "Harmony Living 1",
		"complexBuilding": "B2",
		"address": "12, Main Street, Copenhagen, Denmark, 123456"
	},
	"source_data":
	{
		"properties":
		{
			"apartmentId": "e12e312e12d",
			"apartmentNumber": "10140",
			"floor": "2",
			"address": "12, Main Street, Copenhagen, Denmark, 123456",
			"tenant": ["Jane Doe", "John Doe", "Sparky the dog"]
		},
		"source_data":
		{
			"properties":
			{
				"roomId": "21312",
				"roomName": "Living Room"
			},
			"source_data":
			{
				"properties":
				{
					"name": "MotionSensorCorp Motion Plus Ultra",
					"version": "2",
					"uniqueId": "baevgr8eqdc2ww80f"
				},
				"source_data":
				{}
			}
		}
	}
}
```

This allows for easy composition and forwarding of CloudEvents that is forward-compatible. The motions sensor does not have to be concerned about anything but emitting its event to the gateway. The gateway and the applications that consume and forward this event have to care about correlation.

---

At the Kubecon+CloudnativeCon F2F it was also decided that exactly what will be in that field is out of scope for CloudEvents( maybe in scope for WG Serverless). What is in scope for CloudEvents is if this field is to be added or not.

After the discussion had at Kubecon F2F it was agreed that having this event as an extension would create some "_optional but really mandatory_" extensions -- this is to be avoided. What seemed to be the consensus was to have the 1 `correlation` field added with 2 sub-fields( say `properties` and `source_data`) added to the CloudEvents spec. This would also allow for `source_data` to be signed and/ or encrypted. What each event gateway on the way is doing is adding even more properties to the `properties` field( leaving `source_data` intact) and sending the event forward. Think of nesting dolls.

What this field is proposing is similar to an OpenTracing trace where each span has 1 child. Each CloudEvent would get tracing and the properties discussed would be added as `context` to the trace. Thomas is looking into whether OpenTracing could be used instead of adding the `correlation` field to CloudEvents.

It is still unclear to me if an event may be routed to multiple endpoints with different labels.

Hope I understood this correctly and that the example is clear enough. If I did understood something wrong or misused terms, please let me know.

inlined:
While I accept that any _signature_ system would need to sign ever-increasing blocks of scope, I dislike the nesting-dolls model of adding annotations. This exposes the network infrastructure through which the event was delivered, which can add to fragility. A good test is that adding a well-behaved gateway in the middle should not break any applications. The nesting-doll approach violates this.

Also as a small nit, though we discussed two types of property bags, we didn't discuss (and I don't agree with) putting them under a "correlation" property. Source labels/properties fulfill the requirement that events can be correlated, but aren't limited to that purpose.

I would suspect something similar to:

```
{
  "eventID": "123",
  "source": "homsecurity://uuid",
  "eventType": "homesecurity.motion",
  "labels": {
    "softwareversion": "1.2.3.",
    "displayName": "Joe Livingroom",
    "sensitivity": "HIGH",
  },
  "annotations": {
    "complex": "Paradise Suites",
    "unit": "1337",
  },
  "contentType": "application/json",
  "data": {
    "bestGuessReason": [
      {
        "label": "cat",
        "confidence": 0.9,
      }
    ]
  }
}
```

(Btw, sample events suddenly make this creepy. I'm curious whether the spec should allow for redaction at a gateway too)

Vlaaaaaaad:
Yup, leaking network infrastructure into the event is bad. Both adding properties( be it labels or annotations) and signing the content without a nesting-doll approach is unclear to me. A list of signed stuff seems... dirty. There may be a more elegant approach, but I am missing it.

lfourie:
The use of a base structure with a bag of invariant properties created by the event source, and an annotated structure with a bag of annotations added by event GWs and brokers was also discussed at the F2F. This is similar to what is described in http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-security-v1.0-os.html. A similar structure could be used for correlation labels.

inlined:
@lfourie Though the AQMP link was interesting, was this the section that you intended to send? In a quick scan I didn't grasp how the encryption spec relates to attributes.

lfourie:
@inlined correct reference: http://docs.oasis-open.org/amqp/core/v1.0/amqp-core-messaging-v1.0.html
 - see discussion of immutable payload and properties

duglin:
@cathyhongzhang what's the status of this?

cathyhongzhang:
@duglin Vlad added some suggestion for source label. I would like to discuss them with other suggestions in the meeting. I will close this one after our meeting discussion.

jroper:
I think `correlation` is the wrong name to use here, since correlation id already has a specific meaning used widely across many platforms specific to messaging that I think is different to the meaning here. Correlation ids are used to associate an asynchronous response with a particular request, and they are also used for tracing a single interaction from the outside of a system through multiple calls throughout the system. They are unrelated to an application instance, they are generally an opaque, unique, randomly generated ID, generated when a request comes in. When monitoring the system, the correlation id is passed along with all effects - if that interaction resulted in multiple events being emitted, all of the events emitted will have the same correlation id. Then, a monitoring system that traces these events, and traces chains of side effects through the system, can associate all those effects with the initial interaction that caused them. This allows it to show metrics, like how long it took for all the effects of the request to processed.

What I believe this proposal is about is an entity key, something that groups events from a particular entity. I'm not 100% sure, but it seems the current definition in this proposal is very specific to IoT, more broadly, this should cover any type of entity, whether it's a device, or an entity in a database. A number of message brokers already provide in built support for this, including Kafka, AWS Kinesis, and Azure Event Hubs, and the word they use to call it is "key", a key could be an id of a device, or it could be the id of an entity in a database. Sometimes they call it "partition key", because it's used to establish a causal ordering between like events so that when events are partitioned into multiple queues for the purposes of scaling, events from the same entity will end up on the same queue and so end up in the same order.

This event key concept is what I have proposed in #218.

jroper:
Here's some additional resources to read about correlation ids:

https://blog.rapid7.com/2016/12/23/the-value-of-correlation-ids/
https://blog.arkency.com/correlation-id-and-causation-id-in-evented-systems/
http://hilton.org.uk/blog/microservices-correlation-id

Vlaaaaaaad:
The initial idea was indeed to use a single field with a single value. That proved to be a bit limiting and a bag of labels is currently the idea that is considered. In such a bag of labels there could also be a partition id/ event key( what you proposed in #218 ) or whatever else labels are needed for correlation( which is a word that still needs a proper definition and may be misused because we all have different understandings). Does that work for the workflow you had in mind?

The idea, from what I gather, is not to have a ton of `OPTIONAL` high-level labels added to the CloudEvent. Everything regarding correlation or ordering or workflows should just be put in that bag of labels. The labels might change based on what the app does or what the workflow is.

Still, this is to be discussed in the call today.

cathyhongzhang:
Let me try to clarify a little bit here. The proposal is not really about an entity key that groups events from a particular entity. It is about how to correlate events from different sources which are involved in an application workflow together so that these related events can be processed by the same application workflow instance. Vlad is right that the labels might change based on what the app does or what the workflow is and these labels or unique identifiers should not be optional.

cathyhongzhang:
I would also like to clarify that this is not just for IoT cases. For any serverless use case applications that involve multiple events coming from different sources or the same source, there is a need to correlate/associate those events together for processing, and the events themselves must provide some label/unique identifiers to make the correlation possible.

cathyhongzhang:
I have updated the title and file change based on today's event correlation discussion meeting.

Vlaaaaaaad:
# A more detailed overview of the correlation id/ partition key discussions

Since topics are being re-hashed and discussed again and again at all the meetings, I wrote this wall of text.

## Properties

### What
**It's not into the CloudEvents scope to define what these are.** Just some fields with data that help corelating an event. They may also be part of the payload or they may not.

``` js
"properties": {
    "device": "MotionSensorCorp Motion Plus Ultra",
    "version": "2",
    "address": "12, Main Street, Copenhagen, Denmark, 123456",
    "apartmentNumber": "10140"
}
```

### Where

There were 3 ideas that were floated around this:
- a bag of properties. Add anything you want inside
- two bags, one for the source and one for the transport( gateways, middleware, whatever)
- all the properties added as top-level fields

This is still being debated.

#### Separate lighter bag

An idea was to have another top level key added, `properties` which would hold just string key-value pairs.

Lighter than extensions, no complex parsing of keys and intended for routing and fast processing.

``` js
{
    "cloudEventsVersion" : "0.1",
    "eventType" : "com.motionSenor.motionDetectedEvent",
    "eventTypeVersion" : "1.0",
    "source" : "/mycontext",
    "eventID" : "A234-1234-1234",
    "properties": {
        "device": "MotionSensorCorp Motion Plus Ultra",
        "version": "2",
        "address": "12, Main Street, Copenhagen, Denmark, 123456",
        "apartmentNumber": "10140"
        // The following not allowed as they are not simple key-value pairs:
        //   "tenant": ["Jane Doe", "John Doe", "Sparky the dog"],
        //   "tenant-details": {
        //        "people" : "yes",
        //        "dogs": "yes",
        //        "cats": "no"
        //    }
    },
    "extensions" : {
      "comExampleExtension" : "value"
    },
    "contentType" : "binary",
    "data" : "... encrypted payload A ..."
}
```

#### Added in the extensions bag

An idea was to reuse the already existing `extensions` top-level key for this.

It's unclear what the purpose for the `extensions` key is. I personally assumed it's for more complex structures with nesting and multiple objects and large data. Say a company defining and extension for all the CloudEvents used in their systems( whatever fields and object).

``` js
{
    "cloudEventsVersion" : "0.1",
    "eventType" : "com.motionSenor.motionDetectedEvent",
    "eventTypeVersion" : "1.0",
    "source" : "/mycontext",
    "eventID" : "A234-1234-1234",
    "extensions" : {
        "device": "MotionSensorCorp Motion Plus Ultra",
        "version": "2",
        "address": "12, Main Street, Copenhagen, Denmark, 123456",
        "apartmentNumber": "10140",
        "tenant": ["Jane Doe", "John Doe", "Sparky the dog"],
        "tenant-details": {
            "humans" : "yes",
            "dogs": "yes",
            "cats": "no"
        }
    },
    "contentType" : "binary",
    "data" : "... encrypted payload A ..."
}
```

#### Added as top-level fields

An idea was just to have the properties as top-level fields by themselves.

An advantage of this would be that that newer spec-defined properties will automatically appear to older receivers as extensions and not just ignored.

``` js
{
    "cloudEventsVersion" : "0.1",
    "eventType" : "com.motionSenor.motionDetectedEvent",
    "eventTypeVersion" : "1.0",
    "source" : "/mycontext",
    "eventID" : "A234-1234-1234",
    "device": "MotionSensorCorp Motion Plus Ultra",
    "contentType" : "binary",
    "data" : "... encrypted payload A ..."

    "version": "2",
    "address": "12, Main Street, Copenhagen, Denmark, 123456",
    "apartmentNumber": "10140",
    "tenant": ["Jane Doe", "John Doe", "Sparky the dog"],
    "tenant-details": {
        "humans" : "yes",
        "dogs": "yes",
        "cats": "no"
    }
}
```


### Who

#### Why the who matters

Assuming that the place is defined( refereed to as properties in my next messages), a main question appears: who is allowed to modify that and how? That creates even more questions such as:
- Can I as a developer be sure that my destination will have at least all the properties I defined with the values I defined?
- Can I as a developer be sure that my destination will have at least all the properties I defined, but the values may have been changed?
- Can I as a developer be sure that just the payload may arrive at the destination, with the properties being used for intermediary steps?
- Can I as a developer be sure that my payload will arrive at the destination, with some of my properties and some other properties?

Now, if an event is forwarded( create event -> middleware and event gateways add properties and route the event) the original event emitter can have the properties in the payload too. But the destination might want other properties too which cannot be added by the emitter.
For the motion sensor example, a developer could add the `sensorName` and `sensorVersion` to both the payload and the properties. It then encrypts the payload and sends the event forward. Middleware and gateways may add `roomName`, `apartmentNumber`, `floor` and `address` as properties too which would be very much relevant to the destination.

Small aside: by destination I mean a consumer that is not middleware or an event gateway. Kind of the final consumer of the event. And there can be many final consumers.

#### Eliminating the who - no middleware concept idea

An idea was proposed for having no difference between middleware( be it middleware or event gateways) and actual event consumers.

So there is no middleware. Everything is just a consumer and a new event is always generated if any change is done.

Let's do the motion sensor example. There is an event emitter( source) which creates an event:
``` js
{
    "cloudEventsVersion" : "0.1",
    "eventType" : "com.motionSenor.motionDetectedEvent",
    "eventTypeVersion" : "1.0",
    "source" : "/mycontext",
    "eventID" : "A234-1234-1234",
    "properties": {
        "device": "MotionSensorCorp Motion Plus Ultra",
        "version": "2",
    },
    "extensions" : {
      "comExampleExtension" : "value"
    },
    "contentType" : "binary",
    "data" : "... encrypted payload A ..."
}
```

This first event reaches a apartment-level event gateway. Now, at this level more is known and some more data can be added. But since this is also a consumer a new event has to be created. The same payload is kept( the encrypted payload from the motion sensor), but some more properties are added:
``` js
{
    "cloudEventsVersion" : "0.1",
    "eventType" : "com.motionSenor.motionDetectedEvent",
    "eventTypeVersion" : "1.0",
    "source" : "/mycontext",
    "eventID" : "B987-9876-9876",
    "eventTime" : "2018-04-05T17:42:00Z",
    "properties": {
        "device": "MotionSensorCorp Motion Plus Ultra",
        "version": "2",
        "address": "12, Main Street, Copenhagen, Denmark, 123456",
        "apartmentNumber": "10140",
        "tenant": ["Jane Doe", "John Doe", "Sparky the dog"]
    },
    "extensions" : {
      "comExampleExtension" : "value"
    },
    "contentType" : "binary",
    "data" : "... encrypted payload A ..."
}
```

That works nicely from a properties and payload point of view. But the `eventID` and `eventTime` are lost. A new event is created hence new ID and time.

That might not be acceptable as the initial device says _I created and sent the event A234-1234-1234_ and the actual consumer gets _B987-9876-9876_ and is very much confused. Relation between subsequent events are harder to impossible now.

Solutions for this:
- duplicate relevant data from original envelope in `properties`
- add the original event payload and envelope to the new event payload
- define middleware concept which keeps the original event and is just allowed to add additional `properties`

The simplest and most flexible one seems to define middleware which can modify the event a bit. What does _modify the event a bit_ mean is what we have to define in the event chaining workflow I imagine.

#### Modifying properties

On the other hand we can define middleware as "_ok, middleware is allowed to just add additional properties and no other changes"( maybe extensions too? I don't know). Or some variation of that.

We again have the motion sensor example. The event emitter( source) creates an event:
``` js
{
    "cloudEventsVersion" : "0.1",
    "eventType" : "com.motionSenor.motionDetectedEvent",
    "eventTypeVersion" : "1.0",
    "source" : "/mycontext",
    "eventID" : "A234-1234-1234",
    "properties": {
        "device": "MotionSensorCorp Motion Plus Ultra",
        "version": "2",
    },
    "extensions" : {
      "comExampleExtension" : "value"
    },
    "contentType" : "binary",
    "data" : "... encrypted payload A ..."
}
```

This first event reaches a apartment-level event gateway. Now, at this level more is known and some more data can be added as properties:
``` js
{
    "cloudEventsVersion" : "0.1",
    "eventType" : "com.motionSenor.motionDetectedEvent",
    "eventTypeVersion" : "1.0",
    "source" : "/mycontext",
    "eventID" : "A234-1234-1234",
    "properties": {
        "device": "MotionSensorCorp Motion Plus Ultra",
        "version": "2",
        "address": "12, Main Street, Copenhagen, Denmark, 123456",
        "apartmentNumber": "10140",
        "tenant": ["Jane Doe", "John Doe", "Sparky the dog"]
    },
    "extensions" : {
      "comExampleExtension" : "value"
    },
    "contentType" : "binary",
    "data" : "... encrypted payload A ..."
}
```


The event then is forwarded to 2 consumers: the *security company*( for a burglary system) and the apartment *intelligent light controller*. The security company has a gateway installed in the building and the apartment owners have a server doing the processing for the intelligent light controller.

The security company gateway gets the event from the apartment-level event gateway and wants to forward the event to the security company. Now, what can it do?
- Can it delete the tenants property as that is private data and not necessary?
- Can it add some more properties such as `subscriptionID` or `buildingAdmin` or something like that?

The current consensus around this is that this will be tackled as part of the security discussion( signing, encryption) because this affects the whole CloudEvent definition. Might also be in scope for the WG Serverless function workflows/ event chaining workstreams.


## Partition key

### What

Better descriptions for this are given in https://github.com/cloudevents/spec/pull/218, but let's think of it as a field used for sending the event to a particular Kafka/ Kinesis stream to ensure all events from a producer get sent to the same stream.

``` js
"partitionKey" : "fad11ea2-e2b9-4bb1-b497-22104b354232"
```

### Where

#### As a top-level optional field

An idea expressed in https://github.com/cloudevents/spec/pull/218 is to have a top-level `OPTIONAL`( since this may be added after the event is initially created) field.

``` js
{
    "cloudEventsVersion" : "0.1",
    "eventType" : "com.motionSenor.motionDetectedEvent",
    "eventTypeVersion" : "1.0",
    "source" : "/mycontext",
    "eventID" : "A234-1234-1234",
    "partitionKey": "fad11ea2-e2b9-4bb1-b497-22104b354232",
    "properties": {
        "device": "MotionSensorCorp Motion Plus Ultra",
        "version": "2",
        "address": "12, Main Street, Copenhagen, Denmark, 123456",
        "apartmentNumber": "10140",
        "tenant": ["Jane Doe", "John Doe", "Sparky the dog"]
    },
    "extensions" : {
      "comExampleExtension" : "value"
    },
    "contentType" : "binary",
    "data" : "... encrypted payload A ..."
}
```

#### As another property

As properties and correlation were discussed, the idea that this is just another piece of correlation was floated. If this is just another piece of correlation, it should clearly be another property.
On the other hand, a constant name for this is desired and reserving a name as a property would mean restricting the properties which is not desired.

``` js
{
    "cloudEventsVersion" : "0.1",
    "eventType" : "com.motionSenor.motionDetectedEvent",
    "eventTypeVersion" : "1.0",
    "source" : "/mycontext",
    "eventID" : "A234-1234-1234",
    "properties": {
        "partitionKey": "fad11ea2-e2b9-4bb1-b497-22104b354232",
        "device": "MotionSensorCorp Motion Plus Ultra",
        "version": "2",
        "address": "12, Main Street, Copenhagen, Denmark, 123456",
        "apartmentNumber": "10140",
        "tenant": ["Jane Doe", "John Doe", "Sparky the dog"]
    },
    "extensions" : {
      "comExampleExtension" : "value"
    },
    "contentType" : "binary",
    "data" : "... encrypted payload A ..."
}
```

## Ordinal field
No idea, not really discussed during the calls( or my memory is failing). Issue: https://github.com/cloudevents/spec/pull/191


# Conclusion

What still needs to be discussed:
- reuse `extensions` or a new top-level field? Many top-level fields? Simple string key-value pairs or full JSON objects?
- partition key as another property or separate? Currently the idea is to make the decision based on usage: if this is likely to be used in many cases have it as a top-level field, if not make it just another property.

To be discussed at a later date/ as part of other issues in CloudEvents or WG-Serverless:
- field encryption and signing, the whole data tampering issue AKA who is allowed to add/modify/delete properties
- what is actually put in properties
- what gateways/ middleware can/ should do

---

Hope I made this a bit more clear. If I misunderstood anything please comment here.

spec.md
```diff
@@ -209,6 +209,23 @@ that contains both context and data).
     [RFC 2046](https://tools.ietf.org/html/rfc2046)
 * For Media Type examples see [IANA Media Types](http://www.iana.org/assignments/media-types/media-types.xhtml)
 
+### property-bucket
+* Type: <Key Value>
+* Description: A list of “key : value” pairs in flat strcuture that specify the properties of the event source.
```
johnmccabe: `strcuture` -> `structure`

spec.md
```diff
@@ -209,6 +209,23 @@ that contains both context and data).
     [RFC 2046](https://tools.ietf.org/html/rfc2046)
 * For Media Type examples see [IANA Media Types](http://www.iana.org/assignments/media-types/media-types.xhtml)
 
+### property-bucket
```
markfisher: as mentioned on WG call, this could be simply named "properties" (more consistent with parallel "extensions")

spec.md
johnmccabe: Is it worth making clear that the properties are specific to the source, `specify the properties of the event source` could be interpreted as properties that apply to multiple sources.

spec.md
```diff
@@ -209,6 +209,23 @@ that contains both context and data).
...
+* Type: <Key Value>
+* Description: A list of “key : value” pairs in flat strcuture that specify the properties of the event source.
+  The value could be a string or a JSON object. There should be no duplication of field in this property-bucket. 
+  One example usage of the property-field is to correlate multiple events involved with a serverless application.
```
deissnerk: Are properties only added when the event is created, or can they be changed/added on the event's way through the infrastructure?

inlined:
I think it's worth considering calling this `Correlation-Context` and using the existing WC3 proposal [here](https://github.com/w3c/distributed-tracing/blob/master/correlation_context/HTTP_HEADER_FORMAT.md)

spec.md
Vlaaaaaaad: That might be a thing. In the motion sensors example that sends an event to an apartment-level gateway and then a building-level gateway there would indeed be properties that apply to multiple events( say address). See https://github.com/cloudevents/spec/pull/128#issuecomment-395239721 for some context and examples.

glennblock:
One suggestion I brought up on the last call was around schema for the properties. If the property bag is intended to really be open-ended, it seems like some level of description would be valid. I proposed JSON-LD as one possibility because it has been around a while, is in pretty wide use, and because it is very flexibility, can attach to existing JSON payloads easily.

glennblock:
Aside from that I think the overall idea of properties makes sense. Based on my own experience in similar domains, my preference is it just becomes a top level thing. That makes it really easy for gateways / middleware to pick it up. If you bury it in extensions that makes the logic more complicated of how I consume extensions right?

spec.md
cathyhongzhang: will do

spec.md
cathyhongzhang: Will fix the typo. 
Vlad, Not sure what you mean by multiple events. I would view your example as the "same" event from the eventual event consumer's point of view (e.g. the serverless platform) , with intermediate GWs adding more info or modifying the original event info. As to whether the source is the original sender or the last mile GW, that is a separate discussion.

spec.md
cathyhongzhang: I think they can be changed/added

spec.md
johnmccabe: @cathyhongzhang I deleted a follow-up comment asking if was worth clarifying whether properties were source specific or common to all sources. Deleted it as I'd raised it during the call without digesting @Vlaaaaaaad's detailed summary.

cathyhongzhang:
I don't think it is a good idea to put all the key-value pairs at the top level. That will result in too many user-defined top level attributes. It is better to put these key-value pairs under the "properties" attribute.

glennblock:
@cathyhongzhang just to clarify when I said top-level I meant off of a "properties" field hanging at the top level, not just loose key-value pairs. I agree with your point.

cathyhongzhang:
@glennblock Oh, then we share the same thought :-) Yes, it makes more sense to put them in the "properties"

duglin:
I'm having a really hard time seeing much of a difference between a "properties" field and the "extensions" field. Both are maps. Both allow just about anything. Both describe something about the event (either the event itself, the source, or something else - we don't really place many restrictions). I think its going to be very hard for people picking up the spec to have a very clear understanding of when to use "properties" vs "extensions" - and that confusion will hurt its adoption, and interop.

I see these type of fields being used like this:
- someone wants to add some metadata to the CloudEvent
- they define a new field "coolCompany.myInfo" and (in an ideal world) have a doc/spec defining its semantics and its syntax (type) 
- event producers who want to use this new field add it to their CloudEvents
- event consumers who want to look for this new field will have prior knowledge of what "coolCompany.myInfo" is all about and will look for it, and act upon it

Notice, I didn't say (and didn't have to say) where in the CloudEvent it appears. Deciding whether "coolCompany.myInfo" appears in "properties" or "extensions" feels like an extra step that is unnecessary. Attempting to scope each bucket of properties with extra semantics isn't needed since the definition of this field should already do that.

Additionally, if we do have separate buckets for extensions, we then have to define separate rules for how to serialize each to avoid conflicts between them when they're serialized as HTTP headers. Again, not impossible, but why? It feels like extra processing/thinking w/o a clear purpose.

And... what if one day we want to add this new property to the spec? If we treat it like all other extensions (and do #225) then no one needs to change any code - we just add it to the spec as-is.

Vlaaaaaaad:
I do agree that extensions need to be defined more clearly. I see the difference between extensions and properties from a size/ effort perspective.

In my understanding proprties are simple key-value pairs, no objects or deep nesting. Simple, quick parsing.
I see extensions as more in-depth, both in terms of actual size and in terms of logic.

A properties bag could be:
```js
"properties" : {
    "device": "MotionSensorCorp Motion Plus Ultra",
    "version": "2",
    "address": "12, Main Street, Copenhagen, Denmark, 123456",
    "apartmentNumber": "10140",
}
```

An extension  would be something more complex. Something that extends the CloudEvent for a specific workflow or a company:
```js
"extensions" : {
    "myBigCorpExtension" : {
        "version" : "1.3.2",
        "data": {
            "cert" : " ... base64 encoded stuff ... ",
            "internalIdForThatSpecificThing": "481770fa-70a8-11e8-adc0-fa7ae01bbebc",
            "payloadSchema": "2.9",
            "customer": "12312312312",
            "region": "eu-east-1",
            "priority" : "low"
        }
    },
    "deliveryExtension": {
        "saveAs" : "12312312-file.bin",
        "rights" : "admin",
        "location" : "/data/projectX/bla"
    }
}
```

It would make sense for some values from extensions to also be promoted to the properties bag, but the main idea that `properties` are quick and used for routing would remain. This would kind of clarify where a value should be.

duglin:
A "size" type of argument is odd to me.  In order to do something with the value of a field you need to understand the field (its semantics, syntax, etc...). If someone defines a field that's complex (or "big") and its required in order to do proper routing, I don't think its location in the CloudEvent will matter - the receiver will either want to do that complex logic or not. Its all whether they want to support those semantics (and therefore take on the burden of the complex processing), not location in the Event. As for unknown properties, I don't think size matters since people can skip (or blindly pass-on) the raw JSON w/o parsing/understanding it.

inlined:
IMO, I think extensions are things that an events _framework_ might leverage (e.g. tracing, checksum verification, sampling) whereas event properties are things that an event-based _application_ might leverage (e.g. add affinity by a property or use a property to decide which clients can access events)

duglin:
During the 6/15 f2f we agreed to:
- Cathy will modify the PR so that it includes how this is different than extensions, and how it’s different than tracing, although the language is very similar.

duglin:
Also agreed that:
- the value should just be a "string" for now
- its a flat list of properties
- proposed name is "ApplicationProperties"
- no duplication of property names

glennblock:
I agree with Thomas. Properties are a loose bag of data which the receiver
may or may not use. Extensions I would expect are known and defined in a
spec somewhere. Also they would likely have some behavior attached to them.

duglin:
Actually, extensions and "properties" are similar in that both may be known, or unknown, to the receiver, both could have behavior associated with them (even if just for routing), and both will require out of band knowledge between the producer and receiver to understand them since neither are defined in the spec.  Extensions will be in a non-normative doc, not part of the spec.

IMO, as of the f2f, the only diff between properties and extensions are that properties can only be strings.

cathyhongzhang:
@duglin I am going to add clarification on the definition of property itself.

duglin:
@cathyhongzhang while we won't have a property called "extensions" we will have the notion of extensions and each serialization/binding spec will say where those extensions go - in many cases they will be treated like all of the defined properties of the spec (I hope). But, at the f2f we asked for some text to explain why these extensions are different from the ApplicationProperties - which to some will be viewed as another extension - and that's going to be confusing w/o some clear rationale for why/how they're different.

cathyhongzhang:
@duglin  I understand your point. But going down this "explain difference" path, does it mean that every new attribute to be added should explain the difference from "extensions"?  It seems the root cause is that the "extensions" attribute was NOT defined clearly in the first place and thus the confusion. Given that we will remove the "extension" (right decision ! ), there seems no comparison needed.  How about we forget about "extensions" and let me try to clarify the definition of the "ApplicationProperties"? And let's see if the group has any input on better clarification or is OK with this PR/definition in this week's meeting?

duglin:
@cathyhongzhang can you fix the DCO issue and rebase?

spec.md
```diff
@@ -209,19 +209,25 @@ that contains both context and data).
...
+* Type: <Key Value>
+* Description: A list of “key : value” pairs providing information about the event, which are 
+  used by an event-based application. The “key : value” pairs should be in flat structure.
+  The value should be a string. There should be no duplication of keys.
+  One usage example: an application workflow can use a "key:value" to correlate multiple events 
+  (from different event sources) asscociated with a serverless application workflow.  
 * Constraints:
-  * OPTIONAL
-  * If present, MUST contain at least one entry
+  * MANDATORY
```
duglin: s/MANDATORY/REQUIRED/

spec.md
```diff
@@ -209,19 +209,25 @@ that contains both context and data).
...
 
-### extensions
-* Type: Map <String, Object>
-* Description: This is for additional metadata and this does not have a
-  mandated structure. This enables a place for custom fields a producer or
-  middleware might want to include and provides a place to test metadata before
-  adding them to the CloudEvents specification. TBD - Determine a shorter
-  prefix for this (e.g. OpenAPI uses “x-”)
+### application-properties
+* Type: <Key Value>
```
duglin: I think it would be easier to just define this as a Map and say the "value" MUST be a "string" instead of an "Any". That would remove the need for most of the rules specified below because they would be implicit/obvious.  But if you do keep any of the rules then they should use MUST instead of "should".

spec.md
```diff
@@ -209,19 +209,25 @@ that contains both context and data).
...
-  * OPTIONAL
-  * If present, MUST contain at least one entry
+  * MANDATORY
+  * MUST be in the format of <key value> pair. 
 * Examples:
-  * authorization data
-
+  application-properties
+  { 
+     "building address": "12 Main Street, Copenhagen, Denmark, 123456"
```
duglin: missing `,` at the end of each line

Issue #108, opened 2018-03-01
Need a better description for schema-url


cathyhongzhang:
The description for schema-url is "A link to the schema that the data attribute adheres to". What is meant by "data attribute"?

duglin:
@cathyhongzhang can you keep your title a bit shorter and use the comment field for the longer description :-)  Its kind of awkward to have such long titles for issues.

cathyhongzhang:
The description for schema-url is "A link to the schema that the data attribute adheres to". What is meant by "data attribute"?

clemensv:
The [data attribute is this.](https://github.com/cloudevents/spec/blob/master/spec.md#data-1).

The schema-url points to a document that describes what the structure of the content of that property is. I think the name of the property is fairly good.

yaronha:
@clemensv i thought the schema field can be defining-org/schema-name, think of it like Class/Kind fields in kubernetes (e.g. `azure.microsoft.com/blobstore.update-message`), not a long/full blown URL which will add overhead to every message

clemensv:
@yaronha the field is surely optional. Whether it's worth including a "full blown URL" depends on the event type, how long those events hang around, how the event payload is encoded, and how far they travel.  If the event payload is encoded with Protobuf or Thrift or any other binary format that isn't self-contained, you literally need a schema copy to feed into your decoder. 

Separating a concrete schema hint from the message is certain recipe to having lost track of it when you need to pull the digitally signed event from an audit log in a year. 

Schema references are a distinct concern from the scoping given by "namespace". It is absolutely debatable whether having a schema reference expression is universally important enough for inclusion into the first version of this standard (I'd be ok with cutting it), but we shouldn't conflate its purpose with that of the "namespace" disambiguation function.

cathyhongzhang:
In this scheme definition, "the data attribute" is used. I am confused by the "data attribute" since "metadata attributes" is used in the definition of "context". In the "event" definition, it is stated that "Events include context and data". "contexts" is defined as "A set of consistent metadata attributes included with the event".  So I assume that "schema" here is about the structure of the data, not structure of the context. Suggest to change to "A link to the schema that defines the structure of the event data". @clemensv and others, if you are OK with this change, I can create a PR for this.

duglin:
I think some of our structure is a bit off. The spec says:
- Events include context and data (implying to me "context" and "data" are siblings in an Event object)
- Context is metadata about the Event (again, implying "context" is (sort of) a sibling to  "data")
[...]

PR #131, opened 2018-03-29
move data from "Event Context" to a "Event Data" section

cathyhongzhang:

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -222,6 +222,15 @@ that contains both context and data).
 * Examples:
   * authorization data
 
+## Event Data
+An event includes context and data. This "Event Data" section is the
+sibling to the above "Context Attributes". This is where the original
+event data info goes.
+
+In this section, each serialization/encoding scheme of a Cloud Event should 
+specify how to retrieve interoperability related information from the data
+or where those information appears in the event message.
+
 ## Additional Topics & Questions
 
 * Context Attribute Names - We decided not to spend too much time on property
--- a/spec.md
+++ b/spec.md
@@ -222,15 +222,6 @@ that contains both context and data).
 * Examples:
   * authorization data
 
-### data
-* Type: Arbitrary payload
-* Description: The event payload. The payload depends on the event-type,
-  schema-url and event-type-version, the payload is encoded into a media format
-  which is specified by the content-type attribute (e.g. application/json).
-* Constraints:
-  * OPTIONAL
-
-
 ## Additional Topics & Questions
 
 * Context Attribute Names - We decided not to spend too much time on property
--- a/spec.md
+++ b/spec.md
@@ -227,7 +227,7 @@ An event includes context and data. This "Event Data" section is the
 sibling to the above "Context Attributes". This is where the original
 event data info goes.
 
-In this section, each serialization/encoding scheme of a Cloud Event should 
+In this section, each serialization/encoding scheme of a Cloud Event should
 specify how to retrieve interoperability related information from the data
 or where those information appears in the event message.
 


duglin:
@cathyhongzhang I want to make sure I understand this. Are you saying that we should remove this attribute because its not a "context attribute", and instead we should just say in the serialization portion of our output docs where the actual event itself goes?

cathyhongzhang:
@duglin yes, it should not be a context attribute. Please see our discussion on Issue #108.

sslavic:
Why not move it to another list, e.g. rename the list to "Attributes" and have sublist of "Context attributes"?

ultrasaurus:
I am confused by this.  We certainly need to capture in the specification that there will be event-specific data.  I agree there could be a serialization that would put some attributes in transport-specific headers and others in a standard transport-specific body or alternately names data field, which could be addressed in serialization format.

duglin:
@cathyhongzhang I wonder if it would help if instead of a PR to just remove `data` your PR also included the other text that would need to appear elsewhere explaining where the event data itself would go.

For example, the definition of "Event" says: "Events include context and data."  And then a little bit after that we have a list of "Context Attributes". Perhaps you could add another section, as a sibling to "Context Attributes", called "Event Data" and write a paragraph or two talking about how this is the 2nd half of an "Event" and its where the original event info goes. And each serialization definition of a Cloud Event will explain how, and where, this "Data" would appear in the message.

WDYT?

cathyhongzhang:
@duglin Good suggestion! Will do to avoid confusion.

ultrasaurus:
A lot of events have a very small amount of data... why would we remove it?

cathyhongzhang:
@ultrasaurus  An event includes context and data. We have a "Context Attributes (Event Context)" section. As a sibling to "Context Attributes", we should have a "Event Data" section. 

This "Event Data" section is where the original event data info goes. In this section, each serialization/encoding scheme of a Cloud Event should specify how to get interoperability related information from the data or where those information appears in the event data message.
So I am adding such a section per Doug's suggestion. All data related information should be moved or added to this new section. 
But I am confused by existing definition of "data" and I do not know how that definition can be used. 
I am leaving that to someone else to clarify it and add it into this section.

spec.md
```diff
@@ -222,14 +222,14 @@ that contains both context and data).
...
-* Description: The event payload. The payload depends on the event-type,
-  schema-url and event-type-version, the payload is encoded into a media format
-  which is specified by the content-type attribute (e.g. application/json).
-* Constraints:
-  * OPTIONAL
-
+## Event Data
+An event includes context and data. This "Event Data" section is the
+sibling to the above "Context Attributes". This is where the original
+event data info goes.
```
duglin: LOL I didn't think you could use my words verbatim, I would have word-smithed it a little better.  :-)
How about:
```
The Data portion of a Cloud Event contains the original event data. The
exact format and location in which it will appear when serialized will vary
based on the serialization scheme being used.
```
I think this can probably replace both paragraphs.

ultrasaurus:
@cathyhongzhang I understand what you are trying to do now, which could be a clarification, but may have other implications.  Please rename the PR to "Move..." rather than "Delete..." that was confusing.  Also with "data" being a common attribute name in current representations, I think this may be doing a few things that may or may not be intended.

duglin:
rebase needed.
@cathyhongzhang what's the status of this one?

duglin:
@cathyhongzhang would you be ok if I took a stab at this one?

cathyhongzhang:
@duglin Sure.

PR #134, opened 2018-03-31
Remove "Additional Topics & Questions" section

duglin:
The "Additional Topics & Questions" section appears to be left over from the pre-WG work and I think it should be removed before v0.1. Since its just a list of suggested additional topics I'd like to recommend that we give people until the April 12th WG call to open an issue (or PR) for any item that they would like to see more discussion on. Otherwise, on April 12 we accept the PR and remove the section.  Note that we already have quite a few issues open (from Rachel and Louis)  that cover some of these topics.


ultrasaurus:
Maybe this should be moved somewhere, rather than deleted?

duglin:
@ultrasaurus that's why I suggested giving people 2 weeks to look it over. Anyone who thinks that there are things in that list we don't already have an issue for should open an issue for it. We don't need more than one "todo" or "to consider" list, that's what our github issues are for.

ultrasaurus:
Sorry, I should have elaborated.  This section has many things outside the scope of the specification.  I really would prefer not to have github issues or PRs for things that are beyond the current roadmap.  Although if you could suggest some kind of label or tagging process then that might make things easier to review.

duglin:
that's fine - a "future" or "beyond CE spec" type of label would work for me. I'll wait until someone opens up the first issue though to create it.

duglin:
Directional approval on 4/5 call. Please open an issue for anything in this list that you'd like to have future discussions on. We will vote on this PR next week.

duglin:
Approved during 4/12 call.

PR #135, opened 2018-03-31
Remove the Status section of the spec

duglin:
This section seems unnecessary and old...
- it talks about our scope, but I think we cover that in our "Overview" and "Design Goals" section
- it has a list of TODOs, which we've already done most of, and for the ones we're still working on they would be better tracked via Github issues (and they are)


duglin:
Approved on 4/5 call

PR #138, opened 2018-04-02
Introduce source-labels to help IoT cases.

inlined:
## Overview ##
Add the source's labels as an optional component for Events. This commit has been spliced off of https://github.com/cloudevents/spec/pull/129. For examples/more info on how labels are used today, see:
* [AWS Tags](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html)
* [GCP Labels](https://cloud.google.com/compute/docs/labeling-resources)
* [Azure Tags](https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-using-tags)
* [Docker Labels](https://docs.docker.com/config/labels-custom-metadata/)
* [Kubernetes labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/)

This should make the IoT case much more feasible and avoids a device choosing its correlation ID, which starts to edge back into the world of messaging again (since the sender is coupled
with the receiver). It is reasonable that the source itself is immutable (E.g. a device UUID) but that an Action would respond to a device's events based on where it is deployed. (E.g. a particular user's house).

Labels are a common concept in many existing products. They are set on the source and are orthogonal to any requests that are made against the resource. In the events use cases, they can be particularly useful for non-hierarchical filters or correlation of events. For example, [Istio uses labels to route traffic](https://istio.io/docs/concepts/traffic-management/rules-configuration.html).

### IoT Case ###

This case demonstrates an IoT vendor who may use UUIDs for the sensors they sell and use non-hierarchical attributes for filtering/routing events in an event-connected system. Note that the IoT device's source may not be addressable from the Action which receives the event data.

```json
{
  "source-authority": "iotvendor.net",
  "source-path": "12345678901234-1234-1234-1234-12345678",
  "source-labels": {
    "sensor-type": "vibration",
    "sensor-deployment": "house1234",
    "sensor-location": "window",
  },
  "event-type": "net.iotvendor.window.break"
}
```

### Hosted service ###

This use case focuses on a developer who extends their application by receiving events from a managed cloud service.

```json
{
  "source-authority": "storage.googleapis.com",
  "source-path": "projects/_/buckets/myBucket/objects/foo/bar.jpg",
  "source-labels": {
    "tier": "preprod"
  },
  "event-type": "google.storage.object.archive"
}
```


spec.md
```diff
@@ -313,6 +313,23 @@ that contains both context and data).
...
 
+### source-labels
+* Type: Map <String, String>
+* Description: Labels associated with the resource that emitted the event.
+  This allows filtering or routing based on non-hierarchical source metadata.
+  This is intended to include (but is not limited to) "labels" in cloud
+  platforms like AWS or Kubernetes.
+* Constraints:
+  * OPTIONAL
+  * Keys MUST match the regular expression `[_.a-z0-9]+`
```
duglin: Did you mean for the first char to be allowed to be a '.' ?

clemensv:
This proposal looks like a partial reversal of the consolidation of "source" that just went in.

If the context information is not already encoded in the `source` URI structure (as I discuss [on slide 4ff here](https://github.com/cloudevents/spec/blob/master/share/2018-03-22-TopicsAndSubjects.pptx); topic/subject are just other names for `source`, obvs), then these labels are a good use for the `extensions` object, or for inclusion in the app data.

The Azure tags that are referenced above are user-level grouping constructs for scripts and portal and we don't use them in our eventing infrastructure. They are a construct that allows users to arbitrarily group resources for scripting and in the portal and for billing different from the resource hierarchy.  If we were to include that information, it'd very likely go into the [event data section of the Resource Manager](https://docs.microsoft.com/en-us/azure/event-grid/event-schema-resource-groups) and not into Event Grid core envelope information. If we had to filter on it, we would add a mechanism exactly like the `extensions` object into which such metadata could be promoted by the publisher.

Bottom line: I think this is redundant with `extensions` - that said, I don't like the name `extensions` much myself. I'd prefer something more neutral like `properties`.

duglin:
re: `source-labels` vs `extensions`, see a similar concern here: https://github.com/cloudevents/spec/pull/129#issuecomment-376565023

If we adopt this, we'll have 3 different spots where "other stuff" can go: 1) `extensions`, 2) `source-labels`, and 3) extra top-level json properties (even though we haven't talked about this one yet, until/if we ban it I think people will use it).

I'm trying to think about someone new coming to the spec and looking at this and wondering why we have labels for `source` but not other things like `event-type`. Or when does something go in one of those 3 buckets but not the other? What happens when its in the wrong bucket - does it have a different semantic meaning? Likewise, can the same property name appear in more than one bucket and have different meaning? Probably, but would that be confusing?

On the flip side, K8s has `annotations` and `labels`. I've heard some people say its good to have this split, while others have been confused by it.

No firm position yet, but I am worried about the proliferation of extension points right next to each other.

ultrasaurus:
Extensions were promoted as a place for experimental attributes that would be later (potentially) promoted.  Labels that the source provides are authoritative for that source.  These are completely different concepts and there seems to be quite a bit of evidence of need for this from the IoT use case presentations as well as the references in the overview above.

spec.md
nerdyyatrice: looks like a typo?

duglin: @inlined did you want to fix this?

duglin:
rebase needed

duglin:
I'm not doubting the usefulness of having extra fields to help with routing, but I am wondering why adding:
```
"sourceLabels": {
  "sensor.configuration": "WINDOW"
}
```
is better than:
```
```

inlined:
Did a few small changes to keep the change current:

* I dropped the "source" prefix.
* I updated the examples in the PR description.
* I published a braindump of research into using trace contexts for mutable attributes in #197 

The WG has done well by consolidating on baby steps of agreement and then moving forward. I think there has been a majority view that at least source-provided labels are good. There was a suggestion to call this "properties" instead of "labels;" I have a personal preference to reuse Kubernetes terminology, though I'm happy to use a quick vote to decide.

I think gateway-added labels/attributes/etc. are a valid second discussion. Our options are to either:
1. Create a second, mutable, property bag (proposed: "properties" vs "attributes")
2. Define this attribute to be mutable

I lean moderately towards mutability. Two property bags may not feel as forward compatible (e.g. in the IoT use case, we allow a gateway to be the event source, but what if v2 devices are more powerful and capable of emitting their own events? Should they not for reverse compatibility?) Also, since our example mildly creeped me out, I think a privacy filtering gateway could totally be in spec. FWIW, we had the instinct to look at Distributed Tracing's context object as an inspiration, and that spec has very clear instructions for how software may update even opaque property bags from other clouds.

spec.md
```diff
@@ -328,6 +328,25 @@ that contains both context and data).
 * Constraints:
   * REQUIRED
 
+### labels
+* Type: `Map`
+* Description: Non-heirarchical attributes which describe the `source` of
```
duglin: Do we have to limit it to describing the `source` instead of the `event` in general?

duglin:
To add a little more... I view a lot of our work as being similar to how HTTP defined a set of headers to aid in interop/routing/etc...  and yet they don't create "bags" or "buckets" for extensions. They just let you add new top-level headers as long as there is no name conflict.  So, its not clear to me what value we're adding by trying to force people to pick the appropriate extension point when in the end all people want to do is know if that extension is in the event - and not where it appears. Like I said, if location of the extension changes its semantics then I fear we're adding unnecessary complexity.

PR #139, opened 2018-04-04
add Attribute Categories section

markfisher:
The added section describes the different categories of attributes (required, optional, and extensions) and provides general criteria for an attribute's inclusion in a given category.

The text was originally based off of this suggestion from @duglin: https://github.com/cloudevents/spec/compare/master...duglin:propguidelines

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -194,6 +194,31 @@ concurrently, including being both a producer and a consumer of events.
    For this, the framework will need a suitable metadata discriminator
    that frees it from having to understand the event details.
 
+## Attribute Categories
+
+The number of metadata attributes will be constrained to the minimal set needed
+to achieve the stated goals. To that end, attributes defined by this
+specification will fall into three categories:
+- required
+- optional
+- extensions
+
+As the category names imply, "required" attributes will be the ones that are
+considered vital to all events in all use cases, while "optional" ones will be
+used in a majority of use cases. If an attribute is not common enough to fall
+into those two categories but would still benefit from the level of
+interoperability that comes from being well-defined, then it will be placed into
+the "extensions" category and described in the (extensions)[extensions.md]
+document. The spec will define how these extension attributes will appear within
+a CloudEvent.
+
+Core attributes, whether required or optional, are those that describe the event
+itself. For example those attributes may relate to the source, the content type,
+the time the event occurred, or any identifying information. Attributes that are
+relevant only for certain contexts within which the event may be used will be in
+the extensions category. Examples might include grouping, sampling, correlation,
+and tracing.
+
 ## Status
 At this time the specification is focused on the following scope:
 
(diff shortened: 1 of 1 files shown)


duglin:
@markfisher you'll need to sign your commit so the DCO check passes

duglin:
We're gonna have to figure out how to deal with the use of RFC keywords.... I wonder if an easy out is to put this into the README, or even into the extensions.md doc.  Need to think about this one.

spec.md
```diff
@@ -194,6 +194,31 @@ concurrently, including being both a producer and a consumer of events.
...
    that frees it from having to understand the event details.
 
+## Attribute Categories
+
+The number of metadata attributes will be constrained to the minimal set needed
+to achieve the stated goals. To that end, attributes defined by this
+specification will fall into three categories:
+- required
+- optional
+- extensions
```
inlined: nit: we seem to be dividing extensions into two categories as well:

* Registered extensions that vendors publish to encourage interoperability of a use case
* Private extensions

This concept was taken directly from the JWT spec where [private claims may be registered claims](https://tools.ietf.org/html/rfc7519#section-4.1).

spec.md
```diff
@@ -194,6 +194,31 @@ concurrently, including being both a producer and a consumer of events.
...
+interoperability that comes from being well-defined, then it will be placed into
+the "extensions" category and described in the (extensions)[extensions.md]
+document. The spec will define how these extension attributes will appear within
+a CloudEvent.
+
+Core attributes, whether required or optional, are those that describe the event
+itself. For example those attributes may relate to the source, the content type,
+the time the event occurred, or any identifying information. Attributes that are
+relevant only for certain contexts within which the event may be used will be in
+the extensions category. Examples might include grouping, sampling, correlation,
```
inlined: I'm not sure I take for granted that these are extensions cases, and if they were they should certainly be registered extensions. These are all cases where multiple parts of a distributed stack need to interoperate to work correctly.

spec.md
duglin: true - I guess originally I was thinking that we'd just talk about the stuff that might appear in our docs, but you're right, we should mention that people can do "other extensions" that this WG has no knowledge of.

duglin:
@markfisher will you have time to finish this one up? if not, I can try to find someone else to run with it.

markfisher:
@duglin I can handle it. Did you decide if you'd prefer to keep it here but avoid RFC keywords or whether we should move it to the README?

duglin:
My preference would be to move it out of the spec since explaining what goes into each doc is broader than that one doc (meaning the spec).

duglin:
Perhaps we need some kind of Primer doc as an intro that explains all of the pieces.

duglin:
actually a Primer doc would be a good place for the Usage Scenarios section: https://github.com/cloudevents/spec/blob/master/spec.md#usage-scenarios  - it feels out of place in the spec to me

markfisher:
@duglin if you want to create an initial Primer doc, I can resubmit the PR to add to that instead.

duglin:
if #238 is adopted then I think it should cover this

markfisher:
agreed @duglin

PR #152, opened 2018-04-06
Attribute type system

clemensv:
This PR adds an explicit declaration of the types available for attributes. 

The `data` attribute is modified to be of the  `Object` type, which can be a string, a binary, or a map.


spec.md
```diff
@@ -300,23 +319,23 @@ that contains both context and data).
...
+* Type: `Timestamp` per [RFC 3339](https://tools.ietf.org/html/rfc3339)
 * Description: Timestamp of when the event happened.
 * Constraints:
   * OPTIONAL
   * If present, MUST adhere to the format specified in
     [RFC 3339](https://tools.ietf.org/html/rfc3339)
 
 ### schema-url
-* Type: URI per [RFC 3986](https://tools.ietf.org/html/rfc3986)
+* Type: `URI` per [RFC 3986](https://tools.ietf.org/html/rfc3986)
```
SeanFeldman: Redundant RFC reference since definition of `URI` already contains a reference?

spec.md
```diff
@@ -300,23 +319,23 @@ that contains both context and data).
...
 * Description: A link to the schema that the `data` attribute adheres to.
 * Constraints:
   * OPTIONAL
   * If present, MUST adhere to the format specified in
     [RFC 3986](https://tools.ietf.org/html/rfc3986)
 
 ### content-type
-* Type: String per [RFC 2046](https://tools.ietf.org/html/rfc2046)
+* Type: `String` per [RFC 2046](https://tools.ietf.org/html/rfc2046)
```
SeanFeldman: For consistency sake, should RFC reference be moved up to the `String` definition?

spec.md
```diff
@@ -240,6 +240,25 @@ Messages can be delivered through various industry standard protocol (e.g. HTTP,
...
+- `String` - Sequence of printable Unicode characters.
+- `Binary` - Sequence of bytes.
+- `Map` - `String`-indexed dictionary of `Object`-typed values
+- `Object` - Either a `String`, or a `Binary`, or a `Map`
+- `URI` - String expression as defined in
+  [RFC 3986](https://tools.ietf.org/html/rfc3986)
+- `Timestamp` - String expression as defined in
+  [RFC 3339](https://tools.ietf.org/html/rfc3339)
+
+This specification does not define numeric or logical types.
```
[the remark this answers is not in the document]
clemensv: Only if an urgent need arises, IMO

duglin:
@clemensv did you want to address @SeanFeldman's comments?

spec.md
clemensv: This is a String. And it need to conform to RFC2046 in this instance. So this one is correct,

spec.md
clemensv: I didn't want to touch that bit, but ok.

clemensv:
Addressed @SeanFeldman's comments by dropping redundant definitions.

duglin:
LGTM

spec.md
```diff
@@ -240,6 +240,25 @@ Messages can be delivered through various industry standard protocol (e.g. HTTP,
...
 platform/vendor specific protocols (AWS Kinesis, Azure Event Grid).
 
+## Type System
+
+The following abstract data types are available for use in attributes.
+
+- `String` - Sequence of printable Unicode characters.
+- `Binary` - Sequence of bytes.
+- `Map` - `String`-indexed dictionary of `Object`-typed values
+- `Object` - Either a `String`, or a `Binary`, or a `Map`
```
ultrasaurus: The term `Object` is a bit confusing, since it has a very specific meaning in JavaScript and JSON will be a common encoding

clemensv: The [closest JSON equivalent for this is "value"](https://tools.ietf.org/html/rfc7159#section-3), which I would find even more confusing for mapping into other type systems. 

Since JSON and JavaScript are dynamically typed; there's simply no explicit "could be anything" type. TypeScript therefore [introduces "Any"](http://www.typescriptlang.org/docs/handbook/basic-types.html) 

`Object` is the generic "could be anything" base type in Java, C#, and Python, Golang has an Object Interface, etcetc. 

I'd prefer to align with the concepts of a lot of common languages rather than go out of the way of a particular language, especially also because this is [explicitly addressed in the JSON binding](https://github.com/cloudevents/spec/pull/148/files#diff-541c0c5dfe73eb8ec79427b6bab66748R61)

duglin:
Per our 4/5 call we agreed to tentatively accept this PR for v0.1 assuming it was not controversial. Its been open for 11 days and aside from the  choice of the word "Object" (which if we change it is more of a syntax change then a semantic change) there hasn't been any concerns. So I'm going to place it in the v0.1 milestone bucket.

duglin:
Oops, never mind it already is.  Not sure how I missed it.

spec.md
inlined: My big issue was that we seemed to take a step back when Maps no longer specified what their value-type was. Is there a reason not to templatize Map still?

deissnerk:
LGTM

spec.md
ultrasaurus: I prefer `any` or `variant` to `Object`

PR #154, opened 2018-04-11
Add SHOULD constraint that source has an authority.

inlined:
Events which contain an authority will be easier to route
(authority-specific subscriptions) or process.

It is intentionally undefined what the authority would contain.
It is possible that the authority references an private or public IP
addres, a gateway host, etc.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -287,6 +287,8 @@ that contains both context and data).
   semantics behind the data encoded in the URI is event producer defined.
 * Constraints:
   * REQUIRED
+  * SHOULD include an [authority
+    component](https://tools.ietf.org/html/rfc3986#section-3.2)
 
 ### event-id
 * Type: String


clemensv:
I don't understand how that extra constraint helps with interoperability; "Authority" is not meaningful for a variety of URIs. [RFC3986](https://tools.ietf.org/html/rfc3986#section-1.1.3) is very clear in that URIs can be mere names, and there is [an extensive discussion in the RFC itself about the fact that URIs don't necessarily identify system assets](https://tools.ietf.org/html/rfc3986#section-1.2.2).

inlined:
I definitely misspoke in the meeting; I was spending so much time focused in the middle of the spec that I missed the top level clarification that scheme is necessary. I'll chat with others at Google to figure out how they'd recommend representing our resources in a real URI since it's not clear what scheme to use our resources are accessed on multiple protocols.

PR #155, opened 2018-04-12
HTTP WebHook Specification

clemensv:
This PR complements https://github.com/cloudevents/spec/pull/148 

This specification is a *generic* specification that formalizes delivery of notifications over HTTP, generally known as "Web hooks". The HTTP transport binding for Cloud Events composes with this specification for one-way event delivery.

The specification defines that notifications are delivered by HTTP POST, should use a token-based authNZ scheme either carried in the Authorization header or the query string (whereby "token" can be a simple key), and strongly suggests implementation of a pre-flight handshake (similar to [W3C CORS](www.w3.org/TR/cors/)) that protects a sender from being abused as a distributed denial of service machine gun.

This specification would probably be best proposed to IETF, but here's a good place to start with the draft.

This is not relevant to the 0.1 of the core spec, but is relevant to the interop group

### DIFF
--- a/http-webhook.md
+++ b/http-webhook.md
@@ -68,7 +68,7 @@ This specification does not further constrain the content of the notification,
 and it also does not prescribe the [HTTP target resource][RFC7230-Section-5-1]
 that is used for delivery.
 
-If the delivery target supports and requires [Abuse Protection][#4-abuse-protection],
+If the delivery target supports and requires [Abuse Protection](#4-abuse-protection),
 the delivery request MUST include the `Origin` header. The `Origin` header value
 is a DNS name expression that identifies the sending system.
 
@@ -116,8 +116,10 @@ lean on the OAuth 2.0 Bearer Token [RFC6750][RFC6750] model.
 
 The delivery target MUST support both methods.
 
-The client MAY use any token-based authorization scheme. Challenge-based
-schemes MUST NOT be used.
+The client MAY use any token-based authorization scheme. The token can take any shape,
+and can be a standardized token format or a simple key expression.
+
+Challenge-based schemes MUST NOT be used.
 
 ### 3.1.  Authorization Request Header Field
 
@@ -360,6 +362,7 @@ WebHook-Allowed-Rate: 100
 [429]: https://tools.ietf.org/html/rfc6585#section-4
 [Bearer]:https://tools.ietf.org/html/rfc6750#section-2.1
 [RFC2119]: https://tools.ietf.org/html/rfc2119
+[RFC3986]: https://tools.ietf.org/html/rfc3986
 [RFC2818]: https://tools.ietf.org/html/rfc2818
 [RFC6585]: https://tools.ietf.org/html/rfc6585
 [RFC6750]: https://tools.ietf.org/html/rfc6750
--- a/http-webhook.md
+++ b/http-webhook.md
@@ -1,0 +1,344 @@
+# HTTP 1.1 Web Hooks for Event Delivery
+
+## Abstract
+
+"Web hooks" are a popular pattern to deliver notifications between applications
+and via HTTP endpoints. In spite of pattern usage being widespread, there is no
+formal definition for Web Hooks. This specification aims to provide such a
+definition for use with [CNCF Cloud Events][CE], but is considered generally
+usable beyond the scope of Cloud Events.
+
+## Status of this document
+
+This document is a working draft.
+
+## Table of Contents
+
+1. [Introduction](#1-introduction)
+- 1.1. [Conformance](#11-conformance)
+- 1.2. [Relation to HTTP](#12-relation-to-http)
+- 1.3. [Security](#15-security)
+2. [Delivering notifications](#2-delivering-notifications)
+3. [Authorization](#3-authorization)
+4. [Abuse Protection](#4-abuse-protection)
+5. [References](#4-references)
+
+## 1. Introduction
+
+"Web hooks" are a popular pattern to deliver notifications between applications
+and via HTTP endpoints. Applications that make notifications available, allow
+for other applications to register an HTTP endpoint to which notifications are
+delivered once available.
+
+This specification defines a HTTP method by how notifications are delivered by
+the sender, an authorization model for event delivery to protect the receiver,
+and a registration handshake that protects the sender from being abused for
+flooding arbitrary HTTP sites with requests.
+
+### 1.1. Conformance
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in [RFC2119][RFC2119].
+
+### 1.2. Relation to HTTP
+
+This specification prescribes rules constraining the use and handling of
+specific [HTTP methods][RFC7231-Section-4] and headers.
+
+This specification also applies equivalently to HTTP/2 ([RFC7540][RFC7540]),
+which is compatible with HTTP 1.1 semantics.
+
+## 2. Delivering notifications
+
+### 2.1. Delivery request
+
+Notifications are delivered using a HTTP request. The response indicates the
+resulting status of the delivery.
+
+HTTP-over-TLS (HTTPS) [RFC2818][RFC2818] MUST be used for the connection.
+
+The HTTP method for the delivery request MUST be [POST][POST].
+
+The [`Content-Type`][Content-Type] header MUST be carried and the request MUST
+carry a notification payload of the given content type. Requests without
+payloads, e.g. where the notification is entirely expressed in HTTP headers,
+are not permitted by this specification.
+
+This specification does not further constrain the content of the notification,
+and it also does not prescribe the [HTTP target resource][RFC7230-Section-5-1]
+that is used for delivery.
+
+If the delivery target supports and requires [Abuse Protection][#4-abuse-protection],
+the delivery request MUST include the `Origin` header. The `Origin` header value
+is a DNS name expression that identifies the sending system.
+
+### 2.2. Delivery response
+
+The delivery response MAY contain a payload providing detail status
+information in the case of handling errors. This specification does not attempt
+to define such a payload.
+
+The response MUST NOT use any of the [3xx HTTP Redirect status codes][3xx] and
+the client MUST NOT follow any such redirection.
+
+If a delivery target has been retired, but the HTTP site still exists, the
+site SHOULD return a [410 "Gone"][410] status code and the sender SHOULD refrain
+from sending any further notifications.
+
+If the delivery target is unable to process the request due to exceeding a
+request rate limit, it SHOULD return a [429 Too Many Requests][429] status code
+and MUST include the [`Retry-After`][Retry-After] header. The sender MUST
+observe the value of the Retry-After header and refrain from sending further
+requests until the indicated time.
+
+If the delivery has been accepted and processed, and if the response carries a
+payload with processing details, the response MUST have the [200 OK][200]
+status code. In this case, the response MUST carry a
+[`Content-Type`][Content-Type] header.
+
+If the delivery has been accepted and processed, but carries no payload, the
+response MUST have the [204 No Content][204] status code.
+
+If the delivery has been accepted, but has not yet been processed or if the
+processing status is unknown, the response MUST have the [202 Accepted][202]
+status code.
+
+## 3. Authorization
+
+The delivery request MUST use one of the following two methods, both of which
+lean on the OAuth 2.0 Bearer Token [RFC6750][RFC6750] model.
+
+The delivery target MUST support both methods.
+
+The client MAY use any token-based authorization scheme. Challenge-based
+schemes MUST NOT be used.
+
+### 3.1.  Authorization Request Header Field
+
+The access token is sent in the [`Authorization`][Authorization] request
+header field defined by HTTP/1.1.
+
+For [OAuth 2.0 Bearer][Bearer] tokens, the "Bearer" scheme MUST be used.
+
+Example:
+
+``` text
+GET /resource HTTP/1.1
+Host: server.example.com
+Authorization: Bearer mF_9.B5f-4.1JqM
+```
+
+### 3.2 URI Query Parameter
+
+When sending the access token in the HTTP request URI, the client adds the
+access token to the request URI query component as defined by "Uniform Resource
+Identifier (URI): Generic Syntax" [RFC3986][RFC3986], using the "access_token"
+parameter.
+
+For example, the client makes the following HTTP request:
+
+``` text
+POST /resource?access_token=mF_9.B5f-4.1JqM HTTP/1.1
+Host: server.example.com
+```
+
+The HTTP request URI query can include other request-specific parameters, in
+which case the "access_token" parameter MUST be properly separated from the
+request-specific parameters using "&" character(s) (ASCII code 38).
+
+For example:
+
+    https://server.example.com/resource?access_token=mF_9.B5f-4.1JqM&p=q
+
+Clients using the URI Query Parameter method SHOULD also send a
+Cache-Control header containing the "no-store" option.  Server
+success (2XX status) responses to these requests SHOULD contain a
+Cache-Control header with the "private" option.
+
+Because of the security weaknesses associated with the URI method (see
+[RFC6750, Section 5][RFC6750]), including the high likelihood that the URL
+containing the access token will be logged, it SHOULD NOT be used unless it is
+impossible to transport the access token in the "Authorization" request header
+field or the HTTP request entity-body. All further caveats cited in
+[RFC6750][RFC6750] apply equivalently.
+
+## 4. Abuse Protection
+
+Any system that allows registration of and delivery of notifications to
+arbitary HTTP endpoints can potentially be abused such that someone maliciously
+or inadvertently registers the address of a system that does not expect such
+requests and for which the registering party is not authorized to perform such
+a registration. In extreme cases, a notification infrastructure could be abused
+to launch denial-of-service attacks against an arbitrary web-site.
+
+To protect the sender from being abused in such a way, a legitimate delivery
+target needs to indicate that it agrees with notifications being delivered to it.
+
+Reaching the delivery agreement is realized using the following validation
+handshake. The handshake can either be executed immediately at registration
+time or as a "pre-flight" request immediately preceding a delivery. If the
+handshake outcome is that the delivery is denied, the sender MUST NOT deliver
+events to the target.
+
+It is important to understand is that the handshake does not aim to establish
+an authentication or authorization context. It only serves to protect the
+sender from being told to a push to a destination that is not expecting the
+traffic. While this specification mandates use of an authorization model, this
+mandate is not sufficient to protect any arbitrary website from unwanted
+traffic if that website doesn't implement access control and therefore ignores
+the `Authorization` header.
+
+Delivery targets SHOULD support the abuse protection feature. If a target does
+not support the feature, the sender MAY choose not to send to the target, at
+all, or send only at a very low request rate.
+
+### 4.1.1. Validation request
+
+The validation request uses the HTTP [OPTIONS][OPTIONS] method. The request
+is directed to the exact resource target URI that is being registered.
+
+With the validation request, the sender asks the target for permission to
+send notifications, and it can declare a desired request rate (requests per
+minute).
+
+The delivery target will respond with a permission statement and the permitted
+request rate.
+
+The following header fields MUST be included in the validation request:
+
+#### 4.1.1.2. WebHook-Request-Origin
+
+The `WebHook-Request-Origin` header requests permission to send notifications
+from this sender, and contains a DNS expression that identifies the sending
+system, for example "eventemitter.example.com". The value is meant to summarily
+identify all sender instances that act on the behalf of a certain system, and
+not an individual host.
+
+After the handshake and when permission is granted, the sender MUST use the
+`Origin` request header for each delivery request, with the value matching that
+of this header.
+
+For example:
+
+``` text
+WebHook-Request-Origin: eventemitter.example.com
+```
+
+#### 4.1.1.2. WebHook-Request-Rate
+
+The `WebHook-Request-Rate` header MAY be included in the request and asks for
+permission to send notifications from this sender at the specified rate. The
+value is the string representation of a positive integer number greater than
+zero and expresses the request rate in "requests per minute".
+
+For example, the following header asks for permission to send 120 requests per
+minute:
+
+``` text
+WebHook-Request-Rate: 120
+```
+
+### 4.1.2 Validation response
+
+If and only if the delivery target does allow delivery of the events, it MUST
+reply to the request by including the `WebHook-Allowed-Origin` and
+`WebHook-Allowed-Rate` headers.
+
+If it does not allow delivery of the events or does not expect delivery of
+events and nevertheless handles the HTTP OPTIONS method, the existing response
+ought not to be interpreted as consent, and therefore the handshake cannot
+rely on status codes.
+
+The response SHOULD include the [Allow][Allow] header indicating the [POST][POST]
+method being permitted. Other methods MAY be permitted on the resource, but
+their function is outside the scope of this specification.
+
+#### 4.1.1.2. WebHook-Allowed-Origin
+
+The `WebHook-Allowed-Origin` header MUST be returned when the delivery target
+agrees to notification delivery by the origin service. Its value MUST either be
+the origin name supplied in the `WebHook-Request-Origin` header, or a singular
+asterisk character ('*'), indicating that the delivery target supports
+notifications from all origins.
+
+``` text
+WebHook-Allowed-Origin: eventemitter.example.com
+```
+
+or
+
+``` text
+WebHook-Request-Origin: *
+```
+
+#### 4.1.1.2. WebHook-Allowed-Rate
+
+The `WebHook-Allowed-Rate` header MUST be returned if the request contained
+teh `WebHook-Request-Rate`, otherwise it SHOULD be returned.
+
+The header grants permission to send notifications at the specified rate. The
+value is either an asterisk character or the string representation of a
+positive integer number greater than zero. The asterisk indicates that there is
+no rate limitation. An integer number expresses the permitted request rate in
+"requests per minute". For request rates exceeding the granted notification
+rate, the sender ought to expect request throttling. Throttling is indicated
+by requests being rejected using HTTP status code [429][429].
+
+For example, the following header permits to send 100 requests per
+minute:
+
+``` text
+WebHook-Allowed-Rate: 100
+```
+
+## 5. References
+
+- [RFC2046][RFC2046] Multipurpose Internet Mail Extensions (MIME) Part Two: 
+  Media Types
+- [RFC2119][RFC2119] Key words for use in RFCs to Indicate Requirement Levels
+- [RFC2818][RFC2818] HTTP over TLS
+- [RFC3629][RFC3629] UTF-8, a transformation format of ISO 10646
+- [RFC3986][RFC3986] Uniform Resource Identifier (URI): Generic Syntax 
+- [RFC4627][RFC4627] The application/json Media Type for JavaScript Object
+  Notation (JSON)
+- [RFC4648][RFC4648] The Base16, Base32, and Base64 Data Encodings
+- [RFC6750][RFC6750] The OAuth 2.0 Authorization Framework: Bearer Token Usage
+- [RFC6585][RFC6585] Additional HTTP Status Codes
+- [RFC6839][RFC6839] Additional Media Type Structured Syntax Suffixes
+- [RFC7159][RFC7159] The JavaScript Object Notation (JSON) Data Interchange Format
+- [RFC7230][RFC7230] Hypertext Transfer Protocol (HTTP/1.1): Message Syntax
+  and Routing
+- [RFC7231][RFC7231] Hypertext Transfer Protocol (HTTP/1.1): Semantics and
+  Content
+- [RFC7235][RFC7235] Hypertext Transfer Protocol (HTTP/1.1): Authentication
+- [RFC7540][RFC7540] Hypertext Transfer Protocol Version 2 (HTTP/2)
+
+[CE]: ./spec.md
+[JSON-format]: ./json-format.md
+[Content-Type]: https://tools.ietf.org/html/rfc7231#section-3.1.1.5
+[Retry-After]: https://tools.ietf.org/html/rfc7231#section-7.1.3
+[Authorization]: https://tools.ietf.org/html/rfc7235#section-4.2
+[Allow]: https://tools.ietf.org/html/rfc7231#section-7.4.1
+[POST]: https://tools.ietf.org/html/rfc7231#section-4.3.3
+[OPTIONS]: https://tools.ietf.org/html/rfc7231#section-4.3.7
+[3xx]: https://tools.ietf.org/html/rfc7231#section-6.4
+[429]: https://tools.ietf.org/html/rfc6585#section-4
+[Bearer]:https://tools.ietf.org/html/rfc6750#section-2.1
+[RFC2046]: https://tools.ietf.org/html/rfc2046
+[RFC2119]: https://tools.ietf.org/html/rfc2119
+[RFC2818]: https://tools.ietf.org/html/rfc2818
+[RFC3629]: https://tools.ietf.org/html/rfc3629
+[RFC3986]: https://tools.ietf.org/html/rfc3986
+[RFC4627]: https://tools.ietf.org/html/rfc4627
+[RFC4648]: https://tools.ietf.org/html/rfc4648
+[RFC6585]: https://tools.ietf.org/html/rfc6585
+[RFC6839]: https://tools.ietf.org/html/rfc6839#section-3.1
+[RFC7159]: https://tools.ietf.org/html/rfc7159
+[RFC7230]: https://tools.ietf.org/html/rfc7230
+[RFC7231]: https://tools.ietf.org/html/rfc7231
+[RFC7235]: https://tools.ietf.org/html/rfc7235
+[RFC7230-Section-3]: https://tools.ietf.org/html/rfc7230#section-3
+[RFC7231-Section-4]: https://tools.ietf.org/html/rfc7231#section-4
+[RFC7230-Section-5-1]: https://tools.ietf.org/html/rfc7230#section-5.1
+[RFC7540]: https://tools.ietf.org/html/rfc7540
--- a/http-webhook.md
+++ b/http-webhook.md
@@ -305,6 +305,7 @@ WebHook-Allowed-Rate: 100
 - [RFC7540][RFC7540] Hypertext Transfer Protocol Version 2 (HTTP/2)
 
 [CE]: ./spec.md
+[Content-Type]: https://tools.ietf.org/html/rfc7231#section-3.1.1.5
 [Retry-After]: https://tools.ietf.org/html/rfc7231#section-7.1.3
 [Authorization]: https://tools.ietf.org/html/rfc7235#section-4.2
 [Allow]: https://tools.ietf.org/html/rfc7231#section-7.4.1
--- a/http-webhook.md
+++ b/http-webhook.md
@@ -20,7 +20,7 @@ This document is a working draft.
 2. [Delivering notifications](#2-delivering-notifications)
 3. [Authorization](#3-authorization)
 4. [Abuse Protection](#4-abuse-protection)
-5. [References](#4-references)
+5. [References](#5-references)
 
 ## 1. Introduction
 
--- a/http-webhook.md
+++ b/http-webhook.md
@@ -103,6 +103,11 @@ If the delivery has been accepted, but has not yet been processed or if the
 processing status is unknown, the response MUST have the [202 Accepted][202]
 status code.
 
+If the delivery cannot be accepted because the notification format has not
+been understood, the service MUST respond with status code [415][415].
+
+All further error status codes apply as specified in [RFC7231][RFC7231].
+
 ## 3. Authorization
 
 The delivery request MUST use one of the following two methods, both of which
@@ -317,6 +322,7 @@ WebHook-Allowed-Rate: 100
 [202]: https://tools.ietf.org/html/rfc7231#section-6.3.3
 [204]: https://tools.ietf.org/html/rfc7231#section-6.3.5
 [410]:https://tools.ietf.org/html/rfc7231#section-6.5.9
+[415]: https://tools.ietf.org/html/rfc7231#section-6.5.13
 [429]: https://tools.ietf.org/html/rfc6585#section-4
 [Bearer]:https://tools.ietf.org/html/rfc6750#section-2.1
 [RFC2119]: https://tools.ietf.org/html/rfc2119
--- a/http-webhook.md
+++ b/http-webhook.md
@@ -315,6 +315,7 @@ WebHook-Allowed-Rate: 100
 [Bearer]:https://tools.ietf.org/html/rfc6750#section-2.1
 [RFC2119]: https://tools.ietf.org/html/rfc2119
 [RFC2818]: https://tools.ietf.org/html/rfc2818
+[RFC6585]: https://tools.ietf.org/html/rfc6585
 [RFC6750]: https://tools.ietf.org/html/rfc6750
 [RFC7159]: https://tools.ietf.org/html/rfc7159
 [RFC7230]: https://tools.ietf.org/html/rfc7230
--- a/http-webhook.md
+++ b/http-webhook.md
@@ -5,8 +5,8 @@
 "Web hooks" are a popular pattern to deliver notifications between applications
 and via HTTP endpoints. In spite of pattern usage being widespread, there is no
 formal definition for Web Hooks. This specification aims to provide such a
-definition for use with [CNCF Cloud Events][CE], but is considered generally
-usable beyond the scope of Cloud Events.
+definition for use with [CNCF CloudEvents][CE], but is considered generally
+usable beyond the scope of CloudEvents.
 
 ## Status of this document
 
@@ -92,12 +92,12 @@ observe the value of the Retry-After header and refrain from sending further
 requests until the indicated time.
 
 If the delivery has been accepted and processed, and if the response carries a
-payload with processing details, the response MUST have the [200 OK][200]
-status code. In this case, the response MUST carry a
+payload with processing details, the response MUST have the [200 OK][200] or
+[201 Created][201] status code. In this case, the response MUST carry a
 [`Content-Type`][Content-Type] header.
 
 If the delivery has been accepted and processed, but carries no payload, the
-response MUST have the [204 No Content][204] status code.
+response MUST have the [201 Created][201] or [204 No Content][204] status code.
 
 If the delivery has been accepted, but has not yet been processed or if the
 processing status is unknown, the response MUST have the [202 Accepted][202]
@@ -313,6 +313,7 @@ WebHook-Allowed-Rate: 100
 [OPTIONS]: https://tools.ietf.org/html/rfc7231#section-4.3.7
 [3xx]: https://tools.ietf.org/html/rfc7231#section-6.4
 [200]: https://tools.ietf.org/html/rfc7231#section-6.3.1
+[201]: https://tools.ietf.org/html/rfc7231#section-6.3.2
 [202]: https://tools.ietf.org/html/rfc7231#section-6.3.3
 [204]: https://tools.ietf.org/html/rfc7231#section-6.3.5
 [410]:https://tools.ietf.org/html/rfc7231#section-6.5.9
--- a/http-webhook.md
+++ b/http-webhook.md
@@ -2,7 +2,7 @@
 
 ## Abstract
 
-"Web hooks" are a popular pattern to deliver notifications between applications
+"Webhooks" are a popular pattern to deliver notifications between applications
 and via HTTP endpoints. In spite of pattern usage being widespread, there is no
 formal definition for Web Hooks. This specification aims to provide such a
 definition for use with [CNCF CloudEvents][CE], but is considered generally
@@ -24,15 +24,15 @@ This document is a working draft.
 
 ## 1. Introduction
 
-"Web hooks" are a popular pattern to deliver notifications between applications
-and via HTTP endpoints. Applications that make notifications available, allow
-for other applications to register an HTTP endpoint to which notifications are
-delivered once available.
+["Webhooks"][Webhooks] are a popular pattern to deliver notifications between
+applications and via HTTP endpoints. Applications that make notifications
+available, allow for other applications to register an HTTP endpoint to which
+notifications are delivered.
 
 This specification defines a HTTP method by how notifications are delivered by
-the sender, an authorization model for event delivery to protect the receiver,
-and a registration handshake that protects the sender from being abused for
-flooding arbitrary HTTP sites with requests.
+the sender, an authorization model for event delivery to protect the delivery
+target, and a registration handshake that protects the sender from being abused
+for flooding arbitrary HTTP sites with requests.
 
 ### 1.1. Conformance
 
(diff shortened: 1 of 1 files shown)


http-webhook.md
```diff
@@ -0,0 +1,332 @@
...
+the delivery request MUST include the `Origin` header. The `Origin` header value
+is a DNS name expression that identifies the sending system.
+
+### 2.2. Delivery response
+
+The delivery response MAY contain a payload providing detail status
+information in the case of handling errors. This specification does not attempt
+to define such a payload.
+
+The response MUST NOT use any of the [3xx HTTP Redirect status codes][3xx] and
```
sslavic: Curious what's the rationale for this constraint? To reduce scope, make spec initially more focused?

sslavic:
Would it make sense, in addition to the explaining it with lots of text, to include also OpenAPI based contract stub of such notification endpoint, where HTTP request/message is CloudEvents with JSON binding compatible one?

http-webhook.md
```diff
@@ -0,0 +1,333 @@
...
+observe the value of the Retry-After header and refrain from sending further
+requests until the indicated time.
+
+If the delivery has been accepted and processed, and if the response carries a
+payload with processing details, the response MUST have the [200 OK][200]
+status code. In this case, the response MUST carry a
+[`Content-Type`][Content-Type] header.
+
+If the delivery has been accepted and processed, but carries no payload, the
+response MUST have the [204 No Content][204] status code.
```
sslavic: Would it be a problem to be more permissive? E.g. [Google PubSub Push](https://cloud.google.com/pubsub/docs/push#receive_push) treats 200, 201, 202, 204, or 102 as ack that notification was successfully processed.

tomkerkhove:
Signed-off-by: Tom Kerkhove kerkhove.tom@gmail.com

tomkerkhove:
Apperently my PR breaks the DCO. How can I fix this for you guys?

duglin:
@tomkerkhove when you use "git commit" make sure you use the "-s" flag so its signed. You'll need to make sure your git config has your name and email correct too. See:  https://github.com/cloudevents/spec/blob/master/CONTRIBUTING.md

to fix this... it might be easiest if @clemensv just squashed this down to one commit. The other option is to "--amend" your commit.

tomkerkhove:
@duglin I only found out after the fact, sorry for the inconvenience..

http-webhook.md
clemensv: Protection for the sender. I assume that externally supplied targets may be malicious and may want to send the sender on a neverending journey through redirects, so I disallow those outright. Also, the abuse protection only probes this site and not a site that is being redirected to. With redirects, you can basically sidestep the abuse mechanism with a static site that simply redirects all traffic to a DoS target. So this is security related from both angles.

http-webhook.md
clemensv: I want to put a foot down here and make one profile that is "right". Our services don't implement all the rules here, yet, and we will have to make adjustments so I hope everyone else is also willing to. I allow 200, 202, and 204 here. 102 is not the end of the request, so I don't find that proper. 201 is debatable for adding alongside 200.

http-webhook.md
```diff
@@ -0,0 +1,333 @@
...
+
+If the delivery has been accepted, but has not yet been processed or if the
+processing status is unknown, the response MUST have the [202 Accepted][202]
+status code.
+
+## 3. Authorization
+
+The delivery request MUST use one of the following two methods, both of which
```
sslavic: "following two methods" --> "following two authorization methods" ?

More importantly, is this "MUST" requirement for delivery request and (in next sentence) for delivery target, with practically one option only, really necessary? This alone would eliminate (make non-compliant) many current OSS FaaS solutions, and any notification system which uses k8s+Istio RBAC for implementing security aspect.

http-webhook.md
```diff
@@ -0,0 +1,333 @@
...
+The delivery target will respond with a permission statement and the permitted
+request rate.
+
+The following header fields MUST be included in the validation request:
+
+#### 4.1.1.2. WebHook-Request-Origin
+
+The `WebHook-Request-Origin` header requests permission to send notifications
+from this sender, and contains a DNS expression that identifies the sending
+system, for example "eventemitter.example.com". The value is meant to summarily
```
sslavic: Who/what guarantees this identity really belongs to the request sender?

Would it make sense to build on top of SPIFFE/Spire/Istio?

http-webhook.md
sslavic: App/workload logic is agnostic of the security aspect. Aspect like security is requested declaratively and enforced by Istio, where side-car container injected into protected workloads intercepts incoming and outgoing traffic to uniformly, typically across the k8s cluster, apply cross-cutting concerns like security (identity, authorization, mutual TLS), traffic management (weighted routing), tracing, monitoring. See https://istio.io/docs/concepts/security/rbac.html

http-webhook.md
clemensv: I am trying to [explain here](https://github.com/cloudevents/spec/pull/155/files#diff-c47b1661017c688fecb70618769650f1R183) that Origin is not about identity. Authorization takes care of establishing identity where needed.

There are three roles here: (a) sender, (b) target, and (c) the 3rd party that tells the sender to send to the target. The threat vector here is that (c) tells (a) to send to (b) without (b) being ok to receive events from (a). 

If (b) is an plain website running on nginx without authentication turned on, it's not going to benefit from any fancy identity model. What protects it from unwanted push traffic is that it simply doesn't respond to the abuse handshake with a positive acknowledgement. "Origin" allows a site to differentiate between different senders within the scope of this mechanism. Since the mechanism is designed to mitigate the threat explained above, Origin doesn't need further protection, because (c) can't manipulate it. 

If I am operating (a) as "push.fabrikam.com" and were to steal someone else's identity, say "push.contoso.com" for the origin, because (b) has that name whitelisted but not mine, I'm not gaining any better access because I still need proper credentials. But I just allowed (c) to abuse my service to bombard (b) with invalid requests and therefore risk blacklisting of my originating IP address. So making a false statement about "Origin" is actually self-defeating with what this mechanism is aiming at. 

If `Origin` needs to be further authenticated, that should be done using the existing authNZ model, e.g. you could use TLS client auth to establish the identity, or you could include the Origin value into the bearer token.

http-webhook.md
clemensv: I looked at this in the spec context again and now stand by the existing wording. The request MUST use either of the two options shown. If you have some magic interceptor thing, then that magic interceptor thng must make the request in aggregate compliant with the rule here before it gets to the target. This spec aims to define interop between a sender and a target that use open standards without coupling to some shared security framework.

I see that ISTIO provides a model to establish mutual TLS auth, but I do not understand how that established identity surfaces in an HTTP request other than through whatever proprietary context mechanism a web server gives you. If it's completely transparent also for the receiving endpoint we're effectively talking about unauthenticated HTTP plaintext at both ends that's aided by interceptor magic, and I don't know how to build a secure standard for the endpoints on top of that.

http-webhook.md
clemensv: Adding 201.

clemensv:
@sslavic regarding the OpenAPI spec - yes, I think that would make sense to have. Do you want to take a stab at creating an OpenAPI schema that reflects the doc?

http-webhook.md
```diff
@@ -0,0 +1,339 @@
...
+### 2.2. Delivery response
+
+The delivery response MAY contain a payload providing detail status
+information in the case of handling errors. This specification does not attempt
+to define such a payload.
+
+The response MUST NOT use any of the [3xx HTTP Redirect status codes][3xx] and
+the client MUST NOT follow any such redirection.
+
+If a delivery target has been retired, but the HTTP site still exists, the
```
inlined: Curious how a sender should process a 404 differently from a 410

http-webhook.md
```diff
@@ -0,0 +1,339 @@
...
+site SHOULD return a [410 "Gone"][410] status code and the sender SHOULD refrain
+from sending any further notifications.
+
+If the delivery target is unable to process the request due to exceeding a
+request rate limit, it SHOULD return a [429 Too Many Requests][429] status code
+and MUST include the [`Retry-After`][Retry-After] header. The sender MUST
```
inlined: This may be nitty, though I'm curious whether a distributed sender MUST coordinate/interlock so that whenever one shard gets a 429, all shards must respect that immediately.

http-webhook.md
```diff
@@ -0,0 +1,339 @@
...
+observe the value of the Retry-After header and refrain from sending further
+requests until the indicated time.
+
+If the delivery has been accepted and processed, and if the response carries a
+payload with processing details, the response MUST have the [200 OK][200] or
+[201 Created][201] status code. In this case, the response MUST carry a
+[`Content-Type`][Content-Type] header.
+
+If the delivery has been accepted and processed, but carries no payload, the
+response MUST have the [201 Created][201] or [204 No Content][204] status code.
```
inlined: A number of new web developers send 200 with Content-Length 0. Is there a reason to explicitly call these non-compliant?

http-webhook.md
```diff
@@ -0,0 +1,339 @@
...
+status code.
+
+If the delivery cannot be accepted because the notification format has not
+been understood, the service MUST respond with status code [415][415].
+
+All further error status codes apply as specified in [RFC7231][RFC7231].
+
+## 3. Authorization
+
+The delivery request MUST use one of the following two methods, both of which
```
inlined: Why is OAuth2 the only valid auth? E.g. GitHub uses content signatures in their webhooks. APNs uses connection-level auth by only allowing connections from known certificates.

http-webhook.md
```diff
@@ -0,0 +1,339 @@
...
+or inadvertently registers the address of a system that does not expect such
+requests and for which the registering party is not authorized to perform such
+a registration. In extreme cases, a notification infrastructure could be abused
+to launch denial-of-service attacks against an arbitrary web-site.
+
+To protect the sender from being abused in such a way, a legitimate delivery
+target needs to indicate that it agrees with notifications being delivered to it.
+
+Reaching the delivery agreement is realized using the following validation
+handshake. The handshake can either be executed immediately at registration
```
inlined: nit: the language "either... or" is a false dichotomy. E.g. a K8S style API might call this endpoint both at resource creation _and_ periodically out of band.

http-webhook.md
```diff
@@ -0,0 +1,339 @@
...
+`Origin` request header for each delivery request, with the value matching that
+of this header.
+
+For example:
+
+``` text
+WebHook-Request-Origin: eventemitter.example.com
+```
+
+#### 4.1.1.2. WebHook-Request-Rate
```
inlined: Rate limiting might apply to the endpoint (e.g. the Tweet analyzer subsystem can only support 1500QPS) or to a client as a whole (each twitter user can only send 50req/hr). I wonder if we need an additional header to indicate the scope on which the rate limit applies.

http-webhook.md
inlined: I'm a bit torn here because this specifies a perfectly valid model and I like the idea of dynamic rate limiting. A few questions though:

1. Why not actually reuse the CORS headers here?
2. Many clouds implement this by requiring (c) to prove ownership of (b) before allowing (a) to point to (c). Should such a cloud have a _separate_ auth mechanism just for cloud events?

http-webhook.md
```diff
@@ -0,0 +1,339 @@
...
+permission to send notifications from this sender at the specified rate. The
+value is the string representation of a positive integer number greater than
+zero and expresses the request rate in "requests per minute".
+
+For example, the following header asks for permission to send 120 requests per
+minute:
+
+``` text
+WebHook-Request-Rate: 120
+```
```
inlined: This webhook seems like a reasonable place to negotiate all kinds of features:

1. Valid timeout for processing?
2. Valid data types (e.g. schemas, versioning)
3. Accepted encodings

In general I've held back on these features since we've put the control plane APIs out of scope for now. This PR suddenly puts these (very important) features back in scope.

http-webhook.md
clemensv: 410 is "there was something here but it's gone now". That is different from 404 "there's nothing here". 404 is a likely config error, 410 is outdated config.

http-webhook.md
clemensv: I don't think it's realistic or necessary to expect perfect consistency across multiple nodes that act as if they were one sender. If there are 4 concurrent requests in flight from 4 nodes, each of them is presumably going to get a 429 each and will subsequently back off. For the protocol, multiple senders acting as one doesn't really exist. You'll be smart to have an eventually consistent gossip (or other mechanism) to tell your push agents about a target currently throttling, so that you can move on to other work while the condition persists. That is ultimately what Retry-After helps with.

http-webhook.md
clemensv: Because they are not compliant. HTTP reserves 204 for when you have no payload. I can't relax the rules of HTTP here.

http-webhook.md
clemensv: "The client MAY use any token-based authorization scheme. Challenge-based
schemes MUST NOT be used."

The point of this spec is not to document all variations of WebHooks presently in use, but rather define a straightforward model with a core set of options. You can form a token with a content signature and flow that in the Authorization header under its own scheme and be compliant. 

OAuth 2.0 provides a solid summary of how to do the token flow via Authorization header or via query string and that is why I lean on it.

The reason I'm not including client certificates is that they're notoriously terrible to manage. If/when the group thinks there is concrete need for client cert auth, then we should add it.

http-webhook.md
clemensv: What's the requirement for doing the handshake twice?

http-webhook.md
clemensv: The request rate applies to this exact target resource. If the resource has some composite rules, it ought to do the math and give the sender a simple answer rather than complicating matters by giving two answers that may be conflicting. Keep it simple.

http-webhook.md
clemensv: Addressing both comments at once:

First:
1. I cannot use the CORS headers, because this is a very different problem to solve, and I can't interfere with the actual CORS use-case
2. The point of this mechanism is to provide a common path and not document everything everyone does now. Microsoft will have to make changes for this as well.

Second:

Further negotiations could indeed take place in this handshake, and they should lean on HTTP constructs. We could, for instance, have a reversing HTTP Accept handshake here, meaning that the target gets to declare what format it would like. I don't think that's critical for a first stab, but it would be good to play through and think about what to add. We have the CloudEvents versioning strategy and we do have preliminary media types.

glennblock:
Good to see this @clemensv! Webhooks are a central part of the products I work on. Today whenever the term Webhook comes up it is understood as a general pattern. There are rules of the road, but they haven't really been written, thus this is nice to see. I'll add some comments to the draft as appropriate.

As to the products, [Webtask.io](https://webtask.io) is a free developer tool / environment for creating Serverless endpoints. Webhooks are one of the most common use cases. Webtask is generally the backend for handing notifications from a Service that fires off events, Github, Slack, and Stripe are three of the very common use case both for receiving and invoking.

Second we have a premium product [Extend](https://auth0.com/extend/developers) which is a Serverless Webhook platform which can be "embedded" within a SaaS product to allow users to have a native experience within the product for handling application events.

http-webhook.md
```diff
@@ -0,0 +1,339 @@
+# HTTP 1.1 Web Hooks for Event Delivery
+
+## Abstract
+
+"Web hooks" are a popular pattern to deliver notifications between applications
```
glennblock: I would just go with webhook of Web Hook or WebHook. It is the most common usage these days i.e. [Wikipedia](https://en.wikipedia.org/wiki/Webhook), [Github](https://developer.github.com/webhooks/), [Stripe](https://stripe.com/docs/webhooks), [Slack](https://api.slack.com/incoming-webhooks)

http-webhook.md
```diff
@@ -0,0 +1,339 @@
...
+and via HTTP endpoints. In spite of pattern usage being widespread, there is no
+formal definition for Web Hooks. This specification aims to provide such a
```
glennblock: If it is meant to be generic, why not split it out as its own draft?

http-webhook.md
```diff
@@ -0,0 +1,339 @@
...
+- 1.1. [Conformance](#11-conformance)
+- 1.2. [Relation to HTTP](#12-relation-to-http)
+2. [Delivering notifications](#2-delivering-notifications)
+3. [Authorization](#3-authorization)
+4. [Abuse Protection](#4-abuse-protection)
+5. [References](#5-references)
+
+## 1. Introduction
+
+"Web hooks" are a popular pattern to deliver notifications between applications
```
glennblock: Suggest mentioning Jeff Lindsay's post on this: http://progrium.com/blog/2007/05/03/web-hooks-to-revolutionize-the-web/. He coined it as "Web Hooks" then but since has also adopted the term "Webhook".

http-webhook.md
```diff
@@ -0,0 +1,339 @@
...
+### 2.1. Delivery request
+
+Notifications are delivered using a HTTP request. The response indicates the
+resulting status of the delivery.
+
+HTTP-over-TLS (HTTPS) [RFC2818][RFC2818] MUST be used for the connection.
+
+The HTTP method for the delivery request MUST be [POST][POST].
+
+The [`Content-Type`][Content-Type] header MUST be carried and the request MUST
```
glennblock: What if data is specified via the query string and there is no payload? Are you saying every webhook MUST send a payload? If so, why?

http-webhook.md
glennblock: This makes sense to me. I imagine there's other status codes that don't make sense, and headers. For example does it makes sense to return an ETag? If we assume the webhook use case is for a response to some event, then ETags would serve no purpose.

glennblock: For that matter caching headers in general wouldn't make sense.

http-webhook.md
glennblock: @clemensv according to the HTTP spec any Content-Length greater than or equal to 0 is a valid value. See [here](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.13)

http-webhook.md
```diff
@@ -0,0 +1,339 @@
...
+[201 Created][201] status code. In this case, the response MUST carry a
+[`Content-Type`][Content-Type] header.
+
+If the delivery has been accepted and processed, but carries no payload, the
+response MUST have the [201 Created][201] or [204 No Content][204] status code.
+
+If the delivery has been accepted, but has not yet been processed or if the
+processing status is unknown, the response MUST have the [202 Accepted][202]
+status code.
+
```
glennblock: For 202, maybe go further and require a Location header specifying where the new resource will live OR a status monitor.

glennblock: The reason I am suggesting this is to provide a standard way to handle async webhooks, where the client expects a response, but can handle it out of band.

http-webhook.md
clemensv: @inlined @glennblock the normative spec for that is now RFC7230 which has superseded RFC2616. The [relevant section is here](https://tools.ietf.org/html/rfc7230#section-3.3.2). I agree that the wording distinguishes between an absent and an empty body in the context of allowing for POST with no payload. There is also this SHOULD rule:

"A user agent SHOULD NOT send a Content-Length header field when the request message does not contain a payload body and the method semantics do not anticipate such a body."

The HTTP spec isn't clear here about the recommendation for responses where there is no payload. 

My position is that Content-Length:0 with 200 OK is therefore an edge case I'd like to resolve cleanly for this spec as not being permitted and fof 204 to be used instead.

http-webhook.md
clemensv: It is its own draft. Microsoft would like to take this specific spec to IETF.

http-webhook.md
clemensv: @glennblock you've pointed to the Content-Length rule yourself, Glenn. POST requires a payload body, but that body may be empty and even if it's empty you probably still need to declare it. That said, in the interest of trading choice for interop, I do indeed believe that a payload body can be mandated for a webhook request.

http-webhook.md
glennblock: Thanks for pointing to the updated spec. I agree that it is an edge case. As this is a new standard, there's no problem with setting some firmer ground rules.

http-webhook.md
```diff
@@ -0,0 +1,339 @@
...
+3. [Authorization](#3-authorization)
+4. [Abuse Protection](#4-abuse-protection)
+5. [References](#5-references)
+
+## 1. Introduction
+
+"Web hooks" are a popular pattern to deliver notifications between applications
+and via HTTP endpoints. Applications that make notifications available, allow
+for other applications to register an HTTP endpoint to which notifications are
+delivered once available.
```
duglin: does the "once available" relate to the endpoint or the notification? I wonder if we could drop the "once available" entirely.

http-webhook.md
```diff
@@ -0,0 +1,339 @@
...
+or
+
+``` text
+WebHook-Request-Origin: *
+```
+
+#### 4.1.1.2. WebHook-Allowed-Rate
+
+The `WebHook-Allowed-Rate` header MUST be returned if the request contained
+teh `WebHook-Request-Rate`, otherwise it SHOULD be returned.
```
lfourie: typo: teh

http-webhook.md
```diff
@@ -0,0 +1,339 @@
...
+
+The header grants permission to send notifications at the specified rate. The
+value is either an asterisk character or the string representation of a
+positive integer number greater than zero. The asterisk indicates that there is
+no rate limitation. An integer number expresses the permitted request rate in
+"requests per minute". For request rates exceeding the granted notification
+rate, the sender ought to expect request throttling. Throttling is indicated
+by requests being rejected using HTTP status code [429][429].
```
lfourie: HTTP 429 Too Many Requests

tomkerkhove: I don't really see what is wrong here, it's referring to the spec here so looks ok?

http-webhook.md
```diff
@@ -0,0 +1,339 @@
...
+With the validation request, the sender asks the target for permission to
+send notifications, and it can declare a desired request rate (requests per
+minute).
+
+The delivery target will respond with a permission statement and the permitted
+request rate.
+
+The following header fields MUST be included in the validation request:
+
+#### 4.1.1.2. WebHook-Request-Origin
```
lfourie: section number: 4.1.1.1

http-webhook.md
```diff
@@ -0,0 +1,339 @@
...
+parameter.
+
+For example, the client makes the following HTTP request:
+
+``` text
+POST /resource?access_token=mF_9.B5f-4.1JqM HTTP/1.1
+Host: server.example.com
+```
+
+The HTTP request URI query can include other request-specific parameters, in
```
lfourie: replace "can" with "MAY"

http-webhook.md
```diff
@@ -0,0 +1,339 @@
...
+
+If the delivery has been accepted and processed, but carries no payload, the
+response MUST have the [201 Created][201] or [204 No Content][204] status code.
+
+If the delivery has been accepted, but has not yet been processed or if the
+processing status is unknown, the response MUST have the [202 Accepted][202]
+status code.
+
+If the delivery cannot be accepted because the notification format has not
+been understood, the service MUST respond with status code [415][415].
```
lfourie: 415 Unsupported Media Type

http-webhook.md
```diff
@@ -0,0 +1,339 @@
...
+payload with processing details, the response MUST have the [200 OK][200] or
+[201 Created][201] status code. In this case, the response MUST carry a
+[`Content-Type`][Content-Type] header.
+
+If the delivery has been accepted and processed, but carries no payload, the
+response MUST have the [201 Created][201] or [204 No Content][204] status code.
+
+If the delivery has been accepted, but has not yet been processed or if the
+processing status is unknown, the response MUST have the [202 Accepted][202]
+status code.
```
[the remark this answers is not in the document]
cneijenhuis: After submitting, I saw that the PR description says the spec is for "one-way event delivery".

I didn't find such a claim in the spec itself (unless you interpret "notification" as only applicable to past events). I understand how focusing on one-way delivery makes sense in the context of CloudEvents, but - if this spec is to be used for webhooks in general - I think you should consider the case of the delivery target sending a proper response that the sender uses to change its behaviour.

FWIW, the wikipedia definition of a webhook includes it as well: "A webhook [...] is a method of augmenting or **altering** the behaviour of  a web page, or web application [...]."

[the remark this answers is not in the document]
clemensv: @cneijenhuis If you can get a proper response, that's arguably just another HTTP request. RFC7231 covers that exhaustively. The point if this spec is to constrain RFC7231 to a profile that everyone agrees to for one-way event delivery, i.e. callbacks. 

Here's the place where we all make the spec and then I'm glad if someone summarizes that on Wikipedia - not the other way around :)

http-webhook.md
rperelma: For Adobe I/O Events, we use the client secret, not a client certificate, for producing a content signature. We already manage client secrets generated during clientID generation, so no added burden.

http-webhook.md
```diff
@@ -0,0 +1,339 @@
...
+## 4. Abuse Protection
+
+Any system that allows registration of and delivery of notifications to
+arbitary HTTP endpoints can potentially be abused such that someone maliciously
+or inadvertently registers the address of a system that does not expect such
+requests and for which the registering party is not authorized to perform such
+a registration. In extreme cases, a notification infrastructure could be abused
+to launch denial-of-service attacks against an arbitrary web-site.
+
+To protect the sender from being abused in such a way, a legitimate delivery
```
rperelma: Across the industry (facebook, dropbox, slack, etc.) the most common (and much simpler) approach for URL validation is to send a challenge. I realize you don't want to use a challenge for authorization, but I believe it is perfectly acceptable for validation. If the URL is invalid, or the delivery target doesn't want to accept a webhook call, it simply doesn't respond to the initial challenge, and in that case the sender MUST NOT deliver events to the target.

clemensv: The point here is that we need a standard way to express the opt-in. Everyone does these thibgs, but each in their own way. The opt-in can be a reply or the opt-in can be a callback to a URL; either one is fine and we can also offer both.

http-webhook.md
```diff
@@ -0,0 +1,339 @@
...
+
+Clients using the URI Query Parameter method SHOULD also send a
+Cache-Control header containing the "no-store" option.  Server
+success (2XX status) responses to these requests SHOULD contain a
+Cache-Control header with the "private" option.
+
+Because of the security weaknesses associated with the URI method (see
+[RFC6750, Section 5][RFC6750]), including the high likelihood that the URL
+containing the access token will be logged, it SHOULD NOT be used unless it is
+impossible to transport the access token in the "Authorization" request header
```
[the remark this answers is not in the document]
clemensv: There are plenty of sites where you can configure web hooks that will allow you to paste a URL and nothing else. Having a token right in the URL is simply the easiest way to give someone a reference with embedded access permission. So it's not neccessarily about not having access to the header per-se (which still is an occasional problem w/ some APIs), but about the information being split across two places. The feature is already optional for the client, since that's where those concerns arise. The server shouldn't have a opt-out choice given that it is a client consideration and might be a constraint for which the client has no alternative.

http-webhook.md
clemensv: @rperelma I just added a callback option.

http-webhook.md
```diff
@@ -0,0 +1,339 @@
...
+WebHook-Allowed-Origin: eventemitter.example.com
+```
+
+or
+
+``` text
+WebHook-Request-Origin: *
+```
+
+#### 4.1.1.2. WebHook-Allowed-Rate
```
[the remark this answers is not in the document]
clemensv: I already call out usage of 429 in section 2.2. 

The problem I am trying to address here is that the publisher doesn't suddenly show up with a mortal flood of events that takes out the target outright. 429 is reactive. This here aims to prevent the negotiation apriori.

http-webhook.md
```diff
@@ -0,0 +1,339 @@
...
+If it does not allow delivery of the events or does not expect delivery of
+events and nevertheless handles the HTTP OPTIONS method, the existing response
+ought not to be interpreted as consent, and therefore the handshake cannot
+rely on status codes.
+
+The response SHOULD include the [Allow][Allow] header indicating the [POST][POST]
+method being permitted. Other methods MAY be permitted on the resource, but
+their function is outside the scope of this specification.
+
+#### 4.1.1.2. WebHook-Allowed-Origin
```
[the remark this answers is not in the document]
clemensv: The numbering is off. Fixing.

http-webhook.md
```diff
@@ -0,0 +1,339 @@
...
+site SHOULD return a [410 "Gone"][410] status code and the sender SHOULD refrain
+from sending any further notifications.
+
+If the delivery target is unable to process the request due to exceeding a
+request rate limit, it SHOULD return a [429 Too Many Requests][429] status code
+and MUST include the [`Retry-After`][Retry-After] header. The sender MUST
+observe the value of the Retry-After header and refrain from sending further
+requests until the indicated time.
+
+If the delivery has been accepted and processed, and if the response carries a
```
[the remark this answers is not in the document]
clemensv: Agree. Fixing.

http-webhook.md
clemensv: Numbering is off. Fixing.

http-webhook.md
clemensv: Done

http-webhook.md
clemensv: Done

http-webhook.md
clemensv: Done (adding the reason phrase for clarity)

http-webhook.md
clemensv: thanks

http-webhook.md
rperelma: @clemensv thanks, but not quite what I had in mind :)
I was referring to a simple challenge sent to the target, and the target responding (or not) to that challenge, thereby accepting (or not) further deliveries. This is a very widely supported approach across the industry.

glennblock: Facebook has this for example: https://developers.facebook.com/docs/graph-api/webhooks/getting-started see the section on verification.

http-webhook.md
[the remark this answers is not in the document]
cneijenhuis: > The feature is already optional for the client, since that's where those concerns arise.

I don't agree it is only a concern for the client, and it's been an issue for me several times (e.g. with IronMQ urls, and with Azure Functions - both default to tokens in the URI).

Our clients entrust us with their access tokens, and they should be assured we keep them safe (similar to e.g. passwords, or personal data of our clients customers).
We have policies to never log passwords, or personal data, or similar. Access to our logs is not heavily restricted, allowing many people within our company to quickly gain insights if things go wrong (e.g. support).
If my code logs the access token, I am a) basically violating company policy if you assume access tokens are a form of a password, and b) make the access token available to a broad range of people within my company, that are not allowed to access critical data.

(As a commerce system, stealing an access token may allow the attacker to e.g. trigger a refund, or trigger the shipment of an order before the payment is confirmed - so the threat isn't purely theoretical)

Fair enough, then don't log the URI, right? Turns out it is virtually impossible to assure if you are using an HTTP lib on the JVM (I'll spare you the details...).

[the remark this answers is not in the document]
cneijenhuis: > Having a token right in the URL is simply the easiest way to give someone a reference with embedded access permission.

I totally agree.

Here is a compromise: The delivery target is allowed to register the webhook with the access token in the URI. The sender MUST send the access token, but it SHOULD use the Authorization Request Header Field instead of the URI Query Parameter.

(I guess this is technically already fine via `The delivery target MUST support both methods.`, but it would be surprising for the client)

http-webhook.md
alexellis: I agree with @inlined and made the same comment on the workgroup call. GitHub uses HMAC / symmetric secret and this is adopted by other platforms also like Patreon.

> The delivery request MUST use one of the following two methods

What does this mean? Does it mean that if using auth we must implemented either - or that auth must be done and it must be either? @clemensv

http-webhook.md
```diff
@@ -0,0 +1,373 @@
...
+[415 Unsupported Media Type][415].
+
+All further error status codes apply as specified in [RFC7231][RFC7231].
+
+## 3. Authorization
+
+The delivery request MUST use one of the following two methods, both of which
+lean on the OAuth 2.0 Bearer Token [RFC6750][RFC6750] model.
+
+The delivery target MUST support both methods.
```
alexellis: Why must the target support both? This is imposing implementation detail on the FaaS receiver.

http-webhook.md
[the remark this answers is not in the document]
glennblock: Yes @cneijenhuis, it is not uncommon to have synchronous responses. Auth0 rules, for example, are synchronous. At the time of login, custom logic runs in rules (which are webhooks) which could prevent the login. We have many other synchronous use cases.

http-webhook.md
clemensv: Because only the client knows whether it can or cannot put a token into the header. Headers are the preferred method because they're not dragged through all of the most common logs as query strings are. It's a client choice and therefore the server needs to support the choices.

http-webhook.md
clemensv: @alexellis  I added the following clause yesterday:

"The client MAY use any token-based authorization scheme. The token can take any shape, and can be a standardized token format or a simple key expression."

That means (and meant) using any key or an HMAC using a key is completely legitimate. I don't define or mandate a specific token format here. Azure Functions generates a symmetric key and requires that key for its default implementation that you get from "file new", and that would be conforming. What I need to mandate for interop is that the query argument has a well-known name and that's mostly what I do here.

http-webhook.md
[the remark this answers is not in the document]
clemensv: @cneijenhuis In our service, we have "Connection Strings" that have URLs and AuthNZ material in them, and we then split that information across query string and headers as needed. 

I *like* the rule you stated above, but it's right on the line of getting into the hair of how an implementation does its config. We don't mention the actual subscription gesture in any spec (and per earlier group discussions it'll be quite difficult to harmonize it), and therefore it's difficult to impose constraints on it. 

I already make a pretty strong and normative SHOULD NOT statement against the URI method, including giving caching guidance, and that effectively says what you say here. How about we amend that with:

"Even if the client has been configured with a token embedded in the URI query string conforming with 3.2, the client SHOULD extract and remove the token from the query string and send it via the Authorization request header field whenever feasible."

http-webhook.md
clemensv: @rperelma that's exactly what this mechanism is, though. The client sends a string (WebHook-Request-Origin) and the server echoes that string (WebHook-Allowed-Origin) if and only if it's ok with that source pushing. I do that on OPTIONS, because that's what OPTIONS is for (also see W3C CORS). The rate negotiation is optional.

What I believe I need for a multi-party interop standard is a way for the target to not only echo back some arbitrary string, but for the target to also have a way to make semantic sense out of that string, because the webhook might not just be written for a single client system, but for multiple different client systems and the target may want to be able to differentiate between those.

The origin challenge might be "instance-a35f7d9e.service.example.com" and the target may have a rule to permit "*.example.com" and then grants access by echoing the origin phrase. The implementation of the target might also be ok with anyone pushing to them and then just replies with "*" out of a dumb response rule. 

If I could mandate a tighter authorization model (which I would if this were a product and not a standards design), I would even tie the Origin into the access token. I can't do that here, because there's appears to be majority who work with simple symmetric keys.

http-webhook.md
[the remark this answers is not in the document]
clemensv: @cneijenhuis @glennblock I understand that synchronous responses are being used, but CloudEvents defines a unidirectional and routable event flow (on top of which you can model async responses). The examples you cite are extensibility points where the web hook call performs a command execution in context. Those are not an events; events are facts, commands are intents.

Mind the headline here: "HTTP 1.1 Web Hooks for Event Delivery"

There might quite well be a derivative or parallel spec "HTTP 1.1 Web Hooks for Extensiblity Commands", but I would like to keep this one scoped to the one-way case that we keep symmetric to what we can easily achieve in AMQP, MQTT, NATS, etc.

http-webhook.md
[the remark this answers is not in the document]
cneijenhuis: @clemensv Sounds great 👍 

Minor: `the client SHOULD extract` should probably be `the sender SHOULD extract`?

http-webhook.md
```diff
@@ -0,0 +1,376 @@
...
+5. [References](#5-references)
+
+## 1. Introduction
+
+["Webhooks"][Webhooks] are a popular pattern to deliver notifications between
+applications and via HTTP endpoints. Applications that make notifications
+available, allow for other applications to register an HTTP endpoint to which
+notifications are delivered.
+
+This specification defines a HTTP method by how notifications are delivered by
```
sslavic: Delivery/receiving callback requests is just part of the Web hook mechanism. As web hook was originally covered in @progrium's [blog](http://progrium.com/blog/2007/05/03/web-hooks-to-revolutionize-the-web/)
> To support web hooks, you allow the user to specify a URL where your application will post to and on what events.

So far actual [hook](https://stackoverflow.com/a/467568/381140) API part and related protocol like:
* API how callbacks are to be registered
** how should callback URL attribute be named
** how one defines for which events should callback URL be called (potential CloudEvents attribute usage?)
** negotiating content type for web hook callback
** verifying callback URL is valid and owned by the calling party
** have can all of this information be exchanged (query params, request body, via some 3rd party storage/API like k8s API CRD, ...)
* how can receiving party communicate it's no longer interested in the events
* how can delivery party share and interested party discover for which events hook can be registered, etc.
are not covered, yet.

Would it make sense to explicitly mention that some aspects of Web hook mechanism for reasons have been deliberately left out for now?

http-webhook.md
sslavic: In his [webinar about callback support in OpenAPI 3.0](https://t.co/1XUELjz5G5), @fehguy suggests to use 204 response as one way to signal that receiving party is no longer interested in further updates, that it that callback to be unregistered from web hook (as alternative to explicit API for unregistering callbacks/subscriptions).

http-webhook.md
```diff
@@ -0,0 +1,376 @@
...
+
+If the receiving application does not explicitly support the handshake
+described here, an administrator could nevertheless still find the callback URL
+in the log, and call it manually and therewith grant access.
+
+The delivery target grants permission by issuing an HTTPS GET or POST request against
+the given URL. The HTTP GET request can be performed manually using a browser
+client.
+
+The delivery target MAY include the `WebHook-Allowed-Rate` response in the callback.
```
sslavic: Why is WebHook-Allowed-Rate optional (MAY) for async handshake confirm while in 4.2 it's mandatory (MUST)?

http-webhook.md
```diff
@@ -0,0 +1,376 @@
...
+
+The URL is not formally constrained, but it SHOULD contain an identifier for the
+delivery target along with a secret key that makes the URL difficult to guess so that
```
sslavic: Maybe suggest it should be temporary secret key.

http-webhook.md
```diff
@@ -0,0 +1,376 @@
...
+#### 4.1.2. WebHook-Request-Origin
+
+The `WebHook-Request-Origin` header MUST be included in the validation request
+and requests permission to send notifications from this sender, and contains a
+DNS expression that identifies the sending system, for example
+"eventemitter.example.com". The value is meant to summarily identify all sender
+instances that act on the behalf of a certain system, and not an individual
+host.
+
+After the handshake and if permission has been granted, the sender MUST use the
```
sslavic: Why use `WebHook-Request-Origin` only for abuse protection handshake request? Would it make sense to use `Origin` both for the handshake and later on delivery? The two requests are already different enough, by HTTP method - handshake uses OPTIONS and delivery uses POST.

http-webhook.md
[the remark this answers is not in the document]
cneijenhuis: > I would like to keep this one scoped to the one-way case that we keep symmetric to what we can easily achieve in AMQP, MQTT, NATS, etc.

While I think it wouldn't be too hard to include the two-way cases, I fully understand that you want to keep the scope as small as possible.

If you allow me to make a final request, though, it would be to have the scope more explicitly in the title, e.g. "Web Hooks for Asynchronous Event Delivery", or "Web Hooks for One-Way Event Delivery". Let me try to explain:

> The examples you cite are extensibility points where the web hook call performs a command execution in context. Those are not an events; events are facts, commands are intents.

I have to disagree. All three examples actually send events, or facts, via the webhook.

* Twilio: Caller with ID foobar has called
* Auth0: User foobar has made a login attempt
* Dialogflow: User foobar has said: "Book a flight from LA to London"

These are arguably facts - events that have happened. If the use cases wouldn't require an immediate response, it would also be fine to implement them as asynchronous webhooks.

E.g. for a chatbot, both options work well:
* Chatbot receives event via asynchronous webhook. Chatbot responds by calling API.
* Chatbot receives event via synchronous webhook. Chatbot responds directly.

The event data would be exactly the same in both cases! Therefore, I think it is worth pointing out in the title that the spec is scoped to asynchronous/one-way event delivery.

darrelmiller:
Here is a first stab at capturing the essence of the spec as an OpenAPI V3 document.  I have only done the auth header scenario just to stop it from getting any longer than it already is.

```
openapi: 3.0.0
info: 
  title: Example Web Hook based on Cloud Events spec
  version: 1.0.0
paths:
  '/SubscribeToEvent':  
    'post':    
      summary: Subscription operation. (Not defined by specification)
      parameters:
        - name: deliveryTarget
          in: query
          schema:
            type: string
            format: uri
      responses:
        '201':
          description: Successfully subscribed
      callbacks:
        exampleWithAuthHeader:
          '$request.query.deliveryTarget':
            'post':
              description: Delivery Request
              parameters:
                - name: Origin
                  in: header
                  required: true
              requestBody:
                required: true
              responses:
                '200':
                  description: Delivery Response with payload
                '201':
                  description: Created Notification with or without payload
                '202':
                  description: Accepted Notification Delivery
                '410':
                  description: Unsubscribe
                '415':
                  description: Delivery request payload media type not understood
                '429':
                  description: Too many requests
                  headers:
                    'retry-after':
                      required: true
                      schema:
                        type: integer
              security:
                - callbackToken : []
            'options':
              description: Validation Request
              parameters:
                - name: WebHook-Request-Origin
                  in: header
                  required: true
                - name: WebHook-Request-Callback
                  in: header
                  required: false
                  schema:
                    type: string
                    format: uri
                - name: WebHook-Request-Rate
                  in: header
                  required: false
                  schema:
                    type: integer
                - name: Allow
                  in: header
                  schema:
                    type: string
                    default: POST
              responses:
                200:
                  description: Validated
                  headers:
                    WebHook-Allowed-Origin:
                      required: true
                      schema:
                        type: string
                    WebHook-Allowed-Rate:
                      required: false
                      schema:
                        type: integer                   
components:
  securitySchemes:
    'callbackToken':
      type: http
      scheme: bearer
```

sslavic:
@darrelmiller thanks for sharing. I had a take as well and here's gist of what I have so far: https://gist.github.com/sslavic/b1f36a8fa4ac0a52568b7b0155c3f241

Using OpenAPI it's really hard to define reusable/standard part - it seems only thing possible is to express some examples, like one you shared of event emitter example API and there can be matching example of event consumer API.

nicolaferraro:
I like the attempts to provide a OpenAPI V3 document for the spec. And I really love the idea of adding a subscription API, or better writing a new spec (it seems outside the scope of the webhook delivery).

A standard API to `list` the published event types and handle subscription (`subscribe`, `unsubscribe`, `checkSubscription`). The cloud will be a lot more connected if we reduce fragmentation :smile:.

http-webhook.md
```diff
@@ -0,0 +1,376 @@
...
+which is compatible with HTTP 1.1 semantics.
+
+## 2. Delivering notifications
+
+### 2.1. Delivery request
+
+Notifications are delivered using a HTTP request. The response indicates the
+resulting status of the delivery.
+
+HTTP-over-TLS (HTTPS) [RFC2818][RFC2818] MUST be used for the connection.
```
duglin: While obviously https is a good, this implies that an in-house impl that doesn't use https is non-compliant. Is that the intent?

http-webhook.md
clemensv: @sslavic I don't agree that 204 telegraphs any such intent, ever. If you want to terminate a relationship you need to reach into the 3xx bucket.

http-webhook.md
duglin: While I understand the concerns, isn't this kind of harsh? I come across websites all the time that issue redirects for mindless reasons, like a missing "/" at the end of the URL, and most http clients just blindly handle it for you. Any reason we can't let the sender decide whether they want to support redirects or not?

http-webhook.md
```diff
@@ -0,0 +1,376 @@
...
+is a DNS name expression that identifies the sending system.
+
+### 2.2. Delivery response
+
+The delivery response MAY contain a payload providing detail status
+information in the case of handling errors. This specification does not attempt
+to define such a payload.
+
+The response MUST NOT use any of the [3xx HTTP Redirect status codes][3xx] and
+the client MUST NOT follow any such redirection.
```
duglin: we may want to be consistent in our terms, sender vs client for example. Obviously non-blocking comment :-)

http-webhook.md
clemensv: I am indeed very deliberately leaving quite a few aspects mentioned above, except for the abuse protection handshake, because I don't think we can standardize them at this point.

For instance, in Azure, you are registering webhooks into Event Grid through the Azure Resource Manager API, as a subscription on a topic supplying filters, and that's a single interaction in a large framework of system-wide conventions applying to thousands of APIs. 

It's presently not realistic to expect for that interaction to be superseded by an API standard that this WG comes up with. First, because of the greater platform convention context. Second, because we're far from ready to commit to constraints for such an API.

http-webhook.md
clemensv: I can't make that MUST, because then I can't use a browser to do the confirmation. There may be a clause missing here what the omission of the value means.

http-webhook.md
clemensv: What do you think "temporary" adds that would otherwise be misunderstood by an implementer?

http-webhook.md
clemensv: I don't understand what the 'Origin' header would add for the handshake itself. It is mandated for the delivery request.

http-webhook.md
```diff
@@ -0,0 +1,376 @@
...
+
+If the delivery has been accepted and processed, and if the response carries a
```
duglin: This section appears to be repeating the http specs. While my initial reaction was to suggest that we remove it because repeating specs is usually not a good idea, highlighting certain key aspects for people so they don't have to read the entire set of specs can be very useful. However, can we add a sentence that says any conflict between this spec and the http specs results in the http specs winning?

http-webhook.md
sslavic: IMO these tokens, just as whole subscription request if not confirmed through handshake, should be expiring after some time. Without mentioning these are temporary, it may be assumed that one can confirm at anytime in the future, subscription requests and tokens to be persisted endlessly until confirmed back.

http-webhook.md
sslavic: Currently it's written that:
```
The `WebHook-Request-Origin` header MUST be included in the validation request
and requests permission to send notifications from this sender, and contains a
DNS expression that identifies the sending system, for example
"eventemitter.example.com"

After the handshake and if permission has been granted, the sender MUST use the
`Origin` request header for each delivery request, with the value matching that
of this header.
```

If I understood correctly, in handshake/OPTIONS request event emitter has to send `WebHook-Request-Origin` header with value equal to the hostname of `Origin` header that it will be sending in delivery/POST requests.

Dilemma I have is why not send `Origin` during handshake with same value that will be used later in delivery, why new custom header when standard `Origin` header will be sent anyway during handshake too and probably with same value as during delivery and from it hostname can be extracted if just hostname is needed?

duglin:
I don't think this PR has been changed in about a week. And while there are still some questions I don't see any that are significant enough that the questioner is suggesting that the entire PR is headed in the wrong direction - they appear to be more targeted points of discussion. To that end, I'd like to suggest that we merge this PR and then people can open focused issues/PR to deal with any lingering concerns.

What do people think? If you can not make the call today and object to this suggestion please voice your concern here.

duglin:
Approved on the 5/24 call. Will deal with any open questions/concerns in follow-on PR/Issues.
If you'd like to see a change in this, or any doc, please open a PR with the proposed text you'd like to see.

PR #157, opened 2018-04-12
AMQP 1.0 transport and AMQP type system mapping

clemensv:
Proposed AMQP 1.0 transport binding and a type system mapping for the application properties section, analogous to the same for HTTP: https://github.com/cloudevents/spec/pull/148


amqp-transport-binding.md
```diff
@@ -0,0 +1,287 @@
...
+
+For instance, if the declared `contentType` is
+`application/json;charset=utf-8`, the expectation is that the `data` attribute
+value is made available as [UTF-8][RFC3629] encoded JSON text for use in
+AMQP.
+
+## 3. AMQP Message Mapping
+
+The content mode is chosen by the sender of the event, which is either the
+requesting or the responding party. Gestures that might allow solicitation of
```
lfourie: Gestures ... unclear - please reword

yaronha:
@clemensv do u have an example for an event (like u did for HTTP)?

also wasnt clear if we will be preserving the same notions like in HTTP (bin use AMQP Headers, structured using nested Json)?

Yaron

duglin:
Approved on 6/7 call

PR #161, opened 2018-04-13
Propose the addition of `FollowsFrom` semantics

erikerikson:
Propose the addition of `FollowsFrom` semantics to facilitate event correlation into data flows.

Signed-off-by: Erik Erikson <erik.m.erikson@nordstrom.com>

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -310,6 +310,19 @@ that contains both context and data).
   * MUST be a non-empty string
   * MUST be unique within the scope of the producer
 
+### followsFrom
+* Type: String
+* Description: The eventID of the event that caused this event. The semantics
+  of this string are explicitly undefined to ease the implementation of
+  producers. Enables correlation. The semantics of the relationship between
+  the event identified by followsFrom and the event identified by eventID are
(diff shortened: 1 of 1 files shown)


erikerikson:
Apologies in advance, having not participated in a standards body previously, for any obvious oversights or violation of best practice.

duglin:
@erikerikson what is the relationship between this and https://github.com/cloudevents/spec/pull/128 ? It feels like these two should be merged in some way to allow for a linking of events but with some kind of additional property to indicate the type of relationship between them.

erikerikson:
Ooh.  Still loading here and I missed that one.  Merging seems appropriate.  I will engage there.

duglin:
@erikerikson great - thanks!

spec.md
```diff
@@ -310,6 +310,19 @@ that contains both context and data).
   * MUST be a non-empty string
   * MUST be unique within the scope of the producer
 
+### followsFrom
+* Type: String
+* Description: The eventID of the event that caused this event. The semantics
```
erikerikson: one could argue that a set of eventID(s) could be the cause of an event and that this field should allow for this.  allowing delimited eventID(s) to be supplied could satisfy that fact at the cost of complexity

Issue #125, opened 2018-03-26
Remove TBD in `extensions` attribute


duglin:
In today's version of the spec, for the `extensions` attribute it has this sentence:
- - -
TBD - Determine a shorter prefix for this (e.g. OpenAPI uses “x-”)
- - -
We need to resolve (or remove) this TBD before v1.0

PR #162, opened 2018-04-15
Remove "TBD" comment about extensions

duglin:
We don't need the TBD since each mapping/serialization doc we produce
should state how extensions should appear. E.g. https://github.com/cloudevents/spec/pull/148

Closes: #125


clemensv:
LGTM

duglin:
I consider this more of a clean-up/syntax thing for the spec. If we can get some more reviews on this before the meeting then we can resolve it quickly. thanks

duglin:
Approved on 4/19 call

PR #165, opened 2018-04-19
Roadmap - Add deliverables to 0.1

ac360:
**Note:** This is a clone of #156 that was created to get around DCO issues.

This PR adds deliverables to 0.1:

Include a specification for mapping the CloudEvents specification to HTTP.
Include a specification for mapping the CloudEvents specification to JSON.
Define a type system for CloudEvents values.
This PR also does minor clean-ups (that don't intend to change our intentions).

Signed-off-by: Austen Collins austen@serverless.com

### DIFF
--- a/roadmap.md
+++ b/roadmap.md
@@ -14,23 +14,22 @@ _Note: The ordered lists for each milestone provide a way to reference each item
 *0.1*
 
 1. Draft specification that WG members agree *could* provide interoperability.
+1. Include an initial set of use-cases for CloudEvents.
+1. Define a type system for CloudEvents values.
 1. Document at least 3 sample events that conform to the specification.
 1. Github repo is organized to be approachable to a engineers who might want to
 implement the spec.
-1. Use-cases.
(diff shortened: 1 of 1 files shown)


duglin:
The top comment on this PR doesn't match the changes, but it does match what we agreed to in the last [PR](https://github.com/cloudevents/spec/pull/156)  (and during last week's call).  The original PR had 4 items in v0.1:
```
+* Include an initial set of use-cases for CloudEvents.
+* Include a specification for mapping the CloudEvents specification to HTTP.
+* Include a specification for mapping the CloudEvents specification to JSON.
+* Define a type system for CloudEvents values.
```
And per the notes from the call we:
```
Agreed to moving 2 & 3 of current PR to 0.2 milestone
```
This PR does that. The conflict is the the comment on this PR says "2&3" are part of v0.1, they're part of v0.2.

Merging per the agreement on last week's call. If I missed something, or if we need to revisit, then we can discuss it on today's call or someone can open up a follow-on PR.

PR #170, opened 2018-04-19
Add "Version 0.1" to some md files in prep for release

duglin:
Also added a "Status" section to be consistent across all 3 docs.

This should address @inline's concern about people thinking anything
is set in stone because now it says its a 'working draft'.

One thing we need to decide, later, is whether to have section numbers or not, we not consistent right now.

### DIFF
--- a/http-transport-binding.md
+++ b/http-transport-binding.md
@@ -1,4 +1,4 @@
-# HTTP Transport Binding for CloudEvents
+# HTTP Transport Binding for CloudEvents - Version 0.1
 
 ## Abstract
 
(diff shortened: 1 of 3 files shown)


markpeek:
LGTM

clemensv:
LGTM as-is. 

I would appreciate if we'd formalize spec editor status as in most other bodies of this sort and then reflect that in the docs.

duglin:
Syntax/editorial changes in prep for v0.1 - got 2 LGTMs, merging.

yaronha:
LGTM

PR #172, opened 2018-04-23
Clarify contentType

clemensv:
There has been a good deal of confusion about the role of contentType and whether it designates the contentType of the event per-se or of the event's data attribute. This change clarifies that it is meant to designate the content type of the latter. The content type of the event encoding is defined by the respective event format.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -357,7 +357,7 @@ that contains both context and data).
   rules for how the `data` attribute content is rendered for different
   `contentType` values are defined in the event format specifications; for
   example, the JSON event format defines the relationship in
-  [section 3.1](./json-format.md#3-1-special-handling-of-the-data-attribute).
+  [section 3.1](./json-format.md#31-special-handling-of-the-data-attribute).
 * Constraints:
   * OPTIONAL
   * If present, MUST adhere to the format specified in
--- a/spec.md
+++ b/spec.md
@@ -351,13 +351,13 @@ that contains both context and data).
 * Description: Content type of the `data` attribute value. This attribute
   enables the `data` attribute to carry any type of content, whereby format
   and encoding might differ from that of the chosen event format. For example,
-  an event rendered using the [JSON envelope](./json-format#3-envelope)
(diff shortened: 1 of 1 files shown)


duglin:
LGTM

duglin:
Last discussed on the 4/26 call: https://www.youtube.com/watch?v=PsMF2ddWNbw&t=0s&index=4&list=PLj6h78yzYM2Ph7YoBIgsZNW_RGJvNlFOt

duglin:
Discussed around 43 minutes into the video.

@clemensv  agreed to add some text around the relationship between the transport used and the encoding of `data`

scothis:
@clemensv Can you also clarify the relationship between `contentType` and `eventType`/ `eventTypeVersion`? In HTTP/ReST the responsibility of all three attributes is consolidated within the `Content-Type` header.

There is a distinction between the semantic meaning and the encoding of `Content-Type`. It seems CloudEvents is demoting `contentType` to only be concerned with the encoding of `data` while `eventType` and  `eventTypeVersion` are concerned with the semantic meaning. How should an implementer handle a case where the `contentType` of the event also conveys semantic meaning?

Additionally, `schemaURL` appears to be attempting to convey semantic meaning as well creating a third place where the semantic meaning can exist, and potentially be in conflict. As an implementor, what defines the semantic meaning of the event? Especially if these attributes are in conflict or ambiguous.

clemensv:
@scothis eventType relates to the entire event, the contentType here only to the payload, so these are different scopes.

johnmccabe:
Should the default `contentType` be explicitly defined for cases when its omitted? `application/json` when `data` contains an object, `text/plain` when its a string? Would need added to the `Special Handling of the "data" Attribute` section.

clemensv:
@johnmccabe Let me add a clause for what omitting the field means.

duglin:
All - this PR was updated last week so please review it when you get a chance so we can vote on it at the f2f.

duglin:
Approved during 6/15 f2f

PR #181, opened 2018-04-27
Fix name copy/paste and update Dispatch project link

markpeek:
Signed-off-by: Mark Peek <markpeek@vmware.com>


duglin:
Sure Lee, if you say so  :-)
LGTM

PR #212, opened 2018-05-17
update roadmap

duglin:
Move all of the stuff we've done under 0.1 to be more accurate.

Provide a rough outline for our future milestones, putting the bigger spec change items first.

### DIFF
--- a/roadmap.md
+++ b/roadmap.md
@@ -17,32 +17,53 @@ _Note: The ordered lists for each milestone provide a way to reference each item
 1. Include an initial set of use-cases for CloudEvents.
 1. Define a type system for CloudEvents values.
 1. Document at least 3 sample events that conform to the specification.
-1. Github repo is organized to be approachable to a engineers who might want to
-implement the spec.
+1. Github repo is organized to be approachable to a engineers who might want
+   to implement the spec.
 1. Finalize logo.
 1. Create and deploy a website that features a simple overview, email list and directs visitors to Github.
 1. Store all website assets in the CloudEvents repository, under the governance
 of the working group.
+1. Have at least 2 implementations of the specification that can demonstrate
+   interoperability.
+1. Include a specification for mapping the CloudEvents specification to
+   [HTTP](http-transport-binding.md).
+1. Include a specification for mapping the CloudEvents specification to
+   [JSON](json-format.md).
+1. Changes to the spec to facilitate adoption.
+1. Publicize at conferences
+   ([CloudNativeCon Europe](https://events.linuxfoundation.org/events/kubecon-cloudnativecon-north-america-2018/)).
+1. Interoperability demo.
+    1.  At least one open source implementation of sending and receiving
+	    events, see
+		[community open source](https://github.com/cloudevents/spec/blob/master/community/open-source.md).
+    1. Events are sent by code written by Developer1 and recieved by code
+	   written by Developer2, where Developer1 has no knowledge of Developer2.
 
 *0.2*
 
-1. Have at least 2 implementations of the specification that can demonstrate interoperability.
-1. Include a specification for mapping the CloudEvents specification to HTTP.
-1. Include a specification for mapping the CloudEvents specification to JSON.
-1. Changes to the spec to facilitate adoption.
-1. Draft documentation and developer guide.
-1. Publicize at conferences ([CloudNativeCon Europe](https://events.linuxfoundation.org/events/kubecon-cloudnativecon-north-america-2018/)).
+1. Incorporate learnings and feedback from interop demo to support wider
+   adoption.
+1. Draft documentation, developer and/or user guide.
+1. Resolve all known large design issues (excluding security issues)
+1. Resolve all known "proposed required attributes" issues
+1. Interoperability demo 2
+    1. Details to be determined
+	1. Showcase demo at conferences - perhaps KubeCon NA 2018
+1. Define the set of protocol and serialization mappings we're going to
+   produce for 0.4 milestone
 
 *0.3*
 
-1. Interoperability demo.
-    1. At least one open source implementation of sending and receiving events.
-    1. Events are sent by code written by Developer1 and recieved by code written by Developer2, where Developer1 has no knowledge of Developer2.
-    1. Draft trigger specification (e.g. how events + actions are associated).
-1. Draft documentation and user guides.
+1. Resolve all known "proposed optional attributes" issues
+1. Resolve all known "security related" issues
 
 *0.4*
+1. Complete the proposed set of protocol and serialization mappings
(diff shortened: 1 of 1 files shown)


roadmap.md
```diff
@@ -23,26 +23,38 @@ implement the spec.
...
-*0.3*
-
 1. Interoperability demo.
     1. At least one open source implementation of sending and receiving events.
     1. Events are sent by code written by Developer1 and recieved by code written by Developer2, where Developer1 has no knowledge of Developer2.
     1. Draft trigger specification (e.g. how events + actions are associated).
-1. Draft documentation and user guides.
+
+*0.2*
+
```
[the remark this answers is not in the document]
duglin: I like that - adding...

roadmap.md
```diff
@@ -23,26 +23,40 @@ implement the spec.
 1. Create and deploy a website that features a simple overview, email list and directs visitors to Github.
 1. Store all website assets in the CloudEvents repository, under the governance
 of the working group.
-
-*0.2*
-
 1. Have at least 2 implementations of the specification that can demonstrate interoperability.
 1. Include a specification for mapping the CloudEvents specification to HTTP.
 1. Include a specification for mapping the CloudEvents specification to JSON.
```
ultrasaurus: Add links to these specs

roadmap.md
```diff
@@ -23,26 +23,40 @@ implement the spec.
...
 1. Changes to the spec to facilitate adoption.
```
ultrasaurus: delete? -- the interop demo was done very quickly after 0.1 was adopted so I wouldn't say this really happened, and 0.2 has "Incorporate learnings and feedback from interop demo to support wider adoption."

roadmap.md
```diff
@@ -23,26 +23,40 @@ implement the spec.
...
-1. Draft documentation and developer guide.
 1. Publicize at conferences ([CloudNativeCon Europe](https://events.linuxfoundation.org/events/kubecon-cloudnativecon-north-america-2018/)).
-
-*0.3*
-
 1. Interoperability demo.
     1. At least one open source implementation of sending and receiving events.
```
ultrasaurus: suggest addling link, like:

At least one open source implementation of sending and receiving events, see [community open source](https://github.com/cloudevents/spec/blob/master/community/open-source.md)

roadmap.md
```diff
@@ -23,26 +23,40 @@ implement the spec.
...
     1. Events are sent by code written by Developer1 and recieved by code written by Developer2, where Developer1 has no knowledge of Developer2.
     1. Draft trigger specification (e.g. how events + actions are associated).
```
ultrasaurus: Move to 0.2 -- did I miss this? if so, pls add link... In any case, I believe I was tracking all changes up to 0.1 so would be part of next milestone, it's outside of the 0.1 specification, so leaving it here is a bit confusing.

roadmap.md
```diff
@@ -23,26 +23,40 @@ implement the spec.
...
+1. Interoperability demo 2
+    1. Details to be determined
+	1. Showcase demo at conferences - perhaps KubeCon NA 2018
+1. Define the set of protocol and serialization mappings we're going to
+   produce for 0.4 milestone
+
+*0.3*
+
+1. Resolve all known "proposed optional attributes" issues
+1. Resolve all known "security related" issues
```
ultrasaurus: I would specifically call out Authorization -- would merit a milestone since I would guess implementation will want to checkpoint on that version.  Might be small changes to the spec, but will likely require a bit of discussion and should include validation with implementation

roadmap.md
```diff
@@ -23,26 +23,40 @@ implement the spec.
...
+1. Resolve all known "proposed required attributes" issues
+1. Interoperability demo 2
+    1. Details to be determined
+	1. Showcase demo at conferences - perhaps KubeCon NA 2018
+1. Define the set of protocol and serialization mappings we're going to
+   produce for 0.4 milestone
+
+*0.3*
+
+1. Resolve all known "proposed optional attributes" issues
```
ultrasaurus: I'm not convinced that this needs to be tied to a particular milestone -- from my perspective security and authorization trump this point by a long margin.

roadmap.md
```diff
@@ -23,26 +23,40 @@ implement the spec.
...
+1. Resolve all known "security related" issues
 
 *0.4*
+1. Complete the proposed set of protocol and serialization mappings
+1. Resolve "Process" related issues
```
ultrasaurus: What are these?

roadmap.md
```diff
@@ -23,26 +23,40 @@ implement the spec.
...
+
+*0.5*
 
-1. Incorporate learnings and feedback from interop demo to support wider adoption.
+1. Resolve outstanding clarifications and non-semantic issues
 1. Collaborate on libraries and supporting tools to use CloudEvents and
```
ultrasaurus: this is pretty vague... could mean a huge number of things!  maybe "define and prioritize libraries and supporting tools that will accelerate adoption of CloudEvents" ?

roadmap.md
duglin: Done

roadmap.md
duglin: While the interop demo was completed just after 0.1, I consider it to have been done *for* 0.1 - and I'm not sure the exact ordering really matters when we're just talking about a matter of days.  Besides, saying its done for v0.2 would be even more misleading since by the time 0.2 goes out the 0.1 demo may not even be around - its already pretty close to gone now.

roadmap.md
duglin: Done

roadmap.md
duglin: hmm, since in the original roadmap this was listed under the interop demo line I assumed it was all just related to that - and not suggesting a new spec. I'll remove it since it will just lead to confusion.

roadmap.md
duglin: All work will continue to be done based on when PRs are submitted. So, if PRs for later milestones are submitted prior to use entering that milestone's timeframe they'll still be looked at. As for security itself, suggest an alternative milestone to put those items but for now I assuming trying to get the core required attributes defined was the most important thing. And note that if we think we need a required security attribute then that would fall into that category.

roadmap.md
duglin: I'd prefer to avoid being too specific in the roadmap since it'll just lead to a technical discussion, but if you want to suggest an alternative wording I'm open to that.

roadmap.md
duglin: e.g. https://github.com/cloudevents/spec/pull/16

roadmap.md
duglin: done

PR #218, opened 2018-05-25
Proposal for an event key field as an extension

### LINKED ISSUE #209: Event keys
Event streams often have an ordering which is important for a set of events for the same entity, but not for events of different entities. For example, consider events for a user management system, you might have user created and user deleted events. Here's an example stream:

```
eventID: 1, eventType: com.example.user.created, data: alice
eventID: 2, eventType: com.example.user.created, data: bob
eventID: 3, eventType: com.example.user.deleted, data: bob
```

Now in this case, it matters which order events 2 and 3 are handled in, since they are both for the user called `bob`, user `bob` cannot be deleted before he was created, logically that doesn't make sense. However, event 1 and 2 can be handled in any order, since they are for two users, `alice` and `bob`, it doesn't matter which order they are handled.

For performance and scalability reasons, many message transports, such as Kafka and Kinesis, implement partitioning of event streams. This partitioning however can cause events to be processed out of order. This could happen if two events that need to be processed in order are published to two different partitions. To ensure that message ordering is maintained, these transports typically implement a concept of message keys, where each message has a key associated with it that correlates to the entity that the event that the message carries is for, and all messages with the same key go to the same partition.

This key could be expressed in the CloudEvents spec using an optional `eventKey` field of type `Object`.

A use case where this might be helpful is if there was some middleware doing a generic transformation of events from one source to another, as it would allow the transformation to be done while maintaining the event ordering, without any knowledge of the domain data or underlying transports. The middleware would only need to ensure that the `eventKey` from incoming events was propagated to outgoing events.

Assigned to: @jroper

jroper:
Closes #209.

This adds an eventKey field to the spec, for the purposes of establishing a causal relationship between events.

It would be worth discussing and getting feedback from multiple sources on the type of this field. Originally, I was going to make it `Object`, but if it's an object, it may need additional information around it such as a content type.

Instead I've made it `String` for simplicity, this works well with existing implementations such as AWS Kinesis and Azure Event Hubs, but may present an issue for Kafka which treats its keys as binary, and provides out of the box support for example for 64 bit long encoding and double precision floating point encoding. That said, it's not impossible for Kafka to support this, it just means any automated support for events in Kafka might require base64 encoding them, or potentially including additional metadata to indicate how its been encoded. Generally, this shouldn't be a problem anyway, since the value of the key is typcially opaque, and primarily useful for deciding on a partition, so as long as it is stable and can be hashed, it's fine.

### DIFF
--- a/extensions/partitioning.md
+++ b/extensions/partitioning.md
@@ -1,0 +1,42 @@
+# Partitioning extension
+
+This extension defines an attribute for use by message brokers and their
+clients that support partitioning of events, typically for the purpose of
+scaling.
+
+Often in large scale systems, during times of heavy load, events being received need to be
+partitioned into multiple buckets so that each bucket can be separately processed in order
+for the system to manage the incoming load. A partitioning key can be used to determine
+which bucket each event goes into. The entity sending the events can ensure that events
+that need to be placed into the same bucket are done so by using the same partition key on
+those events.
+
+## Attributes
+
+### partitionkey
+
+* Type: `String`
+* Description: A partition key for the event, typically for the purposes of
+  defining a causal relationship/grouping between multiple events. In cases
+  where the CloudEvent is delivered to an event consumer via multiple hops,
+  it is possible that the value of this attribute might change, or even be 
+  removed, due to transport semantics or business processing logic within 
+  each hop.
+* Examples:
+  * The ID of the entity that the event is associated with
+* Constraints:
+  * REQUIRED
+  * MUST be a non-empty string
+
+## Encoding
+
+### In-memory formats
+
+The partitionkey attribute extension uses the key `partitionkey` for
+in-memory formats.
+
+### Transport format
+
+The Partitioning extension does not customize any transport binding's storage for
+extensions.
+
--- a/documented-extensions.md
+++ b/documented-extensions.md
@@ -48,5 +48,6 @@ their `Attributes`.
 ## Known Extensions
 
 - [Distributed Tracing](extensions/distributed-tracing.md)
+- [Partitioning](extensions/partitioning.md)
 - [Sampling](extensions/sampled-rate.md)
 - [Sequence](extensions/sequence.md)


Vlaaaaaaad:
This seems kind of a duplicate of https://github.com/cloudevents/spec/pull/128, doesn't it? This has a lower scope, but still.

Some context: in the May 24th call the PR that sparked this( https://github.com/cloudevents/spec/pull/191) was discussed and due to the fact that the idea of event and function correlation( be it in simple chaining, more complicated workflows or whatever else) keeps coming up there will be a separate longer meeting to discuss tackling this( see Doodle at https://doodle.com/poll/kqkmdedrznpwgcq6 ). There are also 2 relevant workstreams for this: https://github.com/cncf/wg-serverless/issues/56 for events and https://github.com/cncf/wg-serverless/issues/55 for functions.

jroper:
I should be at the meeting later this week (as long as I remember to set my alarm, it's at 4am my time).

I did look at #128 first, the thing that made me think it's different is that it says the key is for an "application instance". The event key field definitely has nothing to do with applications, and is all about entities, indeed, multiple applications could emit events for the same entity. Though perhaps the wording is bad there, perhaps the wording there is specifically thinking about an IoT world, where application and entity are often the same thing.

If #128 was to fulfill this, then I think the following things would need to be addressed:

* The field itself as specified in #128 is incompatible with existing implementations of partition keys in technologies like Azure Event Hubs, AWS Kinesis and Kafka. The correlation id is specified to be a path in the message, both those 3 technologies have a separate key value that may not be in the message at all (indeed, specifying a path in the message is odd because it pretty much forces the message to be something where paths make sense, like JSON). So I think #128 would need to be updated to say that it is a key value, not a path in the message.
* To be useful for the purposes of maintaining ordering while partitioning, I think #128 would need to explicitly state that messages with the same correlation id must be delivered in order relative to each other. Otherwise, implementations won't know that that's an intended purpose, and won't implement it.

BenBeattieHood:
I agree. A single cause (correlation-id) can have multiple events: eg. the command 'PromoteUser' might emit events 'UserMovedTeams', 'UserChangedTitle', 'UserPayGradeChanged' etc

jroper:
Interestingly, I read #128 as being about correlation to an application, it does say in the description:

> A path string pointing to a field in the event data that can be used to correlate all the events associated with the same application instance.

And then as an example it lists `headers.body.cathy-house-number`. A correlation id for the purposes of tracing a single interaction with the system, or correlating a request with a response, which I think is what you are referring to, is a completely different thing to what is being described in #128. But I think that's a major problem with #128, it's misleading, when people think "correlation-id" in the context of messaging, they think tracing and request/response correlation as you did. So that definitely has the wrong name I think. Existing implementations that have this concept use the word key (in some cases, partition key), so the name here should be key, so as not to be confused with correlation ids used for tracing.

BenBeattieHood:
Thanks for clarifying @Vlaaaaaaad 

I'd disagree that correlation id can be used as a partition key. The correlation id refers to the command or incident caused a batch of events. In contrast, an aggregate considered as the boundary of concurrency,

jroper:
As discussed in the meeting today, one of the main reasons why this would be useful is that there are multiple transports that support this as a first class concept, including Kafka, Azure Event Hubs and AWS Kinesis (and that's just the ones that I know off the top of my head - any cloud based messaging service that intends to handle events at scale is going to need to support some form of partitioning, and hence have a concept of a partition key).

Personally, I'm moving towards calling it a `partitionKey`, as this makes it absolutely clear what its intended purpose is. This could limit the scope to the extent of excluding other useful ways it would be used, but I do think the most important reason for its existence is partitioning, and by naming it partition key, this makes it exactly clear what it is and isn't, which I think is an important thing in a spec that is going to abstract across many different transports. There are many other types of keys out there when considering messaging, and there's a lot of overlap with keys and labels, and in the labels discussion we have decided to explicitly not mandate a particular name for them, rather they're just generic properties, introducing a generic name like `eventKey` for partition keys I think is in conflict with the decision made for labels, I think it needs to be `partitionKey` to make it distinct and give it a reason to exist as a first class concept. Any other uses of keys can be covered by using the `properties` field defined for labels.

duglin:
From the 6/15 f2f notes:
- Because this is about getting this working with Kafka, we can solve this in transport concerns.
- AI Clemens and Doug: Add this non-goal of defining transport

bluemonk3y:
I would agree with @jroper  - and also prefer 'partitionKey' - it maps on to Kafka, Kinesis, and Eventhub concerns and removes ambiguity about its purpose. The constraint of the key being that it needs to ensure correctness in the streams representation. So while it will be pushed into transport concerns it is a field that the user may need to control, it is unlikely that correlationId will meet the ordering/correctness where there is a fan-out, fan-in of events (or other patterns)

duglin:
More clarify from @clemensv about the f2f resolution :This means that the transport binding ought to define how to extract information like this from the event metadata. The partitioning key is a transport artifact.

duglin:
@jroper if the platform decides how the string is to be interpreted then that implies some out of band communication between the two - and not something that a random implementer of the spec can really use, correct?

If so, I wonder if:
1) in order for this to be a useful property for the spec, if we need to actually be more prescriptive about its value and how consumers are to interpret it, or
2) make it a struct where there's a sub-field (e.g. "type") that allows for a URL to indicate how the "value" sub-field is meant to be interpreted. Then our spec offers the common place for information like this to go, but provides the extensibility for people to define their own interpretation rules. And we can even include some well-defined URLs for the common usecases. Or,
3) leave this as an extension because each provider will have their own way they want this to appear. Which means we don't add anything to our spec.

duglin:
@jroper any update on this?

duglin:
@jroper ping

jroper:
Sorry @duglin, I'm on paternity leave at the moment, is it ok if I get to this in a week?

duglin:
Of course and congrats!

duglin:
@jroper hope all is well with you and the little one... any update on this?

duglin:
Couple of questions:
- this attribute will link events together (presumably because they're in different queues) and the consumer is supposed process events with the same partitionKey in proper order, right? If so, what defines the order?
- does this relate to the 'sequence' extension being proposed?

bluemonk3y:
@duglin  - I agree - the event key forms the partition key to ensure the logical stream is maintained and hence can be processed correctly. There should also be an event-time which is further used to determine when the event was created and can be used by processors to ensure correct ordering. The downside to this approach is that it presumes a WAL/clock sync, so it could be used in conjunction with a sequence that is incremented as subsequent events within the stream are generated. The benefit of the event-time is that it helps capture latency within the stream.

duglin:
If we expect the eventTime to be the ordering mechanism then it might be good to say that in there, and say that if partitionKey is present then eventTime MUST also be present.

duglin:
@jroper thank you very much for that example - that really clears it up for me.
The text `...the order of delivery of related events, that is events with the same partition key, is maintained.` read to me like "ordering across queues". 

I wonder if something like this might be ok:
```
A partition key for the event, such that if a transport
partitions events into multiple queues then all events
with the same partition key will be placed into the same
queue to ensure their delivery order will be maintained.
```

duglin:
we could always say `queue/stream`  :-)

I'm actually more wondering about the bigger question... does this belong in the spec or should it be an extension?  Given the direction of the group on yesterday's call w.r.t. our "bar" for extensions, I need to ask... how many people in this group would implement, support or use this property?  That answer will help us determine which direction to go.

bluemonk3y:
yeah - that is a good point, and part of the challenge I guess. It's going to be a mandatory requirement only for transports that support horizontal scaling; (not http)

Similar to how the httpTransport uses a contentType mapping - wouldn't the transport bindings then stipulate it as a mandatory extension field? i.e. KafkaTransportBinding

BenBeattieHood:
Although I agree with the direction of this, I think @duglin in our current wording we're blending two concerns: 
- partitioning (which relates to transport, eg performance and/or availability)
- order (which relates to a causal relationship between events, and therefore has some impact on the logical coupling between producer and consumer)

I think this `partitionKey` value relates only to partitioning here ^; and, although that may aid ordering, it is not for the _purpose_ of ordering.

Would this be another phrasing for the field?:
```
A partition key for the event, used to calculate the logical partition responsible for sharding storage and transport of the event.
In many stream implementations, events within a logical partition are guaranteed to be in order during consumption - but this depends on the underlying stream implementation. If this is a requirement, please see 'Sequence and SequenceType' event metadata, which should be used to determine and guarantee the causal order of events.
```
(nb., logical in the above as distinct from physical, which is an underlying impl responsibility)

bluemonk3y:
This thread is getting pretty long now and as Ben rightly points out we are overloading semantics; I'm also going to change my tune and say that partitionKey seems wrong (sorry guys!). The notion of partitioning should be hidden from the spec, however, the logical grouping of events should be visible in the spec. For want of a better label it would make sense to call it 'stream-key' however I'm not wedded to it - just that it should not contain partition - or event group-key.

I'm in favor of the following: 

> A group-key for the event is for the purposes of defining a causal relationship between multiple events, such that if a transport partitions events into multiple groups then all events with the same group-key will be placed into the same partition to ensure their delivery order will be maintained.

jroper:
The `source` attribute could overlap with this, but to be useful, I think the `source` attribute would need to be changed and constrained. For a start, the `source` attribute says it describes the event producer, but events for the same entity could possibly come from different producers - eg consider an IoT application where you have two different sensors for the same entity, a sensor the detects if the door is opened or closed, and a sensor that reports if the door is locked or unlocked. Events from these sensors would have different `source` URIs according to the spec, but if partitioned, you would want them to go in the same partition so that they can be handled in the order that they were emitted, and to do that, they need the same key. In which case, `source` can't be used to achieve this unless `source` is misused to not describe the producer.

jroper:
@duglin to answer your question:

> how many people in this group would implement, support or use this property?

I think it's safe to say that all implementations that support partitioning in the transport would implement, support and use it, and that would be Kafka, and everything out there with a Kafka style architecture, which includes EventHubs and Kinesis.

For everything else, even if they did implement support for it and use it, obviously something like MQTT or HTTP is going to have support for the property existing, but I doubt there would be any use of it beyond simply making it available in APIs and possibly forwarding it on in routers. But where I think it gets interesting, and where I think the use case for having this in the spec is, is where messages move from one transport to another through some transport agnostic routing approach, for example, messages coming from IoT devices through MQTT, and ending up in Kafka queues. This is where I think the value of such a field is important, since it allows the source of the event to specify the event key, that key then gets forwarded by MQTT even though MQTT doesn't care about it, but then when routed to Kafka, the Kafka consumer sees it there and so is able to use it to select the partition. Without support for the key in cloud events, this won't work automatically, a user would have to write some custom code to decide on what the key is, perhaps by inspecting the event payload, or by applying some semantics to the source attribute to extract an event key out of it, and then use that to select the partition. With support for it in cloud events, this type of high level routing between messaging providers that do or don't support partitioning can be done declaritively with zero custom code, and a causal ordering between events will be maintained.

BenBeattieHood:
I think we could consider using the `source` attribute for this. I understand your point @jroper but I'm not sure I agree entirely - I believe two IoT sensors monitoring one physical object are still independent event sources. 

For example, imagine I had two sensors observing a jet plane, one based on light, the other on sound. The events from the light-based sensor might be:
[10:00:00] jet reached 340m/s
[10:00:01] jet reached 345m/s
[10:00:02] jet reached 350m/s
While those from the sound-based sensor might be:
[10:00:02] jet reached speed of sound

While it's just an example, what I'm trying to say is that unifying the events in order if they were assumed to be from a single `source` would produce nonsense data. 

I believe the sensors which emit the event are responsible for authorising the event - and so are the event's source. 

If a sensor is damaged, and emits bad readings; or if a sensor is in improper environment; do these events speak of the item observed, or of the sensor?

Re partitions: I agree re Kafka-based (topic-based) queues. I also feel there's a lot to be brought in from more of a domain/eventsourcing-oriented approach like geteventstore.com. Each aggregate instance is considered an independent logical stream, under a common root topic. While the topic may be observed for normal consumption, the independent per-aggregate-instance streams may also be observed (and replayed). Partitioning is done per-aggregate-instance-modulo, and (like kafka topics) the events kept in-order per aggregate by reason of each aggregate being only on one logical partition.

BenBeattieHood:
As an aside, although yep I know it happens, I can't for the life of me understand why anyone would put an event in a message queue (eg mqtt), as they don't typically allow broadcast or stream reading. They're much better suited to command queuing. 
Folks confounding events and commands are causing me all sorts of trouble ;)

cneijenhuis:
@jroper For the door-example, I think it is still possible to satisfy both requirements (describe event producer and be used for partitioning), if you allow for the partition to happen on a part of the URI.
```
/buildings/<building-id>/rooms/<room-id>/doors/<door-id>/lockSensor
/buildings/<building-id>/rooms/<room-id>/doors/<door-id>/openSensor
```
If you allow for the partition to happen on `door-id`, it's fine. The advantage is also that it is not the event producer who allows (exactly one) partitioning scheme, but that the consumer can decide where the partitioning needs to happen. E.g. it could also happen on building or room level.

I know this won't be "plug and play" with Kafka, but I'm wondering if it is unrealistic to pull off?

bluemonk3y:
I'm of the opinion that we remain true to the domain model, in that sense, the ```source`` should always be the ```source``` and that ```stream-id``` or ```group-id``` should be an extension (albeit required by all scaling transports). The ```source``` should always be the data provider - i.e. device-id, source-system; the originator of the events. Events emitted from a source system could potentially contain a 1 to many relationships of streams. For example, a trading platform might emit related events for a transaction, a security system might track users entering and passing throughout a building, a clickstream series will contain the source-ip of users accessing a website etc.  

In the above example - the door-id would be extracted as the stream-id when the events enter the system. Note, the producer has to be responsible for specifying the ```stream-id``` in order to ensure correct partition placement of the event - this is contractual by design.

duglin:
I'm assuming this one is not ready for review yet - in case I miss it, don't hesitate to poke me to ensure it gets on the agenda when you think its ready.

jroper:
@cneijenhuis I'm not doubting that the `source` can contain an event key - but I don't think that's that relevant to discuss, because the point of the cloud events spec is to define semantics that don't require any understanding of the event (or its metadata) in order to gain value from them - otherwise we could just tell people to put all this information in the payload and then we don't need the cloud events spec at all.

What is important is whether there is value in processors of events being able to know the event key without having to have any further understanding of the event or its metadata beyond what's defined in the spec. If there is value in that, then putting it in another field and requiring further semantics beyond the spec to extract it doesn't achieve that goal.

And let's make it clear, it is of no value to Kafka itself to have this field defined - Kafka has an event key field in its message whether this spec defines such semantics or not. Kafka doesn't need this. We don't need to have a place to put the event key in the event for Kafka to work. This is about moving from one transport to another, which is one of the major reasons for having a standard at all, so that semantics of events from one transport are maintained when they are passed to another transport, rather than requiring users to define their own mapping every time it happens. I see two reasons why it's valuable - there are multiple transports (not just Kafka) that do partitioning and require an event key, therefore this would aid interoperability when they are routed to each other. The second is to allow the event key to be expressed in a semantic way when events are passed over a transport that doesn't do partitioning, but eventually end up in a transport that does.

cneijenhuis:
@jroper Thank you for taking the time to respond! I think I get what you're trying to achieve more clearly now.

My remaining concern is that the event producer has to choose the partitioning upfront. In your door example, if the producer decides to partition events per door, and your requirement becomes to check if a room (with multiple doors) is locked, that partitioning won't work.

I can see why this isn't a concern where both producer and consumer(s) are developed in-house. However, working on a SaaS product with a diverse set of customers, I don't feel comfortable prescribing how they must partition events on their kafka queue - I may not foresee all use cases, and get it wrong. Or even worse, some customers may have conflicting use cases, and there is no "one size fits all" partitioning.

I understand (now) that it is a trade-off decision, though. It would force the consumers to choose the partitioning scheme (although I think the `source` could be a good default), and the scheme has to repeated for each transport that supports partitioning. I don't have a strong opinion either way.

Tapppi:
@cneijenhuis I understand your concern regarding partitioning keys that conflict with a use-case, but I think you are reading the attribute as more restrictive than it actually is. Disclaimer: my background in partitioned streams is mostly Kafka and that surely affects my view, although I'll try to remain more generic.

### TLDR;

> My remaining concern is that the event producer has to choose the partitioning upfront -- that partitioning won't work.

> I may not foresee all use cases, and get it wrong. Or even worse, some customers may have conflicting use cases, and there is no "one size fits all" partitioning.

This seems to be your main concern, and I'll try to explain why I don't view this as a problem of the proposed attribute, but of partitioned systems in general. *Any* partitioning scheme will constrain the use-cases it can support without repartitioning by virtue of partitioning the stream at all. To me this is a property of partitioning, not the way you implement or surface it's details. If ingesting events to a partitioned system where you want to maintain order within a partition, you **must** partition the events and you **must** know how to order the events, whether an `eventKey` attribute does or does not exist.

### The benefits of `eventKey`

`eventKey` does not, in my view, constrain the partitioning process. It adds information about how the events have *already been partitioned*. Standardisation of that attribute, be it in the spec or an extension, seems like a worthwhile effort to me, as it touches my use-cases and we are all so *special* 😏. If there is no `eventKey` or the partitioning scheme provided does not work for your use-case, you will have to repartition the event stream as a routing step before passing it to the consumer.
 
* If `eventKey` exists, a partitioning scheme is already in-place and **can** be used as is. Well-architected systems will skip repartitioning in most of their streams because the domain specific partitioning has already been applied earlier on in the "pipeline". Events ingested from other systems without a compatible partitioning scheme will have to be repartitioned as they are now.
* If `eventKey` does not exist or is not applicable to your use-case, the ordering must be determined some other way, which is currently **always** true, barring a custom event key extension. The partitioning step will have to be done *every time you are putting events into a partitioned system*. In the worst case, each producer (or producing middleware) must be aware of domain-specific partitioning logic, whether that is part of their architectural responsibility or not.

#### Egress from partitioned systems

If you are running a partitioned system and want to cater to **any** use case imaginable with a globally ordered (i.e. non-partitioned) event API, as I think you were alluding to with your SaaS example, you will have to converge the partitions prior to serving them from the API. This is a fact in partitioned systems that is not made harder, only easier, by the addition of an `eventKey` attribute.

### Ingress and internal events

Let's take `KNative eventing` as an example, with `Kafka` as a backend. For any reasonably sized system your events *will be partitioned whether you like it or not*.
* Internal events will have an `eventKey` set by the producers. Repartitioning is required *IFF* the events cross boundaries where the receiving system no longer can make do with the original partitioning scheme. AFAIK, this is usually a well understood process for partitioned systems.
* For external sources, you might use an applicable `eventKey` present in the event, or a middleware that is configured with enough domain information to do meaningful partitioning as part of event ingestion. It is domain specific logic, which should IMHO not be touched upon in the spec of `eventKey`.

duglin:
poke

gunnarmorling:
Came here via the the Kafka transport PR (https://github.com/cloudevents/spec/pull/337). The notion of event key is definitely needed for this binding. FWIW, I'd be -1 on using the `source` field, as its semantics are different ones than the ones needed here. `partitionKey` kinda works for me, but it exposes a technicality of specific transports (it's not that the key _must_ be used for partitioning for all imaginable transports), so `eventKey` seems best to me, albeit perhaps a bit too generic.

Besides declaring a key verbatim, an alternative I wanted to bring up is to allow for extracting the key from the `data` field via path expressions, e.g. like so: `/customer/id`. It'd be useful in cases where producers are not aware of the notion of "keys" at all but a meaningful key value can be obtained from the payload. The challenge here of this is that bindings would need to have means of extracting the value represented by the given path, and the way of doing this would be dependent on the specific content type of `data`.

Eventually I think having both, a simple defined key and means of extracting the key from the value, would be desirable.

cneijenhuis:
> an alternative I wanted to bring up is to allow for extracting the key from the data field via path expressions

This is partly what I tried to propose in https://github.com/cloudevents/spec/pull/218#issuecomment-424284160, except that the extraction should be based on the `source` field, not on `data`. (`data` can be of any type, and a middleware ideally shouldn't have to parse it. `source` is well-defined, and extracting something out an URI-reference is comparatively straight forward)

See the follow-up comments why this might not be a good idea, especially https://github.com/cloudevents/spec/pull/218#issuecomment-424559947.

bluemonk3y:
> > Besides declaring a key verbatim, an alternative I wanted to bring up is to allow for extracting the key from the `data` field via path expressions

It would be nice, however the problem with this approach is that the CloudEvent is an envelope for the data field. The data field might be avro/protobuf and/or secured. Requiring access to the contents of the data field also raises a security concern (another attack vector).  

If the data doesn't make sense without it (eventKey), and it is also potentially needed for routing concerns - then IMO it should be a top level, explicit property.

clemensv:
As discussed on today's call:

The client is potentially not in a position to decide on the right partitioning strategy for a downstream event distribution system. The partitioning strategy is ultimately a consumer concern that unfortunately leaks through to clients in the Kafka case. Also, the client might not know that an event key is needed.

We should think of transports as the first layer of subscribers to a stream of events emitted by a source. At the routing destination of any transport may be just another routing hop. A device may an event that gets routed to a field gateway via MQTT and the gateway forwards to Kafka. MQTT doesn't need the key and hence you don't get one for your Kafka hop. In a scenario where you have a Kafka route followed by another Kafka route, you might even want a different partitioning pivot. On the first leg you want to group by device locations and in the second leg you want to group by event type.

Thus, instead of putting the burden of producing an event key on the client and then only having one, any transport that requires particular constructs such as keys such define a mechanism by which you can harvest/synthesize a key from an incoming CloudEvent as some sort of transform. The spec doesn't need to prescribe how -- it just needs to say that that's how the key materializes.  A transform that just cooks up a random key might also be valid if that's what you want.

BenBeattieHood:
One of the interesting things about the cloudevents spec is that it's trying to define events - not messages. I think that's important, as an event is more of a logical construct than a data construct.

As a logical construct, an event is only ever partitioned by the producer. By definition (as an event requires intrinsic validation before emission) an event is also logically in-order per producer.

I'd propose the event key is a simple '/producerId/eventSequenceValue'. This is partitionable by any kind of transport you'd want to use.

I'd add that the double-hop scenario regards messages, not events: the 'gateway' consuming the first stream _is_ the producer of the second stream, and so includes validation and is it's own partition. I wouldn't recommend echoing event keys from one stream to a subsequent one, as assumes you'll never filter/validate between the two.

I think we need to be really clear that messages and events are logically different things.

clemensv:
Events aren't messages and precisely because of that, they only provide the context the producer is immediately aware of, and they don't care about who the subscribers are and what the subscriber's context needs might be. Transport bindings are subscribers to the event producer and events are passed to the transport after having been raised.

There are many instances where events are not part of a sequence. Most discrete events that report out individual occurrences of things happening in a system have nothing to anchor a sequence number on. If I create a new blob inside a storage account, I report out that fact as a singular occurrence that has no logical predecessor or successor and there's no good reason for the producer to keep track of some sort of sequence number. 

I'm describing my perspective on the differences between event series and discrete events and the delineation with messaging in this talk: https://redmonk.com/videos/events-data-points-and-messages-clemens-vasters-thingmonk-2017/ 

The core of CloudEvents is specifically useful for discrete events as a prominent architectural notion in the group is that events trigger serverless FaaS actions in a push fashion.

Events series handling typically calls for stateful processors that pull events towards where that state is held, which applies to the Kafka binding with downstream stateful consumers. So it is true that events that run through Kafka will typically have a notion of order, but discrete events do not. But even with series, the producer may not be in the position to produce a monotonically increasing sequence number, because the logical producer identity may be scaled out. For those cases, order or arrival is usually fine as an order criterion, and the key might be the "source" or some context information harvested from the event information.

BenBeattieHood:
@clemensv we kind of agree, but there's something I think that we need to come to more of a shared understanding on.

Events are 'things that have happened'. For something to be determined as 'having happened', a system needs it to be produced by an authority. In order to have authority, a producer must be a trustworthy source that the thing 'having happened' is valid. In order to be valid, it must have validation, - even if only (for iot) the laws of physics. And in order to have validation, there must be some form of state (again for iot this is often provided by the physics of the environment). In short, an event is always 'validated' by either inferred or logical means, and therefore requires a validating state.

In addition to this, in order to validate, an authority must be single. A shared authority in the case of multiple sensors, either a quorum is reached (and so you can think of the quorum witness as the logical authority), or the sensors are producing logically asynchronous (even if physically dependent) events. Following the latter, each sensor - in a system sense - authorises its own events; and if a sensor is defective, the authority behind its events is logically independently questioned. Even if a sensor is unable to produce an ordering key, the system and environment implicitly provide it; and any receiving consumer would assume it.

The main thing I'm getting at here is that events are always validated, by a single authority, - either logically, or by their environment. And as so, in order to be validated, there must be either logical or physical state. And in validating against this logical or physical state, there is inherent (albeit sometimes inferred) ordering in the definition of an event.

Building on this then, relating those events without their inherent order requires more data to allow downstream error correction than logical event. These stateful events (sometimes called state-transfer events, or discrete events) cause, in systems, the same degree of design coupling as direct API (push coupling) calls or queued direct messages. By exposing the underlying authority by exposing the ordering, systems remove transport ordering issues, and allow for less coupling.

So, taking your example of blob storage, if you only ever had a single event 'item created' then I suppose you could avoid the issue because each item is its own stream. But as soon as you wanted events such as 'item downloaded', 'item deleted', 'item permission changed' then you'd require much more downstream coupling (eg assumed creation event if not yet received) than just including ordering.

bluemonk3y:
@clemensv  - the problem we are facing is that there are different types of events for different purposes, discrete, historic, streaming etc. They each have different traits and as such, this affects how they are modeled. 

The key to the eventKey (to use the pun) and as pointed out by @BenBeattieHood  above - is that it is used to capture a sequence of events; more importantly - a sequence of events captures behavior. These are temporal events; they are time sensitive; order dependent. 

You cannot break this order otherwise you corrupt the stream; it becomes meaningless. The eventKey can be a transaction-id, it could be the row-id from an external datastore showing changes to a particular record (CDC event), or from an IoT-device sensor id. 

As pointed out by @BenBeattieHood  - the responsibility of specifying the eventKey is the producer, the authority.

>  The question here is whether we need an extension for a "key", and my position on that is that keys like the ones you need for Kafka and for Azure Event Hubs and for other similar infrastructures are a choice you make based on the partitioning strategy for a particular Topic/Hub and therefore a transport-level configuration concern and nothing the publisher can immediately set.

Sorry @clemensv but I don't agree with this ;( - we all have different experience and background so will naturally see things differently. The eventKey is part of the data model (as above), the fact that the underlying transport uses it to organize delivery is hidden from the producer or consumer; If an event propagates between transports then the transport should preserve it, and it it has the capability to utilize an extension/field etc then it should do so. 

As @gunnarmorling  - suggested extracting it from the data field might work - however.... we don't have security in place yet. I have seen many cases where events are encrypted on a per-message/tenant basis using message headers etc. Apart from being slow, it also provides an attack vector to hackers, not to mention things like key distribution mechanisms etc

@cneijenhuis  wrote
> This is partly what I tried to propose in #218 (comment), except that the extraction should be based on the source field, not on data. (data can be of any type, and a middleware ideally shouldn't have to parse it. source is well-defined, and extracting something out an URI-reference is comparatively straight forward)

I kind of like this as well... but we also have extensions; they seem like the most natural fit: ce_eventKey or something. This way, if other types of events (audit, analytic, versioned etc) are adopted/supported then we have a pattern which is understood and simple.

(this is a super long pr)

BenBeattieHood:
@clemensv can you give some examples of the types of 'discrete' events that do not need a concept of validation at production, and benefit from skipping ordering? IoT I guess you could skip, because it's clear the ordering is per logical and provided by the environment.

@bluemonk3y good coverage of the different aspects playing in here, thanks.

clemensv:
“Blob created” or “VM deployed” or “Smoke detected” are statements of fact relative to a resource and a point in time. They are discrete. I don’t need an event before or after for those to make sense to my app. The resource is captured in “source” and the time is captured in “time”.

clemensv:
Regarding the function of the key: For a partitioned event stream, you can choose your partition based on the event ‘source’, which will give you natural order-of-arrival order and brokers will commonly label your message frame  with a sequence as they arrive in the partition and thus you get sequencing both per source and per partition. If your partitioning criteria is not the ‘source’ or a combination of ‘source’ and ‘key’, it’s likely some information you will need to promote out from the event data. But no matter what you choose, your partitioning strategy is a function of the backend system that receives and routes the events. It is not a function of the producer. When you build a Kafka or Event Hubs publisher, setting the key is an “envoy” function you perform on behalf of the backend to keep the backend out of the business of having to understand the message shape and doing message transforms. I would likely want to set the partition key to ‘source’ or ‘source’+’type’ in most cases and that information is already carried in the event.

duglin:
There may be a time-zone problem but I'll ask anyway, would it help if some of the key folks got on a call to discuss this?  If github is correct, we're looking at Europe and Australia for the main folks - so perhaps early morning Europe time?

bluemonk3y:
This PR has stalled, I like @duglin 's suggestion to jump on a call. Its seems like @BenBeattieHood  @clemensv  @gunnarmorling  and @cneijenhuis and @bluemonk3y are the active participants.

If we can get enough votes then lets jump on a call. I presume this week is out because of kubecon

Pls respond with your dates and timezone. 

Neil @bluemonk3y  - December 17/18/19 BST

cneijenhuis:
Christoph @cneijenhuis - December 17/18/19/20 UTC+1

That said, I am in agreement with what Clemens is saying, and I'm totally fine if the call happens without me.

BenBeattieHood:
I can do Dec 18th, 19th, and 20th, 8am-6pm GMT+11 (Australian Eastern Daylight Time)

BenBeattieHood:
I'd like to add, for the sake of clarity on the call, I still believe we should be addressing *event key* here, and not talking to partitioning: - the single responsibility of the event key being only the logical identity of the event, not dependent on physical storage. 
We need to keep the *logical* partitioning and the *physical* partitioning as separate responsibilities.
If a backend system wants to reuse the event key as a way of physical partitioning, then that's a separate decision, and I feel that should be out-of-scope here. 
Just trying to keep our convo focused here :)

bluemonk3y:
I would appear that we are out of time for a December call. 

Let's move this to next year. W/C Jan 07-11

Neil @bluemonk3y - January 07,08,09,10,11 BST

cneijenhuis:
I agree, we should move it to January. I suggest to do it before or after the regular weekly call.

BenBeattieHood:
Sounds good. I can do Jan 7th-11th, 8am-6pm GMT+11 (Australian Eastern Daylight Time)

duglin:
ping @clemensv for comments about possible call

clemensv:
> Sounds good. Unsure of time of regular weekly call, but I can do Jan 7th-11th, 8am-6pm GMT+11 (Australian Eastern Daylight Time)

I can make 06:30 UTC (07:30 CET, 17:30 AEDT) work on Tue-Thu this week.

clemensv:
> 
> 
> > Sounds good. Unsure of time of regular weekly call, but I can do Jan 7th-11th, 8am-6pm GMT+11 (Australian Eastern Daylight Time)
> 
> I can make 06:30 UTC (07:30 CET, 17:30 AEDT) work on Tue-Thu this week.

Tue will be too tight. Can we try to land on Wed or Thu?

duglin:
Any chance of you folks meetings up early next week to make some progress on this?

clemensv:
@cneijenhuis @BenBeattieHood @bluemonk3y 

Since we have perfect global distribution, here some meeting time options for Wed/Thu this week. If we can't get to a consensus on this, I propose close with no action since the transport bindings can cover the subject, IMO.

https://doodle.com/poll/b8tdpaitqhpdw7q7

Meeting location will be the group Zoom, which I assume is generally open.

duglin:
@clemensv if I'm free I'd like to join and learn more about the issues around this space so please announce the results of the poll here. I won't vote because I don't want to impact the results :-)  I just want to be FYI'd

bluemonk3y:
It looks like the doodle has closed. I cannot meet this week but am open mon-wed next week (EU time) (4th-67th) Regards, Neil

BenBeattieHood:
Is everyone in UTC or CET except me? If so, I very much appreciate the efforts you've gone to try and include AEDT but it seems unfair for me to ask this of you. I'm happy to try and summarize my thoughts before the meeting, and then maybe best to meet without me?

evankanderson:
Happened to notice this due to the call. I'm not sure that the assumption of singular authority holds (for example) in scaled-out cloud systems. Here are a few examples:

## `objectstore.blob.fetched`

In a scale-out blob storage, a read operation may be resolved by multiple different backends, each holding a replica of the data. There is no explicit coordination of these backends in the read path, so events may be emitted in any order (and if the event delivery is performed as a background task, see below).

In Google, this concern could plausibly cover writes as well as reads in several data sources, including BigTable, Firestore, and Spanner, for reasonable definitions of `source` (i.e. a commit which touches key partitions across 3 servers and occurs "at the same time" as a commit in the same table which touches key partitions on 2 distinct servers).

## Non-blocking delivery or bridging

If event delivery (e.g. from scaled out backends) is performed in parallel to the main request (to avoid head-of-line blocking), it is possible for an event which is emitted at time X to be submitted to the next transport at time X+N, possibly later than an event emitted "later". This occurs any time two transports are bridged unless both transports agree on both the `eventkey` and to perform head-of-line blocking. (For example, if an IoT device reconnects to an MQTT bridge and is routed to a new instance, it may submit events which are routed downstream sooner than events which were delivered before the reconnect.)


As such, I'm somewhat sceptical that there is a ["single natural authority"](https://github.com/cloudevents/spec/pull/218#issuecomment-439730697) for all event delivery on an `eventkey`.

I'll try to dial in to the meeting, but I have a conflicting meeting, so I'm not sure if I'll make it.

bluemonk3y:
@clemensv  - I have been traveling and could not join. A bit longer notice on the doodle poll would have been useful; I did try to register but was too late and in the wrong timezone. That being said I still do not agree that this is a properly workable solution and it needs to be a discussion that includes @BenBeattieHood and myself. 

For a call agenda the following would be useful.

1.  agree on 3 different use cases
2. iteration through the options of URL, extensions, suppressed, payload or other mechanisms
3. reach an agreement to either park it until we see more adoption OR actually apply a solution

BenBeattieHood:
1/2 Thanks for the update and work on this @clemensv & all, and I definitely appreciate the efforts made to include AEDT. 

Looking back, we started this PR discussing correlationid, but are now I see we're speaking more about identity. I am truly sorry to say during this convo I had forgotten already we have an `ID` field in the spec, and believe this would handle ordering as needed. Our discussion had related to 'key' and I got unglued by this - very sorry.

This PR being more to do with correlation, perhaps we should move back to `correlationId`; and I am happy to agree that it not be a core part of the spec, and only an extension. But we all have different backgrounds though, and I'm very happy to be wrong on this. 

However, the above both said, I feel we should still keep discussion separate from partitioning, which is an out-of-spec concern.

Re phone calls, I feel it's a stretch for some folks here, and so I truly appreciate all the efforts made to make this convo async and so inclusive. As to more sync convos, @jroper I know you've just had a little baby, I too have two little ones, and I'm guessing others here may also + other needs, so overlaps are hard. So ideally let's continue async for key discussion, but in case it helps the below link covers coordinating for most of the zones (CET being between GMT and EET, and the link only allows for 4 zones): https://www.worldtimebuddy.com/?pl=1&lid=21,213,14,8&h=21

@bluemonk3y BST/GMT
@clemensv CET
@cneijenhuis CET
@duglin EST
@evankanderson PST
@jroper AEDT
@Vlaaaaaaad EET
myself AEDT

Lastly, I am sincerely sorry if I have slowed discussion in this PR, and appreciate what I've learned from you all along the way.

BenBeattieHood:
2/2 As additional discussion re ordering & single-authority has felt fruitful, I'd like to gauge interest in moving the ordering and single-authority aspect of this PR into a separate non-blocking github issue, so we can continue building separately there more of the shared understanding and learning on these tangent issues, without adding additional discussion to the correlation question. Thumbs up/down?

duglin:
All - where are we on this? I really do think we need a call - clearly this async/comment-based discussions are not working as moving us closer to a resolution.

duglin:
In reading through this PR I actually wonder if we're closer to a resolutions than people think. And, forgive me if I'm oversimplifying things, or if I just flat out am not getting it :-) , but let me share my thought process on this:
- there's a desire to enable consumers of events to correlate/group events from a single producer (actually, is it really "single"? dunno, but it probably doesn't matter)
- the reason for this grouping, at least initially, is to keep related events together if events are placed into separate buckets at the consumer
- there appears to be concern that, in the abstract, the spec can't pick a well-defined value (with a concrete semantic meaning for the value) since it seems to be specific to the events/application
- I'm not clear on whether this bit of data has any significance to the application processing or whether it is strictly used for infrastructural purposes (meaning bucket placement). I'm guessing it's more infrastructural because our docs talk about how the CE properties should be (sort of) duplicate info to what's in `data`.  We've had this discussion before... in an environment where no transports are needed would people need the CE wrappering properties/envelope, or should `data` be sufficient for the app to do its jobs? And I believe we landed on "No, `data` should be sufficient".
- If that's still the case, and while technically any transport/middleware could leverage buckets, our current desire for this information seems to be on transports that have buckets as first-class concepts.
- this leads me to think that this isn't a generic concept for the spec but rather something extra (yes an 'extension') that is (sort of) transport specific. By this I mean, calling it something like 'correlationID' might be a mistake because I think a lot of people would mentally jump to some sort of application-level correlation and I don't think that is its real intent.
- so, something more akin to a `key` (as originally proposed) seems to make more sense to me

In the end, I believe more than one expert in this space has commented in this PR that defining this field as an extension would be sufficient - and that's why I think we're close to a resolution. I believe the only open question is whether it should be defined in our `extensions` directory or just within the Kafka transport spec?

My gut is telling me that this field might be useful for more than just Kafka, but with my other "required vs nice to have" comments in mind ( https://github.com/cloudevents/spec/pull/377#issuecomment-466722747 ) it might be best to limit it for now (solve the immediate need) and then consider expanding it later since it costs us nothing to do so. What that means in the context of a concrete proposal is this:
- define `key` as an extension property in the Kafka transport binding spec ( https://github.com/cloudevents/spec/pull/337 )
- when we come across the need for this outside of Kafka then we move it under our `extensions` directory so others are encouraged to use it. This has no impact on implementations of the Kafka transport binding since all we're doing is moving text around in our repo/docs
- if it becomes such a popular property that many transports are using it, we can then consider making it a core optional CE property - again, without any impact on the Kafka implementations

duglin:
ping @jroper @BenBeattieHood @bluemonk3y @Vlaaaaaaad @clemensv

bluemonk3y:
Thank you @duglin. In short, yes there is a data modeling requirement to identify related events.

I agree that it would make sense to start with it in the Kafka specification and then evolve it to extensions when the need arises. By then we should have sufficient understanding of its pros and cons (security, confusion with other parts of the spec etc)

clemensv:
An extension for a "key" might be sufficiently generic for all sorts of cases, but we already have several criteria that classify the event and I don't understand what's semantically missing that ought to go into "key". 

Whether there's an extension or not, the Kafka transport would still have to define a mechanism for how to construct the partition key from the event data, not only based on that field, but probably also on other data. 

As I've already laid out, the partitioning strategyfor an event stream may be different for different stages of a routing cascade, and the partioning strategy for is generally set by the event stream consumer and not by the publisher. That means the event publisher doesn't get to call the shots of what event goes into which partition.

bluemonk3y:
@clemensv  - can you provide use cases on this? I have only seen use cases where the producer provides the eventKey as part of the data modeling exercise (there are many usecases of this nature). 

It should also be noted that there are likely to be many consumers and stream processors, they could each react differently to the same set of events (different processing logic). For them to dictate the partition key would require repartitioning which is costly and should be avoided. This is an antipattern and would make managing topics difficult, break security concerns and potentially overload cluster resources. A topic should be thought of as akin to a database table where keying is part of the data modeling exercise, partition count part of the sizing exercise. It would be the same for any horizontally scaling streaming platform.

clemensv:
@bluemonk3y

Use case derived from DIN92222 effort, if the volume hits levels where partitioning is needed:

Event generated by drive inside robot. Destination is the drive manufacturer‘s predictive maintenance cloud. Routing stages:

1) Unpartitioned to robot control system
2) Partitioned by production cell or line to factory system
3) Partioned by robot model to robot manufacturer cloud
4) Partitioned by component-id to supplier 

The event streams must be pre-sorted by destinations if you want low latency and throughput and hence the topology dictates key choice. Not the publisher.

Also, if we assume the same or similar partition counts for each routing stage, a single key set at origin will degrade in usefulness from stage 2 onwards, because it might hash back to that same partition number that it came from and distribution amongst partitions becomes very uneven.

duglin:
After chatting with a few folks, I'd like to propose the following:
1 - move the definition of this property into the extensions bucket
2 - modify the text slightly:
```
### partionKey

* Type: `String`
* Description: A partition key for the event, typically for the purposes of
  defining a causal relationship/grouping between multiple events. In cases
  where the CloudEvent is delivered to an event consumer via multiple hops,
  it is possible that the value of this attribute might change, or even be 
  removed, due to transport semantics or business processing logic within 
  each hop.
* Examples:
  * The ID ID of the entity that the event is associated with
* Constraints:
  * REQUIRED
  * MUST be a non-empty string
```

This allows us to get experience with this attribute in the field before we consider adding it to the core spec.

ping @jroper @bluemonk3y @clemensv

bluemonk3y:
@duglin  - yes +100

duglin:
ping @jroper @clemensv

duglin:
@clemensv @jroper can you guys comment on the latest proposal? I'd like to see if we can finally put this one behind us.

gunnarmorling:
Sorry for staying silent on this issue lately :(

Regarding the latest proposal above, I think this ignores the point by @clemensv from above:

> a single key set at origin will degrade in usefulness from stage 2 onwards, because it might hash back to that same partition number that it came from and distribution amongst partitions becomes very uneven.

As suggested above, I still think that being able to apply a function which extracts the message key to be used at a given stage would be the most flexible solution. I understand though that it raises concerns in regards to performance and security. So how about this simplified flavor of such function:

The "partitionKey" property wouldn't have the actual key as a value itself, but instead the *name* of another message property. This frees the producer from having to define the partition key (which as argued by others isn't always feasible), it just must define the properties required by later stages to make that decision. So coming back to the robots example, the producer would have to define the fields "production cell", "robot model" and "component-id". Then at the individual stages the value for the "partitionKey" property could be changed from "production cell" to "robot model" to "component-id", and the associated value would be the partition key during that stage.

This approach would give the needed flexibility, without requiring deep parsing/understanding of actual event payloads. WDYT?

duglin:
@gunnarmorling thanks for the comment. If I'm understanding your comment, I actually think we're on the same page w.r.t. the overall concept - that being, each hop that an event can go through might change the `partitionKey` for some reason, right?

Whether the `partitionKey` has the actual value itself or is just a reference to another property with the value seems secondary to me. And on that point, what is the value of this indirection? What if the `partitionKey` needs to be set to something that isn't in the event already? You seem to be suggesting that the sender would need to add a new extension attribute to hold this value so that the `partitionKey` could point to it by name - but, why not just put the value into the `partitionKey` attribute in the first place?

bluemonk3y:
To move this forwards I would favor the simplest approach (by @duglin) and leave extensions and functions for partitionKey extraction until those use cases are better understood. It is also worth noting that the partitionKey extension does not preclude the transport binding from deriving the value as proposed by @gunnarmorling  and @clemensv. This approach would also allow the transport functionality to be developed independently of the spec.

gunnarmorling:
>  that being, each hop that an event can go through might change the partitionKey for some reason, right?

Yes, exactly.

> but, why not just put the value into the partitionKey attribute in the first place?

For the reason above, the key might change at different stages. The suggested indirection would allow for that without having to examine the payload itself, provided the producer has defined properties for everything which might be of interest as a key property at later stages.

clemensv:
Let's get @duglin's proposal into shape as an extension spec and ship it.

duglin:
@jroper - comments? If you're ok with it, do you want to make the edits or shall I?

jroper:
@duglin Edits made. I called the extension "Partitioning", since in future there may be reason to add additional attributes related to partitioning, for example, the id of the partition that the event was on or should be sent to - though I don't think there's any reason to add any others now.

extensions/partitioning.md
```diff
@@ -0,0 +1,35 @@
+# Partitioning extension
+
+This extension defines attributes for use by message brokers and their
+clients that support partitioning of events, typically for the purpose of
+scaling.
+
+## Attributes
+
+### partitionKey
```
duglin: We're lower-casing all of our attribute names now, so `s/K/k/`

duglin:
Aside from the minor syntax comment... LGTM
ping @bluemonk3y @clemensv

gunnarmorling:
LGTM.

bluemonk3y:
LGTM

extensions/partitioning.md
jroper: Updated.

duglin: thanks!

duglin:
if I'm remembering correctly, there were some concerns raised during last week's call but so far no one has added any comments to this. If we don't have any soon then we'll just proceed with the PR as-is.
@cathyhongzhang ??

@clemensv I believe you took an AI to write-up some more background info on how we envision this field (and/or some other fields) but we can do that in a follow-on PR, right?

cathyhongzhang:
We discussed the need to correlate events from different event sources long time ago. If I remember well, the consensus is that the event source/producer itself does not know which key should be used for correlation, instead it is the serverless application or consumer of these events that knows which key/label can be used to correlate these events. 

Is this partition key for multiple events from the same source? 

I would suggest that the author adds definition of event partition so that people will not second-guess what it means and how it can be used.

duglin:
@jroper ^^

duglin:
@cathyhongzhang is this something that can be done in a follow-on PR or do you think it needs to be resolved prior to merging?

cathyhongzhang:
I would like this to be done in this PR so that everyone can be on the same page as to what this partition key means and how to use it.

clemensv:
@cathyhongzhang This mechanism here is for forming partitions, it's not for fine grained event correlation. What this here serves is that you're getting a million events per second and you need to split that up into 100 partitions so that each processor only needs to handle 10000 per second. And you're doing that with a key, so that all events with the same key land at the same processor, but you're not making that choice with the key, but some hashing algorithm makes that choice for you. And if you need to break down the 10000 events/sec again, you need to pick a different key so that there's again even distribution across the 100 upstream partitions that then each ought to get 100 events/sec to handle.

Your notion of event correlation is a very different one. You want to be able to identify, for instance, whether a set of events were raised from different devices that all exist in the same room. That's a completely different motivation. It might compose with the partitioning model (i.e. you might want to keep events from the same place together), but you will also want that correlation also without partitioning even being in play.

duglin:
I like that example, I think it does a good job explaining why and when this type of field is needed in a very easy to follow usecase. And includes the need to change it between hops! Which was a big sticking point for finding a resolution.

@cathyhongzhang , if we added that to the text of this PR would that address your concerns?

cathyhongzhang:
Thanks Clemensv for the explanation! So this is not for the purpose of defining a generic correlation key.  @duglin , yes it will be good to add the explanation, which addresses my concern.

jroper:
Just to make clear, it **is** the event source that knows the event key. Generally, the consumer doesn't care what the key is. The producer sets the event key so that the transport can then partition it correctly.

A more concrete use case might be an e-commerce system. Through the lifetime of a given order, the following events might be produced:

1. Order submitted.
2. Payment received.
3. Order dispatched.

These events must be handled by consumers in order, since it doesn't make sense to handle an order dispatched event before the order submitted event for that order. But in a sufficiently large system, say where there are 100000 orders per second at peak times (eg during Black Friday sales), the events need to be partitioned in order for the system to be able to handle the load. If they were partitioned randomly, then the above messages could all end up on different partitions, and this could result in them being consumed out of order. So instead, when the producer publishes the event, it specifies a key for the event, which would be the ID of the order, and this allows the transport to ensure events with the same key end up on the same partition, and so are guaranteed to be consumed in the correct order. The consumer doesn't usually care about the key (usually, it will already be somewhere in the event anyway), since the transport has already ensured for the consumer that the events are in the correct order - the key has served its purpose by that point.

I'm not sure how much of this should go in the spec - is it the responsibility of the spec to teach people concepts about partitioning events? I would have thought no, this is a specification, not documentation, it shouldn't be teaching concepts. If the reader is unfamiliar with event partition keys, they can search "event partition keys" in Google, and literally every result on the first page of that search is a page that explains the concept (as it happens, most of the results are for Azure event hubs).

duglin:
@jroper thanks for the comment. I agree that our spec isn't the place to "teach" people stuff like that, but there are times when it helps to provide a bit of background on how we expect fields to be used, especially when the terseness of a spec can be misleading to folks who are not experts on that aspect. And that's one of the reasons behind the "primer".  It's for useful info/guidance we want to share but isn't really appropriate for a formal spec.  Now, having said that, since this attribute isn't part of the spec, it's an extension, for now I think it would make sense to add a short bit of descriptive text like you and @clemensv wrote into this extension's md file.  If/when this attribute gets put into the spec itself, then that text can go into the primer.

Does this make sense?

cathyhongzhang:
@jroper Your comment emphasizes that the partition key is to ensure that the events are handled by the consumers in order. But in general, the "guarantee order" part is usually not an attribute associated with a partition key concept. Sequence number is usually used to guarantee the order. Partition key is used a lot in data partition when a big chunk of data is broken up into pieces/segments (e.g. during streaming or for performance improvement of data store) and the key is used to group those segments together. Here is a link to AWS partition key concept https://docs.aws.amazon.com/streams/latest/dev/key-concepts.html

Your concept of partition key serves two purposes: to ensure that events with the same key end up on the same partition, and to guarantee the events to be consumed in the correct order. 

So what is the concept of this partition key in our spec? 

How could the consumer use it? For example, is it only used for grouping events from the same source or could it be used for grouping events from different sources? 

We have had several discussions on event correlation before. My understanding is that this is not positioned for generic event correlation. It will be helpful to the event consumers if we can define/explain this key's concept scope and usage.

duglin:
@cathyhongzhang , I'm sure @jroper or @clemensv will correct me if I'm wrong, but I believe that the difference between the partition key and the sequence number is that the partition key is set by the sender of the event for that particular hop, while the sequence number is more of an implementation detail of the receiving side that will put the message into the appropriate bucket/shard - and it will add/use the sequence number to ensure ordering. While some implementations might choose to expose the sequence number to the applications (e.g. perhaps add it as an extension CE attribute), that is really an implementation choice and not really needed by the receiving application since the infrastructure managing the events to/from the bucket/shard is responsible maintaining the order.

But, you said:
> Your concept of partition key serves two purposes: to ensure that events with the same key end up on the same partition, and to guarantee the events to be consumed in the correct order.

and I don't believe that was @jroper's intent. Yes using the partition key implies there will be some ordering within each bucket, but the partition key itself does do the ordering - it's whatever mechanism the receiver uses to ensure it. Some may use a sequence number, but others might use something else - like an ordered queue. But, that's an implementation detail that we don't need to expose via this extension.  And I think the AWS documentation you referenced is consistent with this.

Net: I actually think you're both in agreement and correct and I think what you're asking for is just some more clarification text.  What if the introductory paragraph was modified to say this:

```
This extension defines an attribute for use by message brokers and their
clients that support partitioning of events, typically for the purpose of
scaling.

Often in large scale systems, during heavy load times, events being received need to be
partitioned into multiple buckets so that each bucket can be separately processed in order
for the system to manage the incoming load. A partitioning key can be used to determine
which bucket each event goes into. The entity sending the events can ensure that events
that need to be placed into the same bucket are done so by using the same partition key on
those events. This is often done to ensure they are processed in the correct order since
events within the same bucket will typically be processed serially. How the receiving entity
ensures this ordering is an implementation detail and out of scope of this specification.
```

@cathyhongzhang Does this help any?  
@jroper @clemensv Did I get this right?

duglin:
ping @cathyhongzhang @jroper @clemensv ^^^

cathyhongzhang:
@duglin Thanks for writing the clarification text. The text definitely helps. As to the sequence or ordering, I don't think it is all the receiving entity's job and it is not just an implementation detail. In many cases, because messages/events might be routed over different network paths which add different latency, the events could arrive at the receiving entity out of the original order. The sending entity has to put a sequence number in the event so that the receiving entity knows the order. As long as we do not put "are guaranteed to be consumed in the correct order" (taken from @jroper's text)  or any "ensure order" text in the partition key description, I am good.

duglin:
Thanks @cathyhongzhang, what if I changed it to be this:
```
This extension defines an attribute for use by message brokers and their
clients that support partitioning of events, typically for the purpose of
scaling.

Often in large scale systems, during heavy load times, events being received need to be
partitioned into multiple buckets so that each bucket can be separately processed in order
for the system to manage the incoming load. A partitioning key can be used to determine
which bucket each event goes into. The entity sending the events can ensure that events
that need to be placed into the same bucket are done so by using the same partition key on
those events. This is often done to ensure the events are processed in the order in which
they are received since events within the same bucket will typically be processed serially.
How the receiving entity ensures this ordering is an implementation detail and out of scope
of this specification.
```

cathyhongzhang:
@duglin Most of the part looks good to me. I feel we can remove this last part ----"This is often done to ensure the events are processed in the order in which they are received since events within the same bucket will typically be processed serially. How the receiving entity ensures this ordering is an implementation detail and out of scopeof this specification."  Adding this part seems to cause more confusion than clarification. My 2-cent.

duglin:
I'm ok with that. @jroper you ok? If so, wanna update the PR? Maybe we can get it in tomorrow????

jroper:
Updated.

duglin:
@jroper thanks!

duglin:
@jroper we discussed this on today's call - and it was exciting  :-)
A couple of things...
1 - could you watch the recording of it once its available? It should be available here: https://www.youtube.com/playlist?list=PLj6h78yzYM2Ph7YoBIgsZNW_RGJvNlFOt  in a few days. And then comment on some of the issues/concerns that Heinz raised.
2 - I put the zoom chat transcript at the end of today's meeting minutes - could you read those? It might give you some insight into some of the concerns even before the video is ready.
3 - could you add some examples (use cases) to the PR? There seemed to be some confusion as to how this will be used, even in the Kafka case. In particular some folks seemed to think the info wasn't necessary and redundant. I think having a crisp explanation (as part of the PR text itself) could help this group as any future readers/users of the extension.

duglin:
@jroper the video is now available here: https://www.youtube.com/watch?v=yCP5K0YkdVA&list=PLj6h78yzYM2Ph7YoBIgsZNW_RGJvNlFOt&index=2&t=0s

Any thoughts?

clemensv:
The recording of the call was time well spent listening. 

We're somewhat going in circles, unfortunately, but I don't think we're that far off from a resolution.

Further up in this thread, @duglin quotes me from the F2F: "This means that the transport binding ought to define how to extract information like this from the event metadata. The partitioning key is a transport artifact."

I agree with Heinz that "partitionKey" as an artifact is completely superfluous if the Kafka SDK client for CloudEvents provides a callback hook into which an app can hook a function that can have full visibility into the event, and then generate a partitionkey as its return value based on any combination of the event metadata and data. A variation of the callback could even determine the partition number outright by some algorithm. 

At the same time, it doesn't hurt for there to be a "partitionKey" extension which the Kafka SDK client for CloudEvents can look for and use when there's no such callback registered. Creating the key while putting together the event is often pretty easy and keeps the app code straightforward in simpler scenarios. And such keys are not exclusive to Kafka - we have partitioned queues and sessions/groups in Service Bus and use partition/session keys to keep related data together in order, and those keys have a very similar function.

- The Kafka binding should mandate a callback mechanism for the implementation that asks the application to provide a key based on a given event. That resulting key never materializes in the CloudEvent and is only used for the Kafka client to pick the partition.
- We should also go forward with this extension that allows for a partitionKey to be specified by the app as it constructs the event, because there are plenty of scenarios where the publisher is aware of the architecture detail and wants to influence partitioning.
- The Kafka binding should first rely on the callback, and if not present rely on the extension, and if not present pick a random partition.

duglin:
@clemensv thanks for that comment/compromise proposal - it also helped me understand some of the concerns that were being expressed.

What do people think?  @jroper @hschaffner @Vlaaaaaaad @bluemonk3y @Tapppi

bluemonk3y:
@clemensv  and @duglin  - LGTM. One minor point - for the transport binding, where the callback is specified, i.e. is it part of the configuration or an anonymous function injected at event-creation time. I see pros/cons for both.

clemensv:
@bluemonk3y The transport binding spec should mandate that the implementation provides a hook for an application-supplied function that generates a key or selects the partition on a per-event basis. The exact shape of that mechanism might vary between implementations; you could make this some extraction rule in a clever framework or just a free form function.

duglin:
This PR was approved on today's call with the following conditions/actions:
- AI Doug: open issue to add examples to this extension - https://github.com/cloudevents/spec/issues/429
- AI Doug: open issue to revisit whether we need this after Kafka PR is done - https://github.com/cloudevents/spec/issues/430
- AI Heinz: join the Kafka PR to talk them out of this extension :-)

Merging!!!!  thanks everyone!

PR #224, opened 2018-05-31
Add an example to the spec and clean-up a little

duglin:
I found it a bit odd that the spec never included an example CloudEvent.
People would have to go to one of the other docs to actually see what
one looked like. For me, its sometimes easier to full understand something
when I can see a real example of what is being defined. Having people leave
our doc to see one felt odd so I just copied one from the JSON mapping spec.

While in there I noticed that our TOC wasn't consistent w.r.t. what we
included so I put all 1st and 2nd level headings in there (excluding the
ones before the TOC).

Also, the TOC included references to two docs w/o any explanation around
them - the usecases and example events from some popular eventing systems.
None of these are normative so they felt really odd to be in our spec's
TOC. However, since they are useful bits of information, I added a pointer
to them from our README instead. This also provides a bit of explanation
about why they exist in our repo at all.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,12 @@ The following specifications are available:
 There is also the [CloudEvents Extension Attributes](https://github.com/cloudevents/spec/blob/master/extensions.md)
 document.
 
+Included within this repository are some documents that helped guide the
+working group's design decisions. They include:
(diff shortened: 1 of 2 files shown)


duglin:
Approved on 6/7

PR #226, opened 2018-05-31
s/property/attribute/ to be consistent

duglin:
Minor fix


duglin:
Approved on 6/7 call

Issue #25, opened 2018-01-25
causation-id context attribute


lfourie:
This is a proposed metadata attribute (from Context Attribute Backlog):

causation-id
    Type: String
    Description: event-id of event that triggered this event to be created (if any). If not triggered in response to an event, then not present. Enables traceability through systems.
    Examples:
        Event A generated with unique event-id. Event handler receives event and generates a new event in response, Event B. Event B copies the event-id of Event A into its own causation-id so that the chain of events can be properly correlated.
    Constraints:
        Optional

duglin:
@cathyhongzhang is this related to your correlation-id issues/PRs? or if not directly related, can it be solved using the same solution - meaning, we provide a place for people to put this an attribute for this info w/o actually defining the attribute itself.

cathyhongzhang:
Yes, I think so.

duglin:
Related to (duplicate of?) #24 and as such copying my comment from there:

Based on our previous discussions, I think we can close this one because it appears that there will be many different ways to do correlation and at this time there doesn't seem to be enough momentary around just one mechanism (or property) to justify adding it to the core spec. We discussed how specific producers can add (via extensions) their own correlation properties to the events though.

I'll add this to the list of possible issues to close for the next call - if you have concerns about this please add a comment.

yordis:
@duglin would be possible to have an official extension similar to the OTEL tracing, except that isn't related to OTEL


Attribute Name | Type | Description
-- | -- | --
correlationid | String | An identifier that groups related events in the same logical flow.
causationid | String | An identifier referencing the event or command that directly caused this event.

Both attributes MUST be strings.

No further constraints are imposed on format (UUID, ULID, etc.)—producers and consumers must agree on conventions.

duglin:
@yordis I can bring this up on our weekly call to see if https://github.com/cloudevents/spec/issues/25#issuecomment-425423215 still applies, but given no one has mentioned this topic in a very long time I'm doubtful.

Is there any reason why you can't just define an extension for your own purposes? That's not much different than what you've suggested... consumers/producers agreeing on what to use/look for.

yordis:
> Is there any reason why you can't just define an extension for your own purposes? 

That is what I have been doing

> but given no one has mentioned this topic in a very long time I'm doubtful.

Never took the time to ask, just did it locally and move on.

The only reason I ask is for alignment, bring the conversation and make sure we all in the same page

duglin:
[...]

PR #227, opened 2018-05-31
Add extension for Distributed Tracing spec.

inlined:
I think this merits the initial statnce of being an extension
for a few reasons:

1. I want to adhere to the CNCF policy against kingmaking even
   though I think the world is better off with one interoperable
   tracing standard.
2. Nothing in the trace state should have an effect on
   application logic.
3. I like exercising our thoughts on extensions.

### DIFF
--- a/extensions.md
+++ b/extensions.md
@@ -30,13 +30,14 @@ platforms like Prometheus are built.
  
 #### traceparent
 * Type: `String`
-* Description: Contains a a trace ID, span ID, and trace options as defined in
+* Description: Contains a trace ID, span ID, and trace options as defined in
   [section 2.2.2](https://w3c.github.io/distributed-tracing/report-trace-context.html#field-value)
 * Constraints
   * REQUIRED
-  * To integrate with Distributed Tracing, this field MUST NOT use standard
-    extension encoding over HTTP(S). `distributedTracing.traceparent` MUST
-    instead be marshaled as the `traceparent` HTTP header.
+  * To integrate with Distributed Tracing, this field MUST NOT use the normal
+    [extension encoding over HTTP(S)](http-transport-binding.md).
+    `distributedTracing.traceparent` MUST instead be marshaled as 
+    the `traceparent` HTTP header.
 
 #### tracestate
 * Type: `String`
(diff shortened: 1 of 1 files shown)


duglin:
Just a few minor thing, but otherwise LGTM.

Reviewers... since the "extensions.md" doc isn't going to be an official deliverable from our WG, the bar is pretty low on this. We don't all need to agree on whether we like the extension or not, rather just that its in scope for the WG (ie. related to cloud events), its written well to be understand by other people and can be implemented.  These are just "experimental" properties with no official standing. 

@inlined thanks for being the first!

jroper:
I'm not that familiar with the distributed tracing spec, so correct me if I'm wrong. I think this is very closely related to #24 and #25. I think if we accept this, then #24 and #25 would just be duplication of fields for achieving the same purpose. Also somewhat vice versa. I don't think CloudEvents should support (whether it's an extension or part of the top level spec) two different mechanisms for attaching trace metadata, that would just be confusing and defeat the purpose because if one party used one, and another used another, then the tracing would not work.

So I think a decision should be made as to whether CloudEvents should be opinionated on supporting and requiring distributing tracing as the tracing mechanism, or whether it should do something differently.

On whether this is an extension or part of the top level spec - I'm not sure that I see the difference between one or the other. The most important thing is that different vendors using this all use the same field names and value format. Whether it's an extension, or part of the spec proper, doesn't seem to make any difference, the consequences of adding it, of maintaining it, of ensuring that it's consistent between each use, are the same either way. I guess what I'm saying is I don't see the point of CloudEvents having any extensions that it documents itself, since I don't see the expectations being any different either way. Extensions should be something that specific transports use that CloudEvents knows nothing about, if it's something that's not transport specific that CloudEvents does document somewhere, then it may as well go in the spec.

inlined:
This is an interesting idea that correlation-id and causation-id would be replaced by this. For correlation, I personally believe that allowing a single ID is a non-goal because we edge into encoding event usage into the event envelope (correlation-id is easily a topic ID which was already decided to be a non-goal). Distributed Tracing also has a "Correlation-Context" which I'm much more open to because it has a property bag and tools can choose which field to route on.

Causation ID is much more interesting. I somehow assumed this was meant for the time series use case. If it's just meant to help correlate the thing that caused an event to the event itself, then I definitely agree in a standards-based approach instead.

duglin:
Approved during 6/15 f2f. We also agreed to:
- Thomas: put each extension in its own doc in an “extensions” dir. Talk about how the RFC2119 keywords only apply IFF the extension is being used
- Clemens: Need an “escape hatch” in our binding specs to allow attributes to override the serialization rules defined.

PR #240, opened 2018-06-20
Add popular IDEs to .gitignore

inlined:
It's very helpful to use IDEs that let us visualize the rendered markdown. It would be more helpful to not worry about local project files getting added to commits.


duglin:
Sure ignore the **best editor ever**... vi ! :-)
LGTM

PR #242, opened 2018-06-20
Factor Distributed Tracing extension into new doc.

inlined:
* Extends extensions.md to clarify that RFC 2119 keywords
  are only for those who implement an extension.
* Notes our resolved rules that each memory format defines
  extensions' placement and each transport binding defines
  a default placement which an extension may override.
* Refactor the contents now in extensions/distributed-tracing.md
  as a first SWAG at a reusable template.

### DIFF
--- a/extensions.md
+++ b/extensions.md
@@ -33,6 +33,6 @@ default placement for extensions, but an extension MAY require special
 representation when transported (e.g. tracing standards that require
 specific headers).
 
-## Known Extensions
+## Documented Extensions
 
 * [Distributed Tracing](extensions/distributed-tracing.md)
--- a/extensions.md
+++ b/extensions.md
@@ -18,34 +18,21 @@ to limit their use of extension attributes to just the ones specified in
 this document. The attributes defined in this document have no official
 standing and might be changed, or removed, at any time.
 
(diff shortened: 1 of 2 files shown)


extensions.md
```diff
@@ -18,34 +18,21 @@ to limit their use of extension attributes to just the ones specified in
...
+Extensions always follow a common placement strategy for in-memory
+formats (e.g. [JSON](json-format.md), XML, Protobuffer) that are
+decided by those representations. Transport bindings (e.g.
+[HTTP](http-transport-binding.md), [MQTT](mqtt-transport-binding.md),
+[AMPQ](amqp-transport-binding.md), [NATS](nats-transport-binding.md)) provide
+default placement for extensions, but an extension MAY require special
+representation when transported (e.g. tracing standards that require
+specific headers).
+
+## Known Extensions
```
duglin: For consistency with your other comment in my PR, can we make this "Documented"?

duglin:
One comment, but otherwise LGTM

extensions.md
inlined: Done

duglin:
Approved on the 6/28 call

PR #243, opened 2018-06-20
Add sampling extension.

inlined:
This pull request supercedes #76 and builds upon #242.
    
Per agreement at the f2f meeting 2018-06-15, this will be ratified
as an extension but not currently as part of the core spec. The
early inclusion of the spec and the ability for side-cars to impose
sampling early in the pipeline make it reasonable to start with
this as an extension.
    
I know the original author hoped to make this a core feature, though
there are two upsides:
1. Due to other decisions on 2018-06-15, this will have zero impact
  on JSON encoding (but will in Proto and may in SDKs)
2. Extensions now get a whole document. This lets us pitch sampling
  in more detail.

Necessary additional changes:
* Added `Integer` to our type system
* Added convention for how scalar extensions should be documented.
  I'm not really excited about this proposal so others are welcome.
    
Special thanks to @maplebed for kicking this effort off.

Closes #76

### DIFF
--- a/extensions.md
+++ b/extensions.md
@@ -31,11 +31,15 @@ decided by those representations. Transport bindings (e.g.
 [AMPQ](amqp-transport-binding.md), [NATS](nats-transport-binding.md)) provide
 default placement for extensions, but an extension MAY require special
 representation when transported (e.g. tracing standards that require
-specific headers).
+specific headers). Extension authors SHOULD only require special
+representation in transport bindings where extensions integrate with
+pre-existing specs; extensions with custom transport bindings are much
+more likely to be dropped by middleware that does not understand the
+extension.
 
 As a convention, extensions of scalar types (`String`, `Binary`, `URI`,
-`Number`) document their format with the attribute name `(self)` in the
-documentation's "Attributes" section.
+`Number`) document their `Value` and structured types document their 
+`Attributes`.
 
 ## Known Extensions
 
--- a/extensions.md
+++ b/extensions.md
@@ -33,6 +33,11 @@ default placement for extensions, but an extension MAY require special
 representation when transported (e.g. tracing standards that require
 specific headers).
 
-## Documented Extensions
+As a convention, extensions of scalar types (`String`, `Binary`, `URI`,
+`Number`) document their format with the attribute name `(self)` in the
+documentation's "Attributes" section.
+
+## Known Extensions
 
 * [Distributed Tracing](extensions/distributed-tracing.md)
+* [Sampling](extensions/sampled-rate.md)
--- a/extensions.md
+++ b/extensions.md
@@ -37,7 +37,7 @@ pre-existing specs; extensions with custom transport bindings are much
 more likely to be dropped by middleware that does not understand the
 extension.
 
-As a convention, extensions of scalar types (`String`, `Binary`, `URI`,
+As a convention, extensions of scalar types (e.g. `String`, `Binary`, `URI`,
 `Number`) document their `Value` and structured types document their 
 `Attributes`.
 
--- a/extensions/sampled-rate.md
+++ b/extensions/sampled-rate.md
@@ -1,17 +1,17 @@
 # Sampling extension
 
 There are many cases in an Event's life when a system (either the system
-creating the event or a system transporting the event) may wish to only emit a
-portion of the events that actually happened. In a high throughput system where
-creating the event is costly, a system may wish to only create an event for
-1/100 of the times that something happened. Additionally, during the
+creating the event or a system transporting the event) might wish to only emit
+a portion of the events that actually happened. In a high throughput system
+where creating the event is costly, a system might wish to only create an event
+for 1/100 of the times that something happened. Additionally, during the
 transmission of an event from the source to the eventual recipient, any step
-along the way may choose to only pass along a fraction of the events it
+along the way might choose to only pass along a fraction of the events it
 receives.
 
 In order for the system receiving the event to understand what is actually
 happening in the system that generated the event, information about how many
-similar events may have happened must be included in the event itself. This
+similar events happened would need to be included in the event itself. This
 field provides a place for a system generating an event to indicate that the
 emitted event represents a given number of other similar events. It also
 provides a place for intermediary transport systems to modify the event when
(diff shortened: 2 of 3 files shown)


duglin:
@inlined I'm confused as to why github doesn't see a conflict when this touches (and overrides) some of the same files from #242 - and in the process I think we'd lose some edits (like the rename of the section to `Documented Extensions`).  Maybe a rebase will help - not sure.

extensions/sampled-rate.md
```diff
@@ -0,0 +1,36 @@
...
+happening in the system that generated the event, information about how many
+similar events may have happened must be included in the event itself. This
+field provides a place for a system generating an event to indicate that the
+emitted event represents a given number of other similar events. It also
+provides a place for intermediary transport systems to modify the event when
+they impose additional sampling.
+
+## Attributes
+
+### (self)
```
duglin: I wonder if using `(self)` is less clear than just putting the actual text here?  I'd kind of prefer to just be explicit about it instead of having a level of indirection.

extensions/sampled-rate.md
```diff
@@ -0,0 +1,36 @@
...
+  For example, if a system sees 30 occurrences and emits a single event, `rate`
+  would be 30 (29 not sent and 1 sent).
+  Consumers SHOULD assume a value of `1` when the extension is omitted.
+* Constraints
+  * The rate MUST be positive.
+
+## Encoding
+
+### In-memory formats
+The Sampling extension uses the key `sampledRate` for in-memory formats
```
duglin: missing period?

extensions/sampled-rate.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Attributes
+
+### (self)
+* Type: `Integer`
+* Description: The rate at which this event has already been sampled. Represents
+  the number of similar events that happened but were not sent plus this event.
+  For example, if a system sees 30 occurrences and emits a single event, `rate`
+  would be 30 (29 not sent and 1 sent).
+  Consumers SHOULD assume a value of `1` when the extension is omitted.
```
duglin: s/SHOULD/MUST/ - although, its kind of weird to have any constraints when the extension isn't used.

Would it be better to say:
A value of `1` is the equivalent of this extension not being used at all.

extensions/sampled-rate.md
```diff
@@ -0,0 +1,36 @@
...
+* Constraints
+  * The rate MUST be positive.
```
duglin: Perhaps "... MUST be greater than zero." ?   Didn't want to get into the possible confusion around whether zero is positive or negative :-)  so let's just be explicit.

extensions/sampled-rate.md
duglin: Why do we mention this for this attribute but not others?

duglin:
@inlined some comments need attention

extensions/sampled-rate.md
inlined: Changed extensions.md to say that structured extensions' documentation will conventionally have an `attributes` section and scalar extensions` documentation will conventionally have a `value` section.

extensions/sampled-rate.md
inlined: I'm ambivalent. Taking the change.

extensions/sampled-rate.md
inlined: I'll change the prose rather than lecture everyone that zero is neither positive nor negative =p

extensions/sampled-rate.md
inlined: What do you mean? The distributed tracing extension doc also says

> The Distributed Tracing extension uses the key `distributedTracing` for in-memory formats

duglin: Didn't realize it did, but I was thinking about the core properties. no biggie either way.

extensions/sampled-rate.md
```diff
@@ -0,0 +1,39 @@
+# Sampling extension
+
+There are many cases in an Event's life when a system (either the system
+creating the event or a system transporting the event) may wish to only emit a
```
duglin: s/may/might/

duglin: since this isn't a normative statement, its just exemplary

extensions/sampled-rate.md
```diff
@@ -0,0 +1,39 @@
...
+portion of the events that actually happened. In a high throughput system where
+creating the event is costly, a system may wish to only create an event for
```
duglin: s/may/might/    since this isn't a normative statement

extensions/sampled-rate.md
```diff
@@ -0,0 +1,39 @@
...
+1/100 of the times that something happened. Additionally, during the
+transmission of an event from the source to the eventual recipient, any step
+along the way may choose to only pass along a fraction of the events it
```
duglin: s/may/can/

extensions/sampled-rate.md
```diff
@@ -0,0 +1,39 @@
...
+receives.
+
+In order for the system receiving the event to understand what is actually
+happening in the system that generated the event, information about how many
+similar events may have happened must be included in the event itself. This
```
duglin: s/may have happened must be/happened would need to be/

duglin:
overall LGTM just a fit syntax things

duglin:
still LGTM

All - please review (and LGTM, or not) for next week's call

Vlaaaaaaad:
LGTM

extensions.md
```diff
@@ -31,8 +31,17 @@ decided by those representations. Transport bindings (e.g.
...
 representation when transported (e.g. tracing standards that require
-specific headers).
+specific headers). Extension authors SHOULD only require special
+representation in transport bindings where extensions integrate with
+pre-existing specs; extensions with custom transport bindings are much
+more likely to be dropped by middleware that does not understand the
+extension.
 
-## Documented Extensions
+As a convention, extensions of scalar types (`String`, `Binary`, `URI`,
```
markpeek: This PR added and Integer type. Should either list it here or add "e.g." to not list all of the scaler types.

duglin: +1 to adding e.g.

duglin:
All, please review this one. I'm putting into the category of "easy/quick approval" for this week's call, so please speak up prior to the call if you see any issues.

duglin:
Approved on 8/16 call

PR #248, opened 2018-06-21
Add links to AMQP transport and event formats

lulf:
@clemensv I'm not sure if it was intentional or not to not add the links to AMQP specs on the front page.


duglin:
LGTM

PR #254, opened 2018-06-26
Qualifying-protocols-and-encodings.md

clemensv:
Per last week's discussion, a proposal for which protocols and encodings are eligible for inclusion in the core set of Cloud Events specifications.

The top priority here must be to drive interoperability, and we don't get that by promoting dozens of underlying application protocol choices.


note-qualifying-protocols-and-encodings.md
```diff
@@ -0,0 +1,47 @@
...
+progress in the last decade in developing a robust and broadly supported protocol 
+foundation, like HTTP 1.1 and HTTP/2 as well as WebSockets or events on the web, 
+or MQTT and AMQP for connection-oriented messaging and telemetry transfers.
+
+Some widely used protocols have become de-facto standards emerging out of strong
+ecosystems of top-level multi-company consortia projects, such as Apache Kafka, 
+and largely in parallel to the evolution of the aforementioned standards stacks.
+
+The CloudEvents effort shall not become a vehicle to even implicitly endorse 
+or promote project- or product-proprietary protocols, because that were
```
duglin: s/were/would be/

note-qualifying-protocols-and-encodings.md
```diff
@@ -0,0 +1,47 @@
...
+ecosystems of top-level multi-company consortia projects, such as Apache Kafka, 
+and largely in parallel to the evolution of the aforementioned standards stacks.
+
+The CloudEvents effort shall not become a vehicle to even implicitly endorse 
+or promote project- or product-proprietary protocols, because that were 
+counterproductive towards CloudEvents' original goals. 
+
+For a protocol or encoding to qualify for a core CloudEvents event format or 
+protocol binding, the protocol must either have formal status as a standard 
+with a widely-recognized multi-vendor protocol standardization body (W3C, IETF,
```
duglin: add "e.g." to the list so people know its not just those.

duglin:
Just a few tweaks, but overall LGTM.

Once/if we get the primer doc added I think this would be good to be included in there instead of as a stand-alone doc.

nerdyyatrice:
Clemens, I remember you mentioned something like has to be implemented by 3 separate parties in the call which I think makes the qualification more quantitative. Any reason it's not mentioned here?

note-qualifying-protocols-and-encodings.md
```diff
@@ -0,0 +1,47 @@
+# Qualifying Protocols and Encodings
+
+The explicit goal of the CloudEvents effort, as expressed in the specification, is 
+"describing event data in a common way" and "to define interoperability of event 
+systems that allow services to produce or consume events, where the producer and 
+consumer can be developed and deployed independently".
```
n3wscott: Technically you get this just using and agreed upon messaging queue. Cloud events is suppose to be for transportability for the workload so you can switch out the producer and not change your consumer code.

note-qualifying-protocols-and-encodings.md
```diff
@@ -0,0 +1,47 @@
...
+with a widely-recognized multi-vendor protocol standardization body (W3C, IETF, 
+OASIS, OMG), or in the case of ecosystem protocols, it must have "de-facto standard" 
+status for its ecosystem category, comparable to Apache Kafka.
+
+Aside from formal status, a key criterion for whether a protocol or encoding shall 
+qualify for a core CloudEvents event format or transport binding is whether the 
+working group agrees that the specification will be of sustained practical benefit 
+for any party that is unrelated to the product or project from which the protocol 
+or encoding emerged. A base requirement for this is that the protocol or encoding 
+is defined in a fashion that allows alternate implementations independent of the
```
nerdyyatrice: I am not sure why we stick with qualitative descriptions instead of a quantitative one.

duglin: Do you mean the number of people using it? If so then I think the defacto  standard sentence above covers it, no?

nerdyyatrice: Here is what I am thinking (basically removed the more descriptive paragraph to make it more clean cut)
-------------------------------------------------------------------------------------------------------------
For a protocol or encoding to qualify for a core CloudEvents event format or protocol binding, it must belong to either one of the following categories
-  The protocol has a formal status as a standard with a widely-recognized multi-vendor protocol standardization body (W3C, IETF, OASIS, OMG)
- The protocol has a "de-facto standard" status for its ecosystem category (a good example would be Apache Kafka in stream-processing category)

duglin: I think if you add `e.g.` before the "W3C" to make it clear that the list isn't exclusive then it looks ok to me.  

I assume this proposal will replace line 31 thru 35, right?

What do other people think of this alternative text?

ultrasaurus: consider a definition of "de-facto" standard as "has an open source implementation and is in use by services or products from independent vendors"

cathyhongzhang: There are other standards. Adding "e.g." is good. I think we need to define the de-facto standard more clearly.

ultrasaurus: we could consider linking to someone else's definition: http://www.linfo.org/de_facto_standard.html

nerdyyatrice: updated the edit

duglin:
@clemensv see if you're ok with @nerdyyatrice's edits

All - please look this over in prep for next week's call. I believe the group was ok with the general direction, there was just an desire twiddle the formatting of the "list of requirements", so it should be an easy one.

note-qualifying-protocols-and-encodings.md
duglin: @clemensv ^^

note-qualifying-protocols-and-encodings.md
duglin: @clemensv ^^

note-qualifying-protocols-and-encodings.md
derekcollison: I believe for the group to achieve its goals it may want to look at a single encoding and transport protocol for all producers. Say JSON and HTTP 1.1 / 2 (Although true push for eventing may provide some challenges). This way would provide a least common denominator and consistency across cloud providers and ecosystems. This is a MUST to honor the spec. 

For specialized consumers that prefer different encodings or transports, they could be referenced here based on status and use in the ecosystem. 

Any provider could produce events in different encodings and transports, but to be compliant they must produce via HTTP and JSON (Or whatever the chosen encoding and transport shall be).

duglin: @derekcollison could you open an issue for your idea? I'd like to track it separately from this PR because I think it applies regardless of whether or not we define a "bar" for new encodings/transports (this PR).  Do I have that right?

note-qualifying-protocols-and-encodings.md
clemensv: Incorporating @nerdyyatrice's comment as-is

duglin:
LGTM thanks @clemensv @nerdyyatrice

note-qualifying-protocols-and-encodings.md
derekcollison: Sure no problem.

derekcollison: Done #288

duglin: thanks!

duglin:
Approved on 8/16 call. Also approved to move it into the primer.

Issue #142, opened 2018-04-05
Compatibility guidelines for event type versions?


deissnerk:
There are some cases in which event type versions might play an important role. 
E.g.:

1. If a standardized event type, that was defined by another organization, is used, it is important for the producer to add the version, as changes in a new version of the event type are not in the producer's hand.

2. If a consumer needs some information that is only available in an event type starting with version x, it might make sense to subscribe to events of the type with version >= x.

Right now the event type version is optional and just a string. So version ranges are not possible. It is also not clear, if a version is a super set of an older one or completely incompatible. I think that adhering to a versioning spec like [Semantic Versioning]( https://semver.org/ ) might bring some advantages for interoperability use cases. 

I could come up with an according PR, but would like to get some opinions first. Should the event-type-version remain as it is? Should there be just some recommendations or even binding rules in the spec?

Assigned to: @deissnerk

deissnerk:
@duglin You can assign this to me.
@inlined This is the issue about event type versioning which I referred to when we were talking at KubeCon.

duglin:
I wonder if it wouldn't be easier to just merge EventType and EventTypeVersion into one string. Then let the author of that string decide how/if/when to version it. We could add text "recommending" that the string be versioned, but ultimately its up to the entity defining it.

As a reference point. during the demo work I asked @clemensv how MSFT's Event Grid had people specify that they wanted CloudEvents as the output format - and he said that they specified something along the lines of "cloudeventv01schema" as the output type. So, that pattern seems to make sense to me - and it makes our life easier.

duglin:
Related: https://github.com/cloudevents/spec/issues/33

inlined:
@deissnerk thanks for the bump!

@duglin Honestly I'm in favor of dropping on EventTypeVersion and just using the data schema. I think eventTypeVersion can lead to a lot of confusion.

Imagine Twitter makes their API eventful and then creates an API v2. When I subscribe for the `com.twitter.tweet.publish` event, I probably want all tweets regardless of which client wrote to that tweet. Instead, I'd want to subscribe to the com.twitter.tweet.publish event and dictate the my client knows how to handle the schema from v1 only.

duglin:
yep - between someone having the option of making the EventType string anything they want (w/ or w/o a version part) and the data schema and content-type it sure seems to me like they have enough ways to express what they want.

deissnerk:
@inlined 
[...]

PR #256, opened 2018-06-27
Removed eventTypeVersion

deissnerk:
This solves https://github.com/cloudevents/spec/issues/142
I added a sentence to the definition of schemaURL claiming that incompatible changes to the schema SHOULD result in a new schema version. Please let me know, if you are fine with this.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -312,15 +312,6 @@ both appear within the same JSON object.
 * Examples
    * com.github.pull.create
 
-### eventTypeVersion
-* Type: `String`
-* Description: The version of the `eventType`. This enables the interpretation
-  of `data` by eventual consumers, requires the consumer to be knowledgeable
-  about the producer.
-* Constraints:
-  * OPTIONAL
-  * If present, MUST be a non-empty string
-
 ### cloudEventsVersion
 * Type: `String`
 * Description: The version of the CloudEvents specification which the event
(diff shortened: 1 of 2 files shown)


spec.md
```diff
@@ -360,6 +351,7 @@ both appear within the same JSON object.
 ### schemaURL
 * Type: `URI`
 * Description: A link to the schema that the `data` attribute adheres to.
+Incompatible changes to the schema SHOULD result in a new schema version.
```
duglin: I wonder if we should say "URI" instead of "version" just to keep it more generic and not imply that we're mandating a version string be part of the URI, even though that would be a good idea.  What do people think? I'm ok either way- it just jumped out at me.

duglin:
@deissnerk can you include an update to the other docs that reference "eventTypeVersion" in this PR?

Otherwise, LGTM

duglin:
you may need to rebase since this include commits that are not yours

duglin:
hmmm you're up to 14 commits now :-)

deissnerk:
@duglin Down to two commits now. git push **-f** did the trick.

duglin:
@deissnerk perfect - thanks.
Still LGTM

duglin:
Approved on 7/5 call

PR #274, opened 2018-07-23
Add JSONSchema Validation

### LINKED ISSUE #272: Consider Providing JSON Schema Documents
http://json-schema.org/

dmlb2000:
to validate events in JSON.

Fix #272

### DIFF
--- a/jsonschema/README.md
+++ b/jsonschema/README.md
@@ -1,0 +1,27 @@
+# CloudEvents JSONSchema
+
+This directory contains [JSONSchema](http://json-schema.org) for the
+spec.
+
+## Main
+
+The core spec JSONSchema document is in [spec.json](spec.json) and
+contains the core definitions for the json schema.
+
+## Testing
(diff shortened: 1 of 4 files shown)


duglin:
Just from a structural perspective, would it be bad if we
- moved the jsonschema/spec.json to the top level and called it spec.json (or json.schema)
- removed the example and misssingdata json files, since they're really more for testing purposes and I'm not sure getting into that yet. Perhaps the SDK work will though.
- move the README text into the json-format.md file
- and then we could get rid of the extra dir

Just to keep it cleaner.  WDYT?

dmlb2000:
That should be fine, I'll see if I can edit this pull request to point to another branch with the requested changes instead of a new pull request... don't want to have github run out of numbers ;)

dmlb2000:
Huh, the edit button doesn't allow me to change the from branch... I've submitted a new pull request #278

PR #278, opened 2018-07-25
Add JSONSchema Validation

### LINKED ISSUE #272: Consider Providing JSON Schema Documents

dmlb2000:
to validate events in JSON.

Fix #272

### DIFF
--- a/spec.json
+++ b/spec.json
@@ -1,0 +1,74 @@
+{
+  "$ref": "#/definitions/event",
+  "definitions": {
+    "cloudEventsVersion": {
+      "type": "string"
+    },
+    "contentType": {
+      "type": "string"
+    },
+    "data": {
+      "type": [
+        "object",
+        "string"
+      ]
+    },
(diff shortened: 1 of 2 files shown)


spec.json
```diff
@@ -0,0 +1,74 @@
...
+          "$ref": "#/definitions/eventType"
+        },
+        "extensions": {
+          "$ref": "#/definitions/extensions"
+        },
+        "source": {
+          "$ref": "#/definitions/source"
+        }
+      },
+      "required": [
```
jonsharratt: Just thought I would feedback that we took the following JSON schema for our cloudevents API gateway endpoint and noticed these fields should be optional rather than required:

* eventTime
* contentType
* data
* extensions

duglin: Excellent - thanks for checking this for us!!

dmlb2000:
Okay resolved the optional bits should be better in #281

PR #279, opened 2018-07-25
Add non-goals for routing information.

inlined:
Extend the non-goals section to clarify what is simply out of scope and
what is in scope but the WG thinks is a bad idea. The latter category
deserves fuller explanations for newcomers.

### DIFF
--- a/primer.md
+++ b/primer.md
@@ -113,6 +113,11 @@ solely for the purpose of proper delivery, and processing, of the message. Data
 that is not intended for that purpose should instead be placed within the event
 (the `data` attribute) itself.
 
+Additionally, it is assumed that the metadata needed by the transport layer
+to deliver the message to the target system is handled entirely by the
+transport and therefore is not included within the CloudEvents attributes.
+See the [Non-Goals](#non-goals) section for more details.
+
 Along with the definition of these attributes, there will also be specifications
 of how to serialize the event in different formats (e.g. JSON) and transports
 (e.g. HTTP, AMQP, Kafka).
@@ -124,11 +129,34 @@ transport specification.
 
 ### Non-Goals
 
-The following will not be part of the specification:
+The following are considered beyond the scope of the specification:
 
 - Function build and invocation process
 - Language-specific runtime APIs
 - Selecting a single identity/access control system
+- Inclusion of transport-level routing information
+
+The CloudEvents spec will not include transport-level routing
+information (e.g. a destination URL to which the event is being sent).
+This is a common suggestion by those new to the concepts of CloudEvents.
+After much deliberation, the working group has come to the conclusion that
+routing is unnecessary in the spec: any transport protocol (e.g. HTTPS,
+MQTT, XMPP, or a Pub/Sub bus) already defines semantics for routing.
+For example, the CloudEvents [HTTPS binding](http-transport-binding.md)
+dictates headers and request body contents. CloudEvents don't need to
+include a destination URL in the spec to be HTTP compatible; the HTTP spec
+already includes one in the
+[Request-Line](https://tools.ietf.org/html/rfc2616#section-5.1).
+
+Routing information is not just redundant, it detracts.
+CloudEvents should increase interoperability and decouple the producer and
+consumer of events. Prohibiting routing information from the events format
+allows CloudEvents to be redelivered to new actions, or to be delivered
+over complex relays that include multiple channels. For example, a
+CloudEvent that was intended for a webhook should be deliverable to a
+dead-letter queue if the webhook address is unavailable. That dead-letter
+queue should be able to feed events to new actions that the original event
+emitter never imagined.
 
 
 ## Architecture


primer.md
```diff
@@ -112,11 +112,36 @@ specifications of how to serialize the event in different formats and
...
 
 ### Non-Goals
-The following will not be part of the specification:
+The following are considered beyond the scope of the specification:
 * Function build and invocation process
 * Language-specific runtime APIs
 * Selecting a single identity/access control system
 
+The following are features that the working group has explicitly decided
+should nto be part of the specification:
```
duglin: s/nto/not/

primer.md
```diff
@@ -112,11 +112,36 @@ specifications of how to serialize the event in different formats and
...
+   For example, the Cloud Events [HTTPS binding](http-transport-binding.md)
+   dictates headers and request body contents. CloudEvents don't need to
+   include a destination URL in the spec to be HTTP compatible; the HTTP spec
+   already includes one in the
+   [Request-Line](https://tools.ietf.org/html/rfc2616#section-5.1).
+   
+   Routing information is not just redundant, it detracts.
+   CloudEvents should increase interoperability and decouple the producer and
+   consumer of events. Prohibiting routing information from the events format
+   allows CloudEvents to be redelivered to new actions to be delivered
```
duglin: I think you're missing something in this sentence. Perhaps ", or " after "actions" perhaps?

duglin:
overall LGTM

primer.md
```diff
@@ -112,11 +112,36 @@ specifications of how to serialize the event in different formats and
...
 ### Non-Goals
-The following will not be part of the specification:
+The following are considered beyond the scope of the specification:
 * Function build and invocation process
 * Language-specific runtime APIs
 * Selecting a single identity/access control system
 
+The following are features that the working group has explicitly decided
+should nto be part of the specification:
+* **routing information** The CloudEvents spec will never include routing
```
Vlaaaaaaad: Since this covers routing, should it be added that data relevant to routing can( should?) be put in `extensions`( or whatever it ends up being named)? Things like partition key or @cathyhongzhang IoT usecase are routing-relevant and the current phrasing seems to completely deny the `extensions` option.

duglin: That's good point. Just because this WG decided that we didn't want the core set of attributes to include routing doesn't mean that people couldn't add them in via extensions. We may, or may not, think its a good idea but people are free to do just about anything they want with extensions.  It is fair for us to provide guidance on why such things might be bad though - which I think he already does.

duglin: That almost begs the question... does it belong in `data` then?

I'm actually not suggesting that the spec should say where it should go either way because I think people should be able to add any extensions they want to our spec, but for use cases where they want to do processing out-of-bounds of the transferring of messages between two components, it feels less like a CloudEvent problem to me. But this goes back to what kind of guidance we provide for people on when to use CE extensions vs use `data`.

cathyhongzhang: You are confusing me:-)  CloudEvents are not just dealing with transporting/transferring of messages between two components. I think one of its goals is to facilitate interpretation and  processing of the event on the event consuming side. Adding this identificationLabels attribute to the main spec.md will help the event consuming side to handle a large number of serverless use cases/applications that involve multiple events, such as those IoT use cases, otherwise the event consumer might need to decrypt and decode the needed info from a big payload (I remember someone raised such a concern before).  Also the needed info might not be in the payload.

BTW, I would suggest that we add a discussion topic in next meeting on whether we allow different classifications/categorizations/bags of event information.

inlined: > Since this covers routing, should it be added that data relevant to routing can( should?) be put in extensions( or whatever it ends up being named)? Things like partition key

I maintain that routing information is _never_ a good idea, even in extensions. This was actually one of the reasons why I shot down "partitionKey" or "correlationId". The event should have enough information that the _application_ can choose what to partition or correlate by.

Now that I can be more concrete, if one looks at KNative eventing, it's totally reasonable that a the Flow/Feed adds a feature for correlation or partitioning. Then, the actual _Flow_ could set a correlation strategy of "event.labels.house" for one application and "event.labels.complex" for another. If the event itself is making the decisions for how it should be routed, these use cases are broken.

cathyhongzhang: Just want to add on Thomas's point. For the correlation usage, the event itself does not know which identification label carried in the event metadata will be used to correlate it with other events. It is up to another entity (could be the use case application developer) to set the correlation strategy as, e.g., correlationID=event.labels.house or correlationID=event.identificationLabels.house assuming we name it "identifcationLabels" instead of "Labels".

cathyhongzhang: Doug: The event producer does not know how the event consumer will use the identification label or for what usage, just like it does not know how the event consumer will use the eventType.  But the event producer knows what Labels can be used to identify this event.

duglin: "can" is the key word. It also "can" be used for correlation, logging, debugging, etc... just about anything.  The producer has no idea how any of these will be used. So trying to put them into classifications doesn't seem very useful since it would just be a guess anyway. As you implied... eventType is in the same position - so why don't we put it into their too?  

Actually, I'm suggesting we do ;-)   Just the other way around.  Put these things into the same "can be used for anything" category as all of the spec-defined CloudEvent properties.

The meaning of the property is defined by its spec - that's all the consumer will care about, not whether we wrap it in some bag, along with other properties they may or may not care about, or that may or may not be used for the same purpose.

fsunaval: @duglin 
>It also "can" be used for correlation, logging, debugging, etc... just about anything.

Here's my sargument for having a bucket/bag/container for additional identification.
Consider the three cases below.  Case1 does not have an identificationLabel while Case2 and Case3 do have identificationLabel.

Functionality wise all three cases are the same for the developer writing the app.  The correlation 
for Case1 is event.taste whilst for Case2/Case3 the correlation is for event.identificationLabel.taste

Now say you are the customer using the app and you can see the events but don't know anythng beyond that.  If you do not know what the key event identifier is where do you start?  Do you classify based on 'color' or 'house'.  But 'house' is not even present in Case3.   
On the other hand, Case2 and Case3 clearly illustrate that the key event identifier is 'taste'. You then classify the events based on 'taste'.  
taste: bitter, total events 131450
taste: sour,  total events 451345
taste: sweet, total events 465000

Case1:
eventID: 0x7145
color: red
house: small
taste:  bitter
boat: big
car: used

Case2:
eventID: 0x7145
animal: horse
house: small
identificationLabel: 
  taste: sour
boat: small
car: new

Case3:
eventID: 0x7145
color: red
animal: dog
owner: yes
record: 1234
identificationLabel: 
  taste: sweet

duglin: @fsunaval Any place people can add "stuff", whether its at the top level or in a bag, means you need to be prepared for "stuff" you don't care about. This means you're assumption that someone will do some kind of semantic processing (e.g. classifying) based on a bag of "random" stuff seems pretty unlikely to me. There could be 100's of properties in that bag that we don't care about, we just don't know, its open ended.  Even @cathyhongzhang agreed that the producer has no idea which properties the consumers will care about, this means they'll be putting all sorts of stuff in there because different consumers might use different properties and they will probably need to include everything that they have just in case someone needs it. 

In order to do something meaningful with the data the consumer must know what to look for - to find the gems in that random bag of stuff. Once that assumption is agreed to, where those properties are isn't really relevant. They can just as easily find them at the top-level as they can in a bag.

Now, let's say someone does really want to have a consumer be unaware of things - like you're suggesting. That's fine. But in that case I would claim that the producer and consumer would need to be in-sync in some way and have agreed on what "bag" to look for and that the only things in that bag will be exactly what the consumer needs and not some random other bits.  At that point I would not call that a "bag" as much as I would call it an extension property whose value is a structure with, most likely, a set of well-defined properties. Or at least some kind of constraints on it to make sure the only things in there are the things the "blind" consumer will need in its processing and not random "stuff" - e.g. a structure w/o an extensibility point. This is similar to what @inlined was asking about on today's call.

But the point is that we can not define this kind of structure in our spec today w/o being so focused on one particular usecase that it would either make this bag unusable for other cases, or lead us down the path of defining a lot of them based on every variant of usecase out there.  I would much prefer to let people define these specialized properties/bags as extensions based on their needs and **if** one day we realized that one of them became so popular that most people seem to be using it then we can promote it as a core spec property.  And because its already at the top-level no code would need to change - as a bonus  :-)

cathyhongzhang: Doug: I do not think that an event producer will put all sorts of stuff in the idenficationLabel bag. It is not a random bag of stuff. Of source we can not prevent anyone to do bad practice and bad practice can be done to any metadata field. When it comes down to each individual event producer which fills in the event metadata specific to that event type, the number of identification labels for that event will not be large. I don't feel your assumption is correct.

duglin: All they have to do is put one item in the bag that some consumer doesn't care about and it makes my point.

primer.md
```diff
@@ -112,11 +112,36 @@ specifications of how to serialize the event in different formats and
...
+   information (e.g. a Pub/Sub topic over which the message is delivered).
+   This is a common suggestion by those new to the concepts of CloudEvents.
+   After much deliberation, the working group has come to the conclusion that
+   routing is unnecessary in the spec: any transport protocol (e.g. HTTPS,
+   MQTT, XMPP, or a Pub/Sub bus) already defines semantics for routing.
+   For example, the Cloud Events [HTTPS binding](http-transport-binding.md)
```
cneijenhuis: s/Cloud Events/CloudEvents/

duglin:
@inlined there are some editorial comments

duglin:
@inlined some trivial syntax fixes are needed

duglin:
@inlined will you have time to make the suggested edits on this one before next week's call? If not I can try to do them for you - they seem minor.

duglin:
thanks @inlined 
LGTM

primer.md
```diff
@@ -112,11 +112,36 @@ specifications of how to serialize the event in different formats and
...
+   [Request-Line](https://tools.ietf.org/html/rfc2616#section-5.1).
+   
+   Routing information is not just redundant, it detracts.
+   CloudEvents should increase interoperability and decouple the producer and
+   consumer of events. Prohibiting routing information from the events format
+   allows CloudEvents to be redelivered to new actions, or to be delivered
+   over complex relays that include multiple channels. For example, a
+   CloudEvent that was intended for a webhook should be deliverable to a
+   dead-letter queue if the webhook address is unavailable. That dead-letter
+   queue should be able to feed events to new Actions that the original event
```
sslavic: Maybe s/Actions/_actions_/ to be consistent with rest of the document.

primer.md
```diff
@@ -112,11 +112,36 @@ specifications of how to serialize the event in different formats and
...
-The following will not be part of the specification:
+The following are considered beyond the scope of the specification:
 * Function build and invocation process
 * Language-specific runtime APIs
 * Selecting a single identity/access control system
 
+The following are features that the working group has explicitly decided
+should not be part of the specification:
+* **routing information** The CloudEvents spec will never include routing
+   information (e.g. a Pub/Sub topic over which the message is delivered).
```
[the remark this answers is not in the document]
deissnerk: I think the idea is to avoid attributes that are specific to transports and I support that intention. Nevertheless routing happens on many layers.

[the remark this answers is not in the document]
duglin: I'm leaning towards resolving this before we merge, but @inlined what do you think?

cneijenhuis:
My impression from today's call was that the term "routing information" may be too broad, or a few (me included) interpret the term more broadly than Thomas does.

E.g. `eventType` or `source` can be used to route events, so one could refer to them as "routing information" as well.

Clemens used the term "address information", which has probably less room for interpretation.

primer.md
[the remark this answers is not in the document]
deissnerk: My quote from above was from the design goals section of the primer, which is just above the non-goals section. So in the goals we state that the event context contains information that is needed to route the event, and in the non-goals we say that the CloudEvents spec will never include routing information. This might confuse someone reading through this for the first time.

deissnerk:
Perhaps we could rephrase to "The CloudEvents spec will never include routing information that is specific to an underlying transport binding".
The CloudEvents attributes are used to route the **event** to consumers. They are **not** used to route a message of some underlying transport, although some middleware may derive that information from them.

duglin:
Are we thinking too much about this? What if we reduced it down to something like:
- - -
The specification authors purposely decided to not include routing information as part of the formal set of attributes defined because this information will typically be included as part of the underlying transport mechanism and thus not required to be repeated within the event itself. 
- - -
As someone mentioned, if a producer wants to include it in the event (as an extension) they're free to do so, but we don't encourage or discourage it - we just talk about what *our* decision is, and why.

deissnerk:
@duglin I agree that the primer is not part of the spec. So it doesn't need to be too formal. Nevertheless, the idea of a primer is to give an introduction to someone looking at CloudEvents for the first time. I could imagine, that my take-away of the design goals section would be that CE attributes are used for event routing. Then, in the next section, I would read that routing information in CE attributes is not considered a good idea.

duglin:
@deissnerk that comment is more about us being inconsistent, and I totally agree with you on that - we should fix as part of this PR too.

duglin:
@inlined will you have a chance to update this one in time for next week's call?

duglin:
@inlined any update on this one?

duglin:
@inlined @deissnerk @cneijenhuis @sslavic  all, I tried to address the comments made on this PR. See what you think.

cneijenhuis:
LGTM

duglin:
Rebased and did some minor wordsmithing - but I think this one is ready for review (again)

duglin:
Approved on the 6/20 call

PR #285, opened 2018-08-03
Add JSONSchema Validation

### LINKED ISSUE #272: Consider Providing JSON Schema Documents

dmlb2000:
to validate events in JSON.

Fix #272

### DIFF
--- a/json-format.md
+++ b/json-format.md
@@ -91,6 +91,12 @@ The following table shows exemplary mappings:
 | data               | Binary   | "Q2xvdWRFdmVudHM="
 | data               | Map      | { "objA" : "vA", "objB", "vB" }
 
+## 2.5. JSONSchema Validation
+
+The CloudEvents [JSONSchema](http://json-schema.org) for the spec is located
+[here](spec.json) and contains the definitions for validating events in JSON.
+
+
 ## 3. Envelope
 
 Each CloudEvents event can be wholly represented as a JSON object.
--- a/spec.json
+++ b/spec.json
@@ -1,0 +1,70 @@
+{
+  "$ref": "#/definitions/event",
+  "definitions": {
+    "cloudEventsVersion": {
+      "type": "string"
+    },
+    "contentType": {
+      "type": "string"
+    },
+    "data": {
+      "type": [
+        "object",
+        "string"
+      ]
+    },
+    "event": {
+      "properties": {
+        "cloudEventsVersion": {
+          "$ref": "#/definitions/cloudEventsVersion"
+        },
+        "contentType": {
+          "$ref": "#/definitions/contentType"
+        },
+        "data": {
+          "$ref": "#/definitions/data"
+        },
+        "eventID": {
+          "$ref": "#/definitions/eventID"
+        },
+        "eventTime": {
+          "$ref": "#/definitions/eventTime"
+        },
+        "eventType": {
+          "$ref": "#/definitions/eventType"
+        },
+        "extensions": {
+          "$ref": "#/definitions/extensions"
+        },
+        "source": {
+          "$ref": "#/definitions/source"
+        }
+      },
+      "required": [
+        "cloudEventsVersion",
+        "eventID",
+        "eventType",
+        "source"
+      ],
+      "type": "object"
+    },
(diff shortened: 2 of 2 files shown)


duglin:
Any reason you don't just amend the previous commits, in the old PRs, instead of creating entirely new PRs?  ;-)

dmlb2000:
Git refuses to push the amended commit to the upstream branch.

duglin:
hmm, did you add "-f" - I usually have to force it:
```
git add ...
git commit -s --amend --no-edit
git push origin xxx -f
```

dmlb2000:
https://stackoverflow.com/questions/253055/how-do-i-push-amended-commit-to-the-remote-git-repository

Now granted this is older but, "Linus told me not to." was good enough for me ;)

duglin:
Well, I know some folks don't like amend, but you still could push additional commits instead of doing a new PR each time.

dmlb2000:
Well, I'm fine with amend. I just couldn't repoint the pull request to another branch name in my fork, which is a GitHub behavior. It could have been solved either way.

duglin:
LGTM

duglin:
All, please review - this should be an easy one for us to approve quickly on the next call

clemensv:
How does the schema look when there's no extensions bag but we allow arbitrary extra content as we're currently discussing, i.e. an open schema?

dmlb2000:
https://tools.ietf.org/html/draft-handrews-json-schema-01#section-8.2.4 (especially check out the examples before sections 8.3)

I think the idea is that this json schema document would be used as the root official minimum definition for describing a CloudEvent. If other json schema documents would be needed to describe other types of events (like those with extensions) the definitions could be referenced from this document to create a new definition document.

The hope is that this json schema document along with additional documents could be hosted at some permanent URL so others could extend the base implementation.

Just a thought. Unless I'm not getting what you mean by "open schema?"

erikerikson:
As mentioned on the call `additionalProperties` controls treatment of undeclared attributes observed in the described object.

> "By default any additional properties are allowed."
from: https://json-schema.org/understanding-json-schema/reference/object.html

dmlb2000:
I'm not on any calls, so if you'd like me to fix things a certain way please be explicit.

Would you like me to add `"additionalProperties": false` to the event definition or not?

duglin:
@dmlb2000 nope you're good - we approved the PR in the call and I'll be merging soon....

duglin:
@erikerikson thanks for finding that info for us!!!

erikerikson:
Apologies for the lack of context @dlmb2000 - a question was raised: what would happen during validation of a compliant object (against this schema) that also contained arbitrary extension attributes as they have been recently discussed?  The group didn't want to put itself in to contradiction with its current trajectory.

dmlb2000:
Just my 2 cents on adding that option...

As your spec changes over time additional attributes might be added in the future to express new features. Would a v1.0 json schema document also be able to validate a v1.1 event with additional attributes as defined by the v1.1 spec?

The question is how do you want the spec to evolve over time, if you say additional attributes aren't allowed then you are stuck in a major version bump of the spec for potentially minor changes.

duglin:
IMO a v1.0 schema needs to be able to validate a v1.1 event because anything new in v1.1 will appear as an optional extension and be ignore-able by a v1.0 receiver, if they wish.  The idea of not allowing a minor version bump to add new optional properties doesn't seem like a good choice to me  - or to turn it around, making the addition of a new optional property be a major version change seems extreme.

duglin:
Approved on 8/16 call

sslavic:
Btw, I just checked json example against schema, used https://www.jsonschemavalidator.net/ - validation fails at source field, since it’s not proper uri.

dmlb2000:
@sslavic my thought is to put in another issue stating the example events in the spec are not valid.

There's two possible issues:

 1. The examples are correct and the code jsonschemavalidator.net is using is being too strict on what a valid URI is, they aren't following the [spec](https://tools.ietf.org/html/rfc3986#section-4.1).
 1. The examples aren't correct and need to be made valid.

Issue #286, opened 2018-08-06
HTTP Header limitations and implications for extensions


cneijenhuis:
On last weeks call, there was a question in the chat if CloudEvents would expect to support thousands of extension values. I replied that I don't think it would be possible to send such an event with the [HTTP Binary Content Mode](https://github.com/cloudevents/spec/blob/master/http-transport-binding.md#31-binary-content-mode), because HTTP middleware limits the headers it accepts.

I decided to follow up with some research. Most servers simply limit the total size of all HTTP headers combined. A few also limit the count of headers, or the size of a single header. Here are a few values I researched, either through documentation (linked) or through experimentation.

- AWS API Gateway: 10kb [docs](https://docs.aws.amazon.com/apigateway/latest/developerguide/limits.html)
- AWS ELB: 64kb total, 8kb per header [docs](https://docs.aws.amazon.com/elasticloadbalancing/latest/userguide/how-elastic-load-balancing-works.html)
- Azure Functions: 64kb
- IIS: 64kb
- Google Cloud Functions: 16kb
- Google Cloud Load Balancer: 16kb
- nginx: 4kb or 8kb [docs](http://nginx.org/en/docs/http/ngx_http_fastcgi_module.html#fastcgi_buffer_size)
- Apache HTTP Server: 8kb, 100 headers [docs](https://httpd.apache.org/docs/2.4/mod/core.html#limitrequestfields)
- akka HTTP: 8kb, 64 headers [docs](https://doc.akka.io/docs/akka-http/current/configuration.html)

I'm a bit surprised that only a few HTTP servers limit the count of headers - but I guess this is good news. If we had to fit within e.g. 64 headers, and reserve space for non-CloudEvent headers as well as non-Extension headers, we'd look at 20 to 30 extensions.

However, size limitations are enforced everywhere as a DOS-protection.

We to reserve some space for normal HTTP headers (i'd guess at least 2kb, especially for Authorization headers) and regular CloudEvent properties (guessing 2kb should be fine).
8kb is widely supported, so that leaves 4kb for Extensions.

4kb is obviously just a ballpark figure, but it sounds fine to me: It is plenty of space to put data that can be used for routing, but it's - in most cases - unlikely that all of the `data` can be put into the extensions.

@duglin If you think this is worth presenting on the call, I can prepare a slide or two. Maybe it can also spark a discussion on similar limits in other (binary) protocols.

duglin:
[...]

Issue #265, opened 2018-07-06
Issue with "-" in extension names


duglin:
See: 
- https://github.com/cloudevents/spec/pull/255#issuecomment-401430602
- https://github.com/cloudevents/spec/pull/255#issuecomment-402989772

In https://github.com/cloudevents/spec/pull/255 we introduced the notion of using "-" to denote stepping into a nested object but we do not ban the use of "-" in a property name. This makes it ambiguous as to whether "-" is part of a name or not.  We should either use some other mechanism or ban the use of "-" in names.

ping @clemensv

cneijenhuis:
I think the issue is a bit broader actually. We have the [attribute naming conventions](https://github.com/cloudevents/spec/blob/master/spec.md#attribute-naming-convention), but they're neither normative, nor do they limit the encoding in any way. E.g. in JSON, UTF-8 fieldnames are allowed, but wouldn't work with HTTP headers.

> ban the use of "-" in names

A simple way to prevent a lot of (similar) issues is to limit attribute names to alphanumeric characters.

We could allow `%` to be used for encoding non-alphanumeric characters, similar to url-encoding. But I'm not sure if that isn't overkill already.

@duglin If it's ok, I'd like to take over and create a PR that:
* allows alphanumeric characters only in [attribute naming conventions](https://github.com/cloudevents/spec/blob/master/spec.md#attribute-naming-convention)
* states that extensions names MUST follow the [attribute naming conventions](https://github.com/cloudevents/spec/blob/master/spec.md#attribute-naming-convention)

deissnerk:
@cneijenhuis What about case-sensitivity? E.g. an event router that receives an event in HTTP binary content mode, is not able to convert an extension attribute back to camel casing.

cneijenhuis:
@deissnerk There is another issue for that: https://github.com/cloudevents/spec/issues/177

I agree they are related. Once it is clearly defined which characters are supported in attribute names, it becomes easier to fix the issue in HTTP headers.

E.g. if the outcome is that only allow alphanumeric characters are allowed, we have many more ASCII characters that can be used (e.g. convert camelCase to snake_case).
However, if the outcome is that all UTF-8 characters are allowed (as in JSON), then base64 encoding (see https://github.com/cloudevents/spec/issues/177#issuecomment-391897958) seems like a good idea.

duglin:
I believe that we can close this one now that we've agreed to adopt #321. Please speak up if you think otherwise.

[...]

PR #287, opened 2018-08-09
Technical motivation for keeping extension attributes minimal

cneijenhuis:
This paragraph should provide technical motivation for keeping the usage of extension attributes in check. For background information, see https://github.com/cloudevents/spec/issues/286

I expect that the following paragraph from https://github.com/cloudevents/spec/pull/277 will precede the paragraph I'm proposing, which provides a more conceptual motivation for limiting the usage:

```
Extension attributes to the CloudEvent specification are meant to
be additional metadata that needs to be included to help ensure proper
routing and processing of the CloudEvent. Additional metadata for other
purposes, that is related to the event itself and not needed in the
transportation of the CloudEvent, should instead be placed within the proper
extensibility points of the event itself.
```

I do think that at a later point in time we can still have a discussion on whether we want to set actual limits (e.g. a CloudEvent MUST NOT have more than #kb of metadata) or guarantees (e.g. a CloudEvent consumer MUST at least accept events with #kb of metadata) in `spec.md`.

However, I believe this change in the primer would not be affected much by that discussion.

### DIFF
--- a/primer.md
+++ b/primer.md
@@ -159,6 +159,12 @@ transportation or processing of the CloudEvent, should instead be placed
 within the proper extensibility points of the event (the `data` attribute)
 itself.
 
+Extension attributes should be kept minimal to ensure the CloudEvent can be
+properly serialized and transported. For example, the
+[HTTP Binary Mode](http-transport-binding.md#31-binary-content-mode) uses HTTP
+headers to transport metadata. Most HTTP servers will reject requests with
+excessive HTTP header data, with limits as low as 8kb.
+
 The specification places no restrictions on the type of the extension
 attributes. Meaning, they may be simple types (e.g. strings, integers),
 complex (e.g. structured) or undefined collection of attributes.
--- a/primer.md
+++ b/primer.md
@@ -161,9 +161,12 @@ itself.
 
 Extension attributes should be kept minimal to ensure the CloudEvent can be
 properly serialized and transported. For example, the
+Event producers should consider the technical limitations that might be
+encountered when adding extensions to a CloudEvent. For example, the
 [HTTP Binary Mode](http-transport-binding.md#31-binary-content-mode) uses HTTP
-headers to transport metadata. Most HTTP servers will reject requests with
-excessive HTTP header data, with limits as low as 8kb.
+headers to transport metadata; most HTTP servers will reject requests with
+excessive HTTP header data, with limits as low as 8kb. Therefore, the aggregate
+size and number of extension attributes should be kept minimal.
 
 The specification places no restrictions on the type of the extension
 attributes. Meaning, they may be simple types (e.g. strings, integers),


duglin:
Seems like reasonable guidance to me.
LGTM

primer.md
```diff
@@ -145,6 +145,12 @@ user-stories to explain the rationale and need for them. This supporting
 information will be added to the [Prior Art](#prior-art) section of this
 document.
 
+Extension attributes should be kept minimal to ensure the CloudEvent can be
```
ultrasaurus: Propose small clarifying change:

Extension attributes required by a single CloudEvent should be kept minimal

e.g. It's ok if the community as a whole as lots of extensions for different domains, but in practice few would be needed for an individual event than needs to be transported on the wire

duglin:
rebase needed

primer.md
duglin: What about:
Event producers should consider the technical limitations that might be encountered when adding extensions to a CloudEvent. For example, the [HTTP Binary Mode](http-transport-binding.md#31-binary-content-mode) uses HTTP headers to transport metadata; most HTTP servers will reject requests with excessive HTTP header data, with limits as low as 8kb. Therefore, the aggregate size and number of extension attributes should be kept minimal.

cneijenhuis:
I'm sorry that I couldn't attend yesterdays call. I watched the recording, at least.

@ultrasaurus You're right, the first sentence was obviously confusing! Does the version that @duglin proposed (which I've adopted, thanks!) sound good to you?


@inlined I believe it was you who said something along the lines of "This PR says: Extensions are bad"? I'd like to clarify that this isn't the intention of my PR. In fact, I believe extensions are good and I'd like that, eventually, CloudEvents middlewares can route based on (custom) extensions.

However, I have two concerns:
* Interoperability: If someone puts a lot of data into extensions, and tests it with e.g. JSON, everything seems fine. But if later someone else tries to transmit that CloudEvent with a different serialization format, it brakes.
* Middleware: Parsing a huge set of extension data will impact performance. But a middleware is trying to achieve high-throughput and a predictable performance. If putting too much data into custom extensions is the norm, then middlewares may decide to not offer features like routing based on them, but effectively treat them like `data`.

I agree that, similar to a leaflet of a drug, it may sound like "Extensions are bad" at first, but the intention is to make sure they are used without hindering interoperability.

duglin:
LGTM

cneijenhuis:
@duglin Yes, it is rebased now.

I'd like to make further contributions - I can get started on https://github.com/cloudevents/spec/issues/265, but I am also happy to pick another issue.

duglin:
@cneijenhuis thanks for the rebase.  Merging....

re: #265 sounds good to me. Please ping @clemensv as he originally wrote the PR that introduced the "-" usage and might have been thinking about possible solutions already.

PR #289, opened 2018-08-16
Eliminate 'extensions' bucket and clarifying extensibility

clemensv:
Dropping the extensions bucket and adding an explanation of the extensibility model.

Proof of concept code that shows an extensions bag isn't needed at the level of this specification is not needed and is rather better accommodated only in the wire-format mappings that have no support for dynamic/open schemas (ex: Thrift and Protobuf) can be found here https://github.com/clemensv/cloudeventsformatdemo

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -124,6 +124,28 @@ The `Object` type is a variant type that can take the shape of either a
 abstract, and therefore it is left to implementations how to represent the
 variant type.
 
+## Extensibility
+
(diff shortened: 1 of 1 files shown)


duglin:
@clemensv what is the relationship between this PR and https://github.com/cloudevents/spec/pull/277  ? This seems like a subset. Are you suggesting this replace that one?

clemensv:
Oops. No. Sorry. I miss you already having that. I will close this one.

PR #291, opened 2018-08-16
Add 'sequence' attribute

duglin:
Fixes: #191

Actually, it picks up the work from #191 since that one appears to be stale.

I didn't address any of the comments in #191 because I wasn't sure how
the WG in general felt about them. So please speak up in this PR if you'd like
to see a change.

I'm not advocating for, or against, this proposal - I'm just moving it to a possible extension attribute per our F2F meeting so the WG can decide if we want to include it.


BenBeattieHood:
Many thanks for picking this up @duglin - I've been away for a bit and unable to respond. I agree that this is better suited as an extension than a core feature if IoC equipment is unable to provide state for incrementing an ordinal field. I do believe this extension is key to designing for events, as otherwise you get forward-coupling between services; but am happy to let adoption work this out.

extensions/ordinal.md
```diff
@@ -0,0 +1,15 @@
+# Ordinal
```
ultrasaurus: Suggest name: Sequence

"ordinal" implies a number; however, this is a string "expressing the relative order of the event" -- this solves a broader range of use cases than just a number, and is definitely a good approach.  Propose we adjust the naming to match the definition.

duglin:
Do we need to define the values of this sting or is it expected that all consumers will know what values each producer will use?  As currently written, w/o some out-of-band communication with the producer, I wouldn't be able to use this extension. I'd like to add some text to the PR telling people about this if that's the case - but I need some guidance on what people's expectations are on this point.

BenBeattieHood:
@Tapppi I'm ok with repetition, so long as the values are contiguous and therefore assertable.
@duglin Yes, if we were constrained to base-10 integers, it would be more obvious; but if the producer used a wider base (eg base 128) this would need to be communicated in order for the consumer to predict the next value.

BenBeattieHood:
Fwiw, I still feel it's worth discussing that this sequence number, when combined with the `source` value, makes the `eventID` unnecessary. eg. a source+sequence would give `IoT/LightSensor5/Event4` would be unique.

duglin:
I switched this to use "sequence" and added some text about how to know what the possible values are. What do people think?

duglin:
In your example, wouldn't the consumer also stop due to it waiting for 'b' to arrive - which it never will according to your example?

Added: RECOMMENDED as contiguous

duglin:
@cneijenhuis I like the idea of two fields (sequence and sequenceType). We can then allow for some well-defined sequenceType values and allow people to define new ones - so its an extensibility point. What do other people think?

BenBeattieHood:
I agree with including the `sequenceType` (in addition to `sequence`) - good suggestion.

(@duglin re 'b': sorry about my example (appears to have disappeared with the updated commit) - you're right I should've used tokens more obviously contiguous - I'm grateful you picked up on the main point I was trying to say.)

duglin:
@cneijenhuis @BenBeattieHood do one of you guys want to suggest some text for what these two attributes would look like?

cneijenhuis:
Here is my first try. Sorry, not wrapped at 80 characters yet 😉 

---

# Sequence

This extension defines an attribute that can be included within a CloudEvent to represent the position of the event relative to the other events from the same event source.

The exact value and meaning of this attribute is defined by the sequence type. 
If the sequence type is missing, or not defined in this specification, event consumers will need to
have some out-of-band communication with the event producer to understand how to interpret the value of the attribute.

## Attributes
### sequence
* Type: `String`
* Description: Value expressing the relative order of the event. This enables
  interpretation of data supercedence.
* Constraints
  * REQUIRED
  * MUST be a non-empty lexicographically-orderable string
  * RECOMMENDED as contiguous

### sequenceType
* Type: `String`
* Description: Defines the semantics of the `sequence` attribute. The value `Integer` is defined in this specification. Other values are allowed.
* Constraints
  * OPTIONAL

#### Integer
If the `sequenceType` is set to `"Integer"`, the `sequence` attribute has the following semantics:
* The values are string-encoded signed 32-bit Integers.
* The first CloudEvent in an `Integer` sequence has the value `1`, the second `2` etc.
* The sequence is contiguous. In other words, if the event consumer receives two CloudEvents with sequence values of `1` and `3`, another CloudEvent with a sequence value of `2` exists, but has not been received yet.
* The sequence wraps around from `2,147,483,647` (2^31 - 1) to `−2,147,483,648` (-2^31).

duglin:
@cneijenhuis thanks!  I just added your suggested text. All - please see what you think

duglin:
oh darn, I just realized that we don't say what the default sequenceType value is if its not present. I'm assuming Integer. Will add that

cneijenhuis:
@duglin I left it out intentionally - if it is missing, the consumer should treat it as an unknown value.

See https://github.com/cloudevents/spec/pull/291/files#diff-7fb2227cb146a7ff3164326d6b46f5a1R10

But I'm also totally fine with making `Integer` the default, and forcing producers to declare some other value. We should change the upper paragraph then, though.

duglin:
@cneijenhuis oh sorry, I missed that line. I reverted my change. While I think defaulting to Integer might be nice, just to provide some base level of interop, as long as we say what's expected when its missing (even if its "use out of band communication") that addresses my biggest concern. I'll leave it to you experts to decide if "out of band" is ok or if we should default to Integer.

extensions/sequence.md
```diff
@@ -0,0 +1,51 @@
...
+## Attributes
+
+### sequence
+* Type: `String`
+* Description: Value expressing the relative order of the event. This enables
+  interpretation of data supercedence.
+* Constraints
+  * REQUIRED
+  * MUST be a non-empty lexicographically-orderable string
+  * RECOMMENDED as contiguous
```
erikerikson: Possibly:
`MUST be monotonically increasing`

erikerikson: at least monotonic

extensions/sequence.md
```diff
@@ -0,0 +1,51 @@
+# Sequence
+
+This extension defines two attributes that can be included within a CloudEvent
+to describe the position of the event relative to the other events from the
```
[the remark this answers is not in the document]
duglin: done

extensions/sequence.md
```diff
@@ -0,0 +1,51 @@
...
+If the `sequenceType` is missing, or not defined in this specification,
+event consumers will need to have some out-of-band communication with the
+event producer to understand how to interpret the value of the attribute.
+
+## Attributes
+
+### sequence
+* Type: `String`
+* Description: Value expressing the relative order of the event. This enables
+  interpretation of data supercedence.
```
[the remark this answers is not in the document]
duglin: I wonder if this would be better done by checking the EventID instead?  If not, have any suggested text?
Do we need to worry about a single event being part of more than one sequence?

extensions/sequence.md
```diff
@@ -0,0 +1,51 @@
...
+* Constraints
+  * REQUIRED
```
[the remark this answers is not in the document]
duglin: This may be something we need to add more text on... but the point here is that while the extension itself is optional, IF it is being used then this property is REQUIRED. See the documented-extension.md doc, where it says "Support for any extension is OPTIONAL. When an extension definition uses RFC 2199 keywords (e.g. MUST, SHOULD, MAY), this usage only applies to events that use the extension."

extensions/sequence.md
duglin: added "monotonically". Wasn't sure about the "increasing" part since that might imply a numeric based sequence and I'm assuming there might be some that don't fit that model. Is that a correct assumption?

extensions/sequence.md
```diff
@@ -0,0 +1,51 @@
...
+  * MUST be a non-empty lexicographically-orderable string
+  * RECOMMENDED as contiguous
+
+### sequenceType
+* Type: `String`
+* Description: Defines the semantics of the sequence attribute.
```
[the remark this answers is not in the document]
duglin: done

extensions/sequence.md
```diff
@@ -0,0 +1,51 @@
...
+
+This specification defines the following values for `sequenceType`.
+Additional values MAY be defined by other specifications.
+
+### Integer
+If the `sequenceType` is set to `Integer`, the `sequence` attribute has
+the following semantics:
+* The values of `sequence` are string-encoded signed 32-bit Integers
+* The first CloudEvent in an `Integer` sequence MUST have the value
+  `1`, the second `2`, etc.
```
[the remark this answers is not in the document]
duglin: I tried to tweak it, let me know if I went too far

extensions/sequence.md
```diff
@@ -20,11 +20,11 @@ event producer to understand how to interpret the value of the attribute.
 * Constraints
   * REQUIRED
   * MUST be a non-empty lexicographically-orderable string
-  * RECOMMENDED as contiguous
+  * RECOMMENDED as montomically contiguous
```
cneijenhuis: There is a typo, maybe `monotonically increasing and contiguous` reads better?

My math is quite rusty, but I'm not sure if a char-based sequence can legitimately be called `monotonically increasing` (e.g. `a, b, ...z, aa, ab, ...`). But, as it is only a recommendation, I think it's fine to get the point across.

duglin: done!

extensions/sequence.md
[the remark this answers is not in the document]
erikerikson: > I wonder if this would be better done by checking the EventID instead?

That's a great question.  As per [the eventID definition](https://github.com/cloudevents/spec/blob/master/spec.md#eventid) there is no monotinicity requirement.  Without a value constraint, one would have to keep track of every `eventID` that has been processed (or at the very least a significant scope of them) to ensure no duplicate processing as opposed to checking the relationship of the currently being processed `eventID` to the last successful checkpoint `eventID`.  I'm not sure that the assumptions I like to make would be appropriate to push onto all users fo the spec so I haven't advocated for that there but would have them adopted in my more-perfect world.

> Do we need to worry about a single event being part of more than one sequence?

I'm taking an event as the tangible message sent by a unique producer (as opposed to the occurrence in reality being read by some sensor).  As I've read this change, it seemed that there was a one-to-one mapping between unique producers of events with this attribute and the series of numbers representing the singular sequence that they produce.  That's a long way to say I don't expect a unique producer to produce multiple sequences OR for multiple unique producers to produce the "same" event (though distinct events could be causally interlinked and contain the exact same content [sans producer id] by the shared reality producing them - e.g. redundant emitters of metrics coming from a single sensor).  I could definitely be reading incorrectly.

extensions/sequence.md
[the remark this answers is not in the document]
erikerikson: Thank you for clarifying.  This makes perfect sense and clears my confusion.

extensions/sequence.md
erikerikson: Answering because you asked.  You've already added "increasing".

I feel a little uncertain and would defer to a mathematician but the `lexicographically orderable` requirement gives an absolute ordering across all unique values.  I would expect "increasing" to make sense in that context and mean "each subsequent value would be ordered/sorted after all preceding values".

For the record, my initial question would disallow the reasonable circular counter described in the `Integer` sequence type and therefore could not be a MUST.

extensions/sequence.md
```diff
@@ -0,0 +1,46 @@
...
+
+This specification defines the following values for `sequenceType`.
+Additional values MAY be defined by other specifications.
+
+### Integer
+If the `sequenceType` is set to `Integer`, the `sequence` attribute has
+the following semantics:
+* The values of `sequence` are string-encoded signed 32-bit Integers.
+* The sequence MUST be contiguous, monotonically increasing by `1` and
+  start with a value of `1`.
```
erikerikson: Sorry to nit.  It seems like it might read more naturally as
```
The sequence MUST start with a value of `1` and increase by `1` for each 
subsequent value (i.e. be contiguous and monotonically increasing).
```

duglin:
Based on the discussions in #308, can those who are interested in possibly using this extension please speak up via a comment in this PR?  @BenBeattieHood I assumed you would be - @Tapppi ? @cneijenhuis ?

cneijenhuis:
Yes, commercetools is strongly interested in this. We already provide an Integer-based sequence for our proprietary events (called [sequenceNumber](https://docs.commercetools.com/http-api-message-types.html#message). For CloudEvents, we'd like to cover this use-case in a non-proprietary way.

In fact, my hope is that an attribute for ordering events makes it into the main spec until 1.0. I think the  extension proposed in this PR is a good first step towards that.

duglin:
Is this one ready for review? @BenBeattieHood @cneijenhuis   you guys ok with the current state?

BenBeattieHood:
Yes, thanks :)

cneijenhuis:
Yes 👍 Thanks for pushing this PR forward!

duglin:
Approved on 9/27 call

Voters who voiced support (per our new "extension bar"): Christoph(commercetools), Jesse(Oracle)

Tapppi:
Thanks for pushing this forward, I'm a bit swamped with product launches at the moment and can barely make it to the calls :) Definitely worthy extension though.

PR #292, opened 2018-08-16
async voting process 

ultrasaurus:
this is based on what I've observed in TOC votes and seems to work well in practice (though I've only followed a few votes)

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -54,6 +54,11 @@ If a vote is taken during a WG meeting, the follow rules will be followed:
   [here](https://docs.google.com/spreadsheets/d/1bw5s9sC2ggYyAiGJHEk7xm-q2KG6jyrfBy69ifkdmt0/edit#gid=0).
   Members must acknowledge their presence verbally, meaning, adding yourself
   to the "Attendees" section of the Agenda document is not sufficient.
+* When a vote is called, votes will be collected over the following week: 
+  * Comment on the PR: "agree, yes, +1", "disagree, no, -1", or "abstain".
+  * Any person is encouraged to vote with a statement of support or dissent to
+    help undecided voters to reach a decision
+  * Voting tally will reflect the above qualification and recorded in PR
 
 ## Release Process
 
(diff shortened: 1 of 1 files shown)


GOVERNANCE.md
```diff
@@ -54,6 +54,11 @@ If a vote is taken during a WG meeting, the follow rules will be followed:
   [here](https://docs.google.com/spreadsheets/d/1bw5s9sC2ggYyAiGJHEk7xm-q2KG6jyrfBy69ifkdmt0/edit#gid=0).
   Members must acknowledge their presence verbally, meaning, adding yourself
   to the "Attendees" section of the Agenda document is not sufficient.
+* When a vote is called, votes will be collected over the following week:
```
duglin: I'd prefer if we didn't mandate a week since its possible all voting members are on the call when the vote is started.  So what about:
* When a vote is called if all voting members are present in the meeting, and the all agree to vote at that time, then votes will be tallied during that meeting.
* If not all voting members are present, or some wish more time, then those members will be allowed one week to vote asynchronously. In the case of a PR vote, they will post a comment to the PR with their vote. In other cases an email to the CloudEvents mailing list is expected.
* In the case of yes/no votes, each vote will be either "yes", "no" or "abstain". In the case of non-yes/no votes, the WG will agree on the valid responses prior to starting the vote.

duglin:
Can you sign your commit? The DCO check is not happy

ultrasaurus:
Clarification: wait +1 week is only if everyone hasn't voted

GOVERNANCE.md
```diff
@@ -54,6 +54,14 @@ If a vote is taken during a WG meeting, the follow rules will be followed:
   [here](https://docs.google.com/spreadsheets/d/1bw5s9sC2ggYyAiGJHEk7xm-q2KG6jyrfBy69ifkdmt0/edit#gid=0).
   Members must acknowledge their presence verbally, meaning, adding yourself
   to the "Attendees" section of the Agenda document is not sufficient.
+* When a vote is called, unless all viting members have voted, the vote will be
```
duglin: s/viting/voting/

GOVERNANCE.md
```diff
@@ -54,6 +54,14 @@ If a vote is taken during a WG meeting, the follow rules will be followed:
...
+  tallied no less than one week after calling the vote.
+* Voting process:
+  * Comment on the PR: "agree, yes, +1", "disagree, no, -1", or "abstain".
```
duglin: Let's make this a bit more concrete and less chance of someone using the word "agree" w/o realizing they're voting.
`Comment on the PR: LGTM, NOT LGTM or ABSTAIN`

duglin:
Approved during the 9/6 call

PR #295, opened 2018-08-22
Add protobuf format

zpencer:
This PR is meant for the spec as it is currently written, where the
extensions is an explicit bag. The focus of this PR is on the protobuf
format itself, not on the extensions proposal.

Signed-off-by: Spencer Fang <spencerfang@google.com>

### DIFF
--- a/protobuf-format.md
+++ b/protobuf-format.md
@@ -12,10 +12,10 @@ This document is a working draft.
 ## 1. Introduction
 
 This specification defines how the [Context
-Attributes](spec.md#context-attributes) Attributes defined in the
-CloudEvents specification MUST be encoded in the protocol buffer
-binary format. Transcoding to and from other formats (e.g. JSON) is
-out of the scope of this document.
+Attributes](spec.md#context-attributes) defined in the CloudEvents
+specification MUST be encoded in the protocol buffer binary
+format. Transcoding to and from other formats (e.g. JSON) is out of
+the scope of this document.
 
 Protocol Buffers are a language-neutral, platform-neutral extensible
 mechanism for serializing structured data. The [Google reference
@@ -23,7 +23,7 @@ implementation of Protocol
 Buffers](https://github.com/protocolbuffers/protobuf) includes support
 for an interface descriptor language (IDL), and this document makes
 use of language level 3 IDL from Protocol Buffers v3.5.0. CloudEvents
-systems using Protocol Buffers are not required to use the IDL or any
+systems using Protocol Buffers are not mandated to use the IDL or any
 particular implementation of Protocol Buffers as long as they produce
 messages which match the binary encoding defined by the IDL.
 
@@ -51,7 +51,8 @@ binary encoding.
 ### 2.1 Definition
 
 Users of Protocol Buffers MUST use a message whose binary encoding is
-identical to the one described by the CloudEventMap message:
+identical to the one described by the [CloudEventMap
+message](./cloudevent.proto):
 
 ```proto
 syntax = "proto3";
@@ -87,8 +88,15 @@ The CloudEvents type system MUST be mapped into the fields of
 | Integer      | int_value
 | Any          | Not applicable. Any is the enclosing CloudEventAny message itself
 
+## 2.2 Relation to HTTP
 
-## Examples
+The [HTTP transport binding](./http-transport-binding.md) defines a
+*structued* and *binary* mode.
+
+For the Protocol Buffer CloudEvent format expressed 
+
+
+## 3. Examples
 
 Below is an example of how to create a CloudEvent Protocol Buffer
 message using the Java Google Protocol Buffers library:
--- a/protobuf-format.md
+++ b/protobuf-format.md
@@ -14,7 +14,7 @@ This document is a working draft.
 CloudEvents is a standardized and transport-neutral definition of the
 structure and metadata description of events. This specification
 defines how the [Context
-Attributes](spec.md#context-attributes)defined in the CloudEvents in
+Attributes](spec.md#context-attributes) defined in the CloudEvents in
 the protocol buffers binary encoding MUST BE encoded. Transcoding to
 and from other formats (e.g. JSON) is out of the scope of this
 proposal.
--- a/protobuf-format.md
+++ b/protobuf-format.md
@@ -81,8 +81,8 @@ The CloudEvents type system MUST be mapped into the fields of
 |--------------|-------------------------------------------------------------
 | String       | string_value
 | Binary       | binary_value
-| URI          | string_value (string expression conforming to URI-reference as defined in RFC 3986 §4.1)
-| Timestamp    | string_value (string expression as defined in RFC 3339.)
+| URI          | string_value (string expression conforming to URI-reference as defined in [RFC 3986 §4.1](https://tools.ietf.org/html/rfc3986#section-4.1))
+| Timestamp    | string_value (string expression as defined in [RFC 3339](https://tools.ietf.org/html/rfc3339))
 | Map          | map_value
 | Integer      | int_value
 | Any          | Not applicable. Any is the enclosing CloudEventAny message itself
--- a/protobuf-format.md
+++ b/protobuf-format.md
@@ -21,12 +21,14 @@ proposal.
 
 Protocol buffers are a language-neutral, platform-neutral extensible
 mechanism for serializing structured data. The [Google reference
-implementation of protobuf](PROTO) includes support for an interface
-descriptor language (IDL), and this document makes use of language
-level 3 IDL from Protocol Buffers v3.5.0. CloudEvents systems using
-protocol buffers are not required to use the IDL or any particular
-implementation of Protocol buffers as long as they produce messages
-which match the binary encoding defined by the IDL.
+implementation of
+protobuf](https://github.com/protocolbuffers/protobuf) includes
+support for an interface descriptor language (IDL), and this document
+makes use of language level 3 IDL from Protocol Buffers
+v3.5.0. CloudEvents systems using protocol buffers are not required to
+use the IDL or any particular implementation of Protocol buffers as
+long as they produce messages which match the binary encoding defined
+by the IDL.
 
 
 ### 1.1. Conformance
--- a/protobuf-format.md
+++ b/protobuf-format.md
@@ -1,9 +1,9 @@
-# Protobuf Event Format for CloudEvents  - Version 0.1
+# Protocol Buffers Event Format for CloudEvents  - Version 0.1
 
 ## Abstract
 
-The Protocol Buffers Format for CloudEvents (CE) defines
-the encoding of events in the protocol buffers binary format.
+The Protocol Buffers Format for CloudEvents (CE) defines the encoding
+of CloudEvents in the Protocol Buffers binary format.
 
 ## Status of this document
 
@@ -11,48 +11,46 @@ This document is a working draft.
 
 ## 1. Introduction
 
-CloudEvents is a standardized and transport-neutral definition of the
-structure and metadata description of events. This specification
-defines how the [Context
-Attributes](spec.md#context-attributes) defined in the CloudEvents in
-the protocol buffers binary encoding MUST BE encoded. Transcoding to
-and from other formats (e.g. JSON) is out of the scope of this
-proposal.
+This specification defines how the [Context
+Attributes](spec.md#context-attributes) Attributes defined in the
+CloudEvents specification MUST be encoded in the protocol buffer
+binary format. Transcoding to and from other formats (e.g. JSON) is
+out of the scope of this document.
 
-Protocol buffers are a language-neutral, platform-neutral extensible
+Protocol Buffers are a language-neutral, platform-neutral extensible
 mechanism for serializing structured data. The [Google reference
-implementation of
-protobuf](https://github.com/protocolbuffers/protobuf) includes
-support for an interface descriptor language (IDL), and this document
-makes use of language level 3 IDL from Protocol Buffers
-v3.5.0. CloudEvents systems using protocol buffers are not required to
-use the IDL or any particular implementation of Protocol buffers as
-long as they produce messages which match the binary encoding defined
-by the IDL.
+implementation of Protocol
+Buffers](https://github.com/protocolbuffers/protobuf) includes support
+for an interface descriptor language (IDL), and this document makes
+use of language level 3 IDL from Protocol Buffers v3.5.0. CloudEvents
+systems using Protocol Buffers are not required to use the IDL or any
+particular implementation of Protocol Buffers as long as they produce
+messages which match the binary encoding defined by the IDL.
 
 
 ### 1.1. Conformance
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
 "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this
-document are to be interpreted as described in RFC2119.
+document are to be interpreted as described in
+[RFC2119](https://tools.ietf.org/html/rfc2119).
 
 ## 2. Protocol Buffers format
 
-Protocol buffers provide a binary data serialization format which is
+Protocol Buffers provide a binary data serialization format which is
 substantially more compact and efficient to parse when compared to XML
 or JSON, along with a variety of language-specific libraries to
-perform automatic serialization and deserialization. The [protocol
-buffers specification defines a well-known encoding
+perform automatic serialization and deserialization. The [Protocol
+Buffers specification defines a well-known encoding
 format](https://developers.google.com/protocol-buffers/docs/encoding)
 which is the basis of this specification. Specifications below may be
-written using the protocol buffers project IDL for readability, but
-the ultimate basis of this specification is the protocol buffers
+written using the Protocol Buffers project IDL for readability, but
+the ultimate basis of this specification is the Protocol Buffers
 binary encoding.
 
 
 ### 2.1 Definition
 
-Users of protocol buffers MUST use a message whose binary encoding is
+Users of Protocol Buffers MUST use a message whose binary encoding is
 identical to the one described by the CloudEventMap message:
 
 ```proto
@@ -75,7 +73,8 @@ message CloudEventAny {
 }
 ```
 
-The CloudEvents type system is mapped into the fields of CloudEventAny as follows:
+The CloudEvents type system MUST be mapped into the fields of
+`CloudEventAny` as follows:
 
 
 | CloudEvents  | CloudEventAny field
@@ -86,3 +85,49 @@ The CloudEvents type system is mapped into the fields of CloudEventAny as follow
 | Timestamp    | string_value (string expression as defined in RFC 3339.)
 | Map          | map_value
 | Integer      | int_value
+| Any          | Not applicable. Any is the enclosing CloudEventAny message itself
+
+
+## Examples
+
+Below is an example of how to create a CloudEvent Protocol Buffer
+message using the Java Google Protocol Buffers library:
+
+```java
+import com.google.common.base.Charsets;
+import com.google.protobuf.ByteString;
+
+
+CloudEventMap event = CloudEventMap.newBuilder()
+  .putValue(
+    "eventType",
+    CloudEventAny.newBuilder()
+      .setStringValue("com.example.emitter.event")
+      .build())
+  .putValue(
+    "cloudEventsVersion",
+    CloudEventAny.newBuilder()
+      .setStringValue("0.1")
+      .build())
+  .putValue(
+    "eventTime",
+    CloudEventAny.newBuilder()
+      .setStringValue("2018-10-25T00:00:00+00:00")
+      .build())
+  .putValue(
+    "source",
+    CloudEventAny.newBuilder()
+      .setStringValue("com.example.source.host1")
+      .build())
+  .putValue(
+    "comExampleCustomextension",
+    CloudEventAny.newBuilder()
+      .setStringValue("some value for the extension")
+      .build())
+  .putValue(
+    "data",
+    CloudEventAny.newBuilder()
+      .setBinaryValue(ByteString.copyFrom("a binary string", Charsets.UTF_8))
+      .build())
+  .build();
+```
--- a/protobuf-format.md
+++ b/protobuf-format.md
@@ -2,36 +2,31 @@
 
 ## Abstract
 
-The Protocol Buffers (protobuf) Format for CloudEvents defines how
-events are to be expressed in protobuf version 3.
+The Protocol Buffers Format for CloudEvents (CE) defines
+the encoding of events in the protocol buffers binary format.
 
 ## Status of this document
 
-This document is a working draft
+This document is a working draft.
 
 ## 1. Introduction
 
 CloudEvents is a standardized and transport-neutral definition of the
 structure and metadata description of events. This specification
-defines how the elements defined in the CloudEvents specification are
-to be represented in the Protocol buffers version 3.
+defines how the [Context
+Attributes](spec.md#context-attributes)defined in the CloudEvents in
+the protocol buffers binary encoding MUST BE encoded. Transcoding to
+and from other formats (e.g. JSON) is out of the scope of this
+proposal.
 
 Protocol buffers are a language-neutral, platform-neutral extensible
-mechanism for serializing structured data. A message is defined once
-using the protobuf interface description language (IDL), and the
-protobuf compiler generates the language specific libraries for
-serializing to and deserializing from the binary representation. The
-protobuf version 3 library can also convert a protobuf message into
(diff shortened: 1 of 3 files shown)


duglin:
@zpencer can you wrap things at 80 columns to be consistent with the other docs we have?

protobuf-format.md
```diff
@@ -0,0 +1,387 @@
...
+If the payload is binary, the implementation MUST store the bytes in
+the `bytes_data` field.
+
+If the payload is a protobuf, the implementation MUST store the
+payload in the `proto_data` field and the `contentType` MUST be
+`application/protobuf`.
+
+### 2.3 Extensions:
+
+`google.protobuf.Struct` ([more info][PROTO_STRUCT]) represents an
```
clemensv: You could indeed take this construct and move it up one level and make the entire event a key/value pair map. Since you seem to be okay with doing that for extensions, why not also do that for the standard attributes? 

If you're principally worried about wire footprint, you could just use special-prefixed keys like "$0" and "$1" as shorthand for the well-known names. 

The complete wire footprint in proto for "$0", including type identifier and length prefix is 4 bytes, literally "x12 x02 x24 x30". We have 8 well-known attributes right now (with at least 1 of them in doubt and only 4 required). If we'd reach into the ASCII control character range instead of resorting to printable chars, I could cut another byte, e.g. `x12 x01 x01`

When we look at the required attributes, that's between 12 and 16 bytes extra "overhead" compared to pure positional encoding, and that's even within the TLS padding range, i.e. if you save those 12-16 bytes, TLS (the AES 128-bit block cipher inside of it) will pad at least some if not all of those savings away with extra zeroes when you put a single event on the wire. 

The effect of all this extra complication may quite literally come out to nothing.

(Mind that I'm not counting whatever the required overhead for the dictionary is because you've got that anyways for the extensions field defined here) 

I do realize that this might constrain the value type choice for the standard dictionary entries at the tooling level, but this spec seems to be unconcerned about that for extension values, which appear to be encoded as JSON strings.

At the wire level, two lower two bits of the preamble of any encoded protobuf field actually indicates the wire type, which means that the format factually can encode polymorphic dictionaries that distinguish between numbers and strings, which is sufficient for the value space we have in CloudEvents.

protobuf-format.md
```diff
@@ -0,0 +1,387 @@
...
+        "email",
+        Value.newBuilder().setStringValue("user@example.com").build())
+    .build()))
+  .setExtensions(
+    Struct.newBuilder()
+      .putFields(
+        "comExampleExtension1",
+        Value.newBuilder().setStringValue("value1").build())
+      .putFields(
+        "comExampleExtension2",
```
cneijenhuis: Maybe it would add some clarity if the field name was `com_example_extension_2`?

If I get https://developers.google.com/protocol-buffers/docs/proto3#json correctly, the snake_case to camelCase conversion is applied to proto field names, but - I guess - the names within a `Struct` aren't touched, as they are not proto fields?

protobuf-format.md
```diff
@@ -0,0 +1,387 @@
...
+  // exts bag to maintain backwards compat
+  .setExtensions(
+    Struct.newBuilder()
+      .putFields(
+        "comExampleMyextension",
+        Value.newBuilder().setStringValue("myvalue").build()))
+  .build();
+```
+
+#### 2.3.1 CloudEvents upgrade process
```
duglin: This feels out of place for this doc. If we need text like this then it should really be in the main spec since this would seem to apply in general, not just for this one serialization, right?

protobuf-format.md
```diff
@@ -0,0 +1,387 @@
...
+
+If a field moves from the extensions bag to a top level field, then
+the producers and consumers of a CloudEvents system SHOULD coordinate
+in the upgrade process:
+
+1. Initially, the producers and consumers are using CloudEvents 1.0
+   and the extension is expressed in the “extensions” bag.
+1. CloudEvents 1.1 is released.
+1. The producers write the extension to both the extensions bag and
```
duglin: Aside from the obvious concern about duplication of data, or worse, these two bits of data being different - can someone provide a URL to any spec out there that has a similar requirement? I can't think of any spec (JSON or otherwise) that requires this type of behavior and so I'd like to see how/why they do it. Since I don't think our spec is doing anything unusual I don't understand why our spec needs to have such special rules.

protobuf-format.md
```diff
@@ -0,0 +1,387 @@
...
+
+| CloudEvents | Protobuf
+|--------------|-------------------------------------------------------------
+| String       | string
+| Binary       | bytes
+| URI          | string
+| Timestamp    | google.protobuf.Timestamp
+| Map          | google.protobuf.Struct
+| Object       | google.protobuf.Value
+| Integer | int32
```
duglin: minor thing, but if you're going to go thru the trouble to align most of the columns, you might as well align all of them  :-)

protobuf-format.md
```diff
@@ -0,0 +1,387 @@
...
+   the type safe top level field. The messages will have
+   "cloudEventsVersion" set to "1.1", but the extension is still
+   readable by 1.0 consumers.
+1. All consumers upgrade to 1.1 and stop reading from the extensions
+   bag and switch to reading from only the type safe top level field.
+1. The producers stop writing to the extensions bag, and only write to
+   the type safe top level field.
+
+
+### 2.4 Relation to CloudEvents JSON format:
```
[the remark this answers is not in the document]
zpencer: If we only allow JSON payloads then the proto message's standard JSON is the same as the CE normative JSON. Binary data would be a base64 encoded JSON string object and protos would be serialized as binary data, so we won't take advantage of proto's more efficient representations of them.

protobuf-format.md
cneijenhuis: I agree that, per convention, extension attributes should be in camelCase. However, this is not a MUST, and a snake_cased attribute must not be transformed to camelCase.

I guess everything works as expected after I dug into the proto json spec, but it wasn't immediately clear to me why e.g. `event_id` is turned into `eventId`, but `com_example_extension` isn't.

zpencer: I see what you mean, I think I misunderstood your original comment. In the case of extensions being in the `Struct`, fields can be arbitrarily cased because it is a map keyed on case sensitive strings.
If we ever introduce a top level attribute that is not lowerCamelCase then we must use the `json_name` option to override the json field name.

cneijenhuis: Thanks for the clarification!

hsaliak:
Here are some reasons why I think Protocol Buffers meets the bar for a qualifying encoding.

Protocol buffers is a widely used format that has been open sourced by Google for a very long time now. 
* It is open sourced under a liberal 3-clause BSD license and has had multiple, significant contributions from folks outside of Google. It recently moved into it’s own org - https://github.com/protocolbuffers/protobuf
* Protocol Buffers is used in multiple RPC systems other than gRPC. A couple of examples are https://github.com/twitchtv/twirp and https://github.com/thesamet/rpcz. 
* Protocol Buffers has several implementations, most of which are not owned by Google. Some examples: https://github.com/dcodeIO/protobuf.js, https://github.com/nanopb/nanopb, https://github.com/protobuf-c/protobuf-c. Demonstrating that alternate implementations are not only possible, but have an important place in certain use cases. 
* gRPC is a CNCF project and is protocol agnostic. A protobuf based spec will not only help gRPC but also other projects that have widely adopted it in this space.

I believe these characteristics, and the strong and [vibrant ecosystem](https://github.com/topics/protocol-buffers) around  protocol buffers implementations and adoption, qualify protobuf as a format for encoding Cloud Events.

ultrasaurus:
When we talked about the language describing a de-facto standard, we specifically talked about *not* requiring something to be donated to a foundation to be considered a de-facto standard -- that was supposed to be an example, not the rule, and we agreed to later word-smith as needed.

Submitted: https://github.com/cloudevents/spec/pull/298/files?utf8=%E2%9C%93&diff=unified&w=1

the main point is to change "of" to "or" -- at the time, naming Kafka seemed to be a reasonable way to describe something that originated at a company and became more widely used, and now it is being used to introduce a requirement which seems to go against the intent of calling something a "de-facto standard"

JemDay:
There's obviously been a lot of discussion on this already but here are my thoughts..

I'd like to see all reference(s) to JSON removed as i think it muddies the definition somewhat if the intention is to create a proto representation of a cloud event.

Is there an argument to follow the spirit of the specification more tightly, by that i mean defining a _message_ to represent a CloudEvent Object (using the _oneof_ construct), then you can use the built-in Map type to represent a CloudEvent Map.

clemensv:
@ultrasaurus Whether something is a de-facto standard rather than a formal standard is a different concern from who owns the technology. Both concerns do apply for Protobuf. The concern that weighs more heavily is that while Protobuf may be widely used, is a single-vendor controlled technology (the current license is irrelevant because the owner can change all terms at any time) and that being the case has legal implications for multivendor standards efforts. Our legal team can talk to yours if that needs further clarification.

bgrant0607:
@clemensv 

The CNCF is not a standards body and does not seek to act like one:
https://github.com/cncf/toc/blob/master/PRINCIPLES.md#not-a-standards-body

Many CNCF projects take dependencies on and/or support things that aren't official standards and/or are non-foundation-owned open-source building blocks. YAML is not a standard, for example. Go is used by several projects. Kubernetes has a .NET client library.

I don't think we want to set the precedent that foundation projects can't support widely used non-official-standard/non-foundation-owned specifications, languages, libraries, etc. that meet our dependency licensing requirements. /cc @caniszczyk @monadic 

It is reasonable to consider what representation(s) in this specification might be most useful for a binary encoding, such as protocol buffers. Since official documentation, stackoverflow posts, and other documentation sources will likely be in the form of the "official" library interfaces, I think it's useful to include for explanatory purposes in addition to the binary representation. Perhaps one of the more popular non-Google implementations would be useful, also, such as https://github.com/dcodeIO/protobuf.js

bgrant0607:
cc @kenowens12

kenowens12:
Agree with @bgrant0607! Let's look at rewording this and adding flexibility.

clemensv:
@bgrant0607 Another principle in the document you're quoting is that projects are self-governing. I understand that CNCF doesn't want to be a standards body, but this particular WG is rather difficult to distinguish from a OASIS TC in how it operates.

We have a rule recognizing de-facto protocol standards emerging out of major foundation OSS projects such as the Apache Kafka protocol. .NET, since you mention it, is run under the umbrella of the .NET Foundation. Go is owned by the "Go authors" per license. YAML is owned by a group of individuals. 

That's all rather different from a toolset fully controlled by a single major vendor that can change and has changed dramatically between revisions or that the vendor could walk away from in favor of the next hot thing tomorrow - and history shows that this isn't a far-fetched scenario.

This group has literally been asked to adapt the JSON event format to the particular idiosyncrasies of that toolset and there's been a fairly direct [threat to sabotage the standard](https://github.com/cloudevents/spec/pull/277#issuecomment-415105008) if the WG does not comply:

"[...] Because proto is an IDL it also defines JSON interfaces as well. We are hoping for a JSON spec for which we can reverse engineer a proto shape to match. Otherwise Google will not be able to keep from releasing an alternate JSON binding. [...]"

I don't think we want to set the precedent that foundation projects must bend to the needs of proprietary projects. 

If Protobuf were to follow gRPC and be donated into CNCF, and the focus of the Protobuf event format were strictly on the binary wire format, along with explanatory references to toolsets for producing that format [as I indicated in my review comment above](https://github.com/cloudevents/spec/pull/295#discussion_r212221591) and which you appear to agree with, my concerns would be addressed.

rachelmyers:
@clemensv I need to address your claim that explaining an implication of a proposed change is a "fairly direct threat to sabotage the standard". It's not. You've misunderstood that comment entirely. And we're not creating a standard.

Here's the substantive paragraph from the comment you linked to:

> Any wire protocol or binding that has struct-like features (e.g. proto or C-structs) will necessarily face a breaking change when an extension is ratified to the main spec. We are looking for a commitment from the working group that they would not consider this a minor version. I've shown this issue without using protobuffers in [https://github.com/inlined/versioningishard](https://github.com/inlined/versioningishard)
Binary proto APIs are written in the proto [Interface Description Language]. Because proto is an IDL it also defines JSON interfaces as well. We are hoping for a JSON spec for which we can reverse engineer a proto shape to match. Otherwise Google will not be able to keep from releasing an alternate JSON binding.

You seem you have misunderstood the last sentence. That's someone explaining that the change forces proto implementations to support the JSON spec generated by the Proto IDL and a JSON spec that matching future versions of CloudEvents. That's not a threat; that's trying to make the spec as widely implementable as possible.

bgrant0607:
Thanks, @rachelmyers.

Since this appears to be getting heated, I'll remind everyone of our code of conduct:
https://github.com/cncf/foundation/blob/master/code-of-conduct.md

To quote from our principles document:

CNCF “specification” materials are not “standards”. It is in future possible that an independent and recognized international standards body takes a CNCF document as “upstream” and evolves it into a standard via (e.g.) the IETF process. The CNCF is morally supportive of independent parties doing this, but does not see this work as its own responsibility. For that matter, to date, no such exercise is in process.

In general CNCF specifications will evolve as “living documents” side by side with the CNCF OSS projects that adopt them. This means they are “dynamic and fast-moving” whereas a typical IETF standard is ”static” and not subject to frequent updates & new releases.

For the avoidance of doubt: Any written “specification” materials in CNCF shall have the same status as CNCF open source software and project technical docs. That is to say that specifications shall be updated and release versioned according to the conventions of a CNCF open source project.

The CNCF IP policy, which includes whitelisted and blacklisted open-source licenses, is in section 11 of the charter:
https://www.cncf.io/about/charter/

There is no foundation-level policy about independence of project dependencies.

Hopefully that was clear before CloudEvents was inducted into the sandbox.

Coming back to the issue of protocol buffers specifically, yes, I agree some reasonable concerns about the form of the document were raised.

However, on the issue of ownership: I don't understand the objective of not having "official" specifications and/or SDK support for widely used formats, or the distinction being made in the YAML case (I thought I'd seen a proposal for an incompatible YAML2, but can't find it right now). Is the desired outcome for fewer formats to be used? Why wouldn't the outcome be fragmentation and inconsistency among the formats used, and/or for adoption to be impeded?

I don't know whether the point raised about the addition of extensions was addressed.

duglin:
I've asked for some off-line discussions to happen around this. I'd like to ask that we hold off further discussions in this PR until those have happened.

duglin:
@clemensv , @rachelmyers , @bgrant0607 and I had a chat the other day concerning this PR. The net results were:
- any legal concerns about CNCF activities referring to protobuf are not for the CloudEvents project to resolve, and therefore will be addressed elsewhere.
- whether or not the CloudEvents spec will reference any other specification, assuming such a reference would not violate any CNCF procedures/governance rules, is up to the CloudEvents members to decide and we can make this decision on a case by case basis.
- this PR will be modified to focus on the binary protobuf format to ensure portability and forward compatibility
- the tool-independent IDL proto file representation may be shown for explanatory purposes

dankohn:
I just want to confirm that CNCF/LF legal is engaged with Microsoft legal to address their concerns.

zpencer:
I'll be sending out an update to this PR in the next couple of days.

duglin:
@zpencer thanks for the update!

glerchundi:
Any update on this @zpencer? Fantastic job by the way!

zpencer:
@glerchundi there are some discussions happening internally about how to address the PR feedback. I will post an update to the PR as soon as it is ready.

glerchundi:
Thanks for the update!

duglin:
@zpencer I see you pushed some more commits, please let us know when you think the PR is ready for another round of reviews.

duglin:
@zpencer thanks!
All - please look this one over for this week's call.

protobuf-format.md
```diff
@@ -0,0 +1,88 @@
+# Protobuf Event Format for CloudEvents  - Version 0.1
+
+## Abstract
+
+The Protocol Buffers Format for CloudEvents (CE) defines
```
duglin: For consistency it seems like this should either be "Protobuf" like the title or the title (on line 1) should be "Protocol Buffers". Or add text to indicate that "Protobuf" is short for "Protocol Buffers".

protobuf-format.md
```diff
@@ -0,0 +1,88 @@
...
+the encoding of events in the protocol buffers binary format.
```
duglin: s/event/CloudEvents/

protobuf-format.md
```diff
@@ -0,0 +1,88 @@
...
+## Status of this document
+
+This document is a working draft.
+
+## 1. Introduction
+
+CloudEvents is a standardized and transport-neutral definition of the
+structure and metadata description of events. This specification
+defines how the [Context
+Attributes](spec.md#context-attributes) defined in the CloudEvents in
```
duglin: This sentence is a bit awkward. Perhaps:
This specification defines how the [Context Attributes](spec.md#context-attributes) defined in the CloudEvents specification MUST be encoded in the protocol buffer binary format.

protobuf-format.md
```diff
@@ -0,0 +1,88 @@
...
+the protocol buffers binary encoding MUST BE encoded. Transcoding to
+and from other formats (e.g. JSON) is out of the scope of this
+proposal.
```
duglin: s/proposal/document/   (or specification)
since once its merged it won't be a proposal any more

protobuf-format.md
```diff
@@ -0,0 +1,88 @@
...
+v3.5.0. CloudEvents systems using protocol buffers are not required to
+use the IDL or any particular implementation of Protocol buffers as
+long as they produce messages which match the binary encoding defined
+by the IDL.
+
+
+### 1.1. Conformance
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this
+document are to be interpreted as described in RFC2119.
```
duglin: can you make the RFC2119 word into a link to https://tools.ietf.org/html/rfc2119  ?

protobuf-format.md
```diff
@@ -0,0 +1,88 @@
...
+
+
+| CloudEvents  | CloudEventAny field
+|--------------|-------------------------------------------------------------
+| String       | string_value
+| Binary       | binary_value
+| URI          | string_value (string expression conforming to URI-reference as defined in RFC 3986 §4.1)
+| Timestamp    | string_value (string expression as defined in RFC 3339.)
+| Map          | map_value
+| Integer      | int_value
```
duglin: Maybe its obvious and not needed, but should "Any" be listed in here just for completeness?

protobuf-format.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+The CloudEvents type system is mapped into the fields of CloudEventAny as follows:
+
+
+| CloudEvents  | CloudEventAny field
+|--------------|-------------------------------------------------------------
+| String       | string_value
+| Binary       | binary_value
+| URI          | string_value (string expression conforming to URI-reference as defined in RFC 3986 §4.1)
```
duglin: Can you make the RFC text into a link?

duglin:
@zpencer While I like how short the doc is, it feels like it ends abruptly when I'm used to seeing some kind of example after the "Definition" section. Maybe it's not needed for someone who knows protobuf and if so then ignore this comment.

protobuf-format.md
```diff
@@ -0,0 +1,88 @@
...
+  oneof value {
+    string string_value = 1;
+    bytes binary_value = 2;
+    uint32 int_value = 3;
+    CloudEventMap map_value = 4;
+  }
+}
+```
+
+The CloudEvents type system is mapped into the fields of CloudEventAny as follows:
```
duglin: s/is mapped/MUST be mapped/

protobuf-format.md
zpencer: Thanks for the improved wording, I will use this phrasing.

protobuf-format.md
zpencer: Added a note on `Any`

protobuf-format.md
```diff
@@ -0,0 +1,133 @@
...
+of CloudEvents in the Protocol Buffers binary format.
+
+## Status of this document
+
+This document is a working draft.
+
+## 1. Introduction
+
+This specification defines how the [Context
+Attributes](spec.md#context-attributes) Attributes defined in the
```
duglin: the 2nd "Attributes" (after the link) is a repeat.

protobuf-format.md
```diff
@@ -0,0 +1,133 @@
...
+binary format. Transcoding to and from other formats (e.g. JSON) is
+out of the scope of this document.
+
+Protocol Buffers are a language-neutral, platform-neutral extensible
+mechanism for serializing structured data. The [Google reference
+implementation of Protocol
+Buffers](https://github.com/protocolbuffers/protobuf) includes support
+for an interface descriptor language (IDL), and this document makes
+use of language level 3 IDL from Protocol Buffers v3.5.0. CloudEvents
+systems using Protocol Buffers are not required to use the IDL or any
```
duglin: s/required/manadated/ to avoid the RFC checker... which reminds me, can you modify the Makefile as part of this PR so that the RFC checker is run on this doc too?

JemDay:
I'd like to see this more closely mirror that of the json-format.md specification, eg..

1. An Envelope section that defines the contentType to be used to allow HTTP transport in structured-mode, eg..
     Such a representation uses the media type `application/cloudevents+<SPECIFY-ME>`.

2. Create a spec.proto file to contain the proto definitions as described in this document.

As a general comment part of me would like to see a formal `message CloudEvent { .. }` defined that contains all of the expected/optional properties as opposed to the generalized `map<>` approach outlined .. i would admit to be internal flip-flopping on that.

duglin:
@zpencer any chance you can address the open question before Tuesday next week? It would be great if we could get this merged for v0.2 - which we're hoping to have for KubeCon Seattle.

protobuf-format.md
zpencer: Done

protobuf-format.md
zpencer: Done

zpencer:
> I'd like to see this more closely mirror that of the json-format.md specification, eg..
> 
> 1. An Envelope section that defines the contentType to be used to allow HTTP transport in structured-mode, eg..
>    ```
>      Such a representation uses the media type `application/cloudevents+<SPECIFY-ME>`.
>    ```
I thought about this some more and I'm not sure structured / binary mode really apply for the proto format. If we had ended up with type safe fields in the proto, and some way to embed a proto inside (like with a `google.protobuf.Any`), then we could make the argument that HTTP binary content mode would let us unwrap the embedded proto and use it for the HTTP request/response as-is. With the current design, the `data` attribute is itself a `CloudEventAny`, so being able to use it as the HTTP body does not buy us very much.

> 2. Create a spec.proto file to contain the proto definitions as described in this document.
> 
Added `spec.proto`

protobuf-format.md
```diff
@@ -0,0 +1,134 @@
...
+
+## 2. Protocol Buffers format
+
+Protocol Buffers provide a binary data serialization format which is
+substantially more compact and efficient to parse when compared to XML
+or JSON, along with a variety of language-specific libraries to
+perform automatic serialization and deserialization. The [Protocol
+Buffers specification defines a well-known encoding
+format](https://developers.google.com/protocol-buffers/docs/encoding)
+which is the basis of this specification. Specifications below may be
```
duglin: s/may/MAY/
What do you mean by "below"? Do you mean later on in the this doc? Which specs are you referring to?

protobuf-format.md
duglin: thanks

JemDay:
@zpencer - Should there at east be some guidance around contentType(s) so that the I could send the  played across other transports such as HTTP. AMQP, etc..

To the best of my knowledge there isn't a registered media-type for a proto object, absent that (if that's still true) we should provide guidance to allow interop.

protobuf-format.md
zpencer: Reworded this section. Apologies for the strange wording.

zpencer:
> @zpencer - Should there at least be some guidance around contentType(s) so that the I could send the played across other transports such as HTTP. AMQP, etc..
> 
> To the best of my knowledge there isn't a registered media-type for a proto object, absent that (if that's still true) we should provide guidance to allow interop.

I believe the closest thing to a proto content type is the [expired IETF draft](https://tools.ietf.org/html/draft-rfernando-protocol-buffers-00) which registers `application/protobuf` and does not define any parameters for expressing the protocol buffer message's type. Between this and the other commonly used `application/x-protobuf` content type, I think we should go with the one from the expired IETF draft.

duglin:
Approved on the Nov 8 call.
@rachelmyers will open a follow-on PR to address: https://github.com/cloudevents/spec/pull/295#issuecomment-435609246

rachelmyers:
Created [PR 346](https://github.com/cloudevents/spec/pull/346) to specify the envelope for messages using the proto spec. cc @JemDay

PR #298, opened 2018-08-30
qualifying projects

ultrasaurus:
de-facto standards often emerge from the community
joining a consortia should not be necessary

### DIFF
--- a/primer.md
+++ b/primer.md
@@ -171,8 +171,10 @@ or events on the web, or MQTT and AMQP for connection-oriented messaging and
 telemetry transfers.
 
 Some widely used protocols have become de-facto standards emerging out of strong
-ecosystems or top-level multi-company consortia projects and largely in parallel
-to the evolution of the aforementioned standards stacks.
+ecosystems of top-level consortia of three or more companies, and some out of
+the strong ecosystems of projects released by a single company, and in either
+case largely in parallel to the evolution of the previously mentioned standards
+stacks.
 
 The CloudEvents effort shall not become a vehicle to even implicitly endorse
 or promote project- or product-proprietary protocols, because that would be
--- a/primer.md
+++ b/primer.md
@@ -158,10 +158,10 @@ protocols, with CloudEvents aiming to provide such an open data format and
 projections of its data format onto commonly used protocols and with commonly
 used encodings.
 
-While each software or service product and project can obviously make its own 
-choices about which form of communication it prefers, its unquestionable that 
-a proprietary protocol that is private to such a product or project does not 
-further the goal of broad interoperability across producers and consumers of 
+While each software or service product and project can obviously make its own
+choices about which form of communication it prefers, its unquestionable that
+a proprietary protocol that is private to such a product or project does not
+further the goal of broad interoperability across producers and consumers of
 events.
 
 Especially in the area of messaging and eventing, the industry has made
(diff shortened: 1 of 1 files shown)


ultrasaurus:
Note on Kafka:  it may have become a de-facto standard before being donated to the consortium. If CloudEvents existed before that point and if many Serverless companies were using the protocol with different implementations, I think we would have (should have?) considered it for inclusion.

duglin:
The change appears to now allow for a `strong ecosystem` w/o a consortia - would we need some kind of measuring stick to make `strong` less subjective?  Perhaps define the expected characteristics? e.g. size, type of governance model, number of companies....

clemensv:
[Changing "of" to "or"](https://github.com/cloudevents/spec/pull/298/files#diff-ed0940ecbc7be65c47b9414d516b3490L174) completely changes the semantics of the sentence you're editing. I understand that you are looking to add a point, but you're breaking up a single term to do so. I believe it will help clarity if you could make your addition more explicit. Simply flipping "of" to "or" ends up completely contradicting what the term says.

Here's how I interpret the intent of your edit:

"Some widely used protocols have become de-facto standards emerging out of strong ecosystems of top-level multi-company consortia projects **, and some out of the strong ecosystems of proprietary projects, and in either case** largely in parallel to the evolution of the aforementioned standards stacks."

The reason I wrote that paragraph is to introduce the notion of de-facto standards relative to consortia standards before discussing them jointly in the following.

primer.md
```diff
@@ -171,17 +171,17 @@ or events on the web, or MQTT and AMQP for connection-oriented messaging and
 telemetry transfers.
 
 Some widely used protocols have become de-facto standards emerging out of strong
-ecosystems of top-level multi-company consortia projects, such as Apache Kafka, 
-and largely in parallel to the evolution of the aforementioned standards stacks.
+ecosystems or top-level multi-company consortia projects and largely in parallel
+to the evolution of the aforementioned standards stacks.
```
cathyhongzhang: Can we quantify "multi-company"?

nerdyyatrice: as I mentioned in last week's call, given what I now know the intent of this change, I am not even sure this part is necessary since it's a very subject description. We have a more objective definition of de-facto standards in the text below.

rachelmyers: `multi-company consortia projects` was already accepted, and isn't being changed. If that needs to be defined to be more specific, I could take a stab at that. 

The change here is that it's sufficient to for a project to have a strong ecosystems built, and it's also sufficient to be built by multi-company consortia. Previously the clauses were each necessary and jointly sufficient; now they're each sufficient. I think it's a good change, because there's a lot of widely used technology that didn't emerge from multi-company consortia that we would ideally still interoperate with.

cathyhongzhang: I support this change. Just feel it will be better to quantify the "multi-company".  You are right that my comment is not to this PR change, but rather to the existing text. It will be great if this PR can take care of this together. Thanks

rachelmyers: How about `s/multi-company consortia/consortia of three or more companies`? In context, with some edits to make the grammar work, that becomes:

> Some widely used protocols have become de-facto standards emerging out of strong ecosystems or projects from top-level consortia of three or more companies, created largely in parallel to the evolution of the previously mentioned standards stacks.

Does that work?

cathyhongzhang: LGTM

rachelmyers: @ultrasaurus, does this wording work for you?

duglin:
@rachelmyers did you want to try to incorporate @clemensv's wording change too?

rachelmyers:
"Propietary" has a different connotation than what I think is meant here, but trying to take the spirit of @clemensv's suggestion, I would be happy with: 

> Some widely used protocols have become de-facto standards emerging out of strong ecosystems of top-level consortia of three or more companies, and some out of the strong ecosystems of projects released by a single company, and in either case largely in parallel to the evolution of the aforementioned standards stacks.

duglin:
What do people think about @rachelmyers's suggested text?

duglin:
I'd like to see if we can resolve this on tomorrow's call - what do people think of @rachelmyers's proposed text in  https://github.com/cloudevents/spec/pull/298#issuecomment-421432533  ?

duglin:
I can live with it.
LGTM

duglin:
@rachelmyers perhaps if no one raises any concern by around noon your time you can update the PR with the proposed text? Or, if you can't do it (directly or via @ultrasaurus ) let me know and I'll give it a try.

clemensv:
The point of the paragraph is ownership and governance and how standards come about. Proprietary provenance is the new option introduced by this edit. The dictionary definition of proprietary is clear. Most services I work on are proprietary; it's a descriptor. People have connotations about that not being the same as FOSS developed under the umbrella of a neutral body and they're right about that.

The change @rachelmyers now suggests makes it a count of companies here and a count of companies over there, which doesn't describe a distinction that I find to matter.

That said, if we really need to make an edit, I clearly prefer @rachelmyers version because it's separating the new clause for proprietary (single company) projects that's being introduced by the edit. 

I'm not going to object to accepting @rachelmyers version.

ultrasaurus:
updated -- sorry for extra whitespace changes, but looks like github lets you ignore them for review

thanks @clemensv @cathyhongzhang @rachelmyers for the feedback and suggested wording!

duglin:
Approved in the 9/20 call

PR #215, opened 2018-05-24
NATS transport binding

ColinSullivan1:
I'd like to propose a [NATS](https://nats.io) transport binding for CloudEvents, similar to the [HTTP](https://github.com/cloudevents/spec/pull/148) binding and proposed [MQTT](https://github.com/cloudevents/spec/pull/158) and [AMQP](https://github.com/cloudevents/spec/pull/157) bindings.  This PR borrows heavily from the MQTT proposal.

Signed-off-by: Colin Sullivan <colin@synadia.com>

PR #158, opened 2018-04-12
MQTT 3.1.1 and 5.0 transport binding

clemensv:
Proposed MQTT 3.1.1 and 5.0 transport binding, analogous to the same for [HTTP](https://github.com/cloudevents/spec/pull/148) and AMQP. This includes the same JSON mapping as the HTTP PR for completeness and the JSON mapping should be discussed there, since HTTP has higher priority.

Signed-off-by: Clemens Vasters <clemensv@microsoft.com>

PR #300, opened 2018-09-01
Kafka transport binding

bnlong4417:
I'd like to propose a [Kafka](https://kafka.apache.org/) transport binding for CloudEvents, similar to the [HTTP](https://github.com/cloudevents/spec/pull/148) binding and proposed [NATS](https://github.com/cloudevents/spec/pull/215), [MQTT](https://github.com/cloudevents/spec/pull/158), [AMQP](https://github.com/cloudevents/spec/pull/157) bindings.

Signed-off-by: Long Bui bnlong@fossil.com

### DIFF
--- a/kafka-transport-binding.md
+++ b/kafka-transport-binding.md
@@ -1,0 +1,277 @@
+# Kafka Transport Binding for CloudEvents
+
+## Abstract
+
+The [Kafka][Kafka] Transport Binding for CloudEvents defines how events are
+mapped to [Kafka messages][Kafka-Message-Format].
+
+## Status of this document
+
+This document is a working draft.
+
+## Table of Contents
+
+1. [Introduction](#1-introduction)
+- 1.1. [Conformance](#11-conformance)
+- 1.2. [Relation to Kafka](#12-relation-to-kafka)
+- 1.3. [Content Modes](#13-content-modes)
+- 1.4. [Event Formats](#14-event-formats)
+- 1.5. [Security](#15-security)
+2. [Use of CloudEvents Attributes](#2-use-of-cloudevents-attributes)
+- 2.1. [contentType Attribute](#21-contenttype-attribute)
+- 2.2. [data Attribute](#22-data-attribute)
+3. [kafka Message Mapping](#3-kafka-message-mapping)
+- 3.2. [Binary Content Mode](#31-binary-content-mode)
+- 3.1. [Structured Content Mode](#32-structured-content-mode)
+4. [References](#4-references)
+
+## 1. Introduction
+
+[CloudEvents][CE] is a standardized and transport-neutral definition of the
+structure and metadata description of events. This specification defines how
+the elements defined in the CloudEvents specification are to be used in the
+Kafka protocol as [Kafka messages][Kafka-Message-Format] (aka Kafka records).
+
+### 1.1. Conformance
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in [RFC2119][RFC2119].
+
+### 1.2. Relation to Kafka
+
+This specification does not prescribe rules constraining transfer or settlement
+of event messages with Kafka; it solely defines how CloudEvents are expressed
+in the Kafka protocol as [Kafka messages][Kafka-Message-Format].
+
+### 1.3. Content Modes
+
+The specification defines two content modes for transferring events:
+*structured* and *binary*.
+
+The *binary* mode *only* applies to Kafka 0.11.0.0 and above, because of Kafka
+0.10.x.x and below are lack of support for message level headers.
+
+In the *structured* content mode, event metadata attributes and event data are
+placed into the Kafka message value section
+using an [event format](#14-event-formats).
+
+In the *binary* content mode, the value of the event `data` attribute is placed
+into the Kafka message's value section as-is, with
+the `kafka_contentType` header value declaring its media type; all other event
+attributes are mapped to the Kafka message's
+[header section][Kafka-Message-Header].
+
+### 1.4. Event Formats
+
+Event formats, used with the *stuctured* content mode, define how an event is
+expressed in a particular data format. All implementations of this
+specification MUST support the [JSON event format][JSON-format].
+
+### 1.5. Security
+
+This specification does not introduce any new security features for Kafka, or
+mandate specific existing features to be used.
+
+## 2. Use of CloudEvents Attributes
+
+This specification does not further define any of the [CloudEvents][CE] event
+attributes.
+
+### 2.1. contentType Attribute
+
+The `contentType` attribute is assumed to contain a media-type expression
+compliant with [RFC2046][RFC2046].
+
+### 2.2. data Attribute
+
+The `data` attribute is assumed to contain opaque application data that is
+encoded as declared by the `contentType` attribute.
+
+An application is free to hold the information in any in-memory representation
+of its choosing, but as the value is transposed into Kafka as defined in this
+specification, core Kafka provides data available as a sequence of bytes.
+
+For instance, if the declared `contentType` is
+`application/json;charset=utf-8`, the expectation is that the `data` attribute
+value is made available as [UTF-8][RFC3629] encoded JSON text.
+
+## 3. Kafka Message Mapping
+
+With Kafka 0.11.0.0 and above, the content mode is chosen by the sender of the 
+event. Protocol usage patterns that might allow solicitation of events using a
+particular content mode might be defined by an application, but are not defined
+ here.
+
+The receiver of the event can distinguish between the two content modes by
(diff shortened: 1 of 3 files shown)


kafka-transport-binding.md
```diff
@@ -0,0 +1,274 @@
+# Kafka Transport Binding for CloudEvents
+
+## Abstract
+
+The [Kafka][Kafka] Transport Binding for CloudEvents defines how events are mapped to [Kafka messages][kafka-Message-Format].
```
duglin: please wrap at 80 cols

duglin:
CI error is ok and will automatically be fixed once merged.

kafka-transport-binding.md
duglin: thanks

kafka-transport-binding.md
```diff
@@ -0,0 +1,277 @@
...
+*structured* and *binary*.
+
+The *binary* mode *only* applies to Kafka 0.11.0.0 and above, because of Kafka
+0.10.x.x and below are lack of support for message level headers.
+
+In the *structured* content mode, event metadata attributes and event data are
+placed into the Kafka message value section
+using an [event format](#14-event-formats).
+
+In the *binary* content mode, the value of the event `data` attribute is placed
```
duglin: I wonder if we need an RFC2119 keyword in here, perhaps s/is placed/MUST be placed/  ??

kafka-transport-binding.md
```diff
@@ -0,0 +1,277 @@
...
+into the Kafka message's value section as-is, with
+the `kafka_contentType` header value declaring its media type; all other event
+attributes are mapped to the Kafka message's
```
duglin: maybe:  s/are mapped/MUST be mapped/

kafka-transport-binding.md
```diff
@@ -0,0 +1,277 @@
...
+For instance, if the declared `contentType` is
+`application/json;charset=utf-8`, the expectation is that the `data` attribute
+value is made available as [UTF-8][RFC3629] encoded JSON text.
+
+## 3. Kafka Message Mapping
+
+With Kafka 0.11.0.0 and above, the content mode is chosen by the sender of the 
+event. Protocol usage patterns that might allow solicitation of events using a
+particular content mode might be defined by an application, but are not defined
+ here.
```
duglin: extra space before "here"

kafka-transport-binding.md
```diff
@@ -0,0 +1,277 @@
...
+
+#### 3.1.2. Event Data Encoding
+
+The [`data` attribute](#22-data-attribute) byte-sequence MUST be used as the
+value of the Kafka message.
+
+#### 3.1.3. Metadata Headers
+
+All [CloudEvents][CE] attributes with exception of `data` MUST be individually
+mapped to and from the Header fields in the Kafka message.
```
duglin: Is there any rule needed w.r.t. the names used? Case of the names?

kafka-transport-binding.md
```diff
@@ -0,0 +1,277 @@
...
+
+##### 3.1.3.1 User Property Names
+
+Cloud Event attributes are prefixed with "cloudEvents_" for use in the
```
duglin: extra space between "Cloud" and "Event"

kafka-transport-binding.md
```diff
@@ -0,0 +1,277 @@
...
+
+The [`data` attribute](#22-data-attribute) byte-sequence MUST be used as the
+value of the Kafka message.
+
+#### 3.1.3. Metadata Headers
+
+All [CloudEvents][CE] attributes with exception of `data` MUST be individually
+mapped to and from the Header fields in the Kafka message.
+
+##### 3.1.3.1 User Property Names
```
duglin: What is the difference between "User Properties" and "Metadata Headers" ?

kafka-transport-binding.md
```diff
@@ -0,0 +1,277 @@
...
+
+
+#### 3.1.4 Examples
+
+This example shows the *binary* mode mapping of an event into the
+Kafka message. The CloudEvents `contentType` attribute is mapped to
+the Kafka `kafka_contentType` header; all other CloudEvents attributes
+are mapped to Kafka Header fields with prefix `cloudEvents_`.
+
+Mind that `cloudEvents_` here does refer to the event `data`
```
duglin: I'm not sure I understand this line, what do you mean by "`cloudEvents_` here" ?

kafka-transport-binding.md
```diff
@@ -0,0 +1,277 @@
...
+``` text
+content-type: application/cloudevents+json; charset=UTF-8
+```
+
+#### 3.2.2. Event Data Encoding
+
+The chosen [event format](#14-event-formats) defines how all attributes,
+including the `data` attribute, are represented.
+
+The event metadata and data is then rendered in accordance with the event
```
duglin: s/is/are/  I think

duglin:
@bnlong4417 can you address the comments in here - then we may be able to resolve this on next week's call

clemensv:
I'm in generally in favor of having such a binding, and haven't yet been able to make the time to review this proposal side-by-side with the Kafka protocol spec, but promise to do so this week.

I would like to use this PR to close some @duglin assigned homework, namely the AI related to issue #218 .

We should have an explicit mechanism in here that states how to derive the partitionKey that Kafka needs and that this proposal here doesn't yet seem to mention except in the example. I don't have a proposal yet for how to do that.

kafka-transport-binding.md
```diff
@@ -0,0 +1,277 @@
...
+
+With Kafka 0.11.0.0 and above, the content mode is chosen by the sender of the 
+event. Protocol usage patterns that might allow solicitation of events using a
+particular content mode might be defined by an application, but are not defined
+ here.
+
+The receiver of the event can distinguish between the two content modes by
+inspecting the `kafka_contentType` property of the Kafka message. If the value
+of the `kafka_contentType` property is prefixed with the CloudEvents media type
+`application/cloudevents`, indicating the use of a known [event
```
jroper: If I'm not mistaken, this has been copied from the HTTP binding, which means we now have two different specs specifying this MIME type. That's not ideal, it should only be specified in one place. So I think this should be added to the main spec to say what its MIME type is, and then both this and the HTTP binding can reference that, plus the JSON spec should be the authoritative spec that says that it is the format for `application/cloudevents+json`, not this and the HTTP binding.

kafka-transport-binding.md
```diff
@@ -0,0 +1,277 @@
...
+The *binary* mode *only* applies to Kafka 0.11.0.0 and above, because of Kafka
+0.10.x.x and below are lack of support for message level headers.
+
+In the *structured* content mode, event metadata attributes and event data are
+placed into the Kafka message value section
+using an [event format](#14-event-formats).
+
+In the *binary* content mode, the value of the event `data` attribute is placed
+into the Kafka message's value section as-is, with
+the `kafka_contentType` header value declaring its media type; all other event
```
jroper: Where does `kafka_contentType` come from? As far as I can work out, this is unique to this spec. While there isn't a standard header in Kafka for content type, it does appear that there are libraries out there that do put a content type header, and it appears to me that there is some consensus around `contentType`. Even without that, I don't think `kafka_contentType` makes sense, since of course it's Kafka. If this header is specific to CloudEvents, and you want to namespace it, then the prefix should be `couldevents`, since a CloudEvents spec like this has no business defining a header in the `kafka` namespace.

duglin:
@bnlong4417 there are some comments/questions - can you try to address those? We can't consider adopting this until those are addressed.

duglin:
@bnlong4417 are you still interested in us adopting this PR? There are some updates needed if so.

bnlong4417:
@duglin My sincere apologies for the slow reply. I am very busy at the moment because our company is preparing for the important holiday sales this year and I forgot to catch this up. I'm totally ok with Avery & Baldwin to carry this forward, thank you for continuing to integrate Kafka into CloudEvents.

kafka-transport-binding.md
bnlong4417: `cloudEvents_` is the prefix of CloudEvents context attributes except `contentType` attribute.

kafka-transport-binding.md
bnlong4417: Currently, there isn't a standard header in Kafka for the content type. So I used the Spring's format for Kafka header key at here:  https://github.com/spring-projects/spring-kafka/blob/996f8649d562cb7a2b15af42ab0bea64b510d961/spring-kafka/src/main/java/org/springframework/kafka/support/KafkaHeaders.java#L32

kafka-transport-binding.md
bnlong4417: Yes, I borrowed it from HTTP specs.  I think it should be moved to main specs too.

duglin:
@bnlong4417 no worries - thanks for getting back to us.... I was a little worried that you vanished :-)

bluemonk3y:
@duglin  - I will have this available next week - I'm currently OOO - regards Neil

duglin:
I'm going to assume we can close this due to #337 is now out there.  I can reopen if someone things I jumped the gun.

PR #301, opened 2018-09-05
Add clarification on including additional context attributes

cathyhongzhang:

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -153,7 +153,9 @@ both appear within the same JSON object.
 
 ### Extension Attributes
 Any context attributes that are not defined in this specification are
-extension attributes.
+extension attributes. See
+[CloudEvent Attributes Extensions](primer.md#cloudevent-attribute-extensions)
+for additional information concerning the use and definition of extensions.
 
 This specification places no restriction on the type or semantics of the
 extension attributes. Each definition of an extensions SHOULD fully
--- a/spec.md
+++ b/spec.md
@@ -167,7 +167,7 @@ use to correlate this event with the other events. If such identity attributes
 happen to be part of the event "data", the event producer SHOULD also add
 the identity attributes to the "context attributes" so that
 event consumers can easily access this information without needing to decode
-and examine the event data. Such identity attributes MAY also be used to
+and examine the event data. Such identity attributes can also be used to
 help intermediate gateways determine how to route the events.
 
 ### eventType
--- a/spec.md
+++ b/spec.md
@@ -129,8 +129,18 @@ variant type.
 Every CloudEvent conforming to this specification MUST include one or more
 of the following context attributes.
 
-It could also include additional attributes in the "context attributes" that
-might be used in ancillary actions related to the processing of the event.
+These attributes, while descriptive of the event, are designed such that they
+can be serialized independent of the event data. This allows for them to be
+inspected at the destination without having to deserialize the event data.
+
+The choice of serialization mechanism will determine how the context
+attributes and the event data will be materialized. For example, in the case
+of a JSON serialization, the context attributes and the event data might
+both appear within the same JSON object.
+
+### Extension Attributes
+CloudEvents producers MAY include additional context attributes in the event
+that might be used in ancillary actions related to the processing of the event.
 For example, in many IoT and enterprise use cases, an event could be used in
 a serverless application that performs actions across multiple types of events.
 To support such use cases, the event producer will need to add additional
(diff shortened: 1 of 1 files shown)


clemensv:
That's a great consolidation of all the extensibility discussions. LGTM.

spec.md
```diff
@@ -129,6 +129,19 @@ variant type.
 Every CloudEvent conforming to this specification MUST include one or more
 of the following context attributes.
 
+It could also include additional attributes in the "context attributes" that
```
cneijenhuis: > CloudEvent producers MAY include additional extension attributes within the
event

spec.md
```diff
@@ -129,6 +129,19 @@ variant type.
...
+might be used in ancillary actions related to the processing of the event.
```
cneijenhuis: > that might be used for any purpose in the processing of the CloudEvent

spec.md
```diff
@@ -129,6 +129,19 @@ variant type.
...
+For example, in many IoT and enterprise use cases, an event could be used in
+a serverless application that performs actions across multiple types of events.
```
cneijenhuis: > [...] such as identifying or correlating event sources.

spec.md
```diff
@@ -129,6 +129,19 @@ variant type.
...
+To support such use cases, the event producer will need to add additional
+identity attributes to the "context attributes" which the event consumers can
+use to correlate this event with the other events. If such identity attributes
+happen to be part of the event "data", it is still suggested that the event
+producer add the identity attributes to the "context attributes" so that event
```
cneijenhuis: `also adds` may be clearer - the attribute shouldn't be removed from "data".

I'm wondering if a more generic version of this applies to all extension attributes - it is completely fine if an attribute is "duplicated" in data and context.

spec.md
cathyhongzhang: I will remove this duplication in the "Extension Attributes".  I think it is better to put into the generic "Context Attribute" so as not to restrict it to extensions

spec.md
cathyhongzhang: Will remove the duplication

spec.md
cathyhongzhang: will remove the duplication

spec.md
cathyhongzhang: Agree it is completely fine if an attribute is duplicated in data and context and this applies to all attributes including extension attributes. Will use "also add" to be clearer.

spec.md
duglin: @cathyhongzhang I think it might make more sense to put this into the other section (the extension) section since this is really about adding extensions.

cathyhongzhang: I am not sure if all future new context attributes belong to extensions. If some context attribute is used by a large number of use cases, then the WG might reach consensus to add it to the main spec.md directly. So I think putting this in the generic "context attributes" section, which includes both attributes endorsed by this WG (i.e. those explicitly defined in spec.md) and those that are not endorsed, is better.  What do you think?

cathyhongzhang: Currently none is defined in the spec.md, so by default, they all belong to extensions.

duglin: I was just focused on the readability of it. This paragraph is really talking about extensions, granted its focused on "identity" extensions, but still extensions so it just felt a little out of place to have it here instead of in the extension section. Plus it then, sort of, reads like we're talking about extensions in 2 places instead of one.

In the end its not a big deal, it just felt a little odd, but I can live with it if others are ok with it.

spec.md
```diff
@@ -152,13 +152,8 @@ of a JSON serialization, the context attributes and the event data might
 both appear within the same JSON object.
 
 ### Extension Attributes
-CloudEvent producers MAY include additional extension attributes within the
-event. This enables event producers, or middleware, to include additional
-metadata that might be used for any purpose in the processing of the
-CloudEvent, such as identifying or correlating event sources. See
-[CloudEvent Attributes Extensions](primer.md#cloudevent-attribute-extensions)
```
cneijenhuis: I think the last sentence, including this link, should remain.

cathyhongzhang: Will add it back to the extension section.

spec.md
```diff
@@ -139,13 +152,10 @@ of a JSON serialization, the context attributes and the event data might
 both appear within the same JSON object.
 
 ### Extension Attributes
-CloudEvent producers MAY include additional extension attributes within the
```
duglin: I think this first sentence (in particular the "MAY") is important.  I'm not sure why this section is being reduced.

cathyhongzhang: It is removed per comment on duplication with what's newly added in previous section

duglin: I agree with removing duplicate text but not the normative statements behind them. I would suggest you try to move your new text into this section since they're both about extensions. Then we don't lose any of the normative language but add the clarifications you're looking for.

cathyhongzhang: The normative statements added in the previous generic "context attributes" section applies to extension attributes too since extension attributes are also "context attributes".  Putting it into "extension" section adds implicit restriction/assumption that any future attribute will be an extension, which is not right. We can not predict the future and should allow the possibility of some future attribute to be added as standard attribute rather than extension attribute.

duglin: @cathyhongzhang there aren't *any* normative statements in your new text though. By "normative" I mean using RFC2119 keywords.

> Putting it into "extension" section adds implicit restriction/assumption that any future attribute will be an extension...

no it doesn't. Any new attribute added to the spec will not be an extension  - it won't be since it's spec-defined. All of the attributes you're talking about in your new text are "extensions" because they are not defined by the spec, which is why it makes more sense to put the text into the "extensions" section. Its location in the spec does not change what identity properties people can add, it just make it more readable to put all extensions text into the same location.

cneijenhuis: I agree with @duglin - and I'm sorry if my review wasn't clear on that. I prefer to merge your proposed text into the extension section.

duglin:
LGTM

duglin:
Approved on the 9/13 call with some minor edits - which are now done.

Issue #167, opened 2018-04-19
Discuss renaming 'Object' to something else in our Attribute type system section


duglin:
See https://github.com/cloudevents/spec/pull/152/files#r181143546 for more info.

Assigned to @clemensv

clemensv:
Sent email to the mailing list:

https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/cncf-wg-serverless/qKW4mYNvZYo/XrPREoppBQAJ

clemensv:
Last call.

inlined:
My suggestion from the meeting still stands. I propose `any` or `var`. These should feel comfortable to JS devs and don't overload the meaning of Object, which is distinct from string or array in JS.

duglin:
Either `Any` or `Object` are ok with me.

duglin:
@clemensv any update on this one?

duglin:
During the 6/15 f2f we agreed to use "Any". Clemens to write-up a PR

PR #305, opened 2018-09-18
#167 rename Object to Any

clemensv:
Fixing #167


clemensv:
Retracting due to Git issue.

PR #310, opened 2018-09-19
Define our versioning scheme across our docs

### LINKED ISSUE #271: Lifecycle of core spec vs serialization/transport-binding specs
We need to decide the versioning scheme for our specs. Are they all totally independent or are they grouped into a single unit when considering versioning?

For example, let's say we publish v1.0 of the core, json and http-binding specs.
Then we update the http-binding spec in non-compatible way - meaning it should be bumped to v2.0.

Do we bump all specs to v2.0 or just the http-binding spec? If just the http-binding spec then how does a receiver of a CloudEvent know which version of the http-binding spec is being adhered to since the CloudEventVersion property will still be at v1.0 ?  This might lead us to group all spec together.

But, if we do that then we may be forced to bump our version number on the core spec more often than we would like to. Any breaking change to ANY of our specs would cause all specs to be bumped. Not ideal either.

However, one thing to note... while a break change does mean the major version number must be incremented, a major version change does not necessarily mean that there were any breaking semantic changes to the spec. In other words... we can choose to go from v1.0 to v2.0 any time we want even if there are NO breaking changes. This would impact implementations because it would make their version compatibility checks harder of course.

duglin:
Closes #271

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -85,7 +85,17 @@ If a vote is taken during a meeting, the follow rules will be followed:
   * Comments are welcome from non-members
   * Voting tally will reflect the above qualification and recorded in PR
 
-## Release Process
+## Release Process and Versioning
+
+The specifications produced will adhere to the following:
+* The versioning scheme used will follow [semver](https://semver.org/)
+* All normative specifications, and the Primer, will be grouped together into
+  a single logical unit and released at the same time, at the same version
+  number. This is true regardless of whether each individual document actually
+  changed during the release cycle.
+
+Note that these rules do not apply to the
+[documented extensions](documented-extensions.md).
 
 To create a new release:
 * Create a PR that modifies the [README](README.md), and all specifications


duglin:
Minor change - added the Primer to the grouping since that might change dramatically as we version the spec. Meaning, a v1.0 primer might look very different from v2.0 primer - and so it should probably be linked with the spec that its trying to help explain.

GOVERNANCE.md
```diff
@@ -85,7 +85,17 @@ If a vote is taken during a meeting, the follow rules will be followed:
   * Comments are welcome from non-members
   * Voting tally will reflect the above qualification and recorded in PR
 
-## Release Process
+## Release Process and Versioning
+
+The specifications produced will adhere to the following:
+* The versioning scheme used will follow [semver](https://semver.org/)
```
erikerikson: While we have been very thankful for and use semver, for the subspace of schemas, I've seen worthwhile semantic differences in [SchemaVer](https://github.com/snowplow/iglu/wiki/SchemaVer).  It would make this more complex but might it be useful to apply those semantics as appropriate?

duglin: I think this is the key section of SchemaVer:
- - - -
That's why we decided to introduce our own schema versioning notion: SchemaVer. SchemaVer is defined as follows: MODEL-REVISION-ADDITION
- MODEL when you make a breaking schema change which will prevent interaction with any historical data
- REVISION when you introduce a schema change which may prevent interaction with some historical data
- ADDITION when you make a schema change that is compatible with all historical data
- - - -

I don't fully understand the diff between MODEL and REVISION when, in the end, they both can introduce breaking changes. And ADDITION sounds like a SemVer "minor" update. So, this equates to just a "Major.Minor" versioning scheme to me.

erikerikson: TL;DR: it's "must" versus "can".

As I understand it, you are right given the current definition of CloudEvents...  ([the official explanation/example](https://github.com/snowplow/iglu/wiki/SchemaVer#revision))

Model is for when the schema introduces breaking changes such as field `a` in the new version must be called field `b`.  This can break current code that depends on the schema's guarantee that the field `a` will be present and is, thereby, necessarily a breaking change.

This is different from the case where one is introducing a field into a schema where the option of having other fields [was allowed](https://json-schema.org/understanding-json-schema/reference/object.html) (search for `additionalProperties`, consider `true` setting) by the previous version of the schema (i.e. the way we have defined extensions).  Because schema compliant data may exist with any attribute (e.g. `c`) not previously in the schema but also not prohibited, adding a new field (e.g. `c`) to the schema creates a circumstance where that field *could* be in conflict with existing data containing that field.  Thus, introducing a new attribute where users could previously comply with the schema but add arbitrary non-conflicting attributes creates a circumstance where additions could (but won't necessarily) cause conflicts.  Adding an attribute under such circumstances would comprise a Revision version increment under SchemaVer.

Perhaps the difference is the delta between a "certainly" (Model) and "possibly" (Revision) breaking change?  The first *breaks* schema users depending on every guarantee and the second only *contingently breaks* users depending on every guarantee.  Effectively the current rules around extensions eliminates the Addition level.

To be clear: I don't mean to say this necessarily ought to be drawn in.  It was something that immediately came to mind when you brought this PR up in last week's call.  I found that reading it put me in a better frame of mind for thinking about the meaning and consequences of schema mutations relative to semver.

duglin: yep - my reading was the MUST vs MAY (or "can") too, but when it comes to trying to guarantee interop and backwards compatibility, I think you have to assume the worst - meaning (to me) the MAY turns into a MUST. Expecting all devs to understand the nuances between them  feels like we're asking them to think too much ;-)

deissnerk:
LGTM

duglin:
Approved on 11/1 call.

PR #315, opened 2018-09-21
Very minor tweaks to README

duglin:
- Update our description since we're not really new any more
- Add a note about when we became a sandbox project, and a ref to
  the TOC minutes in which we were approved - mainly so there's a record
  of it someplace.  I've been asked for this info before and it should be more
  easily found.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,11 +15,11 @@ Enter CloudEvents, a specification for describing event data in a common way.
 CloudEvents seeks to ease event declaration and delivery across services,
 platforms and beyond.
 
-CloudEvents is a new effort and has received a surprising amount of industry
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,11 +15,12 @@ Enter CloudEvents, a specification for describing event data in a common way.
 CloudEvents seeks to ease event declaration and delivery across services,
 platforms and beyond.
 
-CloudEvents is a new effort and has received a surprising amount of industry
+CloudEvents has received a large amount of industry
```
n3wscott: please fix columns.

duglin:
Approved on the 9/27 call

PR #316, opened 2018-09-21
Minor Twist to clarify the attendance count

cathyhongzhang:
Add this clarification per last meeting discussion

### DIFF
--- a/documented-extensions.md
+++ b/documented-extensions.md
@@ -52,3 +52,4 @@ As a convention, extensions of scalar types (e.g. `String`, `Binary`, `URI`,
 
 * [Distributed Tracing](extensions/distributed-tracing.md)
 * [Sampling](extensions/sampled-rate.md)
+* [Sequence](extensions/sequence.md)
--- a/extensions/sequence.md
+++ b/extensions/sequence.md
@@ -1,0 +1,46 @@
+# Sequence
+
+This extension defines two attributes that can be included within a CloudEvent
+to describe the position of an event in the ordered sequence of events produced
+by a unique event source.
+
+The `sequence` attribute represents the value of this event's
+order in the stream of events.  The exact value and meaning of this
+attribute is defined by the `sequenceType` attribute. 
+If the `sequenceType` is missing, or not defined in this specification,
+event consumers will need to have some out-of-band communication with the
+event producer to understand how to interpret the value of the attribute.
+
+## Attributes
+
+### sequence
+* Type: `String`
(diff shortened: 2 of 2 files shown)


nerdyyatrice:
LGTM

GOVERNANCE.md
```diff
@@ -68,6 +68,8 @@ If a vote is taken during a meeting, the follow rules will be followed:
 * A participating company, or nonaffiliated individual, attains voting rights
   by having any of their assigned representative(s) attend 3 out of the last
   4 meetings. They obtain voting rights after the 3rd meeting, not during.
+  If the primary representative can not attend a meeting, his/her
+  delegate's attendance is counted.
```
duglin: Cathy - I agree with the idea but its kind of awkward to have this sentence separate from the one on line 68.
Would you be ok with this instead:
A participating company, or non-affiliated individual, attains voting rights
by having their assigned representative(s), in aggregate, attend 3 out of
the last 4 meetings.

GOVERNANCE.md
```diff
@@ -67,7 +67,9 @@ If a vote is taken during a meeting, the follow rules will be followed:
 * Each participating company can assign a primary and secondary representative.
 * A participating company, or nonaffiliated individual, attains voting rights
   by having any of their assigned representative(s) attend 3 out of the last
-  4 meetings. They obtain voting rights after the 3rd meeting, not during.
+  4 meetings. If the primary representative can not attend a meeting, his/her
+  delegate's attendance is counted. They obtain voting rights
```
n3wscott: please use 80 width columns.

duglin:
All - as @cathyhongzhang and I were making the edits based on this week's call, we realized that we were each looking at the edits as solving different issues:
1 - make it clear that either the primary or secondary person's presence at a call would count for that week's attendance 
2 - companies can change who their representatives as needed, in particular if both primary and secondary are not available

Both of these issues should be cleared up but the previous proposal wasn't clear as to which is was addressing. We worked on some new text to address both - I don't think it changes anything about our current process, it just documents it. While we can't totally avoid game playing, to avoid a company sending different people to the calls each week **just** to keep voting rights, we limit the rotation of people down to once a month - which isn't an issue since most companies have rarely changed this info in the past.

While I don't think the new edits are controversial, I didn't want to merge the edits w/o the group getting a chance to look them over since any edits to a governance type of doc should be well vetted. I've added this item to the agenda for next week's call - please review and speak-up prior to the call if you have any concerns or suggested edits.

PR #322, opened 2018-10-02
Encoding exceptions

clemensv:

Striking one of my homework AIs from the list, this provides an “escape hatch” in our event format and transport specs to allow attributes of extension specs to override the general serialization rules and metadata mappings.

This PR covers the JSON and AMQP event format specs and MQTT/AMQP/HTTP transports (NATS is not affected) and allows for diverging encoding rules defined by an extension _IF_ that extension defines encoding rules for all core formats.

### DIFF
--- a/amqp-format.md
+++ b/amqp-format.md
@@ -50,7 +50,8 @@ system, which this mapping leans on.
 
 ### 2.2. Type System Mapping
 
-The CloudEvents type system is mapped to AMQP types as follows:
+The CloudEvents type system MUST be mapped to AMQP types as follows,
+with exceptions noted below.
 
 | CloudEvents | AMQP
 |-------------|-------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


duglin:
@clemensv does this PR overlap with the text we have in the extensions doc where it says:
- - -
Transport bindings (e.g. HTTP, MQTT, AMPQ, NATS) provide default placement for extensions, but an extension MAY require special representation when transported (e.g. tracing standards that require specific headers). Extension authors SHOULD only require special representation in transport bindings where extensions integrate with pre-existing specs; extensions with custom transport bindings are much more likely to be dropped by middleware that does not understand the extension.
- - -
I don't think there's a conflict, but I'm wondering if we could (should?) cover this "escape hatch" just once in a common location and not have to repeat it for each transport spec.

duglin:
Approved on the Oct 11th call.

Issue #332, opened 2018-10-23
The serialized JSON Example do not pass against the schema


fabiojose:
- CloudEvents JSON Example: https://github.com/cloudevents/spec/blob/master/spec.md#example
- CloudEvents JSON Schema: https://github.com/cloudevents/spec/blob/master/spec.json

Then validation against the schema produces the following error:

```js
[ { keyword: 'format',
    dataPath: '.source',
    schemaPath: '#/definitions/source/format',
    params: { format: 'uri' },
    message: 'should match format "uri"' } ]
```
This is because of `source` value is `/mycontext`, that is an invalid URI.

A valid URI must have this format: `URI = scheme ":" hier-part [ "?" query ] [ "#" fragment ]`. Checkout here: https://www.ietf.org/rfc/rfc3986.txt

According to [RFC 3986, Section 3](https://tools.ietf.org/html/rfc3986#section-3), the `scheme` and `hier-part` are required.

Then, to work properly against the schema validation, I suggest an edit in the serialized JSON example, in `source` value example, changing it to `from:/mycontext` adding the scheme `from:`, as follows:

```json
{
  "cloudEventsVersion":"0.1",
  "eventType":"com.example.someevent",
  "source":"from:/mycontext",
  "eventID":"A234-1234-1234",
  "eventTime":"2018-04-05T17:31:00Z",
  "comExampleExtension1":"value",
  "comExampleExtension2":{
    "otherValue":5
  },
  "contentType":"text/xml",
  "data":"<much wow=\"xml\"/>"
}
```

Or an adjustment in JSON Schema in `source` definition, changing it to another type instead of `URI`.

cneijenhuis:
One could also change the spec to say `URI Reference` (see https://tools.ietf.org/html/rfc3986#section-4.1), which includes relative references such as `/mycontext` (I still think we can improve on that example).

This would also be more inline with current implementations, e.g. https://docs.microsoft.com/en-us/azure/event-grid/cloudevents-schema#cloudevent-schema

I think I'd prefer to force the full URI, though. In many cases, that would map to a URL that can be explored, whether it is a website, a REST API, ...

fabiojose:
@cneijenhuis, personally I'd prefer the full URI too. The problem with relatives references is that require some custom JSON Schema validator, this is common for browsers, but not so common in programming libraries. All of them treats `uri` type as full URI.

I think the best choice is to do some improvements on that examples. I will open a PR!

[...]

PR #333, opened 2018-10-24
Update spec.md

fabiojose:
Some improvements in the event serialized as JSON.

The proposed updates are based in the issue [#332](https://github.com/cloudevents/spec/issues/332).

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -271,7 +271,7 @@ encapsulated within the `data` attribute.
 * Constraints:
   * OPTIONAL
 
-# Examples
+# Example
 
 The following examples shows CloudEvent serialized as JSON.
 
--- a/spec.md
+++ b/spec.md
@@ -275,48 +275,6 @@ encapsulated within the `data` attribute.
 
 The following examples shows CloudEvent serialized as JSON.
 
-## Using relative URI in `source` element
-
-As defined in [RFC 3986, Section 4.2](https://tools.ietf.org/html/rfc3986#section-4.2) it is possible to use relative paths to express URI, but this require the apply of resolution algorithm [RFC 3986, Section 5](https://tools.ietf.org/html/rfc3986#section-5).
-
-``` JSON
-{
-    "cloudEventsVersion" : "0.1",
-    "eventType" : "com.github.pull.create",
-    "source" : "/cloudevents/spec/pull/123",
-    "eventID" : "A234-1234-1234",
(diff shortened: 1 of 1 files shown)


duglin:
@fabiojose  please sign your commits

spec.md
```diff
@@ -273,13 +273,17 @@ encapsulated within the `data` attribute.
 
 # Example
 
-The following example shows a CloudEvent serialized as JSON:
+The following examples shows CloudEvent serialized as JSON.
+
+## Using relative URI in `source` element
+
+As defined in [RFC 3986, Section 4.2](https://tools.ietf.org/html/rfc3986#section-4.2) it is possible to use relative paths to express URI, but this require the apply of resolution algorithm [RFC 3986, Section 5](https://tools.ietf.org/html/rfc3986#section-5).
```
duglin: can you please wrap this at 80 columns to align with our other docs?

spec.md
```diff
@@ -290,3 +294,43 @@ The following example shows a CloudEvent serialized as JSON:
     "data" : "<much wow=\"xml\"/>"
 }
 ```
+
+Regarding the resolution algorithm, lets consider:
```
duglin: This section confuses me. It sounds a bit like text that should be in the main part of the spec, not in an example section, because its trying to explain how to resolve relative URIs - which feels like normative language.  But, it's also just reiterating what RFC3986 says - so why do we need to repeat it?

Also, where is "base URI" coming from?  We don't have that defined as part of the spec. 

Perhaps it would help me if you could explain what you're trying to fix, or clarify, in the spec with this PR.

fabiojose: The PR is based on the issue [332](https://github.com/cloudevents/spec/issues/332) opened by. The [example](https://github.com/cloudevents/spec/blob/master/spec.md#example) of serialized JSON is not valid against the [JSON Schema](https://github.com/cloudevents/spec/blob/master/spec.json), because of `source` element value, which is `/mycontext`.

duglin: Yup - but I think an easier solution is to just change the example to use a real URI or to make the type of "source" into a URI-reference.  Getting into URI resolution seems unnecessary to me.

fabiojose: I totally agree with you @duglin!

I will refactor the PR!

fabiojose:
Closing this to open another one with properly changes.

PR #335, opened 2018-10-30
Add some newbie guidance

### LINKED ISSUE #267: Split Core CloudEvents spec from bindings
This is more of an organizational thing from an outsider. The repository has a bunch of spec documents in the root of the tree, it was difficult for me to find what I need to read to implement something to generate or consume CloudEvents.

As the specification grows, bindings to different transports will become more and more a part of the spec. Maybe those should be logically split from the core CloudEvents spec?

Just a thought.

duglin:
Closes #267 


nerdyyatrice:
LGTM

duglin:
Approved on the Nov 8 call

PR #337, opened 2018-11-01
Kafka transport binding v2

bluemonk3y:
Kafka transport binding for CloudEvents, similar to the HTTP binding and proposed NATS, MQTT, AMQP bindings.

Follow-on PR from https://github.com/cloudevents/spec/pull/300

Signed-off-by: Neil Avery neil@confluent.io

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ The following documents are available:
 | AMQP Transport Binding        | [v0.3](https://github.com/cloudevents/spec/blob/v0.3/amqp-transport-binding.md) | [master](https://github.com/cloudevents/spec/blob/master/amqp-transport-binding.md) |
 | HTTP Transport Binding        | [v0.3](https://github.com/cloudevents/spec/blob/v0.3/http-transport-binding.md) | [master](https://github.com/cloudevents/spec/blob/master/http-transport-binding.md) |
 | JSON Event Format             |      [v0.3](https://github.com/cloudevents/spec/blob/v0.3/json-format.md)       |      [master](https://github.com/cloudevents/spec/blob/master/json-format.md)       |
+| Kafka Transport Binding       | - | [master](https://github.com/cloudevents/spec/blob/master/kafka-transport-binding.md) |
 | MQTT Transport Binding        | [v0.3](https://github.com/cloudevents/spec/blob/v0.3/mqtt-transport-binding.md) | [master](https://github.com/cloudevents/spec/blob/master/mqtt-transport-binding.md) |
 | NATS Transport Binding        | [v0.3](https://github.com/cloudevents/spec/blob/v0.3/nats-transport-binding.md) | [master](https://github.com/cloudevents/spec/blob/master/nats-transport-binding.md) |
 | Protocol Buffers Event Format | [v0.3](https://github.com/cloudevents/spec/blob/v0.3/protobuf-format.md)        | [master](https://github.com/cloudevents/spec/blob/master/protobuf-format.md)        |
--- a/kafka-transport-binding.md
+++ b/kafka-transport-binding.md
@@ -54,7 +54,7 @@ The *binary* mode *only* applies to Kafka 0.11.0.0 and above, because Kafka
 
 In the *binary* content mode, the value of the event `data` attribute MUST be
 placed into the Kafka message's value section as-is, with the
-`ce_contentType` header value declaring its media type; all other
+`ce_datacontenttype` header value declaring its media type; all other
 event attributes MUST be mapped to the Kafka message's 
 [header section][Kafka-Message-Header].
 
@@ -83,13 +83,13 @@ attributes.
 ### 2.1. data Attribute
 
 The `data` attribute is assumed to contain opaque application data that is
-encoded as declared by the `contentType` attribute.
+encoded as declared by the `datacontenttype` attribute.
 
 An application is free to hold the information in any in-memory representation
 of its choosing, but as the value is transposed into Kafka as defined in this
 specification, core Kafka provides data available as a sequence of bytes.
 
-For instance, if the declared `contentType` is
+For instance, if the declared `datacontenttype` is
 `application/json;charset=utf-8`, the expectation is that the `data` attribute
 value is made available as [UTF-8][RFC3629] encoded JSON text.
 
@@ -101,7 +101,7 @@ particular content mode might be defined by an application, but are not defined
 here.
 
 The receiver of the event can distinguish between the two content modes by
-inspecting the `ce_contentType` [Header][Kafka-Message-Header] of the 
+inspecting the `ce_datacontenttype` [Header][Kafka-Message-Header] of the 
 Kafka message. If the value is prefixed with the CloudEvents media type
 `application/cloudevents`, indicating the use of a known
 [event format](#14-event-formats), the receiver uses *structured* mode, otherwise
@@ -125,8 +125,8 @@ efficient transfer and without transcoding effort.
 
 #### 3.2.1. Content Type
 
-For the *binary* mode, the header `ce_contentType` property MUST be
-mapped directly to the CloudEvents `contentType` attribute.
+For the *binary* mode, the header `ce_datacontenttype` property MUST be
+mapped directly to the CloudEvents `datacontenttype` attribute.
 
 
 #### 3.2.2. Event Data Encoding
@@ -202,7 +202,7 @@ hops, and across multiple transports.
 
 #### 3.3.1. Kafka Content-Type
 
-The [Kafka `ce_contentType`] property field MUST be set to the media
+The [Kafka `content-type`] property field MUST be set to the media
 type of an [event format](#14-event-formats).
 
 Example for the [JSON format][JSON-format]:
@@ -240,7 +240,7 @@ Key: mykey
 
 ------------------ headers -------------------
 
-ce_contentType: application/cloudevents+json; charset=UTF-8
+content-type: application/cloudevents+json; charset=UTF-8
 
 ------------------- value --------------------
 
--- a/kafka-transport-binding.md
+++ b/kafka-transport-binding.md
@@ -1,0 +1,281 @@
+# Kafka Transport Binding for CloudEvents - Version 0.4-wip
+
+## Abstract
+
+The [Kafka][Kafka] Transport Binding for CloudEvents defines how events are
+mapped to [Kafka messages][Kafka-Message-Format].
+
+## Status of this document
+
+This document is a working draft.
+
+## Table of Contents
+
+1. [Introduction](#1-introduction)
+- 1.1. [Conformance](#11-conformance)
+- 1.2. [Relation to Kafka](#12-relation-to-kafka)
+- 1.3. [Content Modes](#13-content-modes)
+- 1.4. [Event Formats](#14-event-formats)
+- 1.5. [Security](#15-security)
+2. [Use of CloudEvents Attributes](#2-use-of-cloudevents-attributes)
+- 2.1. [data Attribute](#21-data-attribute)
+3. [Kafka Message Mapping](#3-kafka-message-mapping)
+- 3.1. [Key Attribute](#31-key-attribute)
+- 3.2. [Binary Content Mode](#32-binary-content-mode)
+- 3.3. [Structured Content Mode](#33-structured-content-mode)
+4. [References](#4-references)
+
+## 1. Introduction
+
+[CloudEvents][CE] is a standardized and transport-neutral definition of the
+structure and metadata description of events. This specification defines how
+the elements defined in the CloudEvents specification are to be used in the
+Kafka protocol as [Kafka messages][Kafka-Message-Format] (aka Kafka records).
+
+### 1.1. Conformance
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in [RFC2119][RFC2119].
+
+### 1.2. Relation to Kafka
+
+This specification does not prescribe rules constraining transfer or settlement
+of event messages with Kafka; it solely defines how CloudEvents are expressed
+in the Kafka protocol as [Kafka messages][Kafka-Message-Format].
+
+### 1.3. Content Modes
+
+The specification defines two content modes for transferring events:
+*structured* and *binary*.
+
+The *binary* mode *only* applies to Kafka 0.11.0.0 and above, because Kafka
+0.10.x.x and below lack support for message level headers.
+
+In the *binary* content mode, the value of the event `data` attribute MUST be
+placed into the Kafka message's value section as-is, with the
+`ce_contentType` header value declaring its media type; all other
+event attributes MUST be mapped to the Kafka message's 
+[header section][Kafka-Message-Header].
+
+In the *structured* content mode, event metadata attributes and event data are
+placed into the Kafka message value section
+using an [event format](#14-event-formats).
+
+
+
+### 1.4. Event Formats
+
+Event formats, used with the *structured* content mode, define how an event is
+expressed in a particular data format. All implementations of this
+specification MUST support the [JSON event format][JSON-format].
+
+### 1.5. Security
+
+This specification does not introduce any new security features for Kafka, or
+mandate specific existing features to be used.
+
+## 2. Use of CloudEvents Attributes
+
+This specification does not further define any of the [CloudEvents][CE] event
+attributes.
+
+### 2.1. data Attribute
+
+The `data` attribute is assumed to contain opaque application data that is
+encoded as declared by the `contentType` attribute.
+
+An application is free to hold the information in any in-memory representation
+of its choosing, but as the value is transposed into Kafka as defined in this
+specification, core Kafka provides data available as a sequence of bytes.
+
+For instance, if the declared `contentType` is
+`application/json;charset=utf-8`, the expectation is that the `data` attribute
+value is made available as [UTF-8][RFC3629] encoded JSON text.
+
+## 3. Kafka Message Mapping
+
+With Kafka 0.11.0.0 and above, the content mode is chosen by the sender of the 
+event. Protocol usage patterns that might allow solicitation of events using a
+particular content mode might be defined by an application, but are not defined
+here.
+
+The receiver of the event can distinguish between the two content modes by
+inspecting the `ce_contentType` [Header][Kafka-Message-Header] of the 
+Kafka message. If the value is prefixed with the CloudEvents media type
+`application/cloudevents`, indicating the use of a known
+[event format](#14-event-formats), the receiver uses *structured* mode, otherwise
+it defaults to *binary* mode.
+
+If a receiver finds a CloudEvents media type as per the above rule, but with an
+event format that it cannot handle, for instance
+`application/cloudevents+avro`, it MAY still treat the event as binary and
+forward it to another party as-is.
+
+### 3.1. Key Attribute
+The 'key' attribute is populated by a partitionKeyExtractor function. The 
+partitionKeyExtractor is a transport specific function that contains bespoke logic 
+to extract and populate the value. A default implementation of the extractor will 
+use the [Partitioning](extensions/partitioning.md) extension value. 
+
+### 3.2. Binary Content Mode
+
+The *binary* content mode accommodates any shape of event data, and allows for
+efficient transfer and without transcoding effort.
+
+#### 3.2.1. Content Type
+
+For the *binary* mode, the header `ce_contentType` property MUST be
+mapped directly to the CloudEvents `contentType` attribute.
+
+
+#### 3.2.2. Event Data Encoding
+
+The [`data` attribute](#21-data-attribute) byte-sequence MUST be used as the
+value of the Kafka message.
+
+#### 3.2.3. Metadata Headers
+
+All [CloudEvents][CE] attributes and
+[CloudEvent Attributes Extensions](primer.md#cloudevent-attribute-extensions) 
+with exception of `data` MUST be individually mapped to and from the Header 
+fields in the Kafka message.
+
+##### 3.2.3.1 Property Names
+
+CloudEvent attributes are prefixed with "ce_" for use in the
+[message-headers][Kafka-Message-Header] section.
+
+Examples:
+
+    * `time` maps to `ce_time`
+    * `id` maps to `ce_id`
+    * `specversion` maps to `ce_specversion`
+
+##### 3.2.4.2 Property Values
+
+The value for each Kafka header is constructed from the respective
+header's Kafka representation, compliant with the [Kafka message
+format][Kafka-Message-Format] specification.
+
+
+#### 3.2.5 Example
+
+This example shows the *binary* mode mapping of an event into the
+Kafka message. All other CloudEvents attributes
+are mapped to Kafka Header fields with prefix `ce_`.
+
+Mind that `ce_` here does refer to the event `data`
+content carried in the payload.
+
+``` text
+------------------ Message -------------------
+
+Topic Name: mytopic
+
+------------------- key ----------------------
+
+Key: mykey
+
+------------------ headers -------------------
+
+ce_specversion: "0.4-wip"
+ce_type: "com.example.someevent"
+ce_source: "/mycontext/subcontext"
+ce_id: "1234-1234-1234"
+ce_time: "2018-04-05T03:56:24Z"
+ce_datacontenttype: application/avro
+       .... further attributes ...
+
+------------------- value --------------------
+
+            ... application data ...
+
+-----------------------------------------------
+```
+
+### 3.3. Structured Content Mode
+
+The *structured* content mode keeps event metadata and data together in the
+payload, allowing simple forwarding of the same event across multiple routing
+hops, and across multiple transports.
+
+#### 3.3.1. Kafka Content-Type
+
+The [Kafka `ce_contentType`] property field MUST be set to the media
+type of an [event format](#14-event-formats).
+
+Example for the [JSON format][JSON-format]:
+
+``` text
+content-type: application/cloudevents+json; charset=UTF-8
+```
+
+#### 3.3.2. Event Data Encoding
+
+The chosen [event format](#14-event-formats) defines how all attributes,
+including the `data` attribute, are represented.
+
+The event metadata and data are then rendered in accordance with the [event
+format](#14-event-formats) specification and the resulting data becomes the
+Kafka application [data][data] section.
+
+#### 3.3.3. Metadata Headers
+
+Implementations MAY include the same Kafka headers as defined for the
+[binary mode](#32-binary-content-mode).
+
+#### 3.3.4 Example
+
+This example shows a JSON event format encoded event:
+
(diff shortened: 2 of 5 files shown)


duglin:
CI error is ok since it'll be automagically fixed once merged. But the DCO issue is real.

kafka-transport-binding.md
```diff
@@ -0,0 +1,274 @@
...
+
+This specification does not introduce any new security features for Kafka, or
+mandate specific existing features to be used.
+
+## 2. Use of CloudEvents Attributes
+
+This specification does not further define any of the [CloudEvents][CE] event
+attributes.
+
+### 2.1. contentType Attribute
```
duglin: Do we need this section since it doesn't really mandate that anything special happen for contentType?

kafka-transport-binding.md
```diff
@@ -0,0 +1,274 @@
...
+
+The `contentType` attribute is assumed to contain a media-type expression
```
duglin: Is there anything special to do if contentType is missing - since its OPTIONAL?

kafka-transport-binding.md
```diff
@@ -0,0 +1,274 @@
...
+## 3. Kafka Message Mapping
+
+With Kafka 0.11.0.0 and above, the content mode is chosen by the sender of the 
+event. Protocol usage patterns that might allow solicitation of events using a
+particular content mode might be defined by an application, but are not defined
+here.
+
+The receiver of the event can distinguish between the two content modes by
+inspecting the `cloudEvents_contentType` property of the Kafka message. If the
+value is prefixed with the CloudEvents media type `application/cloudevents`, indicating the use of a known [event
```
duglin: wrap at 80

kafka-transport-binding.md
```diff
@@ -0,0 +1,274 @@
...
+mapped directly to the CloudEvents `contentType` attribute.
+
+#### 3.1.2. Event Data Encoding
+
+The [`data` attribute](#22-data-attribute) byte-sequence MUST be used as the
+value of the Kafka message.
+
+#### 3.1.3. Metadata Headers
+
+All [CloudEvents][CE] attributes with exception of `data` MUST be individually
```
duglin: should we include extensions in this too?

kafka-transport-binding.md
bluemonk3y: The simplest path is to use struct/json like NATs does. However, Kafka doesn't presume the payload format. We could use STRUCT/JSON and Binary/AVRO but this is likely to be too restrictive for many users. CloudEvents_contentType could be used specify something along the lines of `application/cloudevents+proto` or other encodings, or security or other information.

bluemonk3y:
> CI error is ok since it'll be automagically fixed once merged. But the DCO issue is real.

@duglin  - yep I will fix that tomorrow

kafka-transport-binding.md
bluemonk3y: yep

kafka-transport-binding.md
```diff
@@ -0,0 +1,271 @@
...
+#### 3.1.3. Metadata Headers
+
+All [CloudEvents][CE] attributes and
+[CloudEvent Attributes Extensions](primer.md#cloudevent-attribute-extensions) 
+with exception of `data` MUST be individually mapped to and from the Header 
+fields in the Kafka message.
+
+##### 3.1.3.1 Property Names
+
+Cloud Event attributes are prefixed with "cloudEvents_" for use in the
```
matzew: `cloudEvents_` ? instead of "cloudEvents_" ?

matzew: I wonder if we should use some sort of fqn:   `io.cloudevents.` to avoid potential clashes ?

kafka-transport-binding.md
```diff
@@ -0,0 +1,271 @@
...
+This specification does not prescribe rules constraining transfer or settlement
+of event messages with Kafka; it solely defines how CloudEvents are expressed
+in the Kafka protocol as [Kafka messages][Kafka-Message-Format].
+
+### 1.3. Content Modes
+
+The specification defines two content modes for transferring events:
+*structured* and *binary*.
+
+The *binary* mode *only* applies to Kafka 0.11.0.0 and above, because of Kafka
```
gunnarmorling: The second part sounds a bit odd, perhaps:

> because Kafka 0.10.x.x and below don't support message level headers

kafka-transport-binding.md
```diff
@@ -0,0 +1,271 @@
...
+
+In the *binary* content mode, the value of the event `data` attribute MUST be
+placed into the Kafka message's value section as-is, with the
+`cloudEvents_contentType` header value declaring its media type; all other
+event attributes MUST be mapped to the Kafka message's 
+[header section][Kafka-Message-Header].
+
+### 1.4. Event Formats
+
+Event formats, used with the *stuctured* content mode, define how an event is
```
gunnarmorling: stuctured=structured

kafka-transport-binding.md
```diff
@@ -0,0 +1,271 @@
...
+##### 3.1.3.1 Property Names
+
+Cloud Event attributes are prefixed with "cloudEvents_" for use in the
+[message-headers][Kafka-Message-Header] section.
+
+Examples:
+
+    * `eventTime` maps to `cloudEvents_eventTime`
+    * `eventID` maps to `cloudEvents_eventID`
+    * `cloudEventsVersion` maps to `cloudEvents_cloudEventsVersion`
```
matzew: +1 on the general concept of this.

Our group has been doing exactly the same:


Ok, biggest diff is ours is on older internal `RecordHeader(s)` API, while now it should be (speaking in java terms be done with `org.apache.kafka.common.header.Header(s)` API

matzew:
@bluemonk3y we have some POC code for this. and it would be nice to get this into the Cloudevents-sdk for java.

See my proposal here:
https://github.com/cloudevents/sdk-java/issues/5

looking forward working with you on that

kafka-transport-binding.md
```diff
@@ -0,0 +1,271 @@
...
+
+## 3. Kafka Message Mapping
+
+With Kafka 0.11.0.0 and above, the content mode is chosen by the sender of the 
+event. Protocol usage patterns that might allow solicitation of events using a
+particular content mode might be defined by an application, but are not defined
+here.
+
+The receiver of the event can distinguish between the two content modes by
+inspecting the `cloudEvents_contentType` property of the Kafka message. If the
```
gunnarmorling: s/property/header property/?

matzew: header key ? :)

kafka-transport-binding.md
```diff
@@ -0,0 +1,271 @@
...
+content carried in the payload.
+
+``` text
+------------------ Message -------------------
+
+Topic Name: mytopic
+
+------------------- key ----------------------
+
+Key: mykey
```
gunnarmorling: The Kafka message key is where things are still a bit blurry to me. How will its value be obtained from the CloudEvents event? Have you foreseen some extractor function or similar?

kafka-transport-binding.md
bluemonk3y: We are using cloudEvents_ as it has been generally adopted by other transport bindings. I believe we also need to avoid '.' characters. We should be consistent across all bindings.
@duglin  - your thoughts on this?

kafka-transport-binding.md
```diff
@@ -0,0 +1,269 @@
...
+mapped directly to the CloudEvents `contentType` attribute.
+
+#### 3.1.2. Event Data Encoding
+
+The [`data` attribute](#22-data-attribute) byte-sequence MUST be used as the
+value of the Kafka message.
+
+#### 3.1.3. Metadata Headers
+
+All [CloudEvents][CE] attributes and [CloudEvent Attributes Extensions](primer.md#cloudevent-attribute-extensions) with exception of `data` MUST be individually
```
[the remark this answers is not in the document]
bluemonk3y: done

kafka-transport-binding.md
bluemonk3y: lol - true

kafka-transport-binding.md
duglin: I like consistency, but I think we need to be consistent with that users of the transport expect. I did a quick search and came across this: https://streamsets.com/documentation/controlhub/3.5.0/help/pdesigner/datacollector/UserGuide/Origins/KConsumer.html
and it talks about things like `ssl.truststore.location`  and `com.sun.security.auth.module.Krb5LoginModule`.  So Kafka does seem to be ok with '.'.
I don't know enough about Kafka to know if FQNs would be more "expected" or not, I'll leave that to others - I'd be ok with either.

kafka-transport-binding.md
gunnarmorling: Thinking more about that one, perhaps a header key should be defined whose value, if present, will be propagated as key of the Kafka message, e.g. `cloudEvents_messageKey`. Alternatively, the binding, wherever it's running, could be configured with some kind of path expression which gets applied to the `data` value to extract the value, e.g. `/customer/id`.

gunnarmorling: Ah yes, thanks. I've commented over there, too. Seems that'd need resolution first.

kafka-transport-binding.md
bluemonk3y: @clemensv - what is your opinion on this (above)? Do we adopt language packaging/namespace semantics like io.cloudevents or cloudEvents_  It does raise the question about event propagation between transports.

kafka-transport-binding.md
clemensv: The conflict we have here is that the partition key is a transport specific concern (and the requirements may vary across different transport options) and that the publisher ought not to care about what transport the event route gets bound to. We may also have the situation that an event gets first published via MQTT from a little device and then gets put on Kafka by a device gateway. 

I think this binding needs to define a rule by which a key is constructed from the event rather than expecting that the event brings it along.

kafka-transport-binding.md
gwenshap: Just for the record, Apache Kafka never managed to decide on whether "." or "_" are preferred as separators. Configuration parameters are always "." separated (as the Streamset doc shows), but topics are often namespaced with "_" (for example, the internal "__consumer_offsets" topic). I've definitely seen both used in "the wild" (a fact that was sometimes inconvenient as the project had to deal with converting topic names to metric names...)

Either separator will be acceptable by the Apache Kafka community (or rather, either will be objected to, by different sets of people).

kafka-transport-binding.md
gwenshap: I'm not sure if `source` is the right place - I'd expect all events produced by same application to have identical source, why partitioning typically applies to events produced by same source. 

I like the proposal from @clemensv in #218 : 
" instead of putting the burden of producing an event key on the client and then only having one, any transport that requires particular constructs such as keys such define a mechanism by which you can harvest/synthesize a key from an incoming CloudEvent as some sort of transform. The spec doesn't need to prescribe how -- it just needs to say that that's how the key materializes. A transform that just cooks up a random key might also be valid if that's what you want."

KafkaConnect already has "key extraction" transformation, exactly because external records require mapping to Kafka keys, and the logic for doing so varies between use-cases.

duglin: @gwenshap @clemensv that would mean that the kafka transport spec would leave the determination of the "key" as an exercise for the implementer (or admin configuring the transport) ?

matzew:
> I think we'll end up collapsing them to "ce" across the board.

+1

kafka-transport-binding.md
```diff
@@ -0,0 +1,272 @@
...
+
+Topic Name: mytopic
+
+------------------- key ----------------------
+
+Key: mykey
+
+------------------ headers -------------------
+
+cloudEvents_contentType: application/avro
```
kevinbader: should be `avro/binary`, see http://avro.apache.org/docs/current/spec.html#HTTP+as+Transport

JemDay:
It seems this PR is a bit behind the 0.2 spec version ...

kafka-transport-binding.md
gunnarmorling: Just added a comment on this over at https://github.com/cloudevents/spec/pull/218. The idea being that instead of having fully pluggable logic for retrieving the key instead there could be a mechanism to just select the key from a given event field when sending a cloud event to a Kafka Topic.

duglin:
ping @hschaffner  - please review and add any comments w.r.t. why the partitionkey extension should not be needed.

ping @clemensv - please add comments about the call-back mechanism you mentioned in PR #429 

@bluemonk3y now that #429 is merged, can you rebase this one to deal with the merge conflicts? Also, you may need to edit some of the text to deal with recent spec changes - such as the names of the attributes being changed (or lower-cased).

thanks everyone for your patience.

bluemonk3y:
@duglin  - yep as soon as I get a chance.

duglin:
@bluemonk3y sorry to pester :-)  but any chance of getting an update of this one?  I get the sense we're nearing a pretty big milestone and this is one of the bigger outstanding items for us.

bluemonk3y:
@duglin - sure, I'm currently away but will pick it up on Monday.

gunnarmorling:
> now that #429 is merged

@duglin, I reckon you mean #218?

Could you (or someone else) perhaps give a summary of the latest state of the discussion? The closed PR #218 adds the `partitionkey` attribute, but it actually might be removed again as per #430? 

What @clemensv describes in https://github.com/cloudevents/spec/pull/218#issuecomment-490521883 makes sense to me: make this a concern of the transport binding which would allow for a callback function for retrieving the partition/message key on an per-event basis. The spec-defined `partitionkey` attribute would still remain in place as basis for the default behavior in case no custom callback has been configured for the transport and a value can be defined by the producer.

One open question to me is how the callback would be defined. What exactly is it? E.g. a (Java) class implementing a certain interface for extracting the key based on an incoming event? Or would at the spec level just be defined that there is *a* callback mechanism but it'd then be up to specific implementations of this to define the concrete mechanism?

Thanks!

duglin:
@bluemonk3y the travis errors seem real.  Also, can you sign your commits?

kafka-transport-binding.md
bluemonk3y: done

duglin:
Even though I didn't tag this as "v1.0" (it's "try-for-v1.0"), it's been lingering for a while so it'll be first on the PR review list this week - please look it over when you get a chance.

kafka-transport-binding.md
```diff
@@ -0,0 +1,278 @@
...
+The value for each Kafka header is constructed from the respective
+header's Kafka representation, compliant with the [Kafka message
+format][Kafka-Message-Format] specification.
+
+
+#### 3.2.5 Example
+
+This example shows the *binary* mode mapping of an event into the
+Kafka message. All other CloudEvents attributes
+are mapped to Kafka Header fields with prefix `cloudEvents_`.
```
clemensv: ce_ might just be sufficient.

kafka-transport-binding.md
```diff
@@ -0,0 +1,278 @@
...
+fields in the Kafka message.
+
+##### 3.2.3.1 Property Names
+
+CloudEvent attributes are prefixed with "cloudEvents_" for use in the
+[message-headers][Kafka-Message-Header] section.
+
+Examples:
+
+    * `eventTime` maps to `cloudEvents_eventTime`
```
clemensv: needs update

kafka-transport-binding.md
```diff
@@ -0,0 +1,278 @@
...
+
+#### 3.3.1. Kafka Content-Type
+
+The [Kafka `cloudEvent_contentType`] property field MUST be set to the media
+type of an [event format](#14-event-formats).
+
+Example for the [JSON format][JSON-format]:
+
+``` text
+content-type: application/cloudevents+json; charset=UTF-8
```
clemensv: This shows no prefix

kafka-transport-binding.md
```diff
@@ -0,0 +1,278 @@
...
+*structured* and *binary*.
+
+The *binary* mode *only* applies to Kafka 0.11.0.0 and above, because Kafka
+0.10.x.x and below lack support for message level headers.
+
+In the *structured* content mode, event metadata attributes and event data are
+placed into the Kafka message value section
+using an [event format](#14-event-formats).
+
+In the *binary* content mode, the value of the event `data` attribute MUST be
```
duglin: would it make sense to move this paragraph up so it's next to the one on line 52, that also talks about "binary" mode?

kafka-transport-binding.md
duglin: s/eventTime/time/

kafka-transport-binding.md
```diff
@@ -0,0 +1,278 @@
...
+
+##### 3.2.3.1 Property Names
+
+CloudEvent attributes are prefixed with "cloudEvents_" for use in the
+[message-headers][Kafka-Message-Header] section.
+
+Examples:
+
+    * `eventTime` maps to `cloudEvents_eventTime`
+    * `eventID` maps to `cloudEvents_eventID`
```
duglin: s/eventID/id/

kafka-transport-binding.md
duglin: s/cloudEventsVersion/specversion/

duglin:
@hschaffner did you have any comments on this one? In particular around the `key` stuff

duglin:
@bluemonk3y rebase needed - and there are some outstanding comments

kafka-transport-binding.md
```diff
@@ -134,20 +137,20 @@ value of the Kafka message.
...
 [message-headers][Kafka-Message-Header] section.
 
 Examples:
 
-    * `eventTime` maps to `cloudEvents_eventTime`
-    * `eventID` maps to `cloudEvents_eventID`
-    * `cloudEventsVersion` maps to `cloudEvents_cloudEventsVersion`
+    * `time` maps to `ce_time`
+    * `id` maps to `ce_id`
+    * `specVersion` maps to `ce_specVersion`
```
duglin: s/specVersion/specversion/

kafka-transport-binding.md
```diff
@@ -176,12 +179,12 @@ Key: mykey
 
 ------------------ headers -------------------
 
-cloudEvents_contentType: application/avro
-cloudEvents_cloudEventsVersion: "0.1"
-cloudEvents_eventType: "com.example.someevent"
-cloudEvents_eventTime: "2018-04-05T03:56:24Z"
-cloudEvents_eventID: "1234-1234-1234"
-cloudEvents_source: "/mycontext/subcontext"
+ce_specVersion: "0.1"
```
duglin: s/specVersion/specversion/

duglin: and "0.3"

README.md
```diff
@@ -37,6 +37,7 @@ The following documents are available:
 | NATS Transport Binding        | [v0.2](https://github.com/cloudevents/spec/blob/v0.2/nats-transport-binding.md) | [master](https://github.com/cloudevents/spec/blob/master/nats-transport-binding.md) |
 | AMQP Event Format             |      [v0.2](https://github.com/cloudevents/spec/blob/v0.2/amqp-format.md)       |      [master](https://github.com/cloudevents/spec/blob/master/amqp-format.md)       |
 | AMQP Transport Binding        | [v0.2](https://github.com/cloudevents/spec/blob/v0.2/amqp-transport-binding.md) | [master](https://github.com/cloudevents/spec/blob/master/amqp-transport-binding.md) |
+| Kafka Transport Binding | - | [master](https://github.com/cloudevents/spec/blob/master/kafka-transport-binding.md) |
```
duglin: can you put it in alphabetical order? rebase first though

json-format.md
```diff
@@ -51,6 +51,14 @@ the root of the extension MUST be serialized as a top-level JSON property. There
 were many reason for this design decision and they are covered in more detail in
 the [Primer](primer.md#json-extensions).
 
+For clarity, extension attributes are serialized using the same rules as
```
duglin: I think this is a dup of the previous para - bad rebase maybe?

kafka-transport-binding.md
```diff
@@ -0,0 +1,281 @@
+# Kafka Transport Binding for CloudEvents
```
duglin: for consistency with the other specs, can you add: `  - Version 0.4-wip` to the title?

kafka-transport-binding.md
```diff
@@ -0,0 +1,281 @@
...
+Key: mykey
+
+------------------ headers -------------------
+
+ce_contentType: application/cloudevents+json; charset=UTF-8
+
+------------------- value --------------------
+
+{
+    "specVersion" : "0.1",
```
duglin: s/specVersion/specversion/
s/0.1/0.4-wip/

duglin:
I think the rebase got a little wonky

kafka-transport-binding.md
bluemonk3y: yep - done

json-format.md
bluemonk3y: yep - fixed rebase ;)

README.md
bluemonk3y: the list is now ordered

kafka-transport-binding.md
bluemonk3y: yeah - there were quite a few out of date attributes

kafka-transport-binding.md
```diff
@@ -0,0 +1,281 @@
...
+
+## 2. Use of CloudEvents Attributes
+
+This specification does not further define any of the [CloudEvents][CE] event
+attributes.
+
+### 2.1. data Attribute
+
+The `data` attribute is assumed to contain opaque application data that is
+encoded as declared by the `contentType` attribute.
```
duglin: contentType or "datacontenttype" ?

kafka-transport-binding.md
```diff
@@ -0,0 +1,281 @@
...
+
+## 3. Kafka Message Mapping
+
+With Kafka 0.11.0.0 and above, the content mode is chosen by the sender of the 
+event. Protocol usage patterns that might allow solicitation of events using a
+particular content mode might be defined by an application, but are not defined
+here.
+
+The receiver of the event can distinguish between the two content modes by
+inspecting the `ce_contentType` [Header][Kafka-Message-Header] of the
```
duglin: s/ce_contentType/ce_datacontenttype/   I think  (but it's the "T" that jumped out at me)

kafka-transport-binding.md
duglin: Is this missing a `ce_` or is the sentence in the first paragraph of this section incorrect?

kafka-transport-binding.md
```diff
@@ -0,0 +1,281 @@
...
+use the [Partitioning](extensions/partitioning.md) extension value. 
+
+### 3.2. Binary Content Mode
+
+The *binary* content mode accommodates any shape of event data, and allows for
+efficient transfer and without transcoding effort.
+
+#### 3.2.1. Content Type
+
+For the *binary* mode, the header `ce_contentType` property MUST be
```
duglin: "T" -> "t"

kafka-transport-binding.md
duglin: hmm I think I might be wrong, but I wanted to verify.

kafka-transport-binding.md
```diff
@@ -0,0 +1,281 @@
...
+
+### 3.3. Structured Content Mode
+
+The *structured* content mode keeps event metadata and data together in the
+payload, allowing simple forwarding of the same event across multiple routing
+hops, and across multiple transports.
+
+#### 3.3.1. Kafka Content-Type
+
+The [Kafka `ce_contentType`] property field MUST be set to the media
```
duglin: "T" -> "t"

kafka-transport-binding.md
```diff
@@ -0,0 +1,281 @@
...
+
+Topic Name: mytopic
+
+------------------- key ----------------------
+
+Key: mykey
+
+------------------ headers -------------------
+
+ce_contentType: application/cloudevents+json; charset=UTF-8
```
duglin: "T"->"t"

kafka-transport-binding.md
duglin: Actually, I think the "ce_" here is incorrect since this isn't a CE property. Right?

kafka-transport-binding.md
duglin: I think the intro paragraph is wrong

kafka-transport-binding.md
duglin: I thin the "ce_" here is incorrect.

proprietary-specs.md
```diff
@@ -1,4 +1,4 @@
-# CloudEvent Specs for Proprietary Protocols and Encodings
+### CloudEvent Specs for Proprietary Protocols and Encodings
```
duglin: I think this change is invalid

spec.md
```diff
@@ -58,6 +58,14 @@ that does not support that feature will then silently ignore that part of the
 message. The sender needs to be prepared for the situation where a receiver
 ignores that feature.
 
+For clarity, when a feature is marked as "OPTIONAL" this means that it is
```
duglin: I think the edits in this doc are invalid

kafka-transport-binding.md
fabiojose: LGTM

spec.md
bluemonk3y: @duglin  - I believe I have applied the changes correctly ;) - would you mind reviewing?

proprietary-specs.md
bluemonk3y: reverted

kafka-transport-binding.md
bluemonk3y: agree - removed

kafka-transport-binding.md
bluemonk3y: agree

duglin:
@bluemonk3y looks good to me.

Can I get one more LGTM and then I'll merge!!

duglin:
Approved on the 6/20 call with the minor rebase fixes

duglin:
Approved, again :-) , on the 6/28 call - but this time with the rebase fixes

duglin:
@bluemonk3y et al... thanks a ton for your patience and hard work on this one!!!

gabrieljones:
Out of curiosity why was `ce_`(with an underscore) chosen over `ce-`(with a hyphen) for [binary mode header names](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/bindings/kafka-protocol-binding.md#3231-property-names)?

Related: 
- https://github.com/cloudevents/sdk-java/issues/416
- https://github.com/cloudevents/sdk-java/issues/359

duglin:
@gabrieljones https://grok.com/share/bGVnYWN5LWNvcHk%3D_2afcc0f0-4b9e-46c5-80f5-e6f98e26931d see the Kafka Ecosystem Convention part. That aligns with my recollection of the past too.

gabrieljones:
@duglin Thanks so much. This will have to satisfy my curiosity.

Chatgpt and Gemini have both made both claims about kafka record header names:
- That hyphens are preferred over underscores by convention.
- That underscores are preferred over hyphens by convention.

I have been unable to find any source document that corroborates either claim.

duglin:
Yea, I checked with a few folks who were around way way way back then :-)  and the best anyone can remember is basically what I said... while both are valid it was believed at that time that underscore was the more popular option.

PR #338, opened 2018-11-02
Rename URI to URI-reference, fix JSON Schema

cneijenhuis:
Following the discussion in https://github.com/cloudevents/spec/issues/332, we figured out that the spec actually meant to support URI-references all along. However, by defining the type as `URI`, it is easy to make a mistake and assume it is only a URI, not a URI-reference.
This mistake was also made in the JSON Schema.

This PR therefore does three things:
* Rename the type to `URI-reference`, in order to make it easier to grasp for others in the future
* Fix the JSON Schema
* Add an example for a relative URI

The main changes are in `spec.md` and `spec.json`.

### DIFF
--- a/documented-extensions.md
+++ b/documented-extensions.md
@@ -44,9 +44,9 @@ pre-existing specs; extensions with custom transport bindings are much
 more likely to be dropped by middleware that does not understand the
 extension.
 
-As a convention, extensions of scalar types (e.g. `String`, `Binary`, `URI`,
-`Number`) document their `Value` and structured types document their
-`Attributes`.
+As a convention, extensions of scalar types (e.g. `String`, `Binary`,
+`URI-reference`, `Number`) document their `Value` and structured types document
+their `Attributes`.
 
 ## Known Extensions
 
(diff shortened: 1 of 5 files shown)


documented-extensions.md
```diff
@@ -44,9 +44,9 @@ pre-existing specs; extensions with custom transport bindings are much
 more likely to be dropped by middleware that does not understand the
 extension.
 
-As a convention, extensions of scalar types (e.g. `String`, `Binary`, `URI`,
-`Number`) document their `Value` and structured types document their
-`Attributes`.
+As a convention, extensions of scalar types (e.g. `String`, `Binary`,
+`URI-refernce`, `Number`) document their `Value` and structured types document
```
duglin: Missing 'e' in "refernce"

duglin:
thanks!!!

documented-extensions.md
cneijenhuis: Thanks, fixed.

clemensv:
Yes, it meant to do that. We had a lengthy debate about it these being URI references but that got lost in time. LGTM

duglin:
Approved on Nov 8 call

PR #339, opened 2018-11-02
Simplified/Reduced context property names.

### LINKED ISSUE #323: Context attribute naming consistency
In last weeks call a comment was made (Tim from AWS?) relating to the potential for renaming some of the context attributes; i thought it would make sense to raise it for discussion here prior to opening what would be quite a large PR.

The proposal (as i understand it) would be to rename along these lines:

| Current | Proposed |
| --- | --- |
| eventType | type |
| cloudEventsVersion | version |
| eventID | id |
| eventTime | time |

An alternative argument might be to ensure all context attributes follow a consistent naming structure, eg :

| Current | Proposed |
| --- | --- |
| eventType | eventType |
| source | eventSource |
| data | eventData |

JemDay:
Simplified/Reduced context property names.

Closes #323

### DIFF
--- a/community/contributors.md
+++ b/community/contributors.md
@@ -52,6 +52,9 @@ Contributions do not constitute an official endorsement.
   * Chad Arimura - [@carimura](https://github.com/banisadr)
   * Stanley Halka - [@shalka](https://github.com/banisadr)
   * Travis Reeder - [@treeder](https://github.com/banisadr)
+* **PayPal**
+  * Vladimir Bacvanski
+  * Jem Day - [@JemDay](https://github.com/JemDay)
 * **Red Hat**
   * Jim Curtis - [@jimcurtis64](https://github.com/jimcurtis2) 
   * William Markito Oliveira - [@william_markito](https://github.com/markito) 
--- a/spec.md
+++ b/spec.md
@@ -171,7 +171,7 @@ event consumers can easily access this information without needing to decode
 and examine the event data. Such identity attributes can also be used to
 help intermediate gateways determine how to route the events.
 
-### eventtype
+### type
 * Type: `String`
 * Description: Type of occurrence which has happened. Often this
   attribute is used for routing, observability, policy enforcement, etc.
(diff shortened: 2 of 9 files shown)


community/contributors.md
```diff
@@ -52,6 +52,9 @@ Contributions do not constitute an official endorsement.
   * Chad Arimura - [@carimura](https://github.com/banisadr)
   * Stanley Halka - [@shalka](https://github.com/banisadr)
   * Travis Reeder - [@treeder](https://github.com/banisadr)
+* **PayPal**
+  * Vladimir Bacvanski
+  * Jem Day - [@JemDay](https://github.com/JemDay)
```
duglin: EVERYONE - if you've submitted an issue or PR please feel free to open a PR to add your name to this doc.

spec.md
```diff
@@ -273,7 +273,7 @@ encapsulated within the `data` attribute.
 
 ### data
 * Type: `Any`
-* Description: The event payload. The payload depends on the eventtype and
+* Description: The event payload. The payload depends on the `type` and
   the schemaURL. It is encoded into a media format
```
duglin: if you use back-ticks on "type" then for consistency you should on "schemaURL" on the next line, and "contenttype" - and we should be consistent on the case: e.g. schemaURL vs contenttype

duglin:
@JemDay I updated the title of the PR (to be more descriptive) and added the "Closes #323" so it'll automagically close the issue. Hope that's ok.

spec.md
JemDay: Done

rperelma:
Everything LGTM, except for one thing: I would strongly prefer to keep `eventid` instead of `id`

JemDay:
> Everything LGTM, except for one thing: I would strongly prefer to keep `eventid` instead of `id`

NOTED - I think this goes against the _consistency_ theme slightly but is probably a good subject for discussion at the next session (@duglin)

duglin:
yup - I've love to hear what others think, especially since I'm hoping we can resolve that one issue during the call and approve it. So if you can't make the call and have strong opinions on this point please add them here as comments.

clemensv:
LGTM

duglin:
@JemDay can you rebase? or do you want me to try it?

JemDay:
> @JemDay can you rebase? or do you want me to try it?

Let me try ....

duglin:
Approved on the Nov 8 call

PR #341, opened 2018-11-05
Update the examples based on changes of spec.md

fabiojose:

Based on changes made in spec.md, in `source` elements.

### DIFF
--- a/json-format.md
+++ b/json-format.md
@@ -97,7 +97,7 @@ The following table shows exemplary mappings:
 |--------------------|----------|-------------------------------
 | eventtype          | String   | "com.example.someevent"
 | cloudeventsversion | String   | "0.1"
-| source             | URI      | "/mycontext"
+| source             | URI      | "urn:event:from:myapi/resourse"
 | eventid            | String   | "1234-1234-1234"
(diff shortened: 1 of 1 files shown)


duglin:
While it does no harm to make these changes, I will point out that it is not necessary since 'source' is a URI-reference which allows for `/path` type of values.

fabiojose:
@duglin I agree with you, but what do you think if the audience of this spec will get a little bit confusing? They get examples of `source` in `spec.md` and different ones in `json-format.md`. If they try to run the JSON files against the JSON Schema, these files will be invalid.

What do you think if we add additional examples here too?

duglin:
I believe that https://github.com/cloudevents/spec/pull/338 should fix the schema to support uri-reference and help with the confusion because we'll be using URI-Reference throughout the specs.

I actually wouldn't mind if we had a combination of samples that used a urn:... form and a /... form just to reinforce that both are valid. A lot of times people don't read the specs very closely (shocking, I know!) and they look at examples for guidance - so if all they see are urn:... they may jump to the wrong conclusion.

duglin:
But to be clear, I don't think its a major issue either way.

fabiojose:
@duglin Good points! 

Closed!

PR #343, opened 2018-11-06
clarifications to address concerns raised in issue 185

### LINKED ISSUE #185: JSON Event Format processing
JSON Event Format encodes several types of the Cloud Events type system into a JSON string. How is a consumer of a JSON encoded event expected to determine the intended type of a specific field of a Map? For example, how does the consumer determine if a string is a Timestamp, a base 64 encoded Binary, or just a String? 

Is the consumer expected to be equipped with out of band knowledge to make this determination, or is the type of data expected to be described by a document pointed to by schemaUrl? If the latter, is there any prescription around what the schema of the schemaUrl document is?

clemensv:

Closes #185


duglin:
@clemensv rebase needed

json-format.md
```diff
@@ -58,8 +58,8 @@ exceptions noted below.
 | String        | [string][JSON-String]
 | Integer       | [number][JSON-Number], only the `int` component is permitted
 | Binary        | [string][JSON-String], [Base64-encoded][base64] binary
-| URI-reference | [string][JSON-String]
-| Timestamp     | [string][JSON-String]
+| URI-reference | [string][JSON-String] following [RFC 3986][RFC3986]
+| Timestamp     | [string][JSON-String] following [RFC 3339][RFC3339] (ISO 8601)
```
Vlaaaaaaad: From what I see there seem to be small differences between RFC3339 and ISO 8601 ad per the _Usage_ section in https://en.wikipedia.org/wiki/ISO_8601. 

Is supporting them both desired? Should we be more specifics? Do SDKs need to detect and parse?

clemensv: RFC3339: " This document defines a date and time format for use in Internet
   protocols that is a profile of the ISO 8601 standard for
   representation of dates and times using the Gregorian calendar."

RFC3339 is ISO8601, with limits

Vlaaaaaaad: Oooh. Got it. Thank you!

duglin:
Approved on 11/29 call

PR #342, opened 2018-11-06
adding Integer mapping for JSON and AMQP

clemensv:
Signed-off-by: clemensv <clemensv@microsoft.com>

#329

PR #345, opened 2018-11-08
adding Integer support 

clemensv:


duglin:
Replacement for https://github.com/cloudevents/spec/pull/342 since my github skills managed to close that one by mistake while trying to rebase it for @clemensv

duglin:
#342 was approve on the Nov 8 call, which means this one was implicitly approved too :-)

Issue #349, opened 2018-11-21
revert the specversion to cloudeventsversion for compatibility


kevinwengh:
The field cloudeventsversion is the identifier to determine the format and concrete implementation among different versions of the CloudEvents objects incoming. Swagger has long support the terminology of "Discriminator" and naturally identifier field such as **cloudeventsversion** is used as discriminator. Changing the name of a discriminator has huge impact on both service and client side. 

From service side, APIs leverage the discriminator field to deserialize incoming data into typed object instances, even when a call contains events objects wrapped under different version CloudEvents spec (0.1 or 0.2 or both). Internal implementation of APIs should then process the events accordingly by operating on the deserialized object. 

From customer side, they should able to call a single API to publish the events, no matter how many different version of cloudEvents object to be published. Customer should not be forced to call different APIs in order to publish/submit different versions of events. 

In issue #323, the cloudeventsversion is changed to specversion, which will break our implementation at Oracle CloudEvents service, when we add support for future version of cloudEvents. 

So I am propose that we change the **specversion** back to **cloudeventsversion** . and for future reference, we should never change that field name.

duglin:
The reasons you gave for not changing it are correct, but really only apply once we reach (or get close to) v1.0. Pre-1.0 versions of the spec are, by definition, free to have quite radical changes and this has been our assumption in our decision making process.  We have often told people that "we can change our minds later if needed, before we get to v1.0". This gives us the freedom to test out choices w/o feeling locked into them.  I have no strong feeling on whether we change back or not, this is more of a comment on the process or "rules of engagement" I think most people have been operating under.

kevinwengh:
I totally agree with you that we should have smaller room for dramatic change when we get closer to 1.0 version. But before that, I think field like cloudeventsversion is the anchor, which should be changed only when there is very strong justification. I think the justification in #323 is not enough for make the change. 

Changing the cloudeventsversion is essentially an unnecessary punishment for early adapter like Oracle. we published v0.1 kind of officially, so we need some continuity and consistency. 

furthermore, I think cloudeventversion name is concise and clear. the specversion is ambiguous. if we want to be clear, then keep cloudeventsversion. if we want really simple, we should use version. the specversion probably not a good choice in either case.

duglin:
We discussed this on today's call and decided that there was no new information that warranted us re-opening this topic.

PR #350, opened 2018-11-21
change specversion to cloudeventsversion to maintain compatibility

kevinwengh:
as described in issue #349, I am creating this pull request to reinstate the cloudeventsversion for the compatibility issue, as it's the identifier for each version of the cloudevents spec and used as discriminator in swagger. 

Signed-off-by: Kevin, Wen kevin.wen@oracle.com

### DIFF
--- a/amqp-transport-binding.md
+++ b/amqp-transport-binding.md
@@ -174,7 +174,7 @@ Examples:
 
     * `time` maps to `cloudEvents:time`
     * `id` maps to `cloudEvents:id`
-    * `specversion` maps to `cloudEvents:specversion`
+    * `cloudeventsversioin` maps to `cloudEvents:cloudeventsversioin`
 
 ##### 3.1.3.2 AMQP Application Property Values
 
@@ -195,7 +195,7 @@ content-type: application/json; charset=utf-8
 
 ----------- application-properties -----------
 
-cloudEvents:specversion: "0.1"
+cloudEvents:cloudeventsversioin: "0.1"
 cloudEvents:type: "com.example.someevent"
 cloudEvents:time: "2018-04-05T03:56:24Z"
 cloudEvents:id: "1234-1234-1234"
@@ -257,7 +257,7 @@ content-type: application/cloudevents+json; charset=utf-8
 ------------- application-data --------------------------
 
 {
-    "specversion" : "0.1",
+    "cloudeventsversioin" : "0.1",
     "type" : "com.example.someevent",
 
     ... further attributes omitted ...
--- a/amqp-transport-binding.md
+++ b/amqp-transport-binding.md
@@ -174,7 +174,7 @@ Examples:
 
     * `time` maps to `cloudEvents:time`
     * `id` maps to `cloudEvents:id`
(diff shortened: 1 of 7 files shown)


duglin:
Let's see if we can get some comments on this before the next call (Nov 29) to gauge how the group feels about this.

barkerd427:
My preference is for conciseness, so I'd rather keep specversion. This is for v0.2, and if we're already locked into decisions like this, then I'm worried we won't be able to adapt enough when we want simple changes like this. I understand the overhead incurred in these situations as I've used multiple early Apache projects that changed wildly between beta version and caused major rewrites in critical production systems. This was the cost of early adoption. I don't expect a lot of changes in the core tags after the latest update based on the robust discussion around it.

Vlaaaaaaad:
I'd agree with https://github.com/cloudevents/spec/issues/349#issuecomment-440837011 that we need either `version` or `cloudeventversion` as just `specversion` is not that clear.

Can't say I have a preference for either of the two.

duglin:
Just a couple of comments/observations:
- I think the name we pick should be based on what's best for users of the spec, not based on a historical names picked - as painful as it might be for early adopters
- if `specversion` isn't clear, then I'm not sure `version` is either - or stated another way, I view both as equally sufficient or equally vague depending on your POV
- `cloudeventsversion` would become `ce-cloudeventsversion` as an HTTP header - repetitive?

All, please plan on having an opinion on tomorrow's call - I'm not planning to ask for a formal vote, but I will probably pick on people to hear their opinions because I'd like get a sense of what the majority of the group thinks. And I won't know if silence means "ok with current spec" or "anything is fine".

JemDay:
I believe there was some discussion regarding _specversion_ and _version_ which resulted in me suggesting the name we currently have - i'm not wedded to _specversion_ but there was a rationale, i just can't remember what it was.

I would prefer *not* to move back to the long-form name, as @barkerd427 mentioned the current pain is an unfortunate side-affect of early adoption; that said, we should ensure that this property (whatever it ends up being called) must remain constant going forward.

rperelma:
I too would prefer NOT to change it again

vbacvanski:
I prefer to leave it as "specversion". This clearly communicates what the version is about and does not have the redundant "event" in the name, which would also make the naming of properties inconsistent, an issue discussed in #323.

The unfortunate consequence is that "specversion" requires a change in the API. However, I feel at this early stage we are all learning and the changes are expected, particularly around version 0.1. I would not view the change as "punishment" for early adopters, but an iteration towards better specification and consequently the more elegant and clear SDK. The history of an early implementation should not negatively affect the specification. I expect to see the number of such changes asymptotically going down as we mature our ideas.

rohitva:
No other filed has "spec" as a prefix. I feel "cloudeventsversion" or "version" should be a better choice.

JemDay:
To add to my previous comment, i believe there was concern that _"version"_ may be misinterpreted as _version of the event_, thus the tacit agreement to go with _"specversion"_

duglin:
On today's call we discussed this and decided that there was no new information that warranted us re-open this topic.

PR #353, opened 2018-11-24
Add some rationale for our extensions decision

duglin:
The Primer is a good location to document some of our design decisions
without cluttering the spec with non-normative text. This is my first pass
at some of the reasons why we're serializing extensions as top-level
properties in JSON.

Note, as you review this please focus on whether the text accurately
reflects the reasons why we ended up with our current design. This PR
should not be used to question whether it was a good or bad decision.
If you want to question the decision itself then a new issue (or PR) is the
best place to do that - I'm just trying to document history.


duglin:
Approved on 11/29 call

PR #356, opened 2018-11-27
CloudEvents SDK spec proposal

denismakogon:
This spec is the first step towards nicely-shaped CloudEvents SDK spec.

### DIFF
--- a/SDK.md
+++ b/SDK.md
@@ -236,7 +236,7 @@ It means that an HTTP marshaller/unmarshaller is a set of binary and structured 
 
 In `Read`:
 
- - `event` - a placeholder (empty event object), see [event constructor](#Event object constructor API).
+ - `event` - a placeholder (empty event object), see [event constructor](#event-object-constructor-api).
  - `headers` - an HTTP request headers
  - `body` - an HTTP request body
  - `data_unmarshaller` - a function that turns an event data into an object of 
--- a/SDK.md
+++ b/SDK.md
@@ -1,0 +1,173 @@
+# CloudEvents SDK
+
+The intent of this document to describe bare minimum set of requirements for a new SDKs.
+
+## Status of this document
+
+Since [CloudEvent spec](spec.md) still considered as a working draft this document also suppose to be considered as a working draft.
+
+## Contribution acceptance
+
+Being an open source community CloudEvents team is open for a new members as well open to their contribution.
+In order to ensure that an SDK is going to be supported and maintained CloudEvents community would like to ensured that:
+
+ - a person (developer, committer) is going to become a maintainer
+ - a person commits to support ongoing changes to the [CloudEvent spec](spec.md)
+
+## Officially maintained software development kits (SDK)
+
+Software Development Kit (SDK) is a set of helpers designed and implemented to enhance and speed up a CloudEvents integration.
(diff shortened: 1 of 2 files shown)


SDK.md
```diff
@@ -0,0 +1,173 @@
...
+```
+    v01.Event()
+```
+
+#### Event object setters API 
+
+This particular code sample represents bare minimum number of setters:
+```
+    v01.Event().
+    WithContentType("application/json").
```
duglin: Minor thing, but to avoid people getting confused since you use the term "setters", it might be better to use `SetXXX` instead of `WithXXX`.

SDK.md
```diff
@@ -0,0 +1,173 @@
...
+```
+
+Type setter represents an event type setter:
+```
+    WithEventType(type string)
+```
+
+Extensions setter represents an event type setter:
+```
+    WithExtensions(exts map[string]string)
```
duglin: May want to allow for setting an individual extension instead of all at once.

duglin: unless you consider the generic Set() below to do that - in which case ignore my previous comment.

duglin:
Overall, I think its a good first start. Aside from the comments I made in the file, could you:
1 - wrap things at 80 columns to be consistent with the other files in the repo
2 - add a pointer to this file from the main README so people can easily find it.

duglin:
For those who were not in our weekly SDK call today, we talked about how it would be nice to have a document that describes what is expected of new SDKs. That then led to the idea of it being a (sort of) high-level design doc - to push for consistency where possible.

@denismakogon agreed to take a first pass at it - which is this PR.

We don't have a global "SDK" repo for stuff like this - so for now we decided it put it in this repo. If we get more "common SDK" docs then we can consider creating a new repo for them to live in.

SDK.md
denismakogon: yeah,`Set` would take care of a single extension, will a note about that.

SDK.md
```diff
@@ -0,0 +1,281 @@
...
+    ) -> HTTP-Headers, Stream
+    
+```
+
+As you may see, Converter signature follows the signature of marshaller/unmarshaller.
+It means that an HTTP marshaller/unmarshaller is a set of binary and structured converters.
+
+In `Read`:
+
+ - `event` - a placeholder (empty event object), see [event constructor](#Event object constructor API).
```
duglin: I think `#Event object constructor API` needs to be `#event-object-constructor-api`

denismakogon: fixed, thanks

duglin:
SDKers - any comments/thoughts? Good enough for a first draft?

SDK.md
```diff
@@ -0,0 +1,281 @@
...
+```
+
+Source setter represents an event source setter:
+```
+    SetSource(source URL)
+```
+
+Time setter represents event emit time setter:
+```
+    SetEventTime(time RFC3339)
```
matzew: let's use something like `SetTime`, since that actually matches the 0.2 bits ...

denismakogon: I'd rather leave this as it is, the small bonus here is the verbosity, once 0.1 will be deprecated, we can switch to whatever that is.

duglin:
Approved on the 12/6 call

denismakogon:
Cool!

PR #357, opened 2018-11-29
Be explicit about what the specversion string is

duglin:
We never actually said if its 0.1 or v0.1 or something else.


duglin:
Approved on 11/29 call

Issue #330, opened 2018-10-15
What is the recommended way to handle multiple records per event ?


fsunaval:
AWS S3 and many other cloud providers provide multiple records per event.

E.g. the event notification format for S3 is as follows.
https://docs.aws.amazon.com/AmazonS3/latest/dev/notification-content-structure.html

{  
   "Records":[  
      {  
         "eventVersion":"2.0",
         "eventSource":"aws:s3",
         "awsRegion":"us-east-1",
         "eventTime":The time, in ISO-8601 format, for example, 1970-01-01T00:00:00.000Z, when S3 finished processing the request,
         "eventName":"event-type",
         "userIdentity":{  
            "principalId":"Amazon-customer-ID-of-the-user-who-caused-the-event"
         },
         "requestParameters":{  
            "sourceIPAddress":"ip-address-where-request-came-from"
         },
         "responseElements":{  
            "x-amz-request-id":"Amazon S3 generated request ID",
            "x-amz-id-2":"Amazon S3 host that processed the request"
         },
         "s3":{  
            "s3SchemaVersion":"1.0",
            "configurationId":"ID found in the bucket notification configuration",
            "bucket":{  
               "name":"bucket-name",
               "ownerIdentity":{  
                  "principalId":"Amazon-customer-ID-of-the-bucket-owner"
               },
               "arn":"bucket-ARN"
            },
            "object":{  
               "key":"object-key",
               "size":object-size,
               "eTag":"object eTag",
               "versionId":"object version if bucket is versioning-enabled, otherwise null",
               "sequencer": "a string representation of a hexadecimal value used to determine event sequence, 
                   only used with PUTs and DELETEs"            
            }
         }
      },
      {
          // Additional events
      }
   ]
}  

A typical scenario this can happen is when I push records to S3 via kinesis firehose.  It will chunk the records into individual large objects, thus producing just one notification and allowing the user function to parse multiple records with one lambda function call.

My question is:
How can I convert such an event to cloudevent format ?
If the answer is "This is not supported and the event source should not do it" I am fine with that too.
But am open to hearing suggestions from the community.

thanks,
Farhad.

fsunaval:
Yes, complete multiple event records stuffed inside one event notification. 
Splitting them into multiple CloudEvents is easily doable but it will create a massive burst of CloudEvents.
[...]

Issue #282, opened 2018-07-29
Support multiple events for HTTP binding


lennygran:
There are use cases where multiple events should be sent via single web-hook request (aka batching.) Currently it assumes a single event for both, structured and binary modes. Does the HTTP binding in structured mode plan to support it?

rperelma:
See also https://github.com/cloudevents/spec/issues/330

duglin:
@lennygran given #330 what are your thoughts on this?

dmigliori:
I feel like the "conclusion" of many issues that generate a good discussion, but result in no changes to the spec, should still be included in the introductory documentation.  This will enable implementers to have their questions answered as they read the documentation, rather than search closed issues for their answers.

duglin:
@dmigliori yup - feel free to open up PRs to enhance the Primer doc. Any insight into our design choices will only help people understand things.

duglin:
@lennygran given #330 what are your thoughts on this?

cneijenhuis:
To clarify what I said in #330 :
I do not think that a single CloudEvent should contain multiple events. However, a transport layer may offer the possibility to transmit or receive multiple events in a single API call. Once we have the mapping between a (single) CloudEvent to a (single) transport layer event, it is trivial to use the API for transmitting multiple transport layer events with CloudEvents.

The HTTP binding is a transport layer, and therefore my comments in #330 imply that this would be the right layer to implement batching.

---

That said, most other transport layers supporting batching have a middle layer that allows to decouple batching between the producer and the consumer. E.g. the producer may batch 3 messages together, but the consumer batches with 10 messages, or none.

This assumption doesn't necessarily hold for HTTP. The problem I am seeing, in the context of a FaaS that allows to consume HTTP webhooks but doesn't have native support of CloudEvents, is that the producer then effectively reduces the parallelism. E.g. ideally, if there are 10 events, I'd like to have 10 function instances working on them in parallel. However, if the producer batches 10 events into a single HTTP webhook, only one function will be invoked that has to process all 10 events sequentially.

Therefore, I think batching should only be used (by the producer) to the extent the consumer wants to support it. Ideally the consumer is a middleware, and can re-arrange the batch sizes as necessary. But if it isn't, the consumer should be able to receive events one-by-one.

That would either have to go into the initial registration of the webhook, or we could introduce a third mode (e.g. structured-batched).

deissnerk:
[...]

PR #360, opened 2018-12-06
Transports are responsible for batching messages

cneijenhuis:
Follow up to the discussions in #330 and #282 : Since it seems to be a frequently asked question, we should somewhere in the spec write down how to handle multiple messages.

I think the `primer.md` is the place that fits best. I'm not sure if the Design Goal section is a perfect fit, but I didn't find a better place.

### DIFF
--- a/primer.md
+++ b/primer.md
@@ -119,7 +119,8 @@ and transports (e.g. HTTP, AMQP, Kafka).
 
 Batching of multiple events into a single API call is natively supported by some
 transports. To aid interoperability, it is left up to the transports if and how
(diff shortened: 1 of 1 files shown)


duglin:
LGTM

cneijenhuis:
@Tapppi Does https://github.com/cloudevents/spec/pull/360/commits/ed3b17ef8990cf16c50981409515ea2c7a50f0d4 address what you've asked for in last weeks call?

duglin:
Approved on the Jan 17 call

PR #361, opened 2018-12-12
Specify range of Integer type exactly

alanconway:
"32 bit integer" is under-specified, there are multiple ways to encode an
integer in 32 bits with different ranges (signed, unsigned, twos-complement,
ones-complement etc.)

Proposed new wording: "signed 32-bit twos-complement"

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -107,7 +107,7 @@ platform/vendor specific protocols (AWS Kinesis, Azure Event Grid).
 
 The following abstract data types are available for use in attributes.
 
-- `Integer` - A 32-bit whole number.
+- `Integer` - A whole number in the range −2,147,483,647 to +2,147,483,647 inclusive. This is the range of a signed 32-bit twos-complement representation. Event formats do not have to use this encoding, but they MUST only use `Integer` values in this range.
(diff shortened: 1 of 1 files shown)


duglin:
@alanconway can you please sign your commit?

spec.md
```diff
@@ -107,7 +107,7 @@ platform/vendor specific protocols (AWS Kinesis, Azure Event Grid).
 
 The following abstract data types are available for use in attributes.
 
-- `Integer` - A 32-bit whole number.
+- `Integer` - A whole number in the range −2,147,483,647 to +2,147,483,647 inclusive. This is the range of a signed 32-bit twos-complement representation. Event formats do not have to use this encoding, but they MUST only use `Integer` values in this range.
```
duglin: Can you wrap this at 80 columns - so it matches the rest of the doc?

duglin: https://en.wikipedia.org/wiki/32-bit shows it as "...,648" not "...,647" for the low range.

alanconway: Thanks, fixed both. All green now :)

duglin: thanks

duglin:
Approved on the Jan 17 call

PR #363, opened 2019-01-02
s/contenttype/datacontenttype/g

### LINKED ISSUE #354: Rename "contenttype" to "datatype"
From looking at the RocketMQ PR (#352) I think there's going to be some confusion going forward between the CloudEvents "contenttype" property and a transport-level "ContentType" property. I might be mistaken but I believe that in #352 they're using the CE one in place of a transport level one. We need to make it clear the the CloudEvents "contenttype" property is NOT about the encoding of the CloudEvent, rather its about the encoding of the "data" property - nothing else. I believe that using a well-known (and a bit vague) phase like "Content Type" might cause more confusion going forward. To help, I suggest that we rename it to something that more clearly scopes it to its intended usage - some options:
- datatype
- dataencoding
- datacontenttype

I have a slight preference for `datatype`.

This does make me wonder about the relationship between this property and the `schemaurl` property. Conceptually, I understand the two but I wonder if we really need to split them into two separate properties. For example, could the `schemaurl` information not be encoded into the `datatype` as an extra parameter - if needed. It might be worth noting that https://en.wikipedia.org/wiki/List_of_HTTP_header_fields doesn't show a `schemaurl` type of HTTP header.

I'd like to hear some thoughts on this before I open a PR for either, or both, topics.

duglin:
Closes #354

2 reasons for this change:
1 - it just makes sense to me that if you have a `foo` property, then
    the property the describes its type is called `footype`, not `bartype`.
2 - when a transport, like http, already has a `contenttype` property, there
    could be unnecessary confusion when the CloudEvents property is serialized
	along side the transport one. Meaning, people might use the wrong one
	since their names are almost the same.

### DIFF
--- a/json-format.md
+++ b/json-format.md
@@ -113,7 +113,7 @@ The following table shows exemplary mappings:
 | source             | URI-reference | "/mycontext"
 | id                 | String        | "1234-1234-1234"
 | time               | Timestamp     | "2018-04-05T17:31:00Z"
-| contenttype        | String        | "application/json"
+| datacontenttype    | String        | "application/json"
 | data               | String        | "<much wow=\"xml\"/>"
 | data               | Binary        | "Q2xvdWRFdmVudHM="
 | data               | Map           | { "objA" : "vA", "objB", "vB" }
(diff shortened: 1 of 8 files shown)


JemDay:
I believe this makes sense ... we _may_ need communication strategy to address the fact that this is another breaking change which has potential to impact early adopters.

Granted we're still in a sub 1.0 state but we might want to warn people off 0.2 if 0.3 (or 1.0) is considered close.

duglin:
Note: this will be discussed (and possibly resolved) on today's call and since this is a breaking change please speak up if you have any concerns.

json-format.md
```diff
@@ -141,17 +141,18 @@ in [Section 2.3.](#23-mapping-any-typed-attributes), with two additional
...
+
+If the `datacontenttype` value is either ["application/json"][RFC4627] or any
+media type with a [structured +json suffix][RFC6839], the implementation MUST
+translate the `data` attribute value into a [JSON value][JSON-Value], and set
+the `data` attribute of the envelope JSON object to this JSON value.
+
+If the `datacontenttype` value does not follow the
+[structured +json suffix][RFC6839] but is known to use JSON encoding, the
+implementation MUST translate the `data` attribute value into a
+[JSON value][JSON-Value], and set the `data` attribute of the envelope
```
duglin: Just the name change here... but things go pushed past 80 columns so I had to wrap stuff earlier.

duglin:
Per the call yesterday I've changed the name to `DataContentType` - please review for next week's call.

duglin:
Just a poke to remind people to review this one as it's an attribute name change - this will come up today's call.

PR #366, opened 2019-01-10
Add Integer as allowed for Any in description of variant type

Tapppi:
In https://github.com/cloudevents/spec/commit/3248e31cabace2c8d8d8da46a033923d598c1bfb `Integer` was added as an allowed shape of the variant type `Any`, but the modification was left out of the later description of the variant type, which also lists the allowed shapes. Added `Integer` to the list and made it the same textual style as in the type list. So no normative changes.

Another possibility is to just change the text describing the variant type to refer to the type list instead of repeating the list.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -111,19 +111,16 @@ The following abstract data types are available for use in attributes.
 - `String` - Sequence of printable Unicode characters.
 - `Binary` - Sequence of bytes.
 - `Map` - `String`-indexed dictionary of `Any`-typed values.
-- `Any` - Either a `String`, or a `Binary`, or a `Map`, or an `Integer`.
+- `Any` - Either a `Binary`, `Integer`, `Map` or `String`.
 - `URI-reference` - String expression conforming to `URI-reference`
   as defined in
   [RFC 3986 §4.1](https://tools.ietf.org/html/rfc3986#section-4.1).
 - `Timestamp` - String expression as defined in
   [RFC 3339](https://tools.ietf.org/html/rfc3339).
 
-This specification does not define numeric or logical types.
-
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -120,8 +120,8 @@ The following abstract data types are available for use in attributes.
 
 This specification does not define numeric or logical types.
 
-The `Any` type is a variant type that can take the shape of either a
-`String` or a `Binary` or a `Map`. The type system is intentionally
+The `Any` type is a variant type that can take the shape of either a `String`,
+or a `Binary`, or a `Map`, or an `Integer`. The type system is intentionally
```
duglin: @Tapppi we discussed this on today's call and people are ok with the general direction. They did, however, want to make some minor editorial changes - would you be ok with:
... take the shape of either a `Binary`, `Integer`, `Map` or `String`.
Basically, alphabetize, remove the extra "or a"s and the extra comma.

duglin:
@Tapppi will you have a chance to make these minor tweaks?

Tapppi:
@duglin Sure, I think I'll have time tomorrow/wednesday to incorporate the changes. However, do we also want to fix the spelling to be consistent in the data type list above the variant description being changed now? I think it is rather odd to have two wordings for the same thing in a normative spec.

duglin:
LGTM  - thanks @Tapppi 

Can I get one more LGTM and then we can merge since the changes are consistent with what the group agreed to on the last call?

duglin:
Approved on the Jan 24 call with the edit that the "This specification does not define numeric or logical types." line should be removed. Change is now made, so merging.

thanks @Tapppi

PR #370, opened 2019-01-15
HTTP Transport Binding for batching JSON

cneijenhuis:
In the discussion from last week about https://github.com/cloudevents/spec/pull/360 we agreed that a batching option for the HTTP Transport Binding is useful. This is an attempt to introduce it.

During the call, @clemensv suggested to build the batching into the structured mode. I couldn't figure out how to do this generically for any event format. Instead, I did it for the JSON format only.

I also did not include the batching mode as one of the modes the receiver SHOULD support. Take a Function as a Service: Usually, each event should be processed in on an own instance of a function. This is trivial to implement when a HTTP Request always contains a single event. A batch doesn't map nicely onto the core FaaS abstractions.

The HTTP Content-Type is a bit tricky - we want that the receiver can distinguish between the events based on the header, so I've created `application/cloudevents-batch+json`. However, I'm not sure if this clutters the namespace too much.

### DIFF
--- a/http-transport-binding.md
+++ b/http-transport-binding.md
@@ -344,8 +344,10 @@ Content-Length: nnnn
 
 In the *batched* content mode several events are batched into a single HTTP
 request or response body. The chosen [event format](#14-event-formats) MUST
-define how a batch is represented. Currently, the only format supporting
-batching is the [JSON Batch Format][JSON-batch-format].
+define how a batch is represented.
+Based on the [JSON format][JSON-format] (that MUST be supported by any
+compliant implementation), the [JSON Batch format][JSON-batch-format] is an
+event format that supports batching.
 
 #### 3.3.1. HTTP Content-Type
 
@@ -366,7 +368,7 @@ all event attributes, including the `data` attribute, are represented.
 The batch of events is then rendered in accordance with the event format
 specification and the resulting data becomes the HTTP message body.
 
-The batch MAY be empty (typically used in a HTTP response).
+The batch MAY be empty.
 All batched CloudEvents MUST have the same `specversion` attribute. Other
 attributes MAY differ, including the `contenttype` attribute.
 
--- a/http-transport-binding.md
+++ b/http-transport-binding.md
@@ -23,7 +23,7 @@ This document is a working draft.
 3. [HTTP Message Mapping](#3-http-message-mapping)
 - 3.1. [Binary Content Mode](#31-binary-content-mode)
 - 3.2. [Structured Content Mode](#32-structured-content-mode)
-- 3.3. [JSON Batch Content Mode](#33-json-batch-content-mode)
+- 3.3. [Batched Content Mode](#33-batched-content-mode)
 4. [References](#4-references)
 
 ## 1. Introduction
@@ -60,7 +60,7 @@ which is compatible with HTTP 1.1 semantics.
 ### 1.3. Content Modes
 
 This specification defines three content modes for transferring events:
-*structured*, *binary* and *JSON Batch*. Every compliant implementation SHOULD
+*binary*, *structured* and *batched*. Every compliant implementation SHOULD
 support the *structured* and *binary* modes.
 
 In the *binary* content mode, the value of the event `data` attribute is placed
@@ -72,9 +72,9 @@ In the *structured* content mode, event metadata attributes and event data are
 placed into the HTTP request or response body using an [event
 format](#14-event-formats).
 
-In the *JSON Batch* content mode several events are batched into a single HTTP
-request or response body. The payload is built on top of the
-[JSON event format][JSON-format].
+In the *batched* content mode several events are batched into a single HTTP
+request or response body using an [event format](#14-event-formats) that
+supports batching.
 
 ### 1.4. Event Formats
 
@@ -82,6 +82,9 @@ Event formats, used with the *structured* content mode, define how an event is
 expressed in a particular data format. All implementations of this
 specification MUST support the [JSON event format][JSON-format], but MAY
 support any additional, including proprietary, formats.
+
+Event formats MAY additionally define how a batch of events is expressed. Those
+can be used with the *batched* content mode
 
 ### 1.5. Security
 
@@ -129,15 +132,15 @@ The event binding is identical for both HTTP request and response messages.
 The content mode is chosen by the sender of the event, which is either the
 requesting or the responding party. Gestures that might allow solicitation of
 events using a particular mode might be defined by an application, but are not
-defined here. The *JSON Batch* mode MUST NOT be used unless solicited, and the
+defined here. The *batched* mode MUST NOT be used unless solicited, and the
 gesture SHOULD allow the receiver to choose the maximum size of a batch.
 
 The receiver of the event can distinguish between the three modes by inspecting
 the `Content-Type` header value. If the value is prefixed with the CloudEvents
 media type `application/cloudevents`, indicating the use of a known [event
 format](#14-event-formats), the receiver uses *structured* mode. If the value
-is prefixed with `application/cloudevents-batch+json`, the receiver uses the
-*JSON Batch* mode. Otherwise it defaults to *binary* mode.
+is prefixed with `application/cloudevents-batch`, the receiver uses the
+*batched* mode. Otherwise it defaults to *binary* mode.
 
 If a receiver detects the CloudEvents media type, but with an event format that
 it cannot handle, for instance `application/cloudevents+avro`, it MAY still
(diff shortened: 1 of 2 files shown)


duglin:
travis isn't happy :-(   I wonder if it's the case of the "B"

cneijenhuis:
Yep, that was it. Thanks, fixed! :)

deissnerk:
LGTM

clemensv:
@deissnerk @cneijenhuis  batching is about structuring the message payload (one or multiple). The WebHook spec is about how you deliver the payload, independent of how it's structured. 

* Event format: How to encode one event or a batch of events
* Transport: How to bind the encoded message/batch to a transport message
* WebHook (etc): Application contract to deliver the transport message to an endpoint and handle related errors.

http-transport-binding.md
```diff
@@ -330,6 +340,105 @@ Content-Length: nnnn
 
 ```
 
+### 3.3. Batched Content Mode
+
+In the *batched* content mode several events are batched into a single HTTP
+request or response body. The chosen [event format](#14-event-formats) MUST
+define how a batch is represented. Currently, the only format supporting
```
clemensv: You have to change "currently, the only" if/when that changes. It makes sense to make an absolute statement like "The JSON event format that MUST be supported by any compliant implementation supports batching."

http-transport-binding.md
```diff
@@ -330,6 +340,105 @@ Content-Length: nnnn
...
+
+#### 3.3.2. Event Data Encoding
+
+The chosen [event format](#14-event-formats) defines how a batch of events and
+all event attributes, including the `data` attribute, are represented.
+
+The batch of events is then rendered in accordance with the event format
+specification and the resulting data becomes the HTTP message body.
+
+The batch MAY be empty (typically used in a HTTP response).
```
clemensv: Strike "(typically used in a HTTP response)". There might be reasons to push an empty (zero elements) batch.

json-format.md
```diff
@@ -229,7 +230,77 @@ a `Map` or [JSON data](#31-special-handling-of-the-data-attribute) data:
...
+considered as a separate format: a valid implementation of the *JSON Format*
+doesn't need to support it. The *JSON Batch Format* MUST NOT be used when only
+support for the *JSON Format* is indicated.
+
+### 4.1. Mapping CloudEvents
+
+This section defines how a batch of CloudEvents is mapped to JSON.
+
+The outermost JSON element is a [JSON Array][JSON-array], which contains
+CloudEvents rendered in accordance with the [JSON event format][JSON-format]
```
clemensv: ... as elements ...

http-transport-binding.md
cneijenhuis: I didn't implement your suggestion exactly, because it seems to imply that batching MUST be supported by any compliant implementation. In the current proposal, it is however optional, and I think it should stay optional, as I wrote in the PR description:

> I also did not include the batching mode as one of the modes the receiver SHOULD support. Take a Function as a Service: Usually, each event should be processed in on an own instance of a function. This is trivial to implement when a HTTP Request always contains a single event. A batch doesn't map nicely onto the core FaaS abstractions.

Is what I added in https://github.com/cloudevents/spec/pull/370/commits/1191dd226eeb20d77a4d99f47ce6487065e28995 fine?

deissnerk:
> Event format: How to encode one event or a batch of events
Transport: How to bind the encoded message/batch to a transport message
WebHook (etc): Application contract to deliver the transport message to an endpoint and handle related errors.

@clemensv I understand that distinction between transport and web hook. That's why I was suggesting to define the header fields separate from the core web hook spec. Technically I think it makes sense, that the delivery target can specify the (maximum) number of events each message is supposed to contain in addition to the request rate, 
One more thing that might be interesting, is to have additional fields like "Accept:  application/cloudevents-batch+json" in the validation response to signal the support of batching.

cneijenhuis:
I do agree that there is an interesting interaction between request rate and batch size. I even wonder if specifying both statically is useful. E.g. if the request rate is 50/minute, and the batch size is 50:
* I send 1 msg per sec. After 50 sec, I've send 50 msgs and have to stop.
* I send 50 msgs every two sec. After 50 sec, I've send 1250 msgs, and can send 1250 more messages.

Maybe it is better to interpret the request rate as event rate. E.g. for a request rate of 10, you can send me 10 events in 10 requests, or you can send me 10 events in a single batch.

>  One more thing that might be interesting, is to have additional fields like "Accept: application/cloudevents-batch+json" in the validation response to signal the support of batching.

My take is that there is an initial registration step (which is not standardized), and - if the sender offers a choice - the delivery target (administrator) should choose the format and mode.

json-format.md
dmigliori: I like this approach with the use of a JSON array, however I'd like to see an alternate JSON event format that supports a 2D array. 

Time series events are most efficiently contained within a grid (tabular) structure which can be represented by a two-dimensional "array" as an "object".  A CloudEvents "message-level" Batch object can comprise this 2D array.

The column values within the 2D array can correspond to CloudEvent "event-level" attributes (e.g., id, time, type, data), which can include attributes from one or more extensions (e.g., sequence, object, attribute). 

Another "message-level" Schema attribute (type: array as object) can map the column indices to the names of "event-level" attributes:

0, "time"
1, "type"
2, "object"
3, "attribute"
4, "data"

The Schema array could also include constraint columns (e.g. Required) allowing flexibility among producer/consumer ecosystems.

This design has 2 benefits:

1) It compresses a multi-event payload by removing repetitive "names" in name (attribute)-value pairs.

2) It streamlines a consumer's parsing a multi-event payload.

http-transport-binding.md
```diff
@@ -58,26 +59,33 @@ which is compatible with HTTP 1.1 semantics.
...
+
+In the *batched* content mode several events are batched into a single HTTP
+request or response body using an [event format](#14-event-formats) that
+supports batching.
+
 ### 1.4. Event Formats
 
 Event formats, used with the *structured* content mode, define how an event is
 expressed in a particular data format. All implementations of this
 specification MUST support the [JSON event format][JSON-format], but MAY
```
duglin: do we need to make this clearer w.r.t. which json format we mean? single vs batched - or do they have to support both?

http-transport-binding.md
```diff
@@ -330,6 +340,107 @@ Content-Length: nnnn
...
+
+The chosen [event format](#14-event-formats) defines how a batch of events and
+all event attributes, including the `data` attribute, are represented.
+
+The batch of events is then rendered in accordance with the event format
+specification and the resulting data becomes the HTTP message body.
+
+The batch MAY be empty.
+All batched CloudEvents MUST have the same `specversion` attribute. Other
+attributes MAY differ, including the `contenttype` attribute.
```
duglin: datacontenttype

json-format.md
```diff
@@ -229,7 +230,77 @@ a `Map` or [JSON data](#31-special-handling-of-the-data-attribute) data:
...
+## 4. JSON Batch Format
+
+In the *JSON Batch Format* several CloudEvents are batched into a single JSON
+document. The document is a JSON array filled with CloudEvents in the
+[JSON Event format][JSON-format].
+
+Although the *JSON Batch Format* builds ontop of the *JSON Format*, it is
+considered as a separate format: a valid implementation of the *JSON Format*
+doesn't need to support it. The *JSON Batch Format* MUST NOT be used when only
+support for the *JSON Format* is indicated.
```
duglin: I wonder if this last sentence is appropriate since we don't get into how things like this are "indicated".

json-format.md
```diff
@@ -229,7 +230,77 @@ a `Map` or [JSON data](#31-special-handling-of-the-data-attribute) data:
...
+      "specversion" : "0.2",
+      "type" : "com.example.someevent",
+      "source" : "/mycontext/4",
+      "id" : "B234-1234-1234",
+      "time" : "2018-04-05T17:31:00Z",
+      "comexampleextension1" : "value",
+      "comexampleextension2" : {
+          "otherValue": 5
+      },
+      "contenttype" : "application/vnd.apache.thrift.binary",
```
duglin: datacontenttype

json-format.md
```diff
@@ -229,7 +230,77 @@ a `Map` or [JSON data](#31-special-handling-of-the-data-attribute) data:
...
+      "specversion" : "0.2",
+      "type" : "com.example.someotherevent",
+      "source" : "/mycontext/9",
+      "id" : "C234-1234-1234",
+      "time" : "2018-04-05T17:31:05Z",
+      "comexampleextension1" : "value",
+      "comexampleextension2" : {
+          "otherValue": 5
+      },
+      "contenttype" : "application/json",
```
duglin: dct

clemensv:
I have an implementation of this here now: https://github.com/clemensv/cloudevents-sdk-csharp/tree/batching 

The only little pitfall I've found is that a ```contentType.startsWith("application/cloudevents")``` is now the false test for probing for a CloudEvent. You must include the plus sign before the format indicator since application/cloudevents-batch also matches that test.

json-format.md
clemensv: You are proposing a different data layout model. That's not a batch, it's a wholly different encoding. I discuss this here as "Record Sequence with Metadata Preamble"

https://vasters.com/blog/data-encodings-and-layout/ 

If you care about compact time series data transfers, we should make an Apache Avro encoding, because Avro is really good at that.

http-transport-binding.md
duglin: @cneijenhuis what do you think about making this:
```
... MUST support the non-batching [JSON event format][JSON-format], but MAY...
```
so just add `non-batching`  ?

cneijenhuis: I like it! https://github.com/cloudevents/spec/pull/370/commits/d3f2edbd4837dbc6194aa418059d820669392523

duglin: thanks

duglin:
This PR was approved on last week's call with the agreement to make the minor editorial changes that @cneijenhuis has now made. So, LGTM from me, however, I'd like more one LGTM just to make sure one more set of eyes has verified that the latest round of edits didn't introduce some minor bug by mistake.

cneijenhuis:
> My original intention was that with a maximum batch size you could limit memory usage on the receiver side.

Ah, I see. Yes, both limits make sense in their own right.

duglin:
@cneijenhuis @deissnerk anything we need to modify here before we merge?

deissnerk:
@duglin No, I don't think so. This would be a separate issue or PR. I don't really know, where to put it. We wanted to keep the web hook spec CloudEvent agnostic.

duglin:
@deissnerk ok thanks.  Anyone wanna give it one more look-over and an LGTM?

deissnerk:
@duglin LGTM

duglin:
@deissnerk thanks!  merging

PR #372, opened 2019-01-21
minLength for non-empty attributes, add schemaurl

fabiojose:

### DIFF
--- a/spec.json
+++ b/spec.json
@@ -2,7 +2,9 @@
   "$ref": "#/definitions/event",
   "definitions": {
     "specversion": {
-      "type": "string"
+      "type": "string",
+      "minLength": 1,
+      "const": "0.2"
     },
     "contenttype": {
       "type": "string"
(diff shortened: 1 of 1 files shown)


duglin:
nice!
LGTM

spec.json
```diff
@@ -2,7 +2,9 @@
   "$ref": "#/definitions/event",
   "definitions": {
     "specversion": {
-      "type": "string"
+      "type": "string",
+      "minLength": 1,
+      "const": "0.2"
```
Tapppi: Does this mean that no other values than `0.2` are allowed? If so, this also affects how this schema can be used.

Previously one could validate events with schemas across versions, but after this PR one could only use the same version of the spec as the version of the event (as the `const` here wouldn't match across versions). This might not be a problem pre-v1.0 when there are breaking changes in the schema anyway, meaning you will probably need the correct schema for the event version being used anyway, but post-v1.0 I would expect this schema to mostly stay intact across minor versions.

If we want to outline that one should hold the schemas for all versions and validate with the schema of the event version, that's fine and surely has other advantages as well as disadvantages, but it is a process definition that should be discussed.

duglin: ping @fabiojose ^^

duglin:
Approved on the Jan 31st call.

Issue #365, opened 2019-01-10
How should we deal with Deprecated events


mbernard:
It would be nice to have a way to tell the consumer  that an event version is now deprecated or will be in the futur
That way, they can subscribe to a new version of the event.

Do you think this should be part of an extension?

Tapppi:
I think an extension would be appropriate to address the acute issue and then we can see where it goes. A normalised way of notifying consumers of deprecation "on the line" would obviously help them set up alerts on deprecations more easily, even if all producers don't support the extension (at all or in the same exact way). The downside, or worry, is that we would prescribe too much convention before the _"broader solution"_ @duglin referred to gets even a chance of trying to formulate something in this area.

Because of the fine line we'd have to walk here, I think the extension should be merely a well-known, but loosely defined, attribute. More formally, the extension:
* should define a common attribute for signalling deprecation
* should not require prescribed meaningful content for the attribute
* could define a way of discerning different content, for example through an auxiliary attribute similar to [the `sequenceType` of the sequence extension](https://github.com/cloudevents/spec/blob/master/extensions/sequence.md#sequencetype).

duglin:
of those who spoke up, it seems an extension might be the preferred choice (if we do anything). Would someone want to take a stab at a PR to force the discussion/decision?

Tapppi:
@duglin I'll try to make a proposal for an extension, but we'll see whether I have cycles before next weeks call.

duglin:
@Tapppi thanks a ton!!

duglin:
@Tapppi or anyone else... still interested in this?

duglin:
I'm going to suggest we close this since no one seems to have any interest in coming up with a proposal.  Please speak-up before Thursday if you'd like to see it remain open

duglin:
Agreed to close this on the 10/10 call

PR #382, opened 2019-01-30
Deprecation extension proposal

Tapppi:
A draft of an extension defining attributes indicating deprecation of events or properties/versions/schemas etc. I don't feel particularly strong about anything in the draft, except the general direction of leaving things open-ended for development with actual use-cases instead of prescribing specifics. See #365 for more discussion.

Ideas:
* Comment explicitly on middleware and their role in signaling deprecation (of delivery method etc.). I feel like this is so-so on "prescribing specifics"
* Elaborate on versioning by linking to [primer section on versioning]() for the benefit of giving an idea of how this field and versioning might together provide a way to evolve events. Might simplify description of what this field is meant to mean to deprecate (anything) and instead point to primer.

### DIFF
--- a/extensions/deprecation.md
+++ b/extensions/deprecation.md
@@ -1,0 +1,36 @@
+# Deprecation
+
+This extension defines two attributes that can be included within a CloudEvent
+to indicate the deprecation of the event, delivery method, format or other
+properties, such as the `schemaUrl`.
+
+The `deprecated` attribute indicates deprecation of the event delivery. The
+exact value and meaning of this attribute can be defined by the
+`deprecatedType` attribute. As the specification of the attributes is left
+intentionally open-ended, event consumers will need to have some out-of-band
+communication with the event producer to understand how to interpret the value
+of the attribute.
+
+Event producers should use established or well-known patterns to facilitate
+integration. In addition to or instead of integrating specific patterns,
+consumers can also choose to alert on deprecation and rely on human
+intervention.
+
+## Attributes
+
+### deprecated
+* Type: `String`
+* Description: Value indicating the deprecation of the delivery of these events
+* Constraints
+  * REQUIRED
+  * MUST be a non-empty string
+  * RECOMMENDED to provide information on type and timeline of deprecation and
+    a way to find more information
+
+### deprecatedType
+* Type: `String`
+* Description: Used to specify the format and meaning of the `deprecated`
+  attribute
+* Constraints:
+  * OPTIONAL
+  * If present, MUST be a non-empty string


duglin:
Can you provide an example in the doc?

extensions/deprecation.md
```diff
@@ -0,0 +1,36 @@
+# Deprecation
+
+This extension defines two attributes that can be included within a CloudEvent
+to indicate the deprecation of the event, delivery method, format or other
+properties, such as the `schemaUrl`.
```
cneijenhuis:
```suggestion
properties, such as the `schemaurl`.
```

extensions/deprecation.md
```diff
@@ -0,0 +1,36 @@
...
+
+The `deprecated` attribute indicates deprecation of the event delivery. The
+exact value and meaning of this attribute can be defined by the
+`deprecatedType` attribute. As the specification of the attributes is left
```
cneijenhuis:
```suggestion
`deprecatedtype` attribute. As the specification of the attributes is left
```

extensions/deprecation.md
```diff
@@ -0,0 +1,36 @@
...
+### deprecated
+* Type: `String`
+* Description: Value indicating the deprecation of the delivery of these events
+* Constraints
+  * REQUIRED
+  * MUST be a non-empty string
+  * RECOMMENDED to provide information on type and timeline of deprecation and
+    a way to find more information
+
+### deprecatedType
```
cneijenhuis:
```suggestion
### deprecatedtype
```

duglin:
Can you include an example in there to help people understand what kind of strings will be used?

duglin:
@Tapppi did you want to address the comments or wait for some discussion during the call?

clemensv:
When I look at this through a code lens, for example drawing a parallel to the [Obsolete] attribute in C# or the @Deprecated annotation in Java, then I absolutely get the intent, but when I think about an attribute rendered in an encoding that hits the wire, I'm not sure what that extension gives me.

I either receive an attribute or I don't. On the wire and between the code of apps, that choice is typically binary.

This proposal seems to introduce a "you might not get this value tomorrow, but here you have it anyways" middle ground and I don't know what running code is going to do with it? Shall it write a warning into the log? What if I don't even care about the particular attribute, but the publisher includes it and the deprecation note anyways? Can a publisher flood my log that way?

evankanderson:
Additionally, it seems as though the schemaURL might be the correct place to report field information, rather than within each message (which will likely lead to a lot of overhead and duplication, and is unlikely to be useful directly within the consuming code).

duglin:
@Tapppi What's the status of this one? There seems to be open comments

duglin:
Two things...
1 - @Tapppi - did you want to make the edits suggested by @cneijenhuis ? They seems minor/accurate
2 - what does the group think about this extension?

duglin:
@Tapppi what's the status of this one?

duglin:
@Tapppi you still interested in this or should we close it?

duglin:
@Tapppi I'm going to suggest we close this since I don't see a lot of traction or updates being made based on the comments.  We can always reopen later if needed.

duglin:
We agreed to close this on the 8/1 call - we can reopen if necessary

ttripp:
This was closed some time ago.  Are there any newer discussions on this same topic?

duglin:
Nope, not that I can recall

vandewillysilva:
Is there any update on this topic, or has it been addressed in another form within the CloudEvents spec?

duglin:
this was closed due to inactivity (e.g. there were pending change requests and questions ask that were unanswered). If someone wants to revive it that's fine but we'd need someone to address those issue.

vandewillysilva:
I’m reviving the discussion on adding a deprecation specification and have proposed an extension for it. Please take a look and share your thoughts. https://github.com/cloudevents/spec/pull/1307

cc @ttripp @duglin

PR #383, opened 2019-01-30
Fix eventtype -> type reference in it's description

Tapppi:
After renaming of `eventtype` there is a dangling reference to the old name

No normative changes.


duglin:
thanks - I noticed this the other day too, just never got around to PRing it.
LGTM

Merging since it's just a typo - we can revisti if people think otherwise.

PR #386, opened 2019-02-05
Consistency: schemaurl uses URI-reference, protobuf uses URI-reference

### LINKED ISSUE #384: URI vs URI-reference
Found an issue when URI was changed to be URI-reference, the definition of URI was removed but the usage of URI remains in schemaurl.

see https://github.com/cloudevents/spec/blob/master/spec.md#schemaurl for usage of URI

cneijenhuis:
Closes #384

In https://github.com/cloudevents/spec/pull/169 it was decided that inside the specification, we only use `URI-reference`. (From reading the PR, it is not clear to me whether it was intended to change both `schemaurl` and `source`, or only `source`, but the PR effectively changed both).

In https://github.com/cloudevents/spec/pull/338 I renamed the shorthand `URI` to `URI-reference`, because it was easy to wrongly assume only URIs are supported, but not URI references.
I missed changing the `schemaurl` type.
When `schemaurl` was introduced to `spec.json`, that was mixed up as well.

Related, the `protobuf-format.md` type mapping wasn't consistent with the type naming in `spec.md` (I think both PRs ran concurrently), which I'm fixing as well.

---

There is probably a discussion to be had whether https://github.com/cloudevents/spec/pull/169 actually intended to change `schemaurl` - the constraint suggests otherwise - and if we want to have both `URI` and `URI-reference` in the type system.

### DIFF
--- a/protobuf-format.md
+++ b/protobuf-format.md
@@ -78,15 +78,15 @@ The CloudEvents type system MUST be mapped into the fields of
 `CloudEventAny` as follows:
 
 
-| CloudEvents  | CloudEventAny field
-|--------------|-------------------------------------------------------------
-| String       | string_value
-| Binary       | binary_value
-| URI          | string_value (string expression conforming to URI-reference as defined in [RFC 3986 §4.1](https://tools.ietf.org/html/rfc3986#section-4.1))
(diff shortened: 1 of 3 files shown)


duglin:
Approved on the 2/7 call.

Issue #371, opened 2019-01-18
Event bridges cannot tell if structured JSON data is base64 encoded.


alanconway:
I am implementing a bridge between different event transport bindings and formats:
https://godoc.org/github.com/alanconway/scratch/lightning
I don't know how to correctly determine if a JSON structured data attribute is base64 encoded or not. 

Spec says: "If the contenttype media type is known to contain text, the data attribute value is not further interpreted and treated as a text string. Otherwise, it is decoded and treated as a binary value."

I'm not aware of a reliable way to determine that a MIME type is "known to contain text". RFC1341 provides an explicit Content-Transfer-Encoding when data is encoded, it does not expect mail clients to "just know". My nasty hack is to use this regexp to detect "non-text" media-types, clearly not reliable or future-proof:
     /(^$)|(^text$)|(^text/)|(^application.*[/+-.](xml|json|yaml|javascript|html|text|wbxml)[/+-.]([/+-.]|$))/

I think the JSON format should have a "contenttransferencoding" attribute (rfc1341 compliant) - as should future formats that can't handle raw binary data. It doesn't have to be part of the abstract event, but it would reduce re-encoding if it was: base64 encoded data could flow unmodified through different binary and structured event bindings and only be decoded when the body is actually required.

This is related to #261 but I think deserves separate treatment since it affects the core use case of events - accurately forwarding event data - even if you don't care about corner cases like binary values in Maps.

duglin:
I'm kind of surprised that there's no way to encode this information into the Content-Type header.
@clemensv any thoughts on this?

alanconway:
On Thu, Jan 24, 2019 at 8:24 AM Doug Davis <notifications@github.com> wrote:

> I'm kind of surprised that there's no way to encode this information into
> the Content-Type header.
> @clemensv <https://github.com/clemensv> any thoughts on this?
>

The MIME RFC is pretty clear on that. Base64 is a Content-Transfer-Encoding
which is orthogonal to a MIME type. Content-Transfer-Encoding is an extra
layer of encoding for transports with limitations on binary data (e.g. JSON
or email) - it is a feature of how you're sending the data, not the type
that the data represents. Content-Transfer-Encoding is also used for things
like adding compression, which again is not a feature of the type, but of
how you've chosen to transfer it.

clemensv:
As much as it pains me - @alanconway is right.

[...]

PR #387, opened 2019-02-15
Added datacontentencoding

clemensv:
Implements content encoding as raised in issue #371

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -276,20 +276,26 @@ help intermediate gateways determine how to route the events.
 
 ### datacontentencoding
 * Type: `String` per [RFC 2045 Section 6.1](https://tools.ietf.org/html/rfc2045#section-6.1)
-* Description: Describes the content encoding for the `data` 
+* Description: Describes the content encoding for the `data`
   attribute for when the `data` field MUST be encoded as a string,
-  like with structured transport binding modes using the JSON event 
-  format, but the `datacontenttype` indicates a non-string media 
+  like with structured transport binding modes using the JSON event
+  format, but the `datacontenttype` indicates a non-string media
   type. When the `data` field's effective data type is not `String`,
   this attribute MUST NOT be set and MUST be ignored when set.
   
   The "Base64" value for the Base64 encoding as defined in [RFC 2045 Section 6.8](https://tools.ietf.org/html/rfc2045#section-6.8)
   MUST be supported. When set, the event-format-encoded value of the `data` 
-  attribute is a base64 string, but the effective data type of 
+  attribute is a base64 string, but the effective data type of
   the `data` attribute towards the application is the base64-decoded
   binary array.
   
   All other RFC2045 schemes are undefined for CloudEvents.
+
+* Constraints:
+  * The attribute MUST be set if the `data` attribute contains string-encoded
+    binary data. Otherwise the attribute MUST NOT be set.
+  * If present, MUST adhere to [RFC 2045 Section 6.1](https://tools.ietf.org/html/rfc2045#section-6.1)
+
 
 ## Data Attribute
 
--- a/spec.md
+++ b/spec.md
@@ -235,7 +235,7 @@ help intermediate gateways determine how to route the events.
     [RFC 3339](https://tools.ietf.org/html/rfc3339)
 
 ### schemaurl
-* Type: `URI`
+* Type: `URI-reference`
 * Description: A link to the schema that the `data` attribute adheres to.
   Incompatible changes to the schema SHOULD be reflected by a different URL.
   See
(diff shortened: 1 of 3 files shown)


duglin:
ping @alanconway @n3wscott
Any thoughts on this?

alanconway:
The datacontentencoding changes are all good by me.

On a side note: I'm surprised that critical attribute names are still changing, e.g. "contenttype -> datacontenttype". Also since the spec has adopted the cantdecideonwordseparatorssoletshavenone convention, it seems like a not-so-great idea to be adding more words to multi-word names. However, that's another story.

duglin:
Would an example help clarify the need for this to a newbie?  Could be done in a follow-on PR if so.

spec.md
```diff
@@ -276,6 +274,23 @@ help intermediate gateways determine how to route the events.
     [RFC 2046](https://tools.ietf.org/html/rfc2046)
 * For Media Type examples see [IANA Media Types](http://www.iana.org/assignments/media-types/media-types.xhtml)
 
+### datacontentencoding
+* Type: `String` per [RFC 2045 Section 6.1](https://tools.ietf.org/html/rfc2045#section-6.1)
+* Description: Describes the content encoding for the `data` 
+  attribute for when the `data` field MUST be encoded as a string,
+  like with structured transport binding modes using the JSON event 
+  format, but the `datacontenttype` indicates a non-string media 
+  type. When the `data` field's effective data type is not `String`,
```
n3wscott: > When the `data` field's effective data type is not `String`,
>  this attribute MUST NOT be set and MUST be ignored when set.

why not? I want to zip my http binary data payload...

n3wscott: Or I want to base64 encode my binary http payload because I am a weirdo...

n3wscott: Perhaps not, looking at RFC 2045 I do not see anything I would want to use besides base64.  And normal http transports can give me gzip... 

/lgtm

alanconway:
On Wed, Feb 20, 2019 at 8:38 PM Scott Nichols <notifications@github.com>
wrote:

> *@n3wscott* commented on this pull request.
>
> In spec.md
> <https://github.com/cloudevents/spec/pull/387#discussion_r258749647>:
>
> > @@ -276,6 +274,23 @@ help intermediate gateways determine how to route the events.
>      [RFC 2046](https://tools.ietf.org/html/rfc2046)
>  * For Media Type examples see [IANA Media Types](http://www.iana.org/assignments/media-types/media-types.xhtml)
>
> +### datacontentencoding
> +* Type: `String` per [RFC 2045 Section 6.1](https://tools.ietf.org/html/rfc2045#section-6.1)
> +* Description: Describes the content encoding for the `data`
> +  attribute for when the `data` field MUST be encoded as a string,
> +  like with structured transport binding modes using the JSON event
> +  format, but the `datacontenttype` indicates a non-string media
> +  type. When the `data` field's effective data type is not `String`,
>
> Perhaps not, looking at RFC 2045 I do not see anything I would want to use
> besides base64. And normal http transports can give me gzip...
>

Compression is a separate "transfer-encoding" header in HTTP.
Content-encoding is just for transorts that can't store simple binary data,
like 1960s 7-bit-ascii email, or (sigh) JSON in 2019.

spec.md
```diff
@@ -276,6 +274,23 @@ help intermediate gateways determine how to route the events.
...
+  this attribute MUST NOT be set and MUST be ignored when set.
+  
+  The "Base64" value for the Base64 encoding as defined in [RFC 2045 Section 6.8](https://tools.ietf.org/html/rfc2045#section-6.8)
+  MUST be supported. When set, the event-format-encoded value of the `data` 
+  attribute is a base64 string, but the effective data type of 
+  the `data` attribute towards the application is the base64-decoded
+  binary array.
+  
+  All other RFC2045 schemes are undefined for CloudEvents.
```
evankanderson: Add a "Constraints" section, like the other attributes?

I think at the moment, "Constraints" is "Must be the string 'Base64'". If so, would it make more sense to standardize on assuming 'Base64' encoding and excluding the other options?

duglin: @clemensv  ^^^

alanconway:
perhaps "contentencoding" rather than "datacontentencoding"? Given the
(unfortunate IMO) choice of noseparatorswhatsoever for attribute names, I
think it's important to try hard for names that are only one word, and
absolutely ban names that are more than two words.

clemensv:
@alanconway we chose to rename contenttype to datacontenttype because people were confused about what that relates to, and now this also needs to have data in front of it, because it relates to the same attribute.

duglin:
All - This one has been updated - please review and comment

rperelma:
Might be good to add an example

JemDay:
+1 for some examples - but maybe there should be a separate PR to add a few examples based on recent changes.

Generally LGTM but deferring final comment until merge conflicts are resolved.

duglin:
Approved on the 4/4 call

PR #389, opened 2019-02-19
SDK object model structure guidelines

clemensv:
AI: "write-up guidance about serializating & about binary mode into the SDK.md file"

I made this a bit more comprehensive and described how the OM ought to hang together. That also covers how the encoder plugs into the transports.

### DIFF
--- a/SDK.md
+++ b/SDK.md
@@ -44,62 +44,64 @@ Widely used programming language version.
 
 ### Object model structure guidelines
 
-The CloudEvents specification set consists of 
+The CloudEvents specification set consists of
 
 * the base specification that defines the abstract information model and
   associated rules for a CloudEvent,
-* a set of extensions to the base specification that define use-case 
+* a set of extensions to the base specification that define use-case
   specific attributes and associated rules,
 * a set of event format specifications that define how a CloudEvent and  
-  its attributes are encoded in a particular wire transfer format,
+  its attributes are encoded in a particular wire transfer format, and
 * a set of transport bindings that define how a wire-encoded CloudEvent is
-  interoperably transferred from a publisher or intermediary to a target
+  interoperably transported from a publisher or intermediary to a target
   endpoint using a particular application protocol.
 
 The object model of any API projection of CloudEvents MUST reflect the
-specification architecture:
+specification architecture with the following elements:
 
 * A generic CloudEvent class/object/structure that provides access to the
   CloudEvent attributes and enforces data type constraints and all further
   conformance rules defined in the core CloudEvents specification. Instances
   of this class/object/structure MUST always be in a conformant state.
   The generic CloudEvent abstraction SHOULD provide strongly typed access
-  to all well-known attributes, be MUST also provide name-indexed access
+  to all well-known attributes, and MUST also provide name-indexed access
   to any application-specific extension attributes that the event might carry.
   Name-indexed access MUST include well-known attributes from the core
-  specification and from application-activated specification extensions (see 
-  below), and the related rules MUST be enforced when setting the attributed 
-  through the name-indexed access path.
-* A generic abstraction for extensions that allows for one or more active 
-  extensions to be added to the CloudEvent abstraction such that the 
-  extension-defined attributes are attached to the CloudEvent and 
-  conformance can be validated. 
+  specification and from application-activated specification extensions (see
+  below). The related rules MUST be enforced when setting the attribute 
+  values through the name-indexed access path.
+* A generic abstraction for extensions that allows for one or more active
+  extensions to be added to the CloudEvent abstraction such that the
+  extension-defined attributes are attached to the CloudEvent and
+  the conformance of their values can be validated.
   Each SDK MUST implement all extensions defined in the CloudEvents
   repository. "Active" extensions are those that the application explicitly
-  selects for use, meaning that other extensions MUST NOT be added and enforced
-  implicitly. Any application MAY ignore a particular extension and even 
-  use well-known extension attribute names for semantically different purposes.
-* An encoder abstraction and implementation for each event format which turns 
-  a CloudEvent instance and its active extensions into a conformant 
-  wire-representation and vice versa. If defined in the event format, the 
+  selects for use, meaning that other extensions MUST NOT be added and 
+  enforced on a CloudEvent implicitly. Any application MAY ignore a particular
+  extension and even use well-known extension attribute names for semantically
+  different purposes.
+* An encoder abstraction and implementation for each event format which turns
+  a CloudEvent instance and its active extensions into a conformant
+  wire-representation and vice versa. If defined in the event format, the
   encoder MUST be able to render the entire event in a structured transport
   representation, and MUST also be able to render individual attribute values
   for a binary transport representation.
 * At least one implementation for each transport binding which integrates with
   the most commonly used transport implementation for the respective language
-  and runtime. If there are multiple equally popular platform libraries for 
+  and runtime. If there are multiple equally popular platform libraries for
   the same transport, the SDK SHOULD provide implementations for each.
-  The transport binding implementation MUST NOT obscure or encapsulate the 
+  The transport binding implementation MUST NOT obscure or encapsulate the
   regular interaction with the respective platform API for its core send
-  and receive functionality; instead, the SDK SHOULD provide an extension or 
+  and receive functionality; instead, the SDK SHOULD provide an extension or
   complement to the respective platform API that fits the respective API style
-  and allows for mapping a CloudEvent from and to the respective application 
+  and allows for mapping a CloudEvent from and to the respective application
   protocol message/event.
   Because selecting and detecting structured or binary transfer modes is a
-  transport level choice, the respective mapping function MUST be given a 
-  reference to the chosen event format for mapping to the application protocol
-  message/event. The reference MAY also be given through a default configuration 
-  choice. When mapping from the application protocol message/event, the SDK 
+  transport level choice, the respective mapping function MUST be given a
+  reference to the desired event format encoder for mapping to the application
+  protocol message/event. This reference MAY also be given through a default
+  configuration choice.
+  When mapping from the application protocol message/event, the SDK
   SHOULD identify and choose the required event format encoder from the content
   type information of the protocol, if available on the protocol.
 
--- a/SDK.md
+++ b/SDK.md
@@ -76,8 +76,7 @@ specification architecture with the following elements:
   the conformance of their values can be validated.
   Each SDK MUST implement all extensions defined in the CloudEvents
   repository. "Active" extensions are those that the application explicitly
-  selects for use, meaning that other extensions MUST NOT be added and 
-  enforced on a CloudEvent implicitly. Any application MAY ignore a particular
+  selects for use. Any application MAY ignore a particular
   extension and even use well-known extension attribute names for semantically
   different purposes.
 * An encoder abstraction and implementation for each event format which turns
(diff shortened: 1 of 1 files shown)


SDK.md
```diff
@@ -44,6 +44,68 @@ Supports CloudEvents spec milestones and ongoing development version.
...
+specification architecture with the following elements:
+
+* A generic CloudEvent class/object/structure that provides access to the
+  CloudEvent attributes and enforces data type constraints and all further
+  conformance rules defined in the core CloudEvents specification. Instances
+  of this class/object/structure MUST always be in a conformant state.
+  The generic CloudEvent abstraction SHOULD provide strongly typed access
+  to all well-known attributes, and MUST also provide name-indexed access
+  to any application-specific extension attributes that the event might carry.
+  Name-indexed access MUST include well-known attributes from the core
```
duglin: While I like this MUST because it provides consistency to the user of the SDK for all attributes (core or extensions), I wonder if forcing all SDKs to expose the core CE properties via a map is too much of a burden?

SDK.md
```diff
@@ -44,6 +44,68 @@ Supports CloudEvents spec milestones and ongoing development version.
...
+  specification and from application-activated specification extensions (see
+  below). The related rules MUST be enforced when setting the attribute 
+  values through the name-indexed access path.
+* A generic abstraction for extensions that allows for one or more active
+  extensions to be added to the CloudEvent abstraction such that the
+  extension-defined attributes are attached to the CloudEvent and
+  the conformance of their values can be validated.
+  Each SDK MUST implement all extensions defined in the CloudEvents
```
duglin: What does this mean to "implement" an extension? They at least appear in an extension bag (or something), but does this mean you want them to appear as first-class properties too? If they're just in a bag then I'm not sure what an SDK needs to do to satisfy this MUST.

SDK.md
```diff
@@ -44,6 +44,68 @@ Supports CloudEvents spec milestones and ongoing development version.
...
+  repository. "Active" extensions are those that the application explicitly
+  selects for use, meaning that other extensions MUST NOT be added and
```
duglin: Not sure what this "Active" extensions sentence is trying to say or enforce. Can you elaborate on its intent?

SDK.md
```diff
@@ -44,6 +44,68 @@ Supports CloudEvents spec milestones and ongoing development version.
...
+  encoder MUST be able to render the entire event in a structured transport
+  representation, and MUST also be able to render individual attribute values
+  for a binary transport representation.
+* At least one implementation for each transport binding which integrates with
+  the most commonly used transport implementation for the respective language
+  and runtime. If there are multiple equally popular platform libraries for
+  the same transport, the SDK SHOULD provide implementations for each.
+  The transport binding implementation MUST NOT obscure or encapsulate the
+  regular interaction with the respective platform API for its core send
+  and receive functionality; instead, the SDK SHOULD provide an extension or
```
duglin: Is this SHOULD inconsistent with the previous MUST NOT? Seems they should both be SHOULDs or both be MUSTs, no?

SDK.md
```diff
@@ -44,6 +44,68 @@ Supports CloudEvents spec milestones and ongoing development version.
...
+  selects for use, meaning that other extensions MUST NOT be added and 
+  enforced on a CloudEvent implicitly. Any application MAY ignore a particular
+  extension and even use well-known extension attribute names for semantically
+  different purposes.
+* An encoder abstraction and implementation for each event format which turns
+  a CloudEvent instance and its active extensions into a conformant
+  wire-representation and vice versa. If defined in the event format, the
+  encoder MUST be able to render the entire event in a structured transport
+  representation, and MUST also be able to render individual attribute values
+  for a binary transport representation.
```
n3wscott: I am not sure you will understand how to support binary for any given transport. I think you need the transport details.

SDK.md
```diff
@@ -44,6 +44,68 @@ Supports CloudEvents spec milestones and ongoing development version.
...
+* At least one implementation for each transport binding which integrates with
+  the most commonly used transport implementation for the respective language
+  and runtime. If there are multiple equally popular platform libraries for
```
n3wscott: define popular?

n3wscott: Also that is a huge burden on SDK authors.

SDK.md
```diff
@@ -44,6 +44,68 @@ Supports CloudEvents spec milestones and ongoing development version.
...
+  the same transport, the SDK SHOULD provide implementations for each.
+  The transport binding implementation MUST NOT obscure or encapsulate the
+  regular interaction with the respective platform API for its core send
+  and receive functionality; instead, the SDK SHOULD provide an extension or
+  complement to the respective platform API that fits the respective API style
+  and allows for mapping a CloudEvent from and to the respective application
+  protocol message/event.
+  Because selecting and detecting structured or binary transfer modes is a
+  transport level choice, the respective mapping function MUST be given a
```
n3wscott: MUST is wrong here, I can have that function have a default that can be overridden.

SDK.md
```diff
@@ -44,6 +44,68 @@ Supports CloudEvents spec milestones and ongoing development version.
...
+  reference to the desired event format encoder for mapping to the application
+  protocol message/event. This reference MAY also be given through a default
+  configuration choice.
+  When mapping from the application protocol message/event, the SDK
```
n3wscott: in my experience, the event is not enough information to select an encoder, the sdk user needs to have a default or be able to select, as mentioned in the above stipulations.

SDK.md
```diff
@@ -44,6 +44,68 @@ Supports CloudEvents spec milestones and ongoing development version.
...
+  interoperably transported from a publisher or intermediary to a target
+  endpoint using a particular application protocol.
+
+The object model of any API projection of CloudEvents MUST reflect the
+specification architecture with the following elements:
+
+* A generic CloudEvent class/object/structure that provides access to the
+  CloudEvent attributes and enforces data type constraints and all further
+  conformance rules defined in the core CloudEvents specification. Instances
+  of this class/object/structure MUST always be in a conformant state.
```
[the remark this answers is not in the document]
evankanderson: Additionally, looking at the "constructor" and builder pattern in the API recommendation, this looks unlike idiomatic golang; I assume that idiomatic LISP would also look fairly different from the recommendation.

duglin:
Open comments and rebased needed

SDK.md
clemensv: The rationale is that if an extension gets promoted into the core, you break existing code if that extension then vanishes from the indexed access collection. So this rule is for protecting against breaking changes for that scenario -- and for the promotion of a extension into the core I assume that the extension is popular and therefore you break a lot of code.

SDK.md
clemensv: Implementing an extension means that all constraints are checked. And for extensions that we have in the repo, the SDKs should have strongly typed object model and check those constraints.

SDK.md
clemensv: @duglin "active" means just that this is an extension implementation the app uses with a particular CloudEvent instance. If the app doesn't interact with the properties of an extension there's no reason to enforce rules, and extensions can also be in mutual conflict. That rule is poorly worded, though.

SDK.md
clemensv: The second should be a MUST, not the first.

SDK.md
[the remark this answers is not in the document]
clemensv: The constructor should force you to provide all required fields or auto-fill them, IMO.

SDK.md
```diff
@@ -44,6 +44,68 @@ Supports CloudEvents spec milestones and ongoing development version.
...
+  The generic CloudEvent abstraction SHOULD provide strongly typed access
+  to all well-known attributes, and MUST also provide name-indexed access
```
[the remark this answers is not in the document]
clemensv: The C# SDK gives you name indexed access based on the set version of the event. The names change when you change the version. "subject" would thus be an extension in 0.1/0.2. This is exactly for evolving events over time, because with mandatory indexed access you can get at a "2.0" core attribute with the "1.0" SDK even if there's no strongly typed wrapper for it.

SDK.md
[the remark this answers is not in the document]
clemensv: The factory pattern isn't all that uncommon. What we have here is a data structure that has mandatory fields and the data structure is incorrect if those mandatory fields are not set, even in memory.

SDK.md
```diff
@@ -44,6 +44,68 @@ Supports CloudEvents spec milestones and ongoing development version.
...
+  a CloudEvent instance and its active extensions into a conformant
+  wire-representation and vice versa. If defined in the event format, the
+  encoder MUST be able to render the entire event in a structured transport
+  representation, and MUST also be able to render individual attribute values
+  for a binary transport representation.
+* At least one implementation for each transport binding which integrates with
+  the most commonly used transport implementation for the respective language
+  and runtime. If there are multiple equally popular platform libraries for
+  the same transport, the SDK SHOULD provide implementations for each.
+  The transport binding implementation MUST NOT obscure or encapsulate the
```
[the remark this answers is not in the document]
clemensv: The upside of that approach is that you get full access to all the HTTP client capabilities, i.e. you get to set the Authorization header to a custom token format or you get to set custom headers to get past your outbound firewall or negotiate CORS.

SDK.md
duglin: yep - I understand the reason for it, I'm just trying to weigh the pros/cons of actually mandating it with a MUST vs just recommending it with a SHOULD. For example, if there's an SDK that wants to ensure strict compliance to the spec to ensure maximum interop might even go so far as to ban extensions, and therefore might only expose well defined getters/setters for the spec-defined attributes. They would be in violation of this rule even though they're CE-spec compliant.  I'm also not sure how contrived this example is but it's what I'm wondering about...

SDK.md
duglin: I'm interpreting "strongly types object model" as "well defined getters/setters" - is that correct? If so, if we require all SDKs to implement all repo-defined extensions (e.g. `datadef`) are we then not elevating them to the same level as OPTIONAL spec-defined attributes? Our bar for extensions is pretty low and forcing all SDKs to support them sounds like we should take them more seriously than just "experimental" attributes.

SDK.md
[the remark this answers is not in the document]
deissnerk: I agree with @clemensv, but to me it is important that the SDK lets me interact directly with the regular platform API. I don't care, if it provides an encapsulation **in addition** to that.

fabiojose:
@clemensv reading the SDK.md I saw idiomatic (.net) methods signatures in the section "Preferable API signature guidelines". For java developers (like me), a method name following `UpperCamelCase` is so strange, like a horse with basket shoes :horse: :joy:.

Each language has its own idiomatic way of naming.

Example (method naming):

- Java: `lowerCamelCase`
- Python: `snake_camel_case`
- .Net: `UpperCamelCase`
- Golang: `lowerCamelCase`
- Javascript: `lowerCamelCase`

```bash
# Java
v01.event()
  .setContentType("application/json")

# Python
v01.event()
  .set_content_type("application/json")

# .Net
v01.Event()
  .SetContentType("application/json")

# Golang
v01.event()
  .setContentType("application/json")

# JavaScript (without set prefix)
v01.event()
  .setContentType("application/json")
```

I suppose that definition in the "Preferable API signature guidelines" are optional and each SDK could employ their idiomatic way for naming conventions.

If am I right, I would like to suggest an edit with idiomatic examples or any other kind of more explicit information that I am free to employ the language naming conventions.

evankanderson:
One correction:

For Golang, exported methods all start with InitialCaps, not lowercase.

SDK.md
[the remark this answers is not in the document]
n3wscott: Go does not use constructors.

PR #390, opened 2019-02-19
Primer edits - Architecture section 

clemensv:
I have an open AI that has nagged me each week for the better part of the last decade to add some more content to the primer. I want to be a little careful with doing that because we've got a lot of authors on this document already, so I'll be doing smaller chunks. 

This PR adds an architecture summary of how the specs relate, which I don't think we have yet. There's a parallel PR (#389) for the SDK doc that is worded somewhat differently and presents how that same layering ought to be reflected in the SDK class/object/structure model.

### DIFF
--- a/primer.md
+++ b/primer.md
@@ -139,7 +139,7 @@ elements that form a layered architecture model.
 1. The [base specification](spec.md) defines an abstract information model
    made up of attributes (key-value pairs) and associated rules for what 
    constitutes a CloudEvent.
-2. The [extensions](./spec.md#extension-attributes) add use-case specific
+2. The [extensions](./spec.md#extension-context-attributes) add use-case specific
    and potentially overlapping sets of extension attributes and associated
    rules, e.g. to support different tracing standards.
 3. The event format encodings, e.g. [JSON](json-format.md), define how the
--- a/primer.md
+++ b/primer.md
@@ -136,11 +136,12 @@ The following will not be part of the specification:
 The CloudEvents specification set defines four different kinds of protocol
 elements that form a layered architecture model.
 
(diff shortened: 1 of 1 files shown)


primer.md
```diff
@@ -128,6 +129,34 @@ The following will not be part of the specification:
...
 * Selecting a single identity/access control system
 
+
+## Architecture
+
+The CloudEvents specification set defines four different kinds of protocol
+elements that form a layered architecture model.
+
+1. The [base specification](spec.md) defines an abstract information model and
+   associated rules for what constitutes CloudEvent.
```
cneijenhuis: It might be good if `1.` introduces "attributes", otherwise they are a bit abruptly introduced in `2.`.

primer.md
```diff
@@ -128,6 +129,34 @@ The following will not be part of the specification:
...
+2. The [extensions](./extensions) add use-case specific and potentially
```
cneijenhuis: From an architectural standpoint, I agree we should introduce extensions separately.

However, I don't think this link should go to `./extensions`, but to `./spec.md#extension-attributes`. The spec defines what extension attributes are. The extensions folder contains a few "standardized" extensions, but from an architectural point of view, it is irrelevant whether an extension is defined in the repo, or completely custom.

duglin:
LGTM - thanks for doing this!

duglin:
@clemensv did you want to address the comments?

duglin:
ping @clemensv as a remind for the outstanding comments

clemensv:
Addressed the comments

duglin:
LGTM

duglin:
Approved on the 6/20 call

Issue #331, opened 2018-10-18
Clarify scope of eventID's uniqueness


duglin:
On today's call, while discussing #326, it was mentioned that the uniqueness statement in eventID might need some clarity. Right now the spec says that it: `MUST be unique within the scope of the producer` but what constitutes the "producer" could vary and there was a discussion around how combining the `source` and `eventID` properties should produce a globally unique value - but the spec doesn't say this. We should decide what kind of clarity we want to add around this - if any.

alanconway:
My take on this was that source + eventID is unique because eventID is unique in the scope of a producer, and the source attribute identifies the producer. With valid DNS-scoped URIs assigned to producers that won't step on each others toes, that is internet-unique. Applications with lesser requirements can assign source IDs to suit their needs.

 I'm just getting into the specs tho so may be off track.

Tapppi:
I do not believe eventID can be required to be unique in a meaningful way, either in combination with source or any other attribute. The spec does not prescribe what the `source` or `type` (if that was included in the "unique combination") should be (other than `type` SHOULD be prefixed with reverse-DNS and `source` MUST be a `URI-reference`) and especially doesn't say that `source` should be DNS-scoped or scoped uniquely. I also don't think that makes any sense to require considering the complexity of events passing boundaries of internal/external systems, networks etc. The spec describes `source` like this (emphasis mine):
> This describes the event producer. Often this will include information such as the type of the event source, the organization publishing the event, the process that produced the event, and some unique identifiers. **The exact syntax and semantics behind the data encoded in the URI is event producer defined.**

---

As an example, consider an open-source Change Data Capture project producing cloudevents from actions on a database. The relevant fields might look something like this and be compliant with the current spec:
```
{
  "type": "io.CatDogCat.db.row.delete/v1",
  "source": "db-name/orders/order",
  "eventID": "123123123",
  ...
  "data": {
    "before": {"id": "987"},
    "after": null,
    ...
  }
}
```

Even though the project can allow distinct source names for many databases with the same `schema/table` names by prefixing the source with `db-name`, there is no way for them to guarantee or require uniqueness of the database name across deployments by *whoever happens to use that project*. Now, if we require any kind of global uniqueness for any combination of the attributes, the project cannot guarantee to produce compliant CloudEvents without integrating with some kind of repository keeping track of source names or whatever. It could add a `UUID` to the source, but that would destroy usability for users of the project. 

[...]

Issue #326, opened 2018-10-05
make eventId an optional field


jric:
In our organization eventId is an optional field.  Therefore, we do not use the CloudEvent spec as defined.  I would prefer to fix that impedance.  I hope either I can convince cloudevent spec to change or someone here can convince me that we're doing it wrong.

Reasons to make it optional:

1. DRY (don't repeat yourself) principle:  There is usually some combination of fields in any event which can be used to identify a "unique" event.  In our org, we are not interested in any event which could be confused with another but for the id.  We always want to know:  Where did the event originate, at what time, and what type of event was it.  Combined, that information always uniquely defines an event, so we don't need another id.  We just need the data and to know which fields hold that data.  Then, if we want, we can always combine (e.g. hash) that data later to create a unique composite key.
2. KISS (keep it stupid simple):  One of our principles is to keep things as simple as possible, especially at the interface.  No eventId means fewer fields for consumers to manage, and fewer fields for producers to create or worry about generating correctly.
3.  Single source of truth:  If there is both an eventId and another unique key, what if those two don't agree?  On the plus side, the id can serve as a "checksum" of sorts.  However, generating the eventId has to be done correctly.  It must represent the producer's "idea" of what constitutes a unique event, and there is no documentation of that logic, because it's hidden in the algorithm that chooses when to mint a new id, rather than being explicit in the data.  For instance, it would probably be a bug if each communication retry generated a new "eventId", or if two events were generated with the same id, but different payloads, e.g. in a multi-threaded race condition.

Reasons to keep it required:

I'm not sure what to say here.  I think there are a class of tools that will be able to operate much more simply once there is a single field that represents the event.  However, I think those tools could simply require the id at that point, rather than requiring every event to have the same eventId throughout its lifecycle.  In case of an event pipeline where an event forks into multiple events, this id cannot stay static anyway.

jric:
If this proposal is approved, I will submit a PR.

duglin:
@jric  are you able to join one of our weekly calls? They're on Thursdays at 12pm ET.
I'm asking because we typically focus on PRs first but if there's an additional topic (such as an issue like this one) then we can add it to the agenda but its always better to have the proponent of the idea on the call to present their thinking.

You're also free to wait for feedback on the issue itself, but I know folks are pretty busy and so it might be a while before that happens.

cneijenhuis:
Here is my try at convincing you :)

[...]

Issue #290, opened 2018-08-16
Immutability of event context


deissnerk:
In the discussion of #277 I raised the question, if extension attributes can be changed or added by middleware, or if the event context is immutable. I remember that during the f2f meeting in Copenhagen @clemensv proposed to have two property bags, one for properties added by the source, and one for annotations added by middleware. AFAIK this is quite similar to what AMQP 1.0 supports. If we decide to abandon the idea of property bags for extensions, how do we handle the question of information added to the context by middleware? There are quite a few options:

1. Don't support information added by middleware. An event is immutable. If anything is changed, it becomes a new event with a new ID.
2. Allow adding of context attributes by any component on the way.
3. Add a dedicated property bag for information that is added after event creation. This could be part of the context or on the same level as data and context.

cneijenhuis:
A question regarding 1. - if the middleware produces a new event, then the middleware is the (new) event producer. Therefore, wouldn't it be forced to set itself as the `source`?

On one hand, that looks right (one shouldn't be allowed to "impersonate" another event producer), on the other hand it will pose problems to the consumer (expected "example.org/user/...", got "middleware.com/foo").

Vlaaaaaaad:
Regarding option 1( immutable events), if a new event is to be created two options seem to be available:
- make a bad clone of the initial event( reuse time and event id for example, see _Eliminating the who - no middleware concept idea_ section in https://github.com/cloudevents/spec/pull/128#issuecomment-395239721) 
- some type of embedding would have to be done which would leak transport and middleware details into the event


The security( signing and encryption) discussion will also have an impact on this. Also, https://github.com/cloudevents/spec/issues/266 seems to be related

duglin:
Do we even need to make a statement about this at all? Whether a middleware acts like "a new source" or it acts more like a proxy is really up to it. I'm not sure CloudEvents itself needs to say anything either way.

There might be some best practices that middleware might want to think about/do - but for the most part it seems like our spec should not get into the processing model side of things, and stick with just how things appear on the wire.

deissnerk:
@jdstuart I agree with your assessment of the options. My preference would also be option 3. 

@Vlaaaaaaad I also agree to you. I have my doubts about middleware cloning the original event. If an event is propagated through multiple channels, you might end up seeing different "versions" of the same event. I also agree on the nesting option.

@cneijenhuis The questions regarding the event source is a good point. 

@duglin Our design goals state:
[...]

PR #391, opened 2019-02-20
Clarify scope of eventID's uniqueness

alanconway:
This is for issue #331

Signed-off-by: Alan Conway <aconway@redhat.com>

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -165,10 +165,14 @@ The following attributes are REQUIRED to be present in all CloudEvents:
 #### id
 
 - Type: `String`
-- Description: ID of the event. The semantics of this string are explicitly
-  undefined to ease the implementation of producers. Enables deduplication.
+- Description: Identifies the event.
+  Producers MUST ensure that `source` + `id` is unique for each
+  distinct event.  If a duplicate event is re-sent (e.g. due to a
+  network error) it MAY have the same `id`.  Consumers MAY assume that
+  Events with identical `source` and `id` are duplicates.
 - Examples:
-  - A database commit ID
+  - An event counter maintained by the producer
+  - A UUID
 - Constraints:
   - REQUIRED
   - MUST be a non-empty string
@@ -177,18 +181,37 @@ The following attributes are REQUIRED to be present in all CloudEvents:
 #### source
 
 - Type: `URI-reference`
-- Description: This describes the event producer. Often this will include
-  information such as the type of the event source, the organization publishing
-  the event, the process that produced the event, and some unique identifiers.
-  The exact syntax and semantics behind the data encoded in the URI is event
-  producer defined.
+- Description: Identifies the context in which an event
+  happened. Often this will include information such as the type of
+  the event source, the organization publishing the event or the
+  process that produced the event. The exact syntax and semantics
+  behind the data encoded in the URI is defined by the event producer.
+
+  Producers MUST ensure that `source` + `id` is unique for each
+  distinct event.
+
+  An application MAY assign a unique `source` to each distinct
+  producer, which makes it easy to produce unique IDs since no other
+  producer will have the same source. The application MAY use UUIDs,
+  URNs, DNS authorities or an application-specific scheme to create
+  unique `source` identifiers.
+
+  A source MAY include more than one producer. In that case the
+  producers MUST collaborate to ensure that `source` + `id` is unique
+  for each distinct event.
+
 - Constraints:
   - REQUIRED
 - Examples
-  - https://github.com/cloudevents/spec/pull
-  - /cloudevents/spec/pull
-  - urn:event:from:myapi/resource
-  - mailto:cncf-wg-serverless@lists.cncf.io
+  - Internet-wide unique URI with a DNS authority.
+    - https://github.com/cloudevents
+    - mailto:cncf-wg-serverless@lists.cncf.io
+  - Universally-unique URN with a UUID:
+    -  urn:uuid:6e8bc430-9c3a-11d9-9669-0800200c9a66
+  - Application-specific identifiers
+    - /cloudevents/spec/pull/123
+    - /sensors/tn-1234567/alerts
+    - 1-555-123-4567
 
 #### specversion
 


spec.md
```diff
@@ -202,11 +202,11 @@ help intermediate gateways determine how to route the events.
 
 ### source
 * Type: `URI-reference`
-* Description: This describes the event producer. Often this will include
-  information such as the type of the event source, the organization
-  publishing the event, the process that produced the event, and some unique
-  identifiers. The exact syntax and semantics behind the data encoded in the URI
-  is event producer defined.
```
cneijenhuis: Why are you dropping these two sentences? They give a lot of motivation for why a URI is chosen at all. If you don't want to `include information such as...`, then you could also go straight for a UUID and be done with it 😉 

I think it is very valuable to describe what to put into the `source`, and why a hierarchical data structure was chosen.

spec.md
```diff
@@ -202,11 +202,11 @@ help intermediate gateways determine how to route the events.
...
+* Description: Identifies the event producer. If the source includes a
+  URI authority [RFC3986  §3.2](https://tools.ietf.org/html/rfc3986#section-3.2)
+  then it SHOULD be a valid, internet-unique URI for the producer. If there is no
```
cneijenhuis: I'm not sure what `valid` is supposed to say. It MUST be valid according to RFC3986 already.
If you mean valid as in "I should be able to make a request towards this URI, and end up at the producer" - I don't think this is a good idea. Take any GraphQL API.
I'm taking GitHub as the example, assume for a second that there is no web interface. `https://api.github.com/graphql` is then the only valid URL, but it would be a valid choice to have `https://api.github.com/cloudevents/spec/123` for the `source`.

spec.md
```diff
@@ -202,11 +202,11 @@ help intermediate gateways determine how to route the events.
...
+  authority, then it is unique within some application-specific scope defined
```
cneijenhuis: s/it is unique/it SHOULD be unique/

spec.md
```diff
@@ -242,6 +245,7 @@ help intermediate gateways determine how to route the events.
   [Versioning of Attributes in the Primer](primer.md#versioning-of-attributes)
   for more information.
 * Constraints:
+  * MUST be a URL [RFC 3986](https://tools.ietf.org/html/rfc3986#section-1.1.3)
```
cneijenhuis: I'd advise you to keep this PR focused on the eventID and open a separate PR for this line, if you want to push it.

(If you do, I'd also suggest to add `URL` to the type system instead of awkwardly constraining the `URI-reference` type - or alternatively change MUST to RECOMMEND.)

spec.md
```diff
@@ -224,7 +224,10 @@ help intermediate gateways determine how to route the events.
 * Constraints:
   * REQUIRED
   * MUST be a non-empty string
-  * MUST be unique within the scope of the producer
+  * MUST be unique within the scope of the producer. In other words:
+    events with identical "scope" and "id" values MUST be duplicates
```
cneijenhuis: Did you mean `source`, not `scope`?

Anyway, I agree with https://github.com/cloudevents/spec/issues/331#issuecomment-463754907 - scope should be both `source` and `type`.

E.g. `com.github.pull.create` and `/cloudevents/spec/pull/123` together form the scope of the producer.

spec.md
```diff
@@ -224,7 +224,10 @@ help intermediate gateways determine how to route the events.
...
+    that can be discarded without examining any other attribute
+    values.
```
cneijenhuis: The MUST reads strange from a consumer perspective. As a consumer, I can treat them as duplicates, but there is no guarantee they really are. Counter-proposal (including what I've said above):

> In other words: events with identical `id`, `source` and `type` values CAN be discarded without examining any other attribute values.

spec.md
alanconway: To me, the important thing here is that source can be a URI with an internet-unique authority so you don't have to resort to UUIDs for uniqueness. As I read it, the only normative use of the source is to uniquely identify a producer.

The design of source URIs will depend on the application. It might include "information such as type/organization/process..." but it might not. This spec doesn't seem a good place for advice on URI design, which is a topic in its own right.

I'm ok with putting it back if you feel strongly about it.

spec.md
alanconway: "valid" is shoddy, will reword. Definitely did not mean to imply a URL.

spec.md
```diff
@@ -117,7 +117,7 @@ The following abstract data types are available for use in attributes.
 - `Map` - `String`-indexed dictionary of `Any`-typed values.
 - `Any` - Either a `Binary`, `Integer`, `Map` or `String`.
 - `URI-reference` - String expression conforming to `URI-reference`
-  as defined in
+  (a URL, URI or relative reference) as defined in
```
duglin: I believe a URI-reference already covers all 3 doesn't it?

spec.md
duglin: I prefer the existing text because it's a little less strict in that the value doesn't necessarily have to be a valid address on the web. For example, they could use `http://myserver/....` is that is ok for their setup. We purposely didn't want to get too precise here to allow for flexibility.

spec.md
alanconway: WIll fix source/scope typo.

I prefer that source+id be unique and type be orthogonal. 

A source defines exactly the scope within which ids are unique. If you need ids to be unique by type, then you can create a source per type. If you don't, then a single source can emit events of many types. In either case the consumers need only consider source+id for de-duplication.

spec.md
alanconway: Will open a new PR for this, if I still care about it by the time I've applied your other comments :)

spec.md
duglin: not sure source+type is always unique...  `/cloudevents/spec/pull/123`+`com.github.com.comment` is not unique. Wouldn't ID need to be in there some place?

alanconway:
Updated to deal with most of the comments above, left some conversations unresolved where I need more feedback.

spec.md
alanconway: Check my latest text. I want to make it clear that you MAY have an internet-unique authority and URI here but you also MAY also have something of application specific scope. Otherwise it's hard to answer the original question "what is the scope in which source+id is unique" in any meaningful way.

spec.md
alanconway: I agree: source+id is unique. Nothing else should be required/assumed to be unique. type is a separate issue and best not conflated with deduplication. If you want to conflate sources and types you can (e.g. include type info in the source URI and only emit events of that type) but you don't have to. Consumers should not be required to include type in the calculation for de-duplication.

cneijenhuis: @duglin Yes, the ID needs to be in there! I only meant to say source+type=scope, and scope+id is unique.

cneijenhuis: > Consumers should not be required to include type in the calculation for de-duplication.

@alanconway You apparently feel strongly about this. Why? Whether I have to use two or three fields for de-duplication doesn't seem to make a huge difference to me.

spec.md
```diff
@@ -202,11 +202,13 @@ help intermediate gateways determine how to route the events.
...
-  publishing the event, the process that produced the event, and some unique
-  identifiers. The exact syntax and semantics behind the data encoded in the URI
-  is event producer defined.
+* Description: Identifies the event producer.
+  The source MAY include a URI authority
+  [RFC3986  §3.2](https://tools.ietf.org/html/rfc3986#section-3.2)
+  associated with the producer, providing an internet-wide unique identifier.
+  If not, it is unique within some application-specific scope.
+  The details of a source URI are determined by the producer, the consumer
+  SHOULD treat it as an opaque identifier.
```
cneijenhuis: I recommend that you try to stick with the minimal changes needed to achieve eventIDs uniqueness, and not change other things without a clear need.

The last sentence is unnecessary, and - in my opinion - wrong.

Of course a consumer CAN treat the source as a URI, and not as an opaque identifier!
E.g. I can decide that I'm interested in `/cloudevents/spec/*`, but not in `/cloudevents/website/*`, or that I'm interested in `/cloudevents/*/pull/*`, but not in `/cloudevents/*/issue/*`.

From the perspective of deduplication, it is treated as an opaque identifier, but deduplication isn't the only aspect of the spec.

spec.md
duglin: resolved??

spec.md
duglin: @alanconway can you elaborate on what problem you're trying to solve with this change? In the end, aren't these all just opaque identifiers since we're not asking the receiver to do anything with these URI-references - like re-reference them?

spec.md
```diff
@@ -224,7 +226,12 @@ help intermediate gateways determine how to route the events.
 * Constraints:
   * REQUIRED
   * MUST be a non-empty string
-  * MUST be unique within the scope of the producer
+  * MUST be unique within the scope of the producer. A producer MUST
+    NOT generate events with identical "source" and "id" values,
+    unless it is generating an identical duplicate of an event that
+    may have been lost due to a transport failure.  A consumer SHOULD
```
duglin: s/may/might/

duglin: Does the SHOULD put a big burden on the consumer? How long do they need to hold on to the history of events? I'm wondering if it wouldn't be better to be a "MAY" instead so that we're not forcing people to do anything, rather, just making it clear that spec-wise it's safe to drop dups. Although, there's part of me that wonders if this is even something we should get into at all since we've tended to avoid processing semantics. I think I'd feel more comfortable with:
- producer MUST ensure that source+id is unique for each unique event
- a replayed event (or is it "message", @clemensv ?) MAY include a previously used source+id IFF it is semantically the same
- a consumer MAY interpret an event with a previously seen source+id as a replay of that event

spec.md
alanconway: I'm trying to clarify the scope of uniqueness, the "U" in URI. URIs have a standard authority component which provides well-known, internet-wide uniqueness guarantees. I'm trying to clarify that a "source" can be URI-unique if it has a URI authority OR it can be an authority-less reference path which is only unique in some application-defined context. Both are useful uniqueness guarantees for different applications. 

We could leave it unstated since  it's implied by using a URI-reference, but this PR is requesting "Clarify scope of eventID's uniqueness" so it seems to need saying.

duglin: My interpretation of the request for clarity around uniqueness was more around simple string comparison type of things :-)  So, while the semantics behind these values might be interesting in some cases, I think what's more important (from a spec perspective) is that people know if they do the equivalent of `ce.source + ce.id` they'll get a unique value within the scope of that producer, and can use it appropriately.  Getting into the semantics of the strings, while interesting, doesn't really change what the code a consumer would write - I don't think anyway.

spec.md
alanconway: I argue that id and type should be separate both for simplicity and clear semantics

simplicity:
- 2 fields is better than 3 unless there's value in including the 3rd. What's the value?
- when efficiency is a concern you can generate short IDs, but you can't control the length of type names - which may be an orders of magnitude longer.

semantics:
- "id" is for de-duplication, it has no other reason to exist. You also need "source" if there are many sources, because separate sources can't co-operate to generate unique ids.  You don't need anything else.
- "type" is for classifying events for processing. That has nothing at all to do with deduplication. Sources can produce events of one type or many.

I've seen type and id conflated before, I've done it myself :) The early thinking is "it'll be easier for producers to generate unique IDs per type." The reality that hits later is that it is harder for consumers to use IDs that have to be coupled with a type to mean anything. 

Consumers tend to naturally collect messages by source, then they just want a simple hashable ID to check duplicates. If you include the type you force a  bunch of awkward code around:

    type UsableID struct { Type, ID string }

Depending on the language you might have to define hash functions or comparisons to use these as map keys etc  (you can't just concatenate type and id into a single string because that might clash with some other type-id combo, unless you know something about the syntax of type and ID.) It's all just a pain.

Conflating type with ID is certainly implementable, but IMO makes implementation harder and less efficient.

spec.md
alanconway: +1 the important thing here is that the producer MUST NOT generate events with identical IDs unless it is OK for the consumer to ignore one of them. The consumer can make its own decisions about how it handles duplicates.

spec.md
alanconway: +1 restoring original comments.

alanconway:
I've minimized the changes and updated in line with comments - I think the only open point of discussion is whether type is part of the dedup or not. As my changes read now it is not.

spec.md
alanconway: +1 I've restored the original comments on "source" - with the exception of saying "identifies" rather than "describes" and changed the "id" part along the lines you suggested.

spec.md
```diff
@@ -224,7 +224,9 @@ help intermediate gateways determine how to route the events.
 * Constraints:
   * REQUIRED
   * MUST be a non-empty string
-  * MUST be unique within the scope of the producer
+  * The producer MUST ensure that source+id is unique for each distinct event.
```
duglin: I like the overall direction of this PR. Do we need to maintain the concept of the scope of the uniqueness? Meaning, is it globally unique vs just within the scope of the producer?

alanconway: I added some explanatory text/examples on scope of uniqueness for different forms of `source` URI reference. Does that cover it?

cneijenhuis:
```suggestion
  * The producer MUST ensure that source+id is unique for each distinct event within their scope.
```

Otherwise it reads as if the producer MUST have consumed all existing events to be sure that no one else has already produced an event with that source and id.

spec.md
```diff
@@ -202,29 +202,45 @@ help intermediate gateways determine how to route the events.
...
-* Description: This describes the event producer. Often this will include
+* Description: This identifies the event producer. Often this will include
   information such as the type of the event source, the organization
   publishing the event, the process that produced the event, and some unique
-  identifiers. The exact syntax and semantics behind the data encoded in the URI
-  is event producer defined.
+  identifiers.
+
+  The exact syntax of `source`, and the scope in which it is unique,
+  depends on the application. Applications range from a single service
```
cneijenhuis: I don't think we should talk about "application" here, but only about event producer.

See https://github.com/cloudevents/spec/blob/master/primer.md#design-goals especially `and later can be connected to create new applications`.

An (implementation of an) event producer MAY be part of many applications. The author of the event producer may not know what applications it will be part of after a few years.

alanconway:
On Tue, Feb 26, 2019 at 12:03 PM Christoph Neijenhuis <
notifications@github.com> wrote:

> *@cneijenhuis* commented on this pull request.
>
> In spec.md
> <https://github.com/cloudevents/spec/pull/391#discussion_r260388648>:
>
> >    information such as the type of the event source, the organization
>    publishing the event, the process that produced the event, and some unique
> -  identifiers. The exact syntax and semantics behind the data encoded in the URI
> -  is event producer defined.
> +  identifiers.
> +
> +  The exact syntax of `source`, and the scope in which it is unique,
> +  depends on the application. Applications range from a single service
>
> I don't think we should talk about "application" here, but only about
> event producer.
>
It makes no sense to say that 'source' is unique in the scope of a producer
- the source *identifies* the producer. The only way an event consumer can
tell if events are from "the same producer" or "different producers" is to
compare source fields. So uniqueness of source is a consideration for
larger application design.

> See https://github.com/cloudevents/spec/blob/master/primer.md#design-goals
> especially and later can be connected to create new applications.
>
An (implementation of an) event producer MAY be part of many applications.
> The author of the event producer may not know what applications it will be
> part of after a few years.
>
That's exactly why I mention application. We know that scope+id MUST be
unique for events from a single producer. If your application consists of a
single producer and directly connected consumers,  then the source name can
be anything you like and you are done.

To build applications that include multiple producers and loosely coupled
event delivery (routing, store/forward etc.)  producers must use 'source'
names that will be unique across all the producers that might ever have
events routed to the same consumer. Otherwise there is no way for the
*consumer* to know whether events with the same source+id are duplicates,
or distinct events from different producers that happen to use the same
source name.

spec.md
duglin: I think that new text does help - thanks. And I agree with @cneijenhuis on using "event producer" instead of 'app". My last concern is placement of that text since it seems to apply to both `source` and `id` sections. Perhaps just say "The exact syntax of `source`, and the scope in which it (and `id`) is unique,...." to add a forward ref to make it clear the unique text applies to `id` as well.  Just a thought.

cneijenhuis:
@alanconway I think we agree on:
* Having two different events with the same identification is a failure state
* Someone should to be responsible to make sure that doesn't happen

What we don't agree on is who should be responsible. You're saying it should be the Event Producer. I don't agree.

Let me present a few practical examples:

* I'm creating IoT devices to be embedded into rooms. Each IoT device is configured to know what building and room it is in. A `source` may be `http://acmeinc.com/buildings/1/rooms/100/sensors/temperature/1`. How can the IoT device (the event producer) know whether there is another IoT device with the same `source`, or not? It can't. The responsibility to configure all IoT devices correctly does not lie with the IoT device itself.
* I'm creating an Open Source project producing events. In the config, I'm asking for a unique URI. If two groups inside `ACME Inc` decide to deploy separate instances of the project with both `acmeinc.com` as the URI, is it the responsibility of the Open Source project to figure that out?
* I'm a cloud provider. I'm allowing you to completely tear down a service (say, a blob storage bucket) and re-create it from scratch with the same name, as if it never existed. I'm clearly documenting that if you re-create the service with name, events might clash.
  * In a test deployment, I can tearing down all my services, including the event producer (the bucket) and all consumers. Everything is fine. People enjoy that feature and frequently re-create buckets for integration testing.
  * If someone decides to tear down only the bucket, but keep the consumers alive, things will go wrong.
  * Again, it is not the event producer/cloud providers fault that this was done.
* An event producer is restored from backup, with old data. How is the event producer supposed to know which id's have already been sent?

All the event producer can do is to promise that it'll be unique within their scope. You can not make the event producer responsible for the whole application, because the event producer likely isn't in charge of it.

spec.md
evankanderson: It should be possible to ensure that the source+id combination is globally unique using either of the following techniques:

1. By using an authority in the source which is assigned to the source via an authoritative registry (e.g. DNS).
2. By using a GUID as a component of the eventID.

spec.md
evankanderson: If the ID is semantically meaningful (for example, a TrueTime timestamp or a vector clock), then middleware may need to preserve the ID when transforming an event's `type`. In this case, `source` + `id` may need to be non-unique, or we could guide that the middleware should set embed the originating `id` into a longer `id` related to the source in such a way that the recipients know how to extract the originating `id`.

alanconway:
@cneijenhuis I think we agree on everything, but my wording was messy - have a look at the update.

Here's what I'm trying to say:

1. Producers MUST generate unique "id" attributes for their own messages only.
2. Producer identifiers (i.e. "source" attribute) MUST be unique in a wider context that I'm calling the "application". We don't define how that's accomplished, but we give examples of standard URL and URN options, as well as simple application-defined strings or paths.
3. Given 1+2, consumers can safely assume that events with identical source+id are duplicates.

evankanderson:
Thinking about this a bit more, it seems like it should be the eventid+source+id triple that is unique.

Firestore is a good example: for a document update, it may fire some of:

Event Type | Trigger
-- | --
onCreate | Triggered when a document is written to for the first time.
onUpdate | Triggered when a document already exists and has any value changed.
onDelete | Triggered when a document with data is deleted.
onWrite | Triggered when onCreate, onUpdate or onDelete is triggered.

Note that `onWrite` and `onUpdate` may fire with the same source for the same occurrence. Using the same event ID allows downstream consumers to correlate `onWrite` events with creates and updates, for example.

alanconway:
On Sat, Mar 2, 2019 at 2:24 AM Evan Anderson <notifications@github.com>
wrote:

> Thinking about this a bit more, it seems like it should be the
> eventid+source+id triple that is unique.
>

What is "eventid"? My understanding is that the event context attribute
named "id", defined here:
https://github.com/alanconway/cloudevents-spec/blob/master/spec.md#L218
Is an identifier for the event - within the scope of a given producer,
identfied by the attribute named "source".
I don't see "eventid" in the spec.

> Firestore is a good example: for a document update, it may fire some of:
> Event Type Trigger
> onCreate Triggered when a document is written to for the first time.
> onUpdate Triggered when a document already exists and has any value
> changed.
> onDelete Triggered when a document with data is deleted.
> onWrite Triggered when onCreate, onUpdate or onDelete is triggered.
>
> Note that onWrite and onUpdate may fire with the same source for the same
> occurrence. Using the same event ID allows downstream consumers to
> correlate onWrite events with creates and updates, for example.
>
It is the responsibility of a source implementation to ensure unique "id"
values per event. If there are concurrent event streams then there are 2
choices:
1. a single source serializes the concurrent events into a single stream,
and generates unique ids for them all
2. concurrent event streams are represented as separate sources to preserve
concurrency and eliminate the need for synchronization.

What's the benefit of adding a 3rd layer to the scheme?

cneijenhuis:
@alanconway Sorry for the late reply. To follow your number-list:

On 3.: I agree, this is the point where the spec needs to be more specific, and clearly spell out when a consumer can deduplicate a message.

On 1.: Yes. But that is already in the spec.

On 2.: I agree we should provide guidance on that. But I think the primer is a much better place for it. We can only advise readers on how to set up their application properly, so that they don't run into problems given 3.
But putting a MUST into the spec is a pretty high bar. I fear the effect is that event producers will have to jump through additional hoops to comply with such a MUST, despite them (in many situations) having little control over the application as a whole.

cneijenhuis:
Also, to go back to the discussion started in #331 : Like @Tapppi I am not convinced that source + id is a good choice to clarify "scope".

It also goes against the implicit design of the spec, given by the current examples.

The current examples emphasize a unique type, but not necessarily a unique source. Given the examples, I would design a type of `org.example.user.created`, and a source of `/users/123`. I wouldn't go for `https://example.org/users/123`, especially if no such URL exists (maybe the current URL is `https://api.example.org/v2/users/123`, but what happens if I change to `v3` or - like GitHub - deprecate my REST-ish API in favor of GraphQL?).

I think using DNS for uniqueness is a good idea in general, but I favor the convention of using reverse-DNS, as done in many programming languages for this use case. A package name `com.acmeinc` is as unique `http://acmeinc.com`, but my class `Foo` reads much nicer as `com.acmeinc.Foo` than `http://acmeinc.com/Foo`, and it doesn't imply that I could somehow access `Foo` via that URL.

Furthermore, I think it is much more important for the event `type` to be globally unique than for the `source`. I have no data to back this up, but I'm predicting that message routing will happen primarily on `type`, and only secondarily on `source`.
As an example, as a consumer (not middleware!) I'd choose to consume events from `org.example.*` or `com.acmeinc.*` - if they both happen to define a `/user/123`, I won't notice. Even if I consume events from both, I'll likely treat those events completely differently, so the non-globally-unique source won't be a problem in practice.

Summary: I think the spec, as it is currently written, implicitly favors type+source+id as the "unique" scope. Design-wise, I personally prefer trying to make the `type` the thing that ensures globally uniqueness.

duglin:
@alanconway is this one ready for review or did you want to reply to @cneijenhuis's comments first?

spec.md
```diff
@@ -202,29 +202,49 @@ help intermediate gateways determine how to route the events.
...
 
 ### id
 * Type: `String`
-* Description: ID of the event. The semantics of this string are explicitly
-  undefined to ease the implementation of producers. Enables deduplication.
+* Description: Identifies of the event, enables de-duplication.  The
+  format of this string is determined by the producer. Each producer
+  MUST generate unique `id` values for its own events, `id` values
+  from different producers might clash. Consumers MAY assume that
+  events with identical `id` and `source` values are duplicates, and
```
n3wscott: I would say that `eventtype` is also required to understand if the event is a duplicate.

So: `id` + `eventtype` + `source` combine to be unique.

alanconway:
On Thu, Mar 7, 2019 at 12:18 PM Scott Nichols <notifications@github.com>
wrote:

> *@n3wscott* commented on this pull request.
>
> In spec.md
> <https://github.com/cloudevents/spec/pull/391#discussion_r263483818>:
>
> >
>  ### id
>  * Type: `String`
> -* Description: ID of the event. The semantics of this string are explicitly
> -  undefined to ease the implementation of producers. Enables deduplication.
> +* Description: Identifies of the event, enables de-duplication.  The
> +  format of this string is determined by the producer. Each producer
> +  MUST generate unique `id` values for its own events, `id` values
> +  from different producers might clash. Consumers MAY assume that
> +  events with identical `id` and `source` values are duplicates, and
>
> I would say that eventtype is also required to understand if the event is
> a duplicate.
>
> So: id + eventtype + source combine to be unique.
>
>
I don't think type belongs in de-duplication. Type describes a class of
events with similar semantics, it does not define where they come from.
Many independent producers can produce events of the same type.

The only agent that can reliably generate unique IDs is a producer, because
it *produces* the events. A single producer can only ensure it's own IDs
are unique, so to check for duplicates you absolutely must look at:
- the id, because that's the only bit of data that varies on a per-event
basis.
- the source, because that identifies the producer, and different producers
generate ids independently.

There's no reason to bring type into it - a producer that generates unique
IDs is responsible for never using the same ID on non-duplicate events, and
by definition duplicate events will have the same type (and every other
attribute that matters)

That said, if the consensus is to include type we can. I don't see any
benefit to it, but apart from the extra complexity it doesn't cause a
problem.

alanconway:
I think it warrants review to decide if it makes sense as it stands based
on using source+id.

We need to work thru the current discussion about adding type as part of
the identifier. It should be straightforward to update to the text once
we've reached a consensus.

duglin:
@evankanderson wrote:
> it seems like it should be the eventid+source+id triple that is unique.

I think there's a typo in there since `eventid` and `id` are the same, no?  Just `eventid` is the old name for `id`.

---
I tend to agree with @alanconway that including `type` is unnecessary and could actually be problematic if someone chooses to send two messages with the same `source` and `id` but different `type`s. Given that `id` is meant to be unique for the same producer, I'm not sure how someone is supposed to interpret these two messages.

evankanderson:
Sorry, that was a typo, I meant `eventtype + source + id`

cathyhongzhang:
I also agree with @alanconway : 
"There's no reason to bring type into it - a producer that generates unique
IDs is responsible for never using the same ID on non-duplicate events, and
by definition duplicate events will have the same type (and every other
attribute that matters)"

It creates unnecessary complication for the event consumer if we add type. Since the spec already defines that event ID is unique within the scope of the event source, then as long as we ensure source is globally unique, then "source+eventID" will make the event globally unique.   

As to the example of Firestore, it seems the problem can be solved by assigning different event IDs to 
onWrite and onUpdate firing with the same source.

duglin:
Just a reminder - the spec includes this in the definition of `id`: MUST be unique within the scope of the producer

As long as this is true, `type` should not be needed to determine uniqueness for events from the same producer. I opened this clarifying PR: https://github.com/cloudevents/spec/pull/403

n3wscott:
> Just a reminder - the spec includes this in the definition of `id`: MUST be unique within the scope of the producer
> 
> As long as this is true, `type` should not be needed to determine uniqueness for events from the same producer. I opened this clarifying PR: #403

And if there are two producers acting on behalf of a non-cloudevent producing event producer, you will be unable to understand who produced the cloudevent.

duglin:
> And if there are two producers acting on behalf of a non-cloudevent producing event producer, you will be unable to understand who produced the cloudevent.

If there are two CE generators acting on behalf of the same event producer, then each CE generator better make sure the `id`s they generate are unique (per the spec) and a consumer can then de-dup based on that (if `source` is the same). `type` provides no added value in that check.

It seems to me there might be some ambiguity in the spec as to whether each "occurrence" gets a unique ID or each event/CE? And this PR is trying to clarify that it's the event/CE not the occurrence.

n3wscott:
>  each CE generator better make sure the ids they generate are unique (per the spec)

The spec says to pull the id from the event body if possible. So this is conflicting advice. If you don't do this you have to de-dup on the internal id and ignore the cloudevent id. At which point, why have an id?

spec.md
```diff
@@ -202,29 +202,49 @@ help intermediate gateways determine how to route the events.
...
+  * Application-specific path or string identifiers
+    * /cloudevents/spec/pull/123
+    * /sensors/tn-1234567/alerts
+    * 1-555-123-4567
 
 ### id
 * Type: `String`
-* Description: ID of the event. The semantics of this string are explicitly
-  undefined to ease the implementation of producers. Enables deduplication.
+* Description: Identifies of the event, enables de-duplication.  The
```
deissnerk:
```suggestion
* Description: Identifies the event, enables de-duplication.  The
```

spec.md
```diff
@@ -202,29 +202,49 @@ help intermediate gateways determine how to route the events.
...
+  format of this string is determined by the producer. Each producer
+  MUST generate unique `id` values for its own events, `id` values
+  from different producers might clash. Consumers MAY assume that
```
deissnerk: What is the difference between a producer and a source? Is it about unique ìd` values per source?

deissnerk: Sorry, I guess you can ignore this comment. I just checked the definition of source. 😊

n3wscott:
> As to the example of Firestore, it seems the problem can be solved by assigning different event IDs to onWrite and onUpdate firing with the same source.

The answer can't be Firestore is wrong and would have to rewrite to be cloudevents compliant, they are a valid implementing of a working eventing system and, like many other examples we have come up with, use the same id to emit events that are related or causal.

duglin:
I agree that Firestore isn't wrong. However, the Firestore-to-CE mapping is if it uses the same value in two different CE's for the ID attribute.  Per the CE spec, each CE needs to have a unique ID from the same source. Due to this, attempting to use that ID as some kind of mechanism to relate multiple events together would be wrong and impossible. For that people should define/use some kind of "relatesTo" or "originalTransationalID" type of attribute.

evankanderson:
Aren't we defining the CE spec here? So we could change the uniqueness requirements of `id` to cover both producer and event type if we desire.

Logically, this does seem to be a question for #403. _Should we discuss this there, or here?_

I'd like to understand the design concern about specifying which combination of fields defines the uniqueness of the `id`. Right now, we claim that the `id` "MUST be unique within the scope of the producer", but don't define how that scope can be determined. #406 starts to provide clearer semantics, but without a clear motivation for this uniqueness, it's hard to gauge the scope of this property. For example, can I claim that I have different "producers" for event type A and event type B which happen to have overlapping `source` properties? If so, then a consumer may see non-unique `id` values for the same `source`, but that's okay because "The exact syntax and semantics behind the data encoded in the URI is event producer defined." I'm basically trying to find an [adversarial implementation](https://nullprogram.com/blog/2017/05/03/) here to better understand the principles we're trying to provide for event consumers and middleware.

Related to this are #290 (immutability of event context) and the possible context attribute modifications that middleware may make without changing `id`. (If I choose to de-duplicate based on `id` and receive two events which were enriched by different middleware, will the final event properties I react to depend on which event was received first?)

duglin:
The way I tend to look at it like this:
- we shouldn't be too prescriptive on things otherwise it'll limit how our spec can be used
- however, yes we should be clear on our design/intent for each attribute
- I think id being unique for all CEs per source is clear and sufficient for what it's trying to do - just provide a unique ID for the stream of CEs from this source. It is not meant to uniquely identify anything else, like the occurrence or "subject"
- Therefore, I think id+source being unique is clear and sufficient

I think this is running into issues because some people are wanting to use it for other purposes and I'm not keen on overloading the semantics because it'll introduce complexity. If people need a field for those other semantics then I'd prefer to start by defining a new attribute and if it turns out it'll have the exact same semantics/definition/lifecycle/... as 'id' then we can merge the two.  But I'd first like to clearly see the definition and usecases for that attribute.

duglin:
@alanconway I know there's still some high-order discussion going on, but did you want to accept some of the syntax edits proposals?  btw - are you able to join our weekly call? It might be good for you to join the live-chats on this.

alanconway:
I've been on vacation - will accept the syntax edits an get on the next
call.

duglin:
@alanconway can you rebase this?

On last week's call we discussed this a bit and the overall consensus of the group is that `id` + `source` should be sufficient for uniqueness. There didn't appear to be a desire to add `type` into the mix.  However, there was a brief discussion around how using non-unique `source` values could lead to duplicates - so a value like `producer` would not be wise, but a DNS name or UUID would be. I believe the text in this PR here (https://github.com/cloudevents/spec/pull/391/files#diff-958e7270f96f5407d7d980f500805b1bR189) tries to address this.

Overall, what do people think about this? Aside from the merge-conflict I think it's ready for review/consideration - I'll add it to the call this week, but if you have concerns please voice them in the PR.

alanconway:
Rebased and tightened the text a bit.

spec.md
```diff
@@ -231,14 +241,17 @@ within the same JSON object.
 ### id
 
 - Type: `String`
-- Description: ID of the event. The semantics of this string are explicitly
-  undefined to ease the implementation of producers. Enables deduplication.
+- Description: Identifies the event for de-duplication.
+  Producers MUST ensure that each distinct event from the same `source` has a unique `id`  If a duplicate event is re-sent due to a network error, it MAY have the same `id`.
```
duglin: can you wrap stuff at 80 cols?

alanconway:
On Wed, Apr 17, 2019 at 11:26 AM Doug Davis <notifications@github.com>
wrote:

> @alanconway <https://github.com/alanconway> can you rebase this?
>
Done, and tightened up the text a bit.

> On last week's call we discussed this a bit and the overall consensus of
> the group is that id + source should be sufficient for uniqueness. There
> didn't appear to be a desire to add type into the mix. However, there was
> a brief discussion around how using non-unique source values could lead
> to duplicates - so a value like producer would not be wise, but a DNS
> name or UUID would be. I believe the text in this PR here (
> https://github.com/cloudevents/spec/pull/391/files#diff-958e7270f96f5407d7d980f500805b1bR189)
> tries to address this.
>
Yep the new text is:

  An application MUST assign a distinct `source` to each distinct producer.
  The application MAY use UUIDs, URNs, DNS authorities or an
application-specific
  scheme to create unique identifiers.

There are examples of each approach.

> Overall, what do people think about this? Aside from the merge-conflict I
> think it's ready for review/consideration - I'll add it to the call this
> week, but if you have concerns please voice them in the PR.
>

On a related note, I added a plea to
https://github.com/cloudevents/spec/issues/326 - make eventId optional. It
was closed but I feel unjustly :)

alanconway:
Note: the CI build failure above is cause by a broken link to the cloudevent logo, nothing to do with my changes AFAIK.

spec.md
```diff
@@ -190,18 +190,28 @@ within the same JSON object.
...
-  the event, the process that produced the event, and some unique identifiers.
-  The exact syntax and semantics behind the data encoded in the URI is event
-  producer defined.
+- Description: Identifies the event producer. Often this will include
+  information such as the type of the event source, the organization
+  publishing the event, the process that produced the event, and some unique
+  identifiers. The exact syntax and semantics behind the data encoded in the URI
+  is event producer defined.
+
+  An application MUST assign a distinct `source` to each distinct producer.
```
Tapppi: On the call today (25.04.2019) we accepted the terminology in PR #420 for Source, Producer etc. This line contradicts that terminology, because a producer is defined as a specific instance creating events. Each logical system in which the occurrences are recorded/created ("Source" in the terminology), causing events, should have a distinct `source` attribute.

For example, two instances of a "Users service" would be distinct producers, but they *should* still have the same `source` field. Together they are a source and have a distinct `source` attribute compared to another distinct source, say a "Routing service".

Oddly enough, IIUC it should say:
> An application MUST assign a distinct `source` to each distinct source.

Writing this out, I realise this is as much a problem with the new terminology clashing with the attribute name and description. I am leaving a comment on #420 saying as much, and @duglin agreed to not merge it I think.

erikerikson: At least this line and perhaps (but not necessarily) [250](https://github.com/cloudevents/spec/pull/391/files#diff-958e7270f96f5407d7d980f500805b1bR250) appear invalidated by #420.

spec.md
```diff
@@ -190,18 +190,28 @@ within the same JSON object.
...
+
 - Constraints:
   - REQUIRED
 - Examples
-  - https://github.com/cloudevents/spec/pull
-  - /cloudevents/spec/pull
-  - urn:event:from:myapi/resource
-  - mailto:cncf-wg-serverless@lists.cncf.io
+  - Internet-wide unique URI with a DNS authority.
+    - https://github.com/cloudevents/spec/pull/123
```
evankanderson: Suggest that the example here uses `https://github.com/cloudevents/spec` (or some parent path), and move the rest of the URL into the `subject` property.

This makes it much easier for users to write meaningful source-based filters without needing to resort to wildcards.

spec.md
```diff
@@ -236,14 +246,17 @@ within the same JSON object.
...
 - Type: `String`
-- Description: ID of the event. The semantics of this string are explicitly
-  undefined to ease the implementation of producers. Enables deduplication.
+- Description: Identifies the event for de-duplication.
+  Producers MUST ensure that each distinct event from the same `source` has a unique `id`  If a duplicate event is re-sent due to a network error, it MAY have the same `id`.
+.
+  Consumers MAY assume that Events with identical `source` and `id` are duplicates.
 - Examples:
+  - An event counter maintained by the producer
   - A database commit ID
```
duglin: Can we remove this example (DB commit ID)? It's a bad example and will lead to confusion

duglin:
@alanconway on last week's call we agreed with this general direction. Could you address the merge conflict and any outstanding comments? The test associated in here might change based on @deissnerk's PR ( https://github.com/cloudevents/spec/pull/420 ) but we may just have to see which goes in first and then the other will have to adjust accordingly.   Also, we may need to adjust the Primer too.

spec.md
alanconway: Fixed in next update. It reads a little strangely now because the rest of the current master spec is written as though source == producer, but it will make sense when #420 goes in.

spec.md
alanconway: Fixed in next update

spec.md
alanconway: Fixed in next update - will make more sense when #420 is applied.

alanconway:
Updated and rebased. I have re-worded the source definition to use the terms defined in   #420  - so at this point this PR really needs to go in after #420 as currently the source definition talks about sources with many producers, which contradicts other places where producer==source.

spec.md
alanconway: resolving, please check latest and re-raise any issues.

spec.md
alanconway: resolving, please check latest and re-raise any issues.

spec.md
alanconway: resolving, please check latest and re-raise any issues.

spec.md
alanconway: done

spec.md
```diff
@@ -177,18 +181,37 @@ The following attributes are REQUIRED to be present in all CloudEvents:
...
+  Producers MUST ensure that `source` + `id` is unique for each
+  distinct event.
+
+  An application MAY assign a unique `source` to each distinct
+  producer, which makes it easy to produce unique IDs since no other
+  producer will have the same source. The application MAY use UUIDs,
+  URNs, DNS authorities or an application-specific scheme to create
+  unique `source` identifiers.
+
+  A source MAY include more than one producer. In that case the
```
duglin: is the word "include" the right word? Perhaps "use" or "leverage"?  Not a big deal since I think I get the point, but "include" might sound like we're getting include impl details and suggesting that a producer is a sub-component of a source - which it may not be.

duglin:
Just one minor question, otherwise LGTM

alanconway:
On Wed, May 15, 2019 at 3:21 PM Doug Davis <notifications@github.com> wrote:

> *@duglin* commented on this pull request.
>
> In spec.md
> <https://github.com/cloudevents/spec/pull/391#discussion_r284251054>:
>
> > +- Description: Identifies the context in which an event
> +  happened. Often this will include information such as the type of
> +  the event source, the organization publishing the event or the
> +  process that produced the event. The exact syntax and semantics
> +  behind the data encoded in the URI is defined by the event producer.
> +
> +  Producers MUST ensure that `source` + `id` is unique for each
> +  distinct event.
> +
> +  An application MAY assign a unique `source` to each distinct
> +  producer, which makes it easy to produce unique IDs since no other
> +  producer will have the same source. The application MAY use UUIDs,
> +  URNs, DNS authorities or an application-specific scheme to create
> +  unique `source` identifiers.
> +
> +  A source MAY include more than one producer. In that case the
>
> is the word "include" the right word? Perhaps "use" or "leverage"? Not a
> big deal since I think I get the point, but "include" might sound like
> we're getting include impl details and suggesting that a producer is a
> sub-component of a source - which it may not be.
>
>
> I'm fine with another word: "encompass", "comprise", "involve", "contain"?
The source isn't a thing in itself, it is a group of related producers, so
I'm not to sure about "use" or "leverage" - there's no separate source
object that acts on the producers. I don't object however.

duglin:
Let's just leave it as is then unless someone thinks of some better word.

duglin:
Approved on the 5/16 call

PR #402, opened 2019-03-14
Move "extension attributes" section down to end of context attributes

### LINKED ISSUE #401: Document structure leads to confusion between context and extension attributes
I find the current structure of the Context Attributes section a bit confusing. It took me a while to realize that the attributes being described were actually context attributes and not examples of extension attributes since they follow the description of extension attributes.

I think that it would be best to move the the extension attributes description to the end of the Context Attributes section.

duglin:
Closes #401 

Also added changes the title of the sub-section to make it clear we're talking about context attributes

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -140,39 +140,6 @@ attributes and the event data will be materialized. For example, in the case
 of a JSON serialization, the context attributes and the event data might
 both appear within the same JSON object.
 
-### Extension Attributes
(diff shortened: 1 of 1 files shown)


r-marques:
Yes. I believe this is much clearer. Thanks!

duglin:
@r-marques great - thanks for the issue

duglin:
Approved on the 3/21/2019 call

PR #406, opened 2019-03-19
Introducing "subject"

clemensv:
Based on last week's debate on a missing context qualifier and the Knative project apparently now running into similar implementation issues for generic dispatch infra that we're having in Azure Event Grid, I am herewith reintroducing "subject" for consideration.

The (extensive) argument for the "subject" field is in the (closed) issue #112. I'm also talking through it in the audio  track attached to this presentation https://github.com/cloudevents/spec/blob/master/share/2018-03-22-TopicsAndSubjects.pptx and the presentation itself illustrates the topic ("source") and subject split.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1,4 +1,4 @@
-# CloudEvents - Version 0.2
+﻿# CloudEvents - Version 0.2
 
 ## Abstract
 
@@ -186,20 +186,19 @@ within the same JSON object.
   - MUST be a non-empty string
 
 ### source
-
-- Type: `URI-reference`
-- Description: This describes the event producer. Often this will include
-  information such as the type of the event source, the organization publishing
-  the event, the process that produced the event, and some unique identifiers.
-  The exact syntax and semantics behind the data encoded in the URI is event
-  producer defined.
-- Constraints:
-  - REQUIRED
-- Examples
-  - https://github.com/cloudevents/spec/pull/123
-  - /cloudevents/spec/pull/123
-  - urn:event:from:myapi/resourse/123
-  - mailto:cncf-wg-serverless@lists.cncf.io
+* Type: `URI-reference`
+* Description: This describes the event producer. Often this will include
+  information such as the type of the event source, the organization
+  publishing the event, the process that produced the event, and some unique
+  identifiers. The exact syntax and semantics behind the data encoded in the URI
+  is event producer defined.
+* Constraints:
+  * REQUIRED
+* Examples
+    * https://github.com/cloudevents/spec/pull
+    * /cloudevents/spec/pull
+    * urn:event:from:myapi/resource
+    * mailto:cncf-wg-serverless@lists.cncf.io
 
 ### subject
 * Type: `String`
@@ -208,22 +207,26 @@ within the same JSON object.
   a subscriber will typically subscribe to events emitted by a `source`,
   but the `source` identifier alone might not be sufficient as a qualifier for
   any specific event if the `source` context has internal sub-structure.
-* Example: A subscriber might register interest for when new blobs are created
-  inside a blob-storage container. In this case, the event `source` identifies
-  the subscription scope (storage container), the `type` identifies the "blob
-  created" event, and the `id` uniquely identifies the event instance to
-  distinguish separate occurrences of a same-named blob having been created;
-  the name of the newly created blob is carried in `subject`.
 
   Identifying the subject of the event in context metadata (opposed to only in
   the `data` payload) is particularly helpful in generic subscription filtering
   scenarios where middleware is unable to interpret the `data` content. In the
   above example, the subscriber might only be interested in blobs with names
-  ending with '.jpg' or '.jpeg' and the subject attribute allows for constructing
-  a simple and efficient string-suffix filter for that subset of events.
+  ending with '.jpg' or '.jpeg' and the subject attribute allows for 
+  constructing a simple and efficient string-suffix filter for that subset of
+  events.
 * Constraints:
   * OPTIONAL
   * MUST be a non-empty string
+* Example:
+  * A subscriber might register interest for when new blobs are created
+    inside a blob-storage container. In this case, the event `source` identifies
+    the subscription scope (storage container), the `type` identifies the "blob
+    created" event, and the `id` uniquely identifies the event instance to
+    distinguish separate occurrences of a same-named blob having been created;
+    the name of the newly created blob is carried in `subject`:
+    * `source`: https://example.com/storage/tenant/container
+    * `subject`: mynewfile.jpg
 
 ### id
 
@@ -396,7 +399,8 @@ The following example shows a CloudEvent serialized as JSON:
 {
     "specversion" : "0.2",
     "type" : "com.github.pull.create",
-    "source" : "https://github.com/cloudevents/spec/pull/123",
+    "source" : "https://github.com/cloudevents/spec/pull",
+    "subject" : "123", 
     "id" : "A234-1234-1234",
     "time" : "2018-04-05T17:31:00Z",
     "comexampleextension1" : "value",
--- a/spec.md
+++ b/spec.md
@@ -201,6 +201,30 @@ within the same JSON object.
   - urn:event:from:myapi/resourse/123
   - mailto:cncf-wg-serverless@lists.cncf.io
 
+### subject
+* Type: `String`
+* Description: This describes the subject of the event in the context of the
+  event producer (identified by `source`). In publish-subscribe scenarios,
+  a subscriber will typically subscribe to events emitted by a `source`,
+  but the `source` identifier alone might not be sufficient as a qualifier for
+  any specific event if the `source` context has internal sub-structure.
+* Example: A subscriber might register interest for when new blobs are created
+  inside a blob-storage container. In this case, the event `source` identifies
+  the subscription scope (storage container), the `type` identifies the "blob
+  created" event, and the `id` uniquely identifies the event instance to
+  distinguish separate occurrences of a same-named blog having been created;
+  the name of the newly created blob is carried in `subject`.
+
+  Identifying the subject of the event in context metadata (opposed to only in
+  the `data` payload) is particularly helpful in generic subscription filtering
+  scenarios where middleware is unable to interpret the `data` content. In the
+  above example, the subscriber might only be interested in blobs with names
+  ending with '.jpg' or '.jpeg' and the subject attribute allows for constructing
+  a simple and efficient string-suffix filter for that subset of events.
+* Constraints:
+  * OPTIONAL
+  * MUST be a non-empty string
+
 ### id
 
 - Type: `String`
--- a/spec.md
+++ b/spec.md
@@ -212,7 +212,7 @@ within the same JSON object.
   inside a blob-storage container. In this case, the event `source` identifies
   the subscription scope (storage container), the `type` identifies the "blob
   created" event, and the `id` uniquely identifies the event instance to
-  distinguish separate occurrences of a same-named blog having been created;
+  distinguish separate occurrences of a same-named blob having been created;
   the name of the newly created blob is carried in `subject`.
 
   Identifying the subject of the event in context metadata (opposed to only in


spec.md
```diff
@@ -215,6 +215,30 @@ help intermediate gateways determine how to route the events.
...
+
+  Identifying the subject of the event in context metadata (opposed to only in
+  the `data` payload) is particularly helpful in generic subscription filtering
+  scenarios where middleware is unable to interpret the `data` content. In the
+  above example, the subscriber might only be interested in blobs with names
+  ending with '.jpg' or '.jpeg' and the subject attribute allows for constructing
+  a simple and efficient string-suffix filter for that subset of events.
+* Constraints:
+  * OPTIONAL
+  * MUST be a non-empty string
```
n3wscott: **If Present,** MUST be a non-empty string

spec.md
```diff
@@ -215,6 +215,30 @@ help intermediate gateways determine how to route the events.
...
     * mailto:cncf-wg-serverless@lists.cncf.io
 
+### subject
+* Type: `String`
+* Description: This describes the subject of the event in the context of the
+  event producer (identified by `source`). In publish-subscribe scenarios,
+  a subscriber will typically subscribe to events emitted by a `source`,
+  but the `source` identifier alone might not be sufficient as a qualifier for
+  any specific event if the `source` context has internal sub-structure.
+* Example: A subscriber might register interest for when new blobs are created
```
n3wscott: In other fields, Example is after the Constrains section.

spec.md
```diff
@@ -215,6 +215,30 @@ help intermediate gateways determine how to route the events.
     * urn:event:from:myapi/resourse/123
     * mailto:cncf-wg-serverless@lists.cncf.io
 
+### subject
+* Type: `String`
```
n3wscott: Perhaps Subject is a URIRef? and it relates to the source uri?

spec.md
```diff
@@ -215,6 +215,30 @@ help intermediate gateways determine how to route the events.
...
+  but the `source` identifier alone might not be sufficient as a qualifier for
+  any specific event if the `source` context has internal sub-structure.
+* Example: A subscriber might register interest for when new blobs are created
+  inside a blob-storage container. In this case, the event `source` identifies
+  the subscription scope (storage container), the `type` identifies the "blob
+  created" event, and the `id` uniquely identifies the event instance to
+  distinguish separate occurrences of a same-named blog having been created;
+  the name of the newly created blob is carried in `subject`.
+
+  Identifying the subject of the event in context metadata (opposed to only in
```
n3wscott: For source, we have these examples:

> * Examples
    * https://github.com/cloudevents/spec/pull/123
    * /cloudevents/spec/pull/123
    * urn:event:from:myapi/resourse/123
    * mailto:cncf-wg-serverless@lists.cncf.io

Would using subject turn them into:

    * source: https://github.com/cloudevents/spec/
    * subject: pull/123

    * source: /cloudevents/spec/pull/123
    * subject: pull/123

    * source: urn:event:from:myapi/resourse/123
    * subject: UPDATE `<-- ? picked a verb`

    * source: mailto:cncf-wg-serverless@lists.cncf.io
    * subject: Help

JemDay:
I think it might be clearer to have some richer use-case examples added to the 'example' section as opposed to embedding them inside the description of each context attribute.

Does _subject_ now become part of the idempotency set ?

duglin:
I believe we currently have "subject" as part of "source" (even if it's just implied by what's in the spec). Is this PR suggesting we pull it out because in order for generic middleware (e.g. filters) to do their job we can't count on them to understand the format of the source URI to know which parts to extract? And that the filter expression would be too hard to construct even if the entity creating the filter expression did know the format of the source URI?  

Not pushing back, I just want to fully understand the reason for the splitting up of 'source' - plus this would all be good info to put into the Primer.

mrutkows:
Before adding additional fields one should clearly identify the set of use cases it intends to solve to as well exhibit what the field would be used for.  If we are merely separating some partial path from a URI for convenience, it does not seem like an additional field is warranted.  URI (sources) that share the same paths (up to a point) indicate same source and as paths are added/they represent and distinguish sub-areas (components or subject areas) within the same "family" of sources.

evankanderson:
I believe the intent of this is to make it clear what portions of the current `source` field SHOULD be used to identify a unique producer (see also discussion in #391 and #403 about uniqueness of `id` being on a per-event-producer basis).

Three issues I'm aware of with the current definition of `source`:

1. Middleware requires some sort of omniscience (knowledge of all event producers and the schema of their `source` attributes) in order to be able to determine the event producer given a `source` URL. In some cases, the `source` may represent a specific object, like a pull request or an individual object in an object store, but the producing system is GitHub or the object store implementation. In some cases, this may align on hostname boundaries, but in other cases, multiple APIs may be served from a single hostname, or the same API may be served from different hostnames (see Slack).

1. Consumers who wish to consume events from a particular producer need a way to indicate what producer they are interested in (related to point 1) Again, if the producer uses a specific `source` URI of a particular resource in the system (as suggested by our examples), the consumer is then required to specify "the set of sources I'm interested in" as some sort of glob/pattern match/prefix match, depending on the system. If the `source` URI particularly represents the _event producing system_, and the `subject` represents the _individual resource described by the event_. Side note: I have a slight concern about conflating `subject` in this context with `subject` in an access control context; I might prefer `producer` and `source` or `service` and `resource`.

1. Currently, `source` is allowed to be a relative URI ref, but it's not clear what that URI ref is intended to be relative to. If we separate `source` and `subject`, `subject` could logically be a reference relative to the `source`. I realize that this isn't directly in scope of this PR, but it is related. Similarly, we might benefit by suggesting that each event producer should have a (single) unique `source` URI which could be used by #391 and #403 to allow an external observer to make assumptions about the scope of uniqueness of `id` values.

clemensv:
@duglin The "subject" is currently necessarily embedded in "source", and since that's a URI it's hard for us to give hard rules for where the context given by the "subject" sits in that URI and therefore filtering is difficult. In Event Grid, we're making a product decision to put "subject" behind the '#' in the URI and thus effectively treat it as a distinct field. In pub/sub, you subscribe to events from a source and that source itself may raise various types of events on behalf of concepts associated with it, and the "subject" identifies those concepts.

clemensv:
@JemDay it's not part of the idempotency set. For that, 'source' and 'id' remain sufficient. I agree that we need more examples and in-depth explanation of how the attributes hang together.

clemensv:
@mrutkows @evankanderson  If you haven't done so, please read through issue #112 which lays out in lot of detail why I'm reintroducing the field and it also enumerates a lot of examples of prior art in existing messaging and eventing infrastructure.

spec.md
clemensv: ok

spec.md
clemensv: I can shuffle that

spec.md
clemensv: It's some sort of subject identifier. I can be a filename or just some name of a thing that the source knows. I wouldn't constrain that as much as source.

spec.md
clemensv: The last example bring the point home rather nicely. That's exactly what subject is.

spec.md
cathyhongzhang: If the event source has an internal sub-structure, then the event consumer could be interested in any part of the sub-structure. How could the producer know which part of the sub-structure the consumer is interested in, thus which part to put in the subject? Also since it is an internal sub-structure of the source, how could the consumer know how to interpret the semantics of the subject?

clemensv: That is up to the pub/sub infrastructure to provide. In Event Grid we provide filters that let you run a string expression over the subject to filter that down. The publisher doesn’t know, it just publishes for the scope it was asked to deliver events for.

spec.md
```diff
@@ -215,6 +215,30 @@ help intermediate gateways determine how to route the events.
...
+* Description: This describes the subject of the event in the context of the
+  event producer (identified by `source`). In publish-subscribe scenarios,
+  a subscriber will typically subscribe to events emitted by a `source`,
+  but the `source` identifier alone might not be sufficient as a qualifier for
+  any specific event if the `source` context has internal sub-structure.
+* Example: A subscriber might register interest for when new blobs are created
+  inside a blob-storage container. In this case, the event `source` identifies
+  the subscription scope (storage container), the `type` identifies the "blob
+  created" event, and the `id` uniquely identifies the event instance to
+  distinguish separate occurrences of a same-named blog having been created;
```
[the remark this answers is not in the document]
clemensv: There’s a concepts-trap here that caused all sorts of trouble when REST became a thing and that’s obsessing about structure and shape of URIs.

There’s a scope that you subscribe to and then there’s substructure if that scope that you then filter on. The structure of those expressions is completely up to the application.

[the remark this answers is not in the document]
clemensv: There’s a concepts-trap here that caused all sorts of trouble when REST became a thing and that’s obsessing about structure and shape of URIs.

There’s a scope that you subscribe to and then there’s substructure if that scope that you then filter on. The structure of those expressions is completely up to the application.

duglin:
@clemensv can you sign your commits?

clemensv:
@duglin @cneijenhuis used a novel feature which i haven't yet found on Github myself where you can contribute little patches to the patch, and accepting those signs differently:

commit 13241fbf86ac2798c2a6412637228d1a7376dba8 (HEAD -> patch-subject, origin/patch-subject)
Author: Christoph Neijenhuis <christoph.neijenhuis@commercetools.de>
Date:   Wed Apr 3 18:34:19 2019 +0200

    Update spec.md

    Co-Authored-By: clemensv <clemens@vasters.com>

I did a soft reset and re-committed, but the DCO checker should accept these or we should rule out that feature.

cneijenhuis:
Sorry for that. I like the suggested changes feature, because a) the diff view is really nice and b) in theory it should help save time to incorporate suggestions.
I'm guessing you could manually add the `Signed-off-by ...` to the commit message in the GitHub UI, but it makes the process more clunky.

(The problem exists with all edits through the GitHub UI unfortunately...)

duglin:
rebase needed

spec.md
```diff
@@ -1,4 +1,4 @@
-# CloudEvents - Version 0.2
+﻿# CloudEvents - Version 0.2
```
duglin: weird

spec.md
```diff
@@ -210,11 +210,39 @@ help intermediate gateways determine how to route the events.
...
+  event producer (identified by `source`). In publish-subscribe scenarios,
+  a subscriber will typically subscribe to events emitted by a `source`,
+  but the `source` identifier alone might not be sufficient as a qualifier for
+  any specific event if the `source` context has internal sub-structure.
+
+  Identifying the subject of the event in context metadata (opposed to only in
+  the `data` payload) is particularly helpful in generic subscription filtering
+  scenarios where middleware is unable to interpret the `data` content. In the
+  above example, the subscriber might only be interested in blobs with names
+  ending with '.jpg' or '.jpeg' and the subject attribute allows for
```
duglin: s/subject/`subject`/

duglin: I like this paragraph. Would it be helpful to add something about how putting this info into the "source" URI is just as hard to parse/filter-on ? And that's why we creating this property. Before this property I think people were pushed towards merging it into 'source' - if they wanted it outside of 'data'.  Or, if this is too much text for the spec, perhaps in the primer?

spec.md
```diff
@@ -210,11 +210,39 @@ help intermediate gateways determine how to route the events.
...
+  the `data` payload) is particularly helpful in generic subscription filtering
+  scenarios where middleware is unable to interpret the `data` content. In the
+  above example, the subscriber might only be interested in blobs with names
+  ending with '.jpg' or '.jpeg' and the subject attribute allows for 
+  constructing a simple and efficient string-suffix filter for that subset of
+  events.
+* Constraints:
+  * OPTIONAL
+  * MUST be a non-empty string
+* Example:
```
duglin: I like this example because it shows how all of the key properties relate to each other, and better explains what people should expect in each. So, can we move this example out from under the definition of this one property so it can be view in the bigger context?  And then show "id" and "type" too - or even a total CE.

duglin:
@clemensv can you rebase this - then I can merge it.

Approved on the 4/4 call.

spec.md
n3wscott: The updated version adds a extra whitespace to the end of 0.2.

spec.md
n3wscott: URIRef does all:

`this is a valid uri ref`

`http://so.is/this`

spec.md
n3wscott: no pushy?

spec.md
```diff
@@ -186,20 +186,47 @@ within the same JSON object.
...
-  The exact syntax and semantics behind the data encoded in the URI is event
-  producer defined.
-- Constraints:
-  - REQUIRED
-- Examples
-  - https://github.com/cloudevents/spec/pull/123
-  - /cloudevents/spec/pull/123
-  - urn:event:from:myapi/resourse/123
-  - mailto:cncf-wg-serverless@lists.cncf.io
+* Type: `URI-reference`
```
n3wscott: `-` is more clear than `*` that it is suppose to be a list. The rest of the spec uses `-`

PR #408, opened 2019-03-20
Extensions follow attribute naming scheme introduced in #321

cneijenhuis:
#321 made attribute names lower case, but did not apply the new rule to extensions.
This PR cleans that up.


duglin:
good catch.
LGTM

clemensv:
LGTM

duglin:
Approved on 3/21/2019 call.

PR #409, opened 2019-03-21
Ran markdown linter on http-webhook.md

n3wscott:


duglin:
From https://prettier.io/ according to @n3wscott

PR #411, opened 2019-03-21
Ran https://prettier.io/ command on all markdown.

n3wscott:
To get the output, I Ran the following: 

```bash
prettier --write --prose-wrap=always $(find . | grep '.md' | grep -v .github)`
```

Had to reformat `GOVERNANCE.md` a little, it was written with invalid markdown originally.

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -5,121 +5,116 @@ project will manage this repository.
 
 ## Meetings
 
-In order to provide equitable rights to all members,
-the following process will be followed:
+In order to provide equitable rights to all members, the following process will
+be followed:
 
-* Official conference calls will be announced at least a week in advance.
-* Official face-to-face meetings will be announced at least 4 weeks in
-  advance.
-* Proposed changes to any document will be done via a Pull Request (PR).
-* PRs will be reviewed during official meetings, but off-line reviews
-  (LGTMs, NOT LGTMs) and comments are strongly encouraged to gauge the
-  group's opinion on the proposed change prior to the meeting.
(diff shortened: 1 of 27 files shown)


GOVERNANCE.md
```diff
@@ -5,121 +5,116 @@ project will manage this repository.
...
+- Resolving PRs ("merging" or "closing with no action") will be done as a result
+  of a motion made during a meeting, and approved.
+- Reopening a PR can be done if new information is made available, and a motion
+  to do so is approved.
+- Any motion that does not have "unanimous consent" will result in a formal
   vote. See [Voting](#voting).
 
 ## Membership
 
-There are three categories of project membership:
```
n3wscott: fix.

duglin: did you have to manually fix it? That concerning if so

n3wscott: yes I did. The tool tried the best it could.

duglin: that's unfortunate because then we can't automate things.
I wonder if there's a better way to write that line so the tool will like it....

n3wscott: you can turn write off.

n3wscott: I updated the line to be valid markdown.

n3wscott:
Merged with this weeks changes.

markpeek:
BTW, might be good to add that command to the Makefile.

n3wscott:
> BTW, might be good to add that command to the Makefile.

It will fix without regard to intent, so it needs a human to confirm the translation.

duglin:
Approved on 3/28 call

PR #415, opened 2019-04-09
Master represents the future version of the spec, use the future version in the text.

n3wscott:
This change bumps the version used in master to `0.3-wip` of the spec.

This becomes very clear (and confusing) when you are reading the examples of features are are only in `0.3-wip`, like Batch support.

### DIFF
--- a/roadmap.md
+++ b/roadmap.md
@@ -55,7 +55,7 @@ _0.2_ - Completed - 2018/12/06
 1. Define the set of protocol and serialization mappings we're going to produce
    for 0.4 milestone
 
-_0.3_
+_0.3-wip_
 
 1. Resolve all known "proposed optional attributes" issues
 1. Resolve all known "security related" issues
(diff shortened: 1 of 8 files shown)


duglin:
@n3wscott we talked about this on this week's call and there was agreement to change the version to something other than 0.2 but 0.3 didn't seem appropriate since it's not 0.3 yet.  The group seemed to like `0.3.wip` - you ok with that?

n3wscott:
Yeah I can make that update! That works for me.

duglin:
thanks!

roadmap.md
```diff
@@ -55,7 +55,7 @@ _0.2_ - Completed - 2018/12/06
 1. Define the set of protocol and serialization mappings we're going to produce
    for 0.4 milestone
 
-_0.3_
+_0.3-wip_
```
duglin: I think this one needs to be skipped

n3wscott: whoops!

duglin:
LGTM

duglin:
Approved on the 4/18 call

PR #420, opened 2019-04-19
Terminology additions for Source, Consumer, Producer and Intermediary

deissnerk:
I added some definitions to the terminology section of the spec. I know there might be some discussion, especially regarding the producer. I need these terms to better propose rules of when and how the event context can be changed.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -98,6 +98,20 @@ types of information: the [Data](#data) representing the Occurrence and
 [Context](#context) metadata providing contextual information about the
 Occurrence.
 
+#### Producer
+The "producer" is a specific instance, process or device that produces the event
+based on the occurrence.
+
+#### Source
+The "source" is the logical system or service where the occurrence happened. In
+a distributed system it may consist of multiple [Producers](#producer).
+
+#### Consumer
+A "consumer" receives the event and acts upon it.
+
+#### Intermediary
+
+
 #### Context
 
 Context metadata will be encapsulated in the
@@ -162,7 +176,12 @@ within the same JSON object.
 
 The following attributes are REQUIRED to be present in all CloudEvents:
 
+<<<<<<< HEAD
 #### id
+=======
+### specversion
+<<<<<<< HEAD
+>>>>>>> Added producer, consumer and intermediate to terminology section.
 
 - Type: `String`
 - Description: Identifies the event.
@@ -176,7 +195,20 @@ The following attributes are REQUIRED to be present in all CloudEvents:
 - Constraints:
   - REQUIRED
   - MUST be a non-empty string
+<<<<<<< HEAD
   - MUST be unique within the scope of the producer
+=======
+=======
+* Type: `String`
+* Description: The version of the CloudEvents specification which the event
+  uses. This enables the interpretation of the context. Compliant event
+  producers MUST use a value of `0.2` when referring to this version of
+  the specification.
+* Constraints:
+  * REQUIRED
+  * MUST be a non-empty string
+>>>>>>> Added producer, consumer and intermediate to terminology section.
+>>>>>>> Added producer, consumer and intermediate to terminology section.
 
 #### source
 
@@ -412,6 +444,7 @@ Consider the following to prevent inadvertent leakage especially when leveraging
 
 - Context Attributes
 
+<<<<<<< HEAD
   Sensitive information SHOULD NOT be carried or represented in context
   attributes.
 
@@ -429,6 +462,23 @@ Consider the following to prevent inadvertent leakage especially when leveraging
 
   Transport level security SHOULD be employed to ensure the trusted and secure
   exchange of CloudEvents.
+=======
+  Sensitive information SHOULD NOT be carried or represented in context attributes.
+
+  CloudEvent producers, consumers, and intermediaries MAY introspect and log context
+  attributes.
+
+* Data
+
+  Domain specific [data](#data) SHOULD be encrypted to restrict visibility to
+  trusted parties. The mechanism employed for such encryption is an agreement between
+  producers and consumers and thus outside the scope of this specification.
+
+* Transport Bindings
+
+  Transport level security SHOULD be employed to ensure the trusted and
+  secure exchange of CloudEvents.
+>>>>>>> Added producer, consumer and intermediate to terminology section.
 
 # Example
 
--- a/spec.md
+++ b/spec.md
@@ -99,8 +99,9 @@ types of information: the [Data](#data) representing the Occurrence and
 Occurrence.
 
 #### Producer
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -91,13 +91,37 @@ perform a scheduled reboot.
...
+performed based on information contained in the event, but an event will not
+identify a specific routing destination. Events will contain two types of
+information: the [Data](#data) representing the Occurrence and
 [Context](#context) metadata providing contextual information about the
 Occurrence.
 
+#### Producer
+
+The "producer" is a specific instance, process or device that creates the data
+structure describing the event and its context.
```
duglin: the term "data structure" sounds like it's implying the CloudEvent - if so, we should just say that. If not, can you elaborate on what you mean?

spec.md
```diff
@@ -91,13 +91,37 @@ perform a scheduled reboot.
...
+#### Source
+
+The "source" is the logical system or service where the occurrence happened. In
+a distributed system it might consist of multiple [Producers](#producer). If a
+source is not aware of CloudEvents, a producer creates the CloudEvent on behalf
+of the source.
+
+#### Consumer
+
+A "consumer" receives the event and acts upon it. It uses the content and data
```
duglin: not sure what you mean by "content and data" - do you mean "context attribute and data attribute" ?

duglin: Or if you want to just say "Context" instead of "Context attributes" then perhaps make it a "C" instead of "c" and make it an href to the "Context" definition. It's not clear when you mean the CE Context entity vs something else.

deissnerk: Yes, that's a general issue about the terminology. If you look at the definition of _event_, it is already CloudEvent specific, explaining context and data. Does that mean that whenever I mention _event_, I imply that it is a CloudEvent?

spec.md
```diff
@@ -91,13 +91,37 @@ perform a scheduled reboot.
...
+#### Producer
+
+The "producer" is a specific instance, process or device that creates the data
+structure describing the CloudEvent.
+
+#### Source
+
+The "source" is the logical system or service where the occurrence happened. In
+a distributed system it might consist of multiple [Producers](#producer). If a
+source is not aware of CloudEvents, a producer creates the CloudEvent on behalf
```
evankanderson: Suggestion from the call: "an external producer"...

spec.md
```diff
@@ -91,13 +91,37 @@ perform a scheduled reboot.
...
 Occurrence.
 
+#### Producer
+
+The "producer" is a specific instance, process or device that creates the data
+structure describing the CloudEvent.
+
+#### Source
+
+The "source" is the logical system or service where the occurrence happened. In
```
erikerikson: While I will accept the will of the group to accept this PR, this appears to contradict the definition of the `source` attribute as [defined](https://github.com/cloudevents/spec/blob/master/spec.md#source):

```
Description: This describes the event producer.
```

Given this, it seems that either the two should be disambiguated or this PR should be updated to change that text before it is merged.

Tapppi:
Although we agreed that the terminology here makes sense on the call, writing a [comment](https://github.com/cloudevents/spec/pull/391/files#r278658761) on PR #422 that made me realise this terminology is in conflict with both the name and description of the `source` attribute. Reading that comment should make one realise the confusion, but let me try to elaborate here. For clarity, I will capitalise the terminology introduced here.

According to this terminology, two instances of a "Users service" would be distinct PRODUCERS. Together they are a SOURCE (a logical system) and have distinct `source` attribute value*s* compared to other SOURCES. However, the "Users service" as a SOURCE will likely produce events with multiple `source` attribute values for different subsystems/data/tables/etc within the system. When one says "source", it can be difficult to know whether they mean the SOURCE (e.g. "Users service") or a specific `source` attribute value from SOURCE (e.g. `example.com/user/account` or `example.com/user/address`).

The granularity of `source` attribute values will vary depending on system, for example a database change data capture system would probably base it on table/resource names/OIDs while an event-driven microservice would likely assign a value with semantic meaning not necessarily tied to technical implementation.

duglin:
While we accepted this PR on today's call, based on some chats afterwards (and with @deissnerk's agreement) we decided to hold off merging it since there are still some lingering questions. Please add comments!!

spec.md
deissnerk: Thanks for bringing this up. I was aware of this contradiction and wanted to stir up the discussion, because "describes the event producer" seems to me very vague. Why would we call a field, that describes the producer, _source_ and not _producer_? 
I think the fact the source attribute is a URI (an identifier and not descriptor) and the examples like `urn:event:from:myapi/resource` suggest that the source attribute identifies something more logical. I really have a hard time finding better words. Any proposals for a better wording are welcome.

deissnerk:
@Tapppi I was actually quite happy that you brought this up, yesterday! So defining _source_ as a _logical system_ might not be the best wording. I still kind of like logical, though. I also liked that you mentioned _semantic meaning_. 
Bottom line to me is that _producer_ is something technical, while _source_ is more logical or abstract, and with semantic meaning.
I could of course also delete my definition of _source_, as the attribute is already defined, but that would just shift the discussion, as my definition of _producer_ would still be conflicting.
So I would like to know if any of the definitions in this PR make sense in your eyes, or if they should be adjusted. The discussion we had about #391 showed me that some more clarity on the definitions is needed.

spec.md
erikerikson: I'm not sure I have terminology but perhaps its useful to consider that perhaps the "logical" source exists in user/application conceptual design space while producers live more in the space of system implementation?  That seems the strongest argument for the #420 definition of source.

Perhaps a downside is that making it logical reduces the attribute's value to middleware or consumers.  Personally, I had leaned towards preferring the source as a unique producer (guaranteed by application), primarily because of the relationship (assuming ordering guarantees) that ensures a map of producers to lists of ordered events which looks much like a logical clock if not a CRDT.  If the IDs of producers are monotonically increasing and the event IDs are monotonically increasing, one also gets efficient and simple deduplication+idempotency which in my favored scenarios of many multi-versioned consumer groups makes the costs tractable.  I think I, at least (if not fully), failed in my reading of the definition of source because the understanding I retained from some time ago, by bias, weighted that first sentence strongly and broadly ignored the implications of the remainder.

Given the approval of this PR then it seems that solution is to simply remove mention of producers from the definition of `source`.

duglin:
We may need to do a little bit of word-merging with https://github.com/cloudevents/spec/pull/391 in the "source" attribute section but that's minor and I think the two align nicely. LGTM

duglin:
Approved on the 5/16 call

@deissnerk can you rebase this? Then I can merge it.

deissnerk:
@duglin done

duglin:
@deissnerk thanks!
merging

PR #425, opened 2019-04-30
Order the attributes section

duglin:
No semantic changes, just some re-ordering to make it easier to follow/read:
- created REQUIRED and OPTIONAL attributes sub-sections
- moved all attribtues into the appropriate sub-sections
- alphabetized the attributes in each section

I think this makes it easier for people implementing the spec to
1) find the attribute they're looking for since it's alphabetized now,
2) know which attributes they MUST support since they're all in one
section now

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -158,7 +158,50 @@ and the event data will be materialized. For example, in the case of a JSON
 serialization, the context attributes and the event data might both appear
 within the same JSON object.
 
-### type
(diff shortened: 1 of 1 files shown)


erikerikson:
LGTM

clemensv:
LGTM

duglin:
Approved on 5/2 call

Issue #436, opened 2019-06-07
Could Any also take Uri-reference and timestamp attributes


gunnarmorling:
Reading the current wording of the spec it seems `Any` couldn't `Uri-reference` and `timestamp` attributes. Is this intentional? It seems it'd exclude usage of those two in maps for instance. Or if they actually are allowed for `Any`, this should be clarified. I can provide a PR if that's how it is meant.

cneijenhuis:
Good point - I think this is not intentional. Feel free to open a PR.

gunnarmorling:
Done: PR #437.

duglin:
Closed by #454

PR #437, opened 2019-06-07
#436 Clarifying that Any can take the shape of any other type

gunnarmorling:
Also removed the redundancy between the two sentences applying to `Any`.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1,4 +1,4 @@
-# CloudEvents - Version 0.3
+# CloudEvents - Version 0.4-wip
 
 ## Abstract
 
@@ -267,7 +267,7 @@ The following attributes are REQUIRED to be present in all CloudEvents:
 - Type: `String`
 - Description: The version of the CloudEvents specification which the event
   uses. This enables the interpretation of the context. Compliant event
-  producers MUST use a value of `0.3` when referring to this version of the
+  producers MUST use a value of `0.4-wip` when referring to this version of the
   specification.
 - Constraints:
   - REQUIRED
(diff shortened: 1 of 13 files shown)


spec.md
```diff
@@ -184,16 +184,16 @@ as a string.
...
   in [RFC 3986 Section 4.1](https://tools.ietf.org/html/rfc3986#section-4.1).
 - `Timestamp` - Date and time expression using the Gregorian Calendar. 
    - String encoding: [RFC 3339](https://tools.ietf.org/html/rfc3339).
+- `Any` - A variant type that can take the shape of either an `Integer`,
+  `String`, `Binary`, `Map`, `URI-reference` or `Timestamp`.
 
-The `Any` type is a variant type that can take the shape of either a `Binary`,
-`Integer`, `Map` or `String`. The type system is intentionally abstract, and
-therefore it is left to implementations how to represent the variant type.
+The type system is intentionally abstract, and therefore it is left to
```
duglin: I'll let @clemensv comment on the wording change, but I think these 2 lines need to be indented and the blank line removed so they appear under the "Any" section.

gunnarmorling: I left it on a separate line after that list on purpose, as it was like that before and is a rather generic remark on the type-system. But I can also attach it just to that last item on the `Any` type. No strong preference.

duglin: I think the diff here is that since you moved Any to the end of the list (which I like since Any is kind of a "catch-all/everything-else" type of option), the split that was there now looks odd since both sections are "Any" specific and so grouping them together looks nicer to me.

gunnarmorling: Ok, I've pulled the both sections together and force-pushed. While at it, I've also fixed an unrelated typo.

spec.md
```diff
@@ -296,7 +295,7 @@ The following attributes are REQUIRED to be present in all CloudEvents:
 
 ### OPTIONAL Attributes
 
-The following attribtues are OPTIONAL to appear in CloudEvents. See the
+The following attributes are OPTIONAL to appear in CloudEvents. See the
```
duglin: Thanks!!

PR #440, opened 2019-06-08
Add a little clarity around "type"

### LINKED ISSUE #188: Do eventType, eventID describe the type and identity of events, or occurrences?
The specification [says of eventType](https://github.com/cloudevents/spec/blob/v0.1/spec.md#eventtype):

> Type of occurrence which has happened.

But then it [says of eventID](https://github.com/cloudevents/spec/blob/v0.1/spec.md#eventid):

> ID of the event.

As a reader, I had a bit of cognitive dissonance here.  The spec [goes to some lengths to indicate in its terminology section that an occurrence is not an event](https://github.com/cloudevents/spec/blob/v0.1/spec.md#occurrence), but is contained by one, but then here seems to conflate the two again.

duglin:
Closes #188

I think the rest of #188 is already addressed by our recent PRs around
uniqueness of IDs.

/cc @ljnelson

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -95,7 +95,7 @@ routing can be performed based on information contained in the event, but an
 event will not identify a specific routing destination. Events will contain two
 types of information: the [Data](#data) representing the Occurrence and
 [Context](#context) metadata providing contextual information about the
-Occurrence.
+Occurrence. A single occurrence MAY result in more than one event.
 
 #### Producer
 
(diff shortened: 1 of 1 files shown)


duglin:
Based on today's call I redid this one. I moved the sentence about how a single occurrence can result in more than one event up into the terminology section - where I think it flows better.  That should remove the concern that @clemensv had about linking "type" with the idea of multiple events per occurrence.   See whatcha think....

duglin:
ping @clemensv @JemDay for review, since you guys spoke up during the call

duglin:
Approved on the 6/20 call

PR #441, opened 2019-06-08
Add a spot to describe adapters

duglin:
In order to help interop, I'd like to propose that we include some recommendations for how to convert some well-known events into CloudEvents. These are just recommendations and people are free to ignore them if they want and still be spec compliant.

For now I just wrote up what CEs for github, gitlab and AWS SNS events.  We can add more later.

Couple of things:
- as I was going through this the choice of `source` and `subject` wasn't always clear to me. I generally tried to follow the rule that `subject` was the entity being acted upon, and `source` was (sort of) it's owner.  For example, a github issue was the subject and the corresponding github repo was the source for the creation of a new issue.  However, in cases where the issue is edited (and is therefore the subject of the action being taken) does it still remain the "subject" or is it now the "source" since people can subscribe to issues?  Anyway, the point here is that 1) please review what I did very carefully w.r.t. "subject" and "source", and 2) I think we need more text in the primer to help other people with this decision since I personally don't think the choice is always obvious.
- for github and gitlab, I tried to follow the pattern of `domain/type.action`. For example, `github.com/issue.create`. If people think `github.com/issues/create` (note the `/` instead of `.`) makes more sense I'm ok with that, I just like the dot-notation for things that kind of looked like verbs/function calls :-)
- `time` was a bit of a challenge. In some cases there was a clear value to extract from the event. But in other cases it either wasn't clear what to use or it felt like determining the value might be more trouble than it's worth - so I just put "Current time" in there.  Also, the spec defines `time` as "Timestamp of when the event happened", notice it says "event" not "occurrence".  I think we need to be more clear on this. I'll open a new issue/PR.
- there are a couple of spots where I put question-marks, asking for advice.


duglin:
While this one probably isn't technically required for v1.0, I tagged it as "try-for-v1.0" because I think some of the design/clarity questions raised by this work would be really good to get into v1.0

grantr:
> For example, `github.com/issue.create`

The current spec says that event types "SHOULD be prefixed with a reverse-DNS name." Shouldn't these be `com.github/issue.create`? Is that recommendation being dropped?

adapters/aws-sns.md
```diff
@@ -0,0 +1,25 @@
...
+
+All SNS notifications are converted into CloudEvents using the
+same pattern as described in the following table:
+
+| CloudEvents Attribute | Value                                           |
+| :-------------------- | :---------------------------------------------- |
+| `id`                  | "x-amz-sns-message-id" value                    |
+| `source`              | "x-amz-sns-topic-arn" value                     |
+| `specversion`         | `0.3-wip`                                       |
+| `type`                | `aws.com/sns/` + "x-amz-sns-message-type" value |
```
grantr: Why `aws.com` here? Seems like the actual domain is `aws.amazonaws.com`.

duglin:
> > For example, `github.com/issue.create`
> 
> The current spec says that event types "SHOULD be prefixed with a reverse-DNS name." Shouldn't these be `com.github/issue.create`? Is that recommendation being dropped?

Good catch! I'll fix that

adapters/aws-sns.md
duglin: yep - I'll fix - thanks!  and do reverse-dns too :-)

adapters/aws-sns.md
```diff
@@ -0,0 +1,25 @@
...
+
+AWS SNS notification documentation:
+https://docs.aws.amazon.com/sns/latest/api/Welcome.html
+
+All SNS notifications are converted into CloudEvents using the
+same pattern as described in the following table:
+
+| CloudEvents Attribute | Value                                           |
+| :-------------------- | :---------------------------------------------- |
+| `id`                  | "x-amz-sns-message-id" value                    |
```
n3wscott: We have talked at length that `id` is suppose to be a new uuid even if the upstream source provides an id.

adapters/aws-sns.md
```diff
@@ -0,0 +1,25 @@
...
+| `source`              | "x-amz-sns-topic-arn" value                     |
+| `specversion`         | `0.3-wip`                                       |
+| `type`                | `aws.com/sns/` + "x-amz-sns-message-type" value |
+| `datacontentencoding` | Omit                                            |
+| `datacontenttype`     | "Content-Type" value                            |
+| `schemaurl`           | Omit                                            |
+| `subject`             | "x-amz-sns-subscription-arn" value              |
+| `time`                | Current time                                    |
+| `data`                | Content of HTTP request body                    |
+
```
n3wscott: Any extensions that are common to SNS that should become CE extensions in the context?

adapters/github.md
```diff
@@ -0,0 +1,672 @@
...
+based on the specified event.
+
+### CheckRunEvent
+
+| CloudEvents Attribute | Value                                                                            |
+| :-------------------- | :------------------------------------------------------------------------------- |
+| `id`                  | "X-GitHub-Delivery" HTTP header value                                            |
+| `source`              | "repository.url" value                                                           |
+| `specversion`         | `0.3-wip`                                                                        |
+| `type`                | `github.com/check_run.` + "action" value                                         |
```
n3wscott: reverse dns: `com.github.check_run`

adapters/github.md
```diff
@@ -0,0 +1,672 @@
...
+| `datacontentencoding` | Omit                                                                             |
+| `datacontenttype`     | `application/json`                                                               |
+| `schemaurl`           | Omit                                                                             |
```
n3wscott: it would be great to have an example that uses `schemaurl`

adapters/aws-sns.md
grantr: Is [this](https://github.com/cloudevents/spec/blob/master/spec.md#id) no longer accurate then?

> Producers MUST ensure that source + id is unique for each distinct event. If a duplicate event is re-sent (e.g. due to a network error) it MAY have the same id. Consumers MAY assume that Events with identical source and id are duplicates.

Upstream sources may also re-send events due to network errors. Doug's proposal allows those duplicates to be detectable downstream.

duglin: We only need to create a UUID if the `x-amz-sns-message-id` is not unique per SNS event - I was assuming it was.  The docs say this:
```
A Universally Unique Identifier, unique for each message published. For a notification that
Amazon SNS resends during a retry, the message ID of the original message is used. 
```
So I think that should be ok.  @timbray would you agree?

adapters/github.md
duglin: fixed

adapters/github.md
duglin: yep

adapters/aws-sns.md
duglin: Nothing jumped out at me

cneijenhuis:
I like the idea to define adapters for existing event producers 👍 

However, I am quite confused by the attempt to treat AWS SNS as an event producer and not as a middleware/intermediary. In my opinion, we should rather attempt to define a (proprietary ?) transport binding for SNS.

The reason is two-fold:
* Users will want to use SNS as an intermediary to transport CloudEvents. This adapter will do no good in such cases, as it'll force-set several attributes to values that are not in the publishers control (e.g. `subject`, `type`)
* The mapping is unstable/unpredictable:
  * If I write a CloudEvents consumer, I can not easily port it one SNS setup to SNS another. Even though the underlying event producer is the same, the CloudEvents end up having different `subject`, `source` etc.
  * If I use SNS to fan-out events to another SNS topic, the event is still the same logically. However, due to the dependency on topic name, ARN, and the id, the adapted CloudEvent looks very different and can not be de-duplicated.

That said, I think it makes sense to define adapters for popular AWS products. Maybe S3 events are a good starting point for an adapter? See https://docs.aws.amazon.com/AmazonS3/latest/dev/notification-content-structure.html

duglin:
@cneijenhuis excellent point. I removed it and added S3 :-)  
There's a question in that one .

adapters.md
```diff
@@ -0,0 +1,12 @@
...
+Not all event producers will produce CloudEvents natively. As a result,
+some "adapter" might be needed to convert these events into CloudEvents.
+This will typically mean extracting metadata from the events to be used as
+CloudEvents attributes. In order to promote interoperability across multiple
+implementations of these adpaters, the following documents show the proposed
+algorithms that should be used:
+
+* [AWS S3](adapters/aws-s3.md)
+* [Github](adapters/github.md)
+* [Gitlab](adapters/gitlab.md)
```
jamesramsay:
```suggestion
* [GitLab](adapters/gitlab.md)
```

adapters/gitlab.md
```diff
@@ -0,0 +1,177 @@
+# Gitlab CloudEvents Adapter
```
jamesramsay:
```suggestion
# GitLab CloudEvents Adapter
```

adapters/gitlab.md
```diff
@@ -0,0 +1,177 @@
...
+
+This document describes how to convert
+[Gitlab webhook events](https://docs.gitlab.com/ee/user/project/integrations/webhooks.html#events)
```
jamesramsay:
```suggestion
[GitLab webhook events](https://docs.gitlab.com/ee/user/project/integrations/webhooks.html#events)
```

adapters/gitlab.md
```diff
@@ -0,0 +1,177 @@
...
+into a CloudEvents.
+
+Gitlab webhook event documentation:
```
jamesramsay:
```suggestion
GitLab webhook event documentation:
```

duglin:
> Thanks @duglin for the ping on https://gitlab.com/gitlab-org/gitlab-ce/issues/62698 - these look sensible to me! 🦊
> 
> Only changes proposed are capitalization. I think GitHub also prefers a capital **H** but I'll leave that to you.

@jamesramsay thanks for the review!  Just fixed the **H** and **L** casing!

duglin:
Some GitLab have reviewed the gitlab mapping and are ok with it. You can see the thread here: https://gitlab.com/gitlab-org/gitlab-ce/issues/62698

adapters/aws-s3.md
```diff
@@ -0,0 +1,39 @@
+# AWS Simple Storage Service CloudEvents Adapter
+
+This document describes how to convert
+[AWS S3 events](https://docs.aws.amazon.com/AmazonS3/latest/dev/notification-content-structure.html)
+into a CloudEvents.
```
cneijenhuis:
```suggestion
into CloudEvents.
```

adapters/aws-s3.md
```diff
@@ -0,0 +1,39 @@
...
+| `id`                  | Generate a new unique value, e.g. a UUID        |
+| `source`              | "eventSource" value + `.` + "awsRegion" value + `.` + "s3.buckets.name" value  |
+| `specversion`         | `0.4-wip`                                       |
+| `type`                | `com.amazonaws.s3.` + "eventName" value         |
+| `datacontentencoding` | Omit                                            |
+| `datacontenttype`     | "Content-Type" value                            |
+| `schemaurl`           | Omit                                            |
+| `subject`             | "s3.object.key" value                           |
+| `time`                | "eventTime" value                               |
+| `data`                | Content of HTTP request body                    |
```
cneijenhuis: What does `HTTP request body` refer to? The original S3 event?

S3 events aren't delivered via HTTP, they're put into SQS, SNS or invoke a Lambda function: https://docs.aws.amazon.com/AmazonS3/latest/dev/NotificationHowTo.html#how-to-enable-disable-notification-intro

adapters/aws-s3.md
```diff
@@ -0,0 +1,39 @@
...
+  the S3 event format it is expected that this value will not be
+  `aws:s3` for them - it should be something specific to their environment.
+- Consumers of these events will therefore be able to know if the event
+  is an S3 type of event (regardless of whether it is coming from S3 or
+  an S3-compatible provider) by detecting the `com.amazonaws.s3` prefix
+  on the `type` attribute.
+
+
+Question:
+- should `id` be "responseElements.x-amz-request-id" ?
```
cneijenhuis: Maybe it needs to be combined with the "x-amz-id-2" to be unique? Generally, I'm in favor of using it.

adapters.md
```diff
@@ -0,0 +1,12 @@
...
+
+Not all event producers will produce CloudEvents natively. As a result,
+some "adapter" might be needed to convert these events into CloudEvents.
+This will typically mean extracting metadata from the events to be used as
+CloudEvents attributes. In order to promote interoperability across multiple
+implementations of these adpaters, the following documents show the proposed
+algorithms that should be used:
+
+* [AWS S3](adapters/aws-s3.md)
+* [Github](adapters/github.md)
```
cneijenhuis:
```suggestion
* [GitHub](adapters/github.md)
```

adapters/aws-s3.md
duglin: fixed

adapters.md
duglin: fixed

adapters.md
duglin: fixed

adapters/aws-s3.md
duglin: ok I think i fixed this.

adapters/aws-s3.md
duglin: ok - using both now.

cneijenhuis: With my reading of the S3 docs, that should be unique (per source) - but it's worth double-checking with someone from AWS.

cneijenhuis:
The general section and the S3 adapters LGTM 👍 

(Keep in mind I'm not an expert on S3 events, though...)

duglin:
Per the latest spec, I changed the gitlab events to use "Current time" for `time` since I couldn't consistently get the event time from the gitlab event.

duglin:
change my mind - reverted `time` change

clemensv:
I'm all for getting this merged.

duglin:
all - did make some minor changes today but nothing big. After this latest round of checks I feel pretty good about its state (I was worried before), so please review this before our next call (July 11) and let's see if we can get this one in.  I think the implied guidance for people on how to fill in certain fields (like source vs subject) can be useful.  I'll think about possibly adding some text to the primer to make sure people know to look at it.

duglin:
Approved on 7/11 call

PR #442, opened 2019-06-08
A draft of an extension defining attributes indicating correlation-id…

lennygran:
A draft of an extension defining attributes indicating correlation-id of events for auditing and other similar purposes. I see the need and wide use of this functionality, especially for enterprise applications, but specific semantic can be changed based on the discussion.

### DIFF
--- a/documented-extensions.md
+++ b/documented-extensions.md
@@ -52,3 +52,4 @@ their `Attributes`.
 - [Partitioning](extensions/partitioning.md)
 - [Sampling](extensions/sampled-rate.md)
 - [Sequence](extensions/sequence.md)
+- [Correlation](extensions/correlation.md)
--- a/extensions/correlation.md
+++ b/extensions/correlation.md
@@ -1,0 +1,45 @@
+# Correlation 
+
+This extension defines two attributes that can be included within a CloudEvent
+to describe an identifier for correlating multiple events produced by one or more  
+event sources in order to identify related events for auditing purposes. 
+
+The `correlationid` attribute represents the value of this event's correlation 
+attribute. The exact value and meaning of this attribute is defined by the 
+`correlationtype` attribute. If the `correlationtype` is missing, event 
+consumers MUST assume the default value of `uuid`. If the `correlationtype` is not 
+defined in this specification, event consumers will need to have some out-of-band
+communication with the event producer to understand how to interpret the value
+of the attribute.
+
+## Attributes
+
+### correlationid
+
+- Type: `String`
+- Description: Value expressing the correlation identifier of the event. 
+- Constraints:
+  - REQUIRED, if this extension is used
+  - MUST be a non-empty string in the format defined by the `correlationtype`. 
+
+### correlationtype
+
+- Type: `String`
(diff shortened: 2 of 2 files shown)


duglin:
Can you sign your commits?

duglin:
If the correlationID is just a string, why is it not sufficient to simply do a strcmp to see if it matches some other event's correlationID? ie. why do we need the correlationType?  The only reason I can think of that we may need the "type" is if the value has some semantic meaning outside of the strcmp/grouping - and if so, then it seems like "type" needs to be far more descriptive than something like "uuid". It seems like it would need to be more like a URL to some spec that defines the meaning of the ID. Similar to how a schema URL tells us about the shape/meaning of `data`.

But, then at that point is there a fundamental difference between:
```
ce-correlationID: 123-asd
ce-correlationType: url-to-spec-that-I-have-out-of-band-knowledge-of
```
and
```
ce-myCorrelationID: 123-asd
```
where I have similar out of band knowledge to know what "myCorrelationID" means?

I'm not necessarily against the idea of a correlationID type of attribute, but I tend to think that a generic term like `correlationID` doesn't really help much and a more semantically meaningful name would be more useful. This would then also allow for more than one correlationID type of attribute to be used at the same time.

lennygran:
Thank you for great feedback. I see and share your point. I’ll think about it and update the PR.

lennygran:
@duglin I simplified the extension based on your suggestion. I believe it would be easier to use and understand now.

clemensv:
We have an extension for the W3C Trace Context, which is a correlation id. https://github.com/cloudevents/spec/blob/master/extensions/distributed-tracing.md

Is your use-case significantly different? What is auditing in your view?

duglin:
@lennygran I guess I wasn't clear enough in my comment  :-)

I tend to think that defining an attribute called `correlationid` isn't very helpful because by itself it has no real meaning. Take our `id` attribute. The name alone isn't helpful. The description of it is where it becomes useful. It clearly states what it's value is meant to represent and how it is to be used. This is why we just went thru some long discussions around uniqueness checking... some people thought that `id` could be used for other purposes and that was causing confusion and was potentially going to add complexity to the spec to resolve it.  `correlationid` has the same potential.

`correlationid`, as currently defined, has no real meaning and therefore no real value. Correlation can mean many different things - it could be used to group events (but by what?), it could be used to relate to some common entity (e.g. the occurrence if there are multiple events for that occurrence), it could point to some other event that caused this event, it could point to the predecessor event if the events are ordered, etc...  And, what if there is more than one type of relationship people need to express at the same time? Which one gets to use `correlationid` and which then has to be renamed? And how will the consumers know this?

My point is that I think when an event producer wants to describe a relationship between the event and something else a well-defined/named attribute and description should be defined. Then there's no ambiguity around what its value means and how it is to be used. It doesn't change based on the producer's current needs, which may change over time or be different from other producers.

Net: I think we either need to define a single meaning/value-space for this attribute (which I don't think we should do), or not define it at all.

I would prefer that we don't define it at all and instead add some text to the Primer explaining why we don't have this attribute (and how producers should define their own well-defined extensions for this purpose) since this has come up before and that probably means we need to write it down.

PR #444, opened 2019-06-08
A draft of an extension defining attribute for a Tenant Identifier

lennygran:
This extension defines an attribute that can be included within a CloudEvent
to describe a tenant identifier for the event produced by an event source in
multi-tenant systems. This attribute allows intermediaries to apply functional
tenant-based logic, such as routing, scheduling or processing priorities, etc.
without a need to understand the event body.

### DIFF
--- a/documented-extensions.md
+++ b/documented-extensions.md
@@ -52,3 +52,4 @@ their `Attributes`.
 - [Partitioning](extensions/partitioning.md)
 - [Sampling](extensions/sampled-rate.md)
 - [Sequence](extensions/sequence.md)
+- [Tenant Identifier](extensions/tenant.md)
--- a/extensions/tenant.md
+++ b/extensions/tenant.md
@@ -1,0 +1,36 @@
+# Tenant Identifier 
+
+This extension defines an attributes that can be included within a CloudEvent
+to describe a tenant identifier for the event produced by an event source in 
+multi-tenant systems. This attribute allow intermediaries to apply functional 
+tenant-based logic, such as routing, scheduling or processing priorities, etc. 
+without a need to understand the event body.   
+
+The `tenantid` attribute represents the value of this event's Tenant Identifier 
+attribute. 
+
+## Attributes
+
+### tenantid
+
+- Type: `String`
+- Description: Value expressing the tenant identifier of the event. 
+- Examples:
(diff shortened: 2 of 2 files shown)


JemDay:
I understand the purpose of this but i think it may need a bit more guidance on usage, eg..

- If events flow across multiple provider platforms is the expectation that tenant-id is preserved, overwritten, augmented, etc ?
- Is the assumption that a 'tenant-id' would be added at the point-of-ingress into a provider platform.
- Should Id's be stripped at point of egress (i'm not sure how that would happen).

lennygran:
@JemDay I addressed your comment with the commit above. The rationale is the following: because the tenant-id is defined by the event producer, it should stay with the event unchanged, hence the middleware at multiple provider platforms should just propagate it through.

cneijenhuis:
My understanding was that the tenant id is part of the `source` attribute. It "identifies the context in which an event happened." To me, the tenant id is an important part of the context.

An example source from the spec is: `/cloudevents/spec/pull/123` - I'd say either only `cloudevents` or `cloudevents/spec` is the tenant id.

An example source from Azure (see https://docs.microsoft.com/en-us/azure/event-grid/cloudevents-schema#cloudevent-schema): 
`/subscriptions/{subscription-id}/resourceGroups/{resource-group}/providers/Microsoft.Storage/storageAccounts/{storage-account}#blobServices/default/containers/{storage-container}/blobs/{new-file}`
Either `storage-account` or `storage-container` are tenant ids.

My hope is that, if the tenant id is part of the standardized source attribute, performing "routing, scheduling or processing priorities" will be supported by more intermediaries than if the tenant id is an extension attribute.

I see where the PR comes from though, as this isn't clearly spelled out in the `source` description. Maybe you could instead open up a PR that clarifies that a tenant id should be part of it?

lennygran:
Yes, @cneijenhuis, the lack of semantic clarity on the `source` attribute is contributed to this new extension. I also see a value of clarifying an existing attribute that can carry the same information instead of loading spec with more extensions.

clemensv:
"source" is a URI and as such can identify a tenant easily, for instance: 

urn:tenant.myservice.net:context

(using RFC2141 as one way to do that)

PR #449, opened 2019-06-11
Address Persistence in Primer

### LINKED ISSUE #143: Integrity requirement: how do we verify that no bits have been changed?
**DISCLAIMER**: This will likely not be included in v0.1 of the specification, barring further discussion on this issue.

When an event is published, should the publisher be required to include fields related to the integrity of the event envelope, payload, or both (such as checksums), that allow consumers to verify that the event has not changed? There are two primary concerns related to this issue:
* Malicious middleware that intentionally changes payload values
* Bits have been dropped/mangled during transport or conversion between formats

### LINKED ISSUE #276: CloudEvent Persistence?
As I discussed during community time it seems that there is a certain assumption of transience in this specification and the discussions occurring.  In other words, that relevant concerns are resolved within the context of the connection over which the CloudEvent content is sent and not addressed by CloudEvents.  Concrete examples include authentication as a property of an HTTP connection and event transmission fidelity validated by a TCP checksum.

This issue is opened to facilitate some discussion about the topic.

Certainly some of these concerns should not be solved by CloudEvents and It may be best not to solve for some others given that they comprise a good deal of complexity as well as an ever evolving set of best practices.  That acknowledged, if a CloudEvent is to be written without alteration to a producer and consumer intermediary such as a Kafka or Kinesis the producer's connection context, without event augmentation, is not available to consumers that may have requirements necessitating that information.

A concrete example of this might involve a Kafka stream operated and archived by a banking organization that is used to record, analyze, and actuate inter-account transfers of capital.  While it would be appropriate for me to place an authorization of capital transfer from my account to another, it would be inappropriate for another account holder to have placed the same event on the topic.  Whereas I would be unlikely to directly place events onto such a log, a transfer authorizations service may well do so while a personal information storage service should not.  As an example solution mechanism, a user-specific (and, perhaps, additionally a system-specific) private key signed hash of the CloudEvent could be validated using the associated public key(s) by consumers of the event that required some guarantee of the identity and authorization of the producer.

Obviously any such information could be added to the CloudEvent's [data attribute](https://github.com/cloudevents/spec/blob/master/spec.md#data-attribute) or added as an [extension](https://github.com/cloudevents/spec/blob/master/spec.md#extensions).  However, these needs are present across domains.  Possible alternatives to addressing these concerns within CloudEvents include adding envelopes around persisted CloudEvents or custom solutions within appropriate attribute values of CloudEvents.

Potential questions:
If any, what contextual attributes from connections should be considered for inclusion?
What bases would be helpful in distinguishing between context attributes that should or should not be included?
What adverse outcomes would adding context attributes to the CloudEvents specification cause?

### LINKED ISSUE #378: Producer Identity
As part of #276, information about the identity of the producer can be lost once the connection across which the event is sent gets terminated.  Is addressing this need in certain systems within scope for CloudEvents?

erikerikson:
Briefly detail the importance and challenges of persistence and the measured approach to addressing it and its requirements within the CloudEvents specification.

Closes #143
Closes #276
Closes #378

Signed-off-by: Erik Erikson <erik.erikson@gmail.com>

### DIFF
--- a/primer.md
+++ b/primer.md
@@ -135,6 +135,7 @@ The following are considered beyond the scope of the specification:
 - Language-specific runtime APIs
 - Selecting a single identity/access control system
 - Inclusion of transport-level routing information
+- Event persistence processes
 
 The CloudEvents spec will not include transport-level routing
 information (e.g. a destination URL to which the event is being sent).
@@ -157,6 +158,19 @@ CloudEvent that was intended for a webhook should be deliverable to a
 dead-letter queue if the webhook address is unavailable. That dead-letter
 queue should be able to feed events to new actions that the original event
 emitter never imagined.
+
+The CloudEvents that are produced and consumed within and across systems trigger
+behaviors that derive value. As such, archiving and/or replaying events can be
+valuable for debugging or replication purposes. However, persisting an event
+removes the contextual information available during transmission such as the
+identity and rights of the producer, fidelity validation mechanisms, or
+confidentiality protections. Additionally, persistence can add complexity and
+challenge to meeting user's requirements. For example, the repeated use of a
+private key for encryption or signing purposes increases the information
+available to attackers and thereby reduces security. It is expected that
+attributes may be defined that facilitate meeting persistence requirements but
+it is expected that these will continuously evolve along with industry best
+practices and advancements.
 
 
 ## Architecture


erikerikson:
Closing #143 with this may be controversial.  I can easily accept removing the closure statement for that issue.

clemensv:
I don't understand the particular concern here. The sender context is commonly evaluated when the event is accepted by an event broker and then events are "persisted" either in a volatile or durable store where they are held either for pull pickup or push delivery.

erikerikson:
@clemensv recording the results or artifacts of the evaluation of the context is not provided for in the CloudEvent.  Examples: creator identity, confidentiality, irrefutability.

Consider [event sourcing](https://martinfowler.com/eaaDev/EventSourcing.html) and particularly the permanent storage of the event for repeated reprocessing.  This declares we're not yet solving the problems of persisting the context.

Note that the context can be vital for business logic.  Consider an event demanding a transfer of funds from your account to mine.  The identity of the creator and their rights are crucial for correctly performing the transfer or detecting an attempt at fraud.

Effectively, this declares that CloudEvent users will have to solve those problems for themselves.  Users will have to supply that information in the `data` field or augment the CloudEvent prior to persistence.  This section makes that explicit.

Does this help you understand the concern?  If so, perhaps you can identify what helped and I can clarify the text.  If not, can you be more specific or detail what part of the narrative seems to break down or fail to communicate a useful purpose?

duglin:
"you're on your own" is how interpreted the PR  :-)

clemensv:
@erikerikson I might just be puzzled that this needs explaining. I also think your explanation in the comment here is better than the text.

erikerikson:
I think that we would be (within the context of the primer) not explaining so much as declaring.  In particular declaring that persistence concerns are currently out of scope while recognizing they are important and that the community of users may want to solve some of those concerns.  If nothing else, the statements may help to reduce the number of annoying contributors like me. ;). I mean...  I still would like those things solved for me (I'm lazy, albeit willing to work very hard to be so) but at least with this I can accept that my concerns have been considered rather than being off the radar.  For the percentage that don't read the primer, they can be pointed here when they ask for about such things which not only declares the out-of-scope'ness but suggests a productive way to join the community to expend their energy towards a solution (using perhaps standardized mechanisms supported via extensions).
It may be that this PR should just be closed and no or a different remark made. I would accept that opinion of the group.

duglin:
Approved on 7/11 call

Issue #448, opened 2019-06-11
Complete or remove "Cloudbursting" primer section


erikerikson:
The [primer retains the following](https://github.com/cloudevents/spec/blame/master/primer.md#L625):

```
Cloudbursting

TBD
```

This should likely be completed or removed before the 1.0 release.

PR #450, opened 2019-06-11
Remove blank section

duglin:
We can always recreate the section if someone has text they want to add.

Thank to @erikerikson for noticing

Closes: #448


duglin:
Thanks @Vlaaaaaaad 
Merging since it's just a syntax clean-up PR.

PR #455, opened 2019-06-17
Fix README table - missing protobuf and reorder

duglin:
It was noticed that we were missing the protobuf spec - while in there I alphabetized things


duglin:
If I can get a review/LGTM then I can merge since this is just administrivia

duglin:
thanks @Vlaaaaaaad

PR #457, opened 2019-06-18
Added data payload

### LINKED ISSUE #261: Map and Object type problems
This is related to #184. Currently the spec introduces a lot of confusion around Maps and Objects when it comes to JSON.

So, the data attribute is an Object, which means it can be Map, String or Binary. Map is a map of name to Object pairs. So, if I have a data attribute of:

```
"data": {"foo": true}
```

It's not a Map. Why? Because `true` is not a valid Object according to the CloudEvenvts spec (a valid object must be a Map, String or Binary), and the CloudEvents spec says that Maps are mappings of Strings to Objects. So the above is actually invalid, if you wanted to pass the above, you would have to encode it like this:

```
"data": "{\"foo\": true}"
```

Now the data is a String, and so it's a valid CloudEvents Object, and it can be decoded to JSON. However this contradicts the JSON encoding spec, which says that a JSON object can be directly embedded into the Event json, which would effectively make the above event unencodable as soon as you tried to convert it from the JSON encoding to another encoding.

I think there are a number of points in the spec where JSON and the Map type are confused, particularly in the way JSON events are encoded. Either, the CloudEvents spec should support JSON as a top level concept, replacing Map with JSON, or, the Object type should be redefined to be either String or Binary, but not Map, and force JSON to be encoded as a String. If this doesn't happen, the CloudEvents implementations I think will forever be plagued with issues arising from ambiguities of when to treat something as JSON and when to treat it as a String, the result being things like something going in as JSON, but coming out as a String, or vice versa, especially when events go through different encodings, for example, through JSON encoding, then HTTP, then MQTT, etc.

jroper:
Fixes #261

This change causes data to not be thought of as an attribute, rather to be considered an opaque blob described by the content type. The name given is data payload. The wording in the JSON encoding has been modified to reflect this.

### DIFF
--- a/json-format.md
+++ b/json-format.md
@@ -91,9 +91,6 @@ respective CloudEvents type when the mapping rules are fulfilled.
 
 ### 2.3. Mapping Any-typed Attributes
 
-The CloudEvents `data` attribute is `Any`-typed, meaning that it holds a value
-of any valid type. `Map` entry values are also `Any` typed.
-
 If an implementation determines that the actual type of an `Any` is a `String`,
 the value MUST be represented as [JSON string][json-string] expression; for
 `Binary`, the value MUST represented as [JSON string][json-string] expression
@@ -114,9 +111,6 @@ The following table shows exemplary mappings:
 | id              | String        | "1234-1234-1234"                |
 | time            | Timestamp     | "2018-04-05T17:31:00Z"          |
 | datacontenttype | String        | "application/json"              |
-| data            | String        | "<much wow=\"xml\"/>"           |
-| data            | Binary        | "Q2xvdWRFdmVudHM="              |
-| data            | Map           | { "objA" : "vA", "objB", "vB" } |
 
 ### 2.5. JSONSchema Validation
 
@@ -136,40 +130,45 @@ the [type system mapping](#22-type-system-mapping).
 
 ### 3.1. Special Handling of the "data" Attribute
 
-The mapping of the `Any`-typed `data` attribute follows the rules laid out in
-[Section 2.3.](#23-mapping-any-typed-attributes), with two additional rules:
+The `data` attribute has no specified type, and so is treated specially,
+according to the following rules. In cases where multiple rules apply to the
+same data, the top most rule MUST be used.
 
-First, if an implementation determines that the type of the `data` attribute is
-`Binary` or `String`, it MUST inspect the `datacontenttype` attribute to
-determine whether it is indicated that the data value contains JSON data.
+If the data attribute is known by the implementation to be JSON, the
+implementation MUST translate the `data` attribute value into a
+[JSON value][json-value], and set the `data` attribute of the envelope JSON
+object to this JSON value. An implementation MUST consider events with the
+following `datacontenttype` values as JSON:
 
-If the `datacontenttype` value is either ["application/json"][rfc4627] or any
-media type with a [structured +json suffix][rfc6839], the implementation MUST
-translate the `data` attribute value into a [JSON value][json-value], and set
-the `data` attribute of the envelope JSON object to this JSON value.
+* [`"application/json"`][rfc4627]
+* Any value with a [structured `+json` suffix][rfc6839]
 
-If the `datacontenttype` value does not follow the [structured +json
-suffix][rfc6839] but is known to use JSON encoding, the implementation MUST
-translate the `data` attribute value into a [JSON value][json-value], and set
-the `data` attribute of the envelope JSON object to this JSON value. Its typical
-examples are, but not limited to, `text/json`,
-[`application/json-seq`][json-seq] and
-[`application/geo+json-seq`][json-geoseq].
+An implementation MAY use other means to know if an event is representable as
+JSON, for example, using other well known `datacontentype` values such as
+`text/json` or `application/json-seq`][json-seq], or through inspection of
+types passed to an API.
 
-Unlike all other attributes, for which value types are restricted to strings per
-the [type-system mapping](#22-type-system-mapping), the resulting `data` member
-[JSON value][json-value] is unrestricted, and MAY also contain numeric and
-logical JSON types.
+If the top-level type of `datacontenttype` is `text` (eg, `text/plain`), the
+implementation MUST translate the `data` attribute into a
+[JSON string][json-string] (taking the `datacontentencoding` attribute into
+consideration if the value is originally binary) and set the `data` attribute
+of the envelope JSON object to this JSON string.
 
-Second, whether a Base64-encoded string in the data attribute is treated as
-`Binary` or as a `String` is also determined by the `datacontenttype` value. If
-the `datacontenttype` media type is known to contain text, the data attribute
-value is not further interpreted and treated as a text string. Otherwise, it is
-decoded and treated as a binary value.
+If the `datacontenttype` is anything else, or is not set, the implementation
+MUST translate the value to a [Base64-encoded][base64] 
+[JSON string][json-string], and set the `data` attribute of the envelope JSON
+object to this JSON string.
+
+Note that even if an implementation knows that the type of the `data` attribute
+is a `String`, if the `datacontenttype` top-level type is not `text`, the value
+MUST still be [Base64-encoded][base64]. This so that, when decoding, a consumer
+might not have the same knowledge that that `data` attribute is a `String`, and
+so without the indication of `text` from the `datacontenttype`, cannot know not
+to Base64 decode the String into `Binary`.
 
 ### 3.2. Examples
 
-Example event with `String`-valued `data`:
+Example event with `text`-valued `data`:
 
 ```JSON
 {
@@ -187,7 +186,7 @@ Example event with `String`-valued `data`:
 }
 ```
 
-Example event with `Binary`-valued data
+Example event with `Binary`-valued data:
 
 ```JSON
 {
@@ -205,8 +204,7 @@ Example event with `Binary`-valued data
 }
 ```
 
-Example event with JSON data for the "data" member, either derived from a `Map`
-or [JSON data](#31-special-handling-of-the-data-attribute) data:
+Example event with JSON valued data:
 
 ```JSON
 {
--- a/spec.md
+++ b/spec.md
@@ -448,10 +448,15 @@ encapsulated within the `data` attribute.
 
 ### data
 
-- Type: `Any`
+- Type: Unspecified, but MUST be encodable as `Binary`.
 - Description: The event payload. The payload depends on the `type` and the
-  `schemaurl`. It is encoded into a media format which is specified by the
-  `datacontenttype` attribute (e.g. application/json).
+  `schemaurl`. The `Binary` encoding of the `data` is specified by the
+  `datacontenttype` attribute (e.g. application/json). Implementations of this
+  spec, including transport bindings, format encodings, and APIs, MAY represent
+  the `data` attribute in whatever form they choose. For example, the JSON
+  binding MAY represent JSON data as a JSON value. If the type of the `data` is
+  not natively representable by the implementation or binding, then the
+  implementation or binding MUST represent it in its `Binary` encoding.
 - Constraints:
   - OPTIONAL
 


json-format.md
```diff
@@ -134,42 +132,38 @@ become members of the JSON object, with the respective JSON object member name
...
-
-If the `datacontenttype` value does not follow the [structured +json
-suffix][rfc6839] but is known to use JSON encoding, the implementation MUST
-translate the `data` attribute value into a [JSON value][json-value], and set
-the `data` attribute of the envelope JSON object to this JSON value. Its typical
-examples are, but not limited to, `text/json`,
+If the `datacontenttype` is [`"application/json"`][rfc4627], any structured
+media type with a [structured `+json` suffix][rfc6839], any other media type
+that is known by the implementation to use JSON encoding, or if through some
+other means the implementation knows that the payload is JSON, the data payload
```
jroper: Note the "some other means" here is intended to allow for programmatic APIs, where the payload might allow any type, or where a polymorphic type is used that allows JSON AST, to set the data to JSON, even if the user hasn't explicitly set the content type to a JSON content type.

clemensv:
We already explicitly call out that the data attribute is the payload:

Description: The event payload. The payload depends on the type and the schemaurl. It is encoded into a media format which is specified by the datacontenttype attribute (e.g. application/json).

The objection here seems to be whether data may be called an attribute. In the JSON structured format and any other derivatives I can picture (CBOR; MsgPack, Avro, etc) the data attribute is just another key in the event's attribute map.  

Only for the binary transfer modes we break the payload out and special-case it. 

I don't see how this change is a substantial improvement over what we have.

jroper:
It's not about whether it's called an attribute, it's whether it is an attribute. An attribute has a value that has a type of one of the CloudEvent event types. But the data payload is not one of the CloudEvent types, and in fact in most language API implementations, it can be any type, see the extensive discussion in #261 on why this is so. And therefore since it doesn't fit into the CloudEvents type system, it can't be an attribute.

In JSON, data is specially cased, there's several paragraphs on how the data payload is treated. Yes it gets encoded into the JSON as a member like all other attributes, but its value has a number of very particular rules, under what circumstances embedded directly as JSON, under what circumstances it's embedded as a string, and under what circumstances it's embedded as binary. It needs this precisely because the payloads type doesn't fit the CloudEvent data types.

spec.md
```diff
@@ -202,12 +202,12 @@ attributes designated as REQUIRED and MAY include one or more OPTIONAL context
 attributes.
 
 These attributes, while descriptive of the event, are designed such that they
-can be serialized independent of the event data. This allows for them to be
-inspected at the destination without having to deserialize the event data.
+can be serialized independent of the data payload. This allows for them to be
+inspected at the destination without having to deserialize the data payload.
```
cneijenhuis: I think `event data` referred to an abstract notion (attributes of event vs event data). Removing "event" does not improve this paragraph IMO.

Either:
* Keep it as it is
* Name it `event data payload`

spec.md
```diff
@@ -202,12 +202,12 @@ attributes designated as REQUIRED and MAY include one or more OPTIONAL context
...
 
 The choice of serialization mechanism will determine how the context attributes
-and the event data will be materialized. For example, in the case of a JSON
-serialization, the context attributes and the event data might both appear
+and the data payload will be materialized. For example, in the case of a JSON
+serialization, the context attributes and the data payload might both appear
```
cneijenhuis: See `event data` above.

spec.md
```diff
@@ -433,26 +432,25 @@ that performs actions across multiple types of events. To support such use
 cases, the event producer will need to add additional identity attributes to the
 "context attributes" which the event consumers can use to correlate this event
 with the other events. If such identity attributes happen to be part of the
-event "data", the event producer SHOULD also add the identity attributes to the
+data payload, the event producer SHOULD also add the identity attributes to the
```
cneijenhuis: See `event data` above.

spec.md
```diff
@@ -433,26 +432,25 @@ that performs actions across multiple types of events. To support such use
...
 "context attributes" so that event consumers can easily access this information
-without needing to decode and examine the event data. Such identity attributes
+without needing to decode and examine the data payload. Such identity attributes
```
cneijenhuis: See `event data` above.

spec.md
```diff
@@ -433,26 +432,25 @@ that performs actions across multiple types of events. To support such use
...
+## Data Payload
 
 As defined by the term [Data](#data), CloudEvents MAY include domain-specific
-information about the occurrence. When present, this information will be
-encapsulated within the `data` attribute.
+information about the occurrence. When present, this information SHALL be
+encapsulated by the data payload.
 
-### data
+The data payload is optional, some events MAY have no corresponding
```
cneijenhuis: `OPTIONAL`?

spec.md
```diff
@@ -433,26 +432,25 @@ that performs actions across multiple types of events. To support such use
...
+domain-specific information, rather, all information relevant to the Occurance
+SHOULD be found in the Context, such as the `type` attribute.
```
cneijenhuis: In general, I don't understand why it is stated three times that the payload is optional:

"CloudEvents MAY include"
"payload is OPTIONAL"
"events MAY have no"

> all information relevant to the Occurance SHOULD be found in the Context, such as the `type` attribute.

I find that this section nudges the reader towards leaving data empty, and encourages to cram more into the context. I think it is at odds with what we say in the primer.

---

TL;DR: What value does this paragraph bring? What do we miss if we remove lines 447 to 449?

amqp-transport-binding.md
```diff
@@ -236,12 +236,12 @@ content-type: application/cloudevents+json; charset=UTF-8
...
 
-The chosen [event format](#14-event-formats) defines how all attributes,
-including the `data` attribute, are represented.
+The chosen [event format](#14-event-formats) defines how all attributes and the
+data payload are represented.
 
-The event metadata and data is then rendered in accordance with the event format
-specification and the resulting data becomes the AMQP application [data][data]
-section.
+The event metadata and payload  is then rendered in accordance with the event
```
cneijenhuis: `s/payload  /payload /`

I'm also not sure why you've dropped `data` here, shouldn't it be `data payload`?

http-transport-binding.md
```diff
@@ -103,33 +103,33 @@ identically to [HTTP over TLS]([RFC2818][RFC2818]).
 This specification does not further define any of the [CloudEvents][ce] event
 attributes.
 
-Two of the event attributes, `datacontenttype` and `data` are handled specially
-and mapped onto HTTP constructs, all other attributes are transferred as
-metadata without further interpretation.
+The `datacontenttype` attribute is handled specially and mapped onto HTTP
+constructs, all other attributes are transferred as metadata without further
+interpretation.
```
cneijenhuis: I think it's still worth mentioning that the data payload is mapped onto HTTP constructs.

Maybe:

> The data payload and the `datacontenttype` attribute is handled specially and mapped onto HTTP constructs, all other attributes are transferred as metadata without further interpretation.

❓

mqtt-transport-binding.md
```diff
@@ -215,10 +214,10 @@ source: /mycontext/subcontext
 
 ### 3.2. Structured Content Mode
 
-The _structured_ content mode keeps event metadata and data together in the
-payload, allowing simple forwarding of the same event across multiple routing
-hops, and across multiple transports. This is the only supported mode for MQTT
-3.1.1
+The _structured_ content mode keeps event metadata and data payload together in
+the payload, allowing simple forwarding of the same event across multiple
```
cneijenhuis: See `event data` discussion in `spec.md`.

This sentence is confusing now: `keeps [...] data payload in the payload`?

I recommend against this change.

nats-transport-binding.md
```diff
@@ -103,17 +103,17 @@ With NATS, the content mode is always _structured_ and the NATS message payload
 MUST be the [JSON event format][json-format] serialized as specified by the
 [UTF-8][rfc3629] encoded JSON text for use in NATS.
 
-The _structured_ content mode keeps event metadata and data together, allowing
-simple forwarding of the same event across multiple routing hops, and across
-multiple transports.
+The _structured_ content mode keeps event metadata and data payload together,
```
cneijenhuis: See `event data` discussion in `spec.md`.

nats-transport-binding.md
```diff
@@ -103,17 +103,17 @@ With NATS, the content mode is always _structured_ and the NATS message payload
...
+across multiple transports.
 
 ### 3.1 Event Data Encoding
 
 The chosen [event format](#14-event-formats) defines how all attributes,
 including the payload, are represented.
 
-The event metadata and data MUST then be rendered in accordance with the event
-format specification and the resulting data becomes the payload.
+The event metadata and data payload MUST then be rendered in accordance with
```
cneijenhuis: See `event data` discussion in `spec.md`.

clemensv:
> It's not about whether it's called an attribute, it's whether it is an attribute. An attribute has a value that has a type of one of the CloudEvent event types. But the data payload is not one of the CloudEvent types

We define Binary and we define String.

The `data` attribute value is generally a `datacontenttype`-qualified binary or string (potentially further qualified as binary via `datacontentencoding`) or, in the special case of the JSON encoding, an inlined JSON value for simplicity. There's no reason why it can't just be a Timestamp expression.

We already separate the `data` attribute from the context attributes. As such, it is already different.

I still don't see a justification for such a sweeping set of edits.

duglin:
Just my 2 cents... I think I understand where @jroper is coming from. We're, sort of, using the term "attribute" in two different ways. First as a generic term to mean "some data that carries some semantics meaning" - and in this sense we have the freedom to use it for "normal" attributes like `time` and well as special attributes like `data`, even though `data` doesn't always get serialized as an "attribute". The second way we use it is as a "normal" attributes, in particular in the structured format where it's the property (ala attribute) name, and `data` is only this kind of "attribute" in the structured format.  Hopefully, this is correct and I'm not missing his point.

So, from a purist perspective I can see why it might be nice to have two different terms, perhaps property (for the abstract concept) and attribute (when it really is an attribute/key at serialization time). And I guess @jroper decided to attack problem this by calling `data` a `payload` instead of an attribute.

But to me it comes down to whether what's currently in the spec is confusing and will lead to interop issues - even if it does take some liberties with the definition of "attribute". And, TBH, I'm not sure I see it being that big of an issue - or at least not one that couldn't be fixed with a sentence or two added at the right spot (like in the type system section to indicated `data` is special).

jroper:
I think I've been misunderstood here. The name is not the important thing. We can change that back, I don't mind. The problems that I'm trying to address are the holes in the spec - part of the reason for changing the name is that I think that treating it as an attribute is in part what causes the confusion that led to these holes, but the name is not important to solve, the holes are, so let's stop talking about the name for a minute, I'll change it back if it's so controversial.

Currently, the type of the data attribute is `Any`. Which means it can be a `String`, a `Binary`, an `Integer` or a `Map`. So, here's some questions:

* In any of the encodings that have a binary content mode, if the data attribute is an `Integer`, how should it be encoded? The spec doesn't answer this. Should it be a decimal? Hexadecimal? Twos complement big/little endian? And how should it be communicated to the consumer of the event that the value is an Integer so that it can decode it correctly back into an attribute of type `Integer`? That's also completely unspecified at the moment. Right now the most likely thing to happen, depending on how the person implementing it decides to handle this, would be that the producer encodes it as a String, and then the consumer will consume it as either a `String` or `Binary` depending on what `datacontenttype`, so from the producer to the consumer, the data attribute will have changed type, and a domain application, which when it asks the event for the data `Integer`, will not get an `Integer`, it'll get a corrupted blob of bytes or characters with no way of knowing how to convert them back to an `Integer`.
* In any of the encodings that have a binary content mode, if the data attribute is a `Map`, how should it be encoded? Again, the spec doesn't answer this. You might say JSON, I'll cover why JSON doesn't work in the next point, but if JSON, when it gets decoded at the other end (eg, let's say we sending an HTTP request, and the server decodes it), how does the server know to decode it as a Map, rather than treat it as binary, or a string? It can't set the `Content-Type` to `application/json` because `datacontenttype` is already mapped to that, and let's say the `datacontenttype` is already set to `foo/bar`, and the body is of type `Map`, nothing in the spec says that's illegal. How does it know to convert it back to a `Map`?
* In the JSON encoding, if the data attribute is a `Map`, how should it be encoded? `Map` and JSON are not compatible, since the type of the values in `Map` are `Any`, which means any member in the JSON could be a `Binary`, and JSON doesn't support `Binary`. Yes, you can encode the `Binary` to a base64ed String, but when you decode, you end up losing information, because you don't know that it was originally a binary, so it stays as String. So round trips corrupt the data.

The problem is that by trying to ascribe a type to the data attribute, we are over specifying it, and it's causing the above mentioned holes in the spec - questions for which the spec doesn't have an answer. And any time there is a question that the spec doesn't have an answer, implementers will have to still make a decision, and they are not all guaranteed to make the same decision, and the result will be incompatible implementations.

So we need to stop saying that the type of `data` is `Any`. It shouldn't have any type. It's value has no semantic meaning to the spec. This is what makes it different to the rest of the attributes, attributes give semantic meaning to the event, that's why they have and need a data type like timestamp or integer or map or string, because to provide that value different implementations need to agree on how the value is to be interpreted. But data has no semantic meaning, and different interpretations don't interpret it at all, it is 100% domain specific. So the data attribute is an opaque value that can be represented as bytes, it doesn't have a type. Certain implementations can treat it specially if through some means they understand more about the data, eg, the JSON encoding will embed it as JSON if it knows its JSON, or as a String if it knows its text. But beyond that, it's opaque.

So, do we agree that these are problems in the spec that need to be fixed?

duglin:
@jroper thanks for that write-up, it helps.  It brings up other questions for me but I'll focus on just one point.... "data shouldn't have a type".  I think that aspect makes a lot of sense and TBH part of assumed that "Any" was supposed to be an opaque blob as you suggested. We just made a mistake by listing a few of the possible choices.  What do you think about focusing the PR more on the smaller set of changes to remove the "data == Any" in the spec?  And then we can see how the group feels about that.

jroper:
@duglin Updated as requested. Essentially I've only modified the definition of the `data` attribute, as well as the description of how it should be encoded to JSON (since the JSON encoding was about how to map its `Any` typed value to JSON).

In the interest of keeping the PR small, I haven't done this change, but I think it could be done, that is, all of the transport bindings that have a binary mode currently say something like this:

```
The `data` attribute is assumed to contain opaque application data that is
encoded as declared by the `datacontenttype` attribute.

An application is free to hold the information in any in-memory representation
of its choosing, but as the value is transposed into HTTP as defined in this
specification, the assumption is that the `data` attribute value is made
available as a sequence of bytes.

For instance, if the declared `datacontenttype` is
`application/json;charset=utf-8`, the expectation is that the `data` attribute
value is made available as [UTF-8][rfc3629] encoded JSON text to HTTP.
```

This has always been a bit odd to me, why are the transport bindings making an assumption that isn't in the main spec? And if so many of them make the same assumption, then shouldn't this be in the main spec? Anyway, with this change, they no longer need to assume anything, since I explicitly say:

```
- Type: Unspecified, but MUST be encodable as `Binary`.
```

Therefore, the above paragraphs can be modified to say something like:

```
The `data` attribute MUST be mapped to the HTTP message body using its `Binary` encoding.
```

Let me know if you think that change should be made.

json-format.md
```diff
@@ -136,40 +130,45 @@ the [type system mapping](#22-type-system-mapping).
...
+[JSON string][json-string] (taking the `datacontentencoding` attribute into
+consideration if the value is originally binary) and set the `data` attribute
+of the envelope JSON object to this JSON string.
+
+If the `datacontenttype` is anything else, or is not set, the implementation
+MUST translate the value to a [Base64-encoded][base64] 
+[JSON string][json-string], and set the `data` attribute of the envelope JSON
+object to this JSON string.
+
+Note that even if an implementation knows that the type of the `data` attribute
```
clemensv: Some of the original text here should have taken the introduction of `datacontentencoding` into account when that was introduced. A String is a binary if and only if `datacontentencoding` says so. Otherwise it's not. So this particular clause is already covered.

jroper: I'm a bit confused actually about the `datacontentencoding` attribute. So, let's say I (an application developer writing a program that emits events) produce an event that is `application/x-csh` by passing a string of bytes as the data attribute value to the cloud events implementation that I'm using and setting the `datacontenttype` to `application/x-csh`. The implementation I'm working with has no idea what `application/x-csh`, it just treats it as a binary blob, like it would `application/octet-stream`, and sends it on to the next producer. But then it gets passed to another implementation, and that implementation knows that that's a c-shell script, which is text. But there's no `datacontentencoding` set, because the previous implementation that didn't know that it was text and so didn't know to ensure that there was a `datacontentencoding` attribute didn't set one.. So, what should it do? Treat it as text? Or as binary?

Another thing, how about `application/xml`? XML actually has an embedded charset parameter (ie `<?xml version="1.0" charset="utf-8"?>`). Should the `datacontentencoding` be set in that case? Are we therefore requiring all CloudEvents implementations, when they receive XML, to parse the XML, extract the charset, and set the `datacontentencoding` to it? I would have thought not, I would have thought they can just pass the binary blob as is without the `datacontentencoding`.

Anyway, this clause is not just redundant for the reason you stated, but it's also redundant because the behavior for when something should be `String` or `Binary` is completely specified above it. The reason I included it is because I can imagine many implementors making the mistake of encoding something like `application/xml` as a `String` when it should be base64ed `Binary`. Afterall, xml is text, so it's intuitive to handle it as text. So I thought this would be a good place to add that reminder to ensure that mistake isn't made.

duglin:
@jroper missed you on the call today :-)  would love it if you could join the next one to be part of the back-n-forth discussions.

We discussed this for quite a while on the call today. And my view of what was said is this...
- it wasn't clear that this PR changed much. Some folks seemed to think it was just a rewording of what's already there
- there was some sympathy (not sure if it was unanimous or not tho) for not using "Any" for `data`, and just saying it's "whatever you want as long as it is encoded per the `datacontenttype` and `datacontentencoding`". And if those properties are not set then `data` needed to be encoded to align with the rest of envelope being used.

Folks on the call... correct me if I got that wrong.

@clemensv said he'd work with you on this to see if there's a proposal to address everyone's concerns.

@jroper it might help if you could list one or two examples that you think are problems. I know you did so in the original issue but I believe those were written a while ago and may have been done so before we had `datacontenttype` and `datacontentencoding` attributes - which people seem to think helps with your concerns. So if you can show how even with those attributes there's still an interop issue I think that would really help.

rperelma:
@duglin I think you summarized today's discussion accurately, and I agree with the direction per your second bullet

jroper:
@duglin I listed 3 examples [in my comment written 6 days ago](https://github.com/cloudevents/spec/pull/457#issuecomment-504633360) that are problems, explaining in detail why they are problems. I'm not sure how to be more specific than that. `datacontentype` and `datacontencoding` says nothing about how would encode a body of type `Integer` such that when the receiving end received it, it would decode it back into the same `Integer`. It doesn't even say how a body of type `Map` should be encoded, what if I had the following protobuf encoding (written hopefully in a pseudocode that makes sense) of a CloudEvent:

```
Map(
  datacontenttype -> String("foo/bar"),
  data -> Map(
    somebinaryvalue -> Binary(0x01020304)
  )
)
```

Now, the above is a valid cloud event (minus a few required attributes like type and schema) according to the spec right? We have a `data` of type `Map`, it has declared a `datacontenttype` of `foo/bar`. Now, let's say we transcode to JSON, how do we do that? The spec actually doesn't say. It says that if the `datacontenttype` is `application/json` or anything that ends with `+json` it should be JSON, but this has a `datacontenttype` of `foo/bar`. So, it should get encoded as `Binary`. But how? There is no defined encoding for `Map` to `Binary`. So let's try and use JSON anyway. We get this:

```json
{
  datacontenttype: "foo/bar",
  data: {
    somebinaryvalue: "AQIDBA=="
  }
}
```

Ok, now lets convert back to the protobuf encoding. We get:

```
Map(
  datacontenttype" -> "foo/bar",
  "data" -> Map(
    "somebinaryvalue" -> String("AQIDBA==")
  )
)
```

But that's not what we originally encoded, the value of `somebinaryvalue` has changed from `Binary` to `String`. It's corrupted. And of course, you might say, why would anyone do the above? They wouldn't, but that's my point, the spec has a partial specification for things that people would never do. And while that might be fine today if we just say "ok everyone, ignore the fact that the spec says that the data can be an Integer", we don't know what use cases people are going to come up with tomorrow. Maybe there might be a good reason why someone might try the above in future. Specs need to be resilient to future possibilities, if they want to guarantee interoperablility not just for todays use cases, but also tomorrows use cases.

Another alternative is that we could fix the fact that the encoding of the data payload is only partially specified. We could say "when an Integer is encoded to binary, it should be 32 bit twos complement little endian encoded, and an additional attribute, independent of `datacontenttype` (because that's set by the user and could be anything), should be set to indicate that it was originally an integer so that we know to apply the same encoding to decode it again". Then at least we have a complete specification, but we're specifying things that we never envision anyone to use. So that's why I'm saying we need to not say that the data attribute can be of type Any.

clemensv:
@jroper If the event has a  ´datacontenttype´ of "foo/bar", and if you don't know how to handle "foo/bar", you can't interpret the body. At that point it's irrelevant what's in it. 

What we define is that if the contenttype is JSON and the event encoding is structured JSON, `data` may contain valid JSON inlined into the event, which means that the two cases that lead issue #261 are both completely legitimate. The first is a JSON object and the second is a JSON string. The JSON encoding already exempts the ´data´ field from the CloudEvents type system. 

What we do allow is that if you omit `datacontenttype` (we spell this out), you can use the CE type system and its mapping into event formats for 'data', whereby the event format chosen for the event counts. Only types with defined mappings are legitimate in that case, for instance you don't get to use boolean since we didn't define it in our abstract type system. If we remove 'Any', we would lose the ability to make a minimal event without ´datacontenttype` that just gives a simple string or a couple of key/value pairs as data. (@duglin I've rethought my position on this since yesterday) 

For when you define a `datacontenttype`, and that is a text-based type and the `data` field is a string in the underlying encoding, you just process the text. If it's a binary in the underlying encoding, you use the default character encoding for the media-type or the character-encoding given in the content-type as a parameter to turn that binary into text. If `datacontenttype` is a type that is binary, any type other than binary can obviously not be interpreted.

jroper:
> @jroper If the event has a ´datacontenttype´ of "foo/bar", and if you don't know how to handle "foo/bar", you can't interpret the body. At that point it's irrelevant what's in it.

This should be true, but I believe the spec as it currently stands forces you to interpret the body in some way, and that's what this PR is attempting to address, it's trying to make it so that an intermediary (neither the producer or consumer, something that doesn't understand the event but is forwarding it, typically transcoding from one encoding to another, such as might happen when bridging event streams between two different clouds that provide different event transport technologies) can forward the body as is without needing to interpret the body.

So let's go to the `Integer` data type. If I have a CloudEvent for which the data payload is the `Integer` 2, and I send it, it shouldn't matter what intermediaries it goes through, how many different times it gets transcoded from JSON to HTTP binary mode to protobuf to Kafka, whatever, at the other end, I should be able to receive a CloudEvent with a data attribute that is of type `Integer` and has a value of 2. Do we agree on that?

If so, explain to me, how will that happen, according to the spec as it currently stands? Because I don't think it's possible with the spec as it stands, I believe that certain transport encodings (in particular, HTTP binary mode) will result in either an implementation failing to encode it because the spec doesn't say how, or it will make its own non spec instructed decision, that will result in a corruption - that is, that will result in the receiver not getting an `Integer` of value 2, and, depending on what implementation specific decisions were made along the way, you could end up with things such as a `String` value of `AgAAAA==` (32 bit little endian twos complement encoding of 2 that's been base64ed) or `Mg==` (ascii decimal encoding of 2 that's been base64ed).

duglin:
Let's take your example, data = `2`, and let me ask you this, even w/o CloudEvents what do you think the HTTP request would look like in a binary format?  In particular, what do you think the HTTP headers around encoding and content-type would look like? I'm assuming the body would just be one byte: `2`

jroper:
@duglin 2 is not a byte, it's a number. Do you mean an ascii 2? That's the byte `0x32`. Or do you actually mean a byte with value `0x02`? Because then you're using a 1 byte integer encoding, which is only useful for values up to 256, or 127 if you're twos complement encoding.

duglin:
I meant ascii `2` or byte `0x32`.  But if you prefer, we can also talk about the byte `0x02`. I suspect a very similar answer for all - but I could be wrong.  Also don't assume that a single byte doesn't mean that leading zeros (in some form) weren't just omitted for brevity.

jroper:
My assumption actually for this is that since the spec already talks about valid integer values being in the range -2^31 to 2^31 - 1, ie the range for twos complement encoded integers, that integers should be twos complement encoded when encoded to binary. And I brought word width because we do really need to be specific about this - sure I can assume that leading zeros are dropped when encoding 2, but not when encoding 128, because `0x80` is -128, not 128, and `0x0080` could be -32768 or 128 depending on endianness.

The point is, none of this stuff is specified. There is nothing saying, when encoded to binary, an `Integer` should look like this, and a `Map` should look like that. There is a mapping from each type to `String`. But not to `Binary`. And there's nothing saying that the mapping to `String` should be used when encoding to `Binary`. And then there's nothing saying how a type that has been mapped from a `Integer` or `Map` to `String` should indicate that it is a `String`. And there's also nothing indicating how, at the other end, the consumer should know that it was originally an `Integer` or `Map`. And you can't use `datacontenttype` to indicate this. Why? Because the user is free to set `datacontenttype` to whatever they want. I can say `datacontenttype` is `foo/bar`, and set the value of `data` to an `Integer` of 2. This can be represented easily as JSON:

```json
{
    "datacontenttype": "foo/bar",
    "data": 2
}
```

No problems there, it's my domains prerogative to say what a media type of `foo/bar` means when the data is an Integer. So when it gets transcoded to HTTP binary mode, the spec requires that `Content-Type` will be set to `foo/bar`, which of course nothing but the eventual end consumer that the user controls understands, so what happens to the `2`? The spec says nothing.

duglin:
ok - it took me a while to finally get what I think you're saying....
1 - if data is json then we're ok when using structured because that json can be data's value
2 - if data is a string (content-type = text) then we're ok because that can be a JSON string
3 - anything else get converted to base64 encoding since the best we can do is treat it as binary and just pass it along.

Do I have that right?  Sorry for being slow.

jroper:
Not quite.

If the encoding is JSON, we're fine (sort of, I don't think we're fine for `Map` but let's leave that aside). This is because JSON has a type system, a property is either string, array, object, integer, boolean, or null, and we can and have mapped that to the CloudEvents type system. You can put the value `2` in there, and JSON knows its a number, so the other end will unambiguously know that that body should be decoded to a CloudEvents `Integer` 2. We have some rules, based on `datacontenttype` for translating `Binary`, so even though JSON doesn't have a binary type, we've made it work.

But with HTTP bodies - there's no type system. There's only bytes. So if you put 2 into an HTTP body, you need to define two rules, one for how to translate from the number 2 to and from bytes, and the other for how it is indicated what the type is so that the other end knows to use the above translation to ensure that it gets back the number 2. Currently, we define neither - or rather, we have a partial definition of how it should be done, for some types, but we have holes where it's not specified for others. We don't define how 2 gets translated to bytes, and we don't define how it's indicated that those bytes should be translated using the codec we used to decode them. Usually, indicating how things are encoded in HTTP is done using the `Content-Type` header, but we've already specified that the `Content-Type` header must contain whatever the user supplied as the `datacontentype` Cloud Events attribute. So we can't rely on the `Content-Type` header. In some cases, when the user has specified a `datacontenttype` that we understand, eg `text`, AND that matches the type of the data attribute they are sending, eg `String`, then it just works. But if the user specifies a `datacontenttype` that is not understood, then it doesn't.

jroper:
To be a little more concrete, let's say I have the following CloudEvent encoded in JSON:

```json
{
  "datacontenttype": "application/octet-stream",
  "data": 2
}
```

Now, lets say it goes through an intemediary that transcodes to HTTP binary mode. What am I going to get? Let's say the implementors of that decide that Integers should be encoded as ascii decimal numbers. What request are they going to produce? Perhaps this:

```
Content-Type: application/octet-stream

2
```

Now, it goes to another intermediary, and this intermediary decides to transcode back to JSON. What's it going to end up as? How does that intermediary know that that "2" is an Integer? It doesn't. It sees the octet `0x32`, it sees the `Content-Type` of `application/octet-stream`, and it says "this body is binary, I have to read the bytes, and base64 them, and set the result to the data". So, it transcodes back to this:

```json
{
  "datacontenttype": "application/octet-stream",
  "data": "Mg=="
}
```

The round trip has changed the event. I'm not sure about you, but in my opinion, that's not acceptable. Even if somehow the intermediary knew to treat the body as a `String`, eg, if the first transcoding introduced a `datacontentencoding` of `us-ascii` (the spec doesn't say to do this), so let's say the HTTP transcoding was:

```
Content-Type: application/octet-stream
datacontenttype: us-ascii

2
```

We would end up with this event:

```json
{
  "datacontenttype": "application/octet-stream",
  "datacontentencoding": "us-ascii",
  "data": "2"
}
```

That's still a corruption in my opinion, because the original event had data set to the number `2`, the round tripped event has data set to the String `"2"`.

So what this PR does is it stops trying to ascribe any types to the `data`. It says, "your data can be anything you want, as long as it has a binary encoding". And then we have some rules for when certain `datacontenttype`'s don't get expressed as base64ed binary in JSON, so that in same cases, it's just a raw string, and in others, its a json map. So, the following CloudEvent:

```json
{
  "datacontenttype": "application/octet-stream",
  "data": 2
}
```

Is actually invalid to start with according to this PR, since the JSON `data` member must be either an object or a string. If a user wants to send the integer 2, with a `datacontenttype` of `application/octet-stream`, it's their responsibility to decide what the byte representation of that will be. If they choose an ascii-decimal 2, they could send this:

```json
{
  "datacontenttype": "application/octet-stream",
  "datacontentencoding": "us-ascii",
  "data": "Mg=="
}
```

Which when transcoded to HTTP, gets translated to:

```
Content-Type: application/octet-stream
datacontentencoding: us-ascii

2
```

And then, when that gets translated back, they will get:

```json
{
  "datacontenttype": "application/octet-stream",
  "datacontentencoding": "us-ascii",
  "data": "Mg=="
}
```

It's identical. So, it worked.

Alternatively, the user may decide that they actually don't want it to appear base64 encoded in the JSON, and the spec allows them to control that by selecting a different `datacontenttype`, one that indicates it's `text`:

```json
{
  "datacontenttype": "text/integer",
  "data": "2"
}
```

Now when transcoded to HTTP, you get:

```
Content-Type: text/integer

2
```

And when that gets transcoded back to JSON, you get:

```json
{
  "datacontenttype": "text/integer",
  "data": "2"
}
```

It's identical, so, it worked.

This all works because we've stopped trying to specify the type of `data`, and we've just said that it's raw bytes that in some cases get handled specially if `datacontenttype` allows us to.

clemensv:
> ```json
> {
>   "datacontenttype": "application/octet-stream",
>   "data": 2
> }
> ```

The premise is incorrect, @jroper. Your starting example is invalid. The data field can only be a JSON number if the `datacontenttype` indicates JSON encoding or if `datacontenttype` is omitted while the event format is JSON. We literally only allow that integer literal if the content of  `data` is unambiguously JSON. 

If "datacontenttype" is "application/octet-stream", the only valid encoding is binary because the media-type itself says it's binary. See https://tools.ietf.org/html/rfc2046#section-4.5.1

jroper:
@clemensv The rules you say there about JSON encoding are written in the spec, as I understanding, as being *additional* rules on top of the regular encoding rules, which say that `Integer` gets mapped to a JSON number, not instead of. But I can see how it could be read to mean your definition, and I think that only further illustrates why typing `data` as `Any` is causing confusion.

But there is a question that I have asked several times that you haven't answered. If `data` is an `Integer` (allowed by the spec because it is of type `Any` and `Integer` is a valid type), and the `datacontenttype` is `application/octet-stream`, how should it be encoded to binary? Or JSON? Or any of our specified encodings? And please link to where in the spec you got your answer from.

clemensv:
@jroper 

> But there is a question that I have asked several times that you haven't answered. If data is an Integer (allowed by the spec because it is of type Any and Integer is a valid type), and the datacontenttype is application/octet-stream, how should it be encoded to binary?

If `datacontenttype` is "application/octet-stream", `data` MUST be binary because it is so declared. The only eligible data type of the available set from 'Any' is 'Binary' because only that can hold the declared content. **`Any` is not a license to pick and incompatible data type with what you declare in `datacontenttype`.**

Whether a media type is binary or text based and thus ought to be mapped to `Binary` or `String` is either obvious from the media type [registry group](https://www.iana.org/assignments/media-types/media-types.xhtml#text) (text or not) or can be understood from the respective spec that defines the media type. 

We don't impose any novel character encoding rules here, meaning that if the `data` content is text, but it MUST be encoded differently from either the event format or the transport message text encoding, it MUST be carried as binary and the encoding MUST be declared in the content-type. That's not specific to CloudEvents, though. HTTP only allows US-ASCII/ISO-8859-1 for any of its content and Unicode is not permitted for "text/plain" unless you explicitly declare it "text/plain; charset=utf-8" and once you do that, the entity body is no longer immediately interpretable as text but you must first run it through the UTF-8 decoder.  

The JSON literal '2' is valid ONLY IF the datacontenttype is either declaring some JSON content-type or if it's empty and the event format is JSON (the latter combo is only ever valid in structured mode). And in those cases it's unambiguously a number 2.

If `data` isn't declared to be of a specific encoding (which can only happen in binary mode when the Content-Type is omitted), the only valid representation for a standalone Integer as a payload is our canonical format and that is a always a string, meaning it never gets interpreted as anything but text at the CloudEvents level.

jroper:
> Any is not a license to pick an incompatible data type with what you declare in datacontenttype.

But the spec doesn't say that! The spec says that `Integer` is a valid subtype of `Any`. It doesn't say that it must be chosen in combination with a `datacontenttype` that will encode it, the spec says `datacontenttype` is optional. Reading this, I think it would be a very reasonable assumption for a user or implementor of the spec to assume that if they send an `Integer`, that they'll get an `Integer` on the other end. And the fact that multiple people have raised this in various issues indicates that I'm not alone in this, so even if by some reading of the spec in combination with reading other RFCs you're understanding is right, the spec has a problem because a spec should make it unambiguously clear how these things work.

The thing is, I think we're in agreement of what the outcome should be, a user is not free to just specify an abstract type, they must specify a type that has an accompanying encoding, using a combination of `datacontenttype` and supplying the right bytes if needed, with exceptions for when its not needed, (eg, when encoding to JSON, and the content type is either json or text, it's not needed). This PR makes that unambiguous, by removing the abstract type, so now rather than CloudEvents having some concept of type in addition to `datacontenttype`, which I think is the biggest source of confusion there because how that abstract `Any` type applies is very confusing, there is just `datacontenttype`. That's all that's needed. And nothing actually changes in the way things get encoded with this PR, only the wording of the spec is being changed to remove the confusion that having two different type concepts is causing.

clemensv:
@jroper the spec says that `datacontenttype` defines "Content type of the data attribute value" and for  "application/octet-stream" you can find in RFC2046 that this means that the content must be binary. Which means that your leading example is in the prior comments false because your `data` content is obviously not conformant to how we express binary and therefore it's misdeclared. If you're choosing a type that makes it impossible for the `datacontenttype` declaration to be truthful, that's a reason to throw.

If `datacontenttype` is omitted, they can specify an Integer, correct. The canonical format for that is a string, so if that goes in as a JSON literal, but pops out as a string on the other end, the assurance we give through the rule set here, is that that string will be made up of characters that can be parsed into an Internet value on your current runtime/architecture if that parser understands JSON integer number encoding. 

That's exactly the same story for data as it is for all attributes.

jroper:
>  If you're choosing a type that makes it impossible for the datacontenttype declaration to be truthful, that's a reason to throw.

But why does the spec even allow it in the first place if it's an error? And, if it is an error why doesn't the spec state that it's an error? And why should the spec say anything about the type of `data` other than it's encoded as described by `datacontenttype`? Isn't `datacontenttype` enough? It's enough for HTTP, it's enough for most protocols that I know of, why is CloudEvents differing here from the rest of the world and trying to ascribe abstract types to the data payload? It's just confusing, the length of this thread and the number of other issues raised in relation to the type system is evidence of that.

>  If datacontenttype is omitted, they can specify an Integer, correct. The canonical format for that is a string,

Ok, so in the spec we have this:

> The following abstract data types are available for use in attributes. Each of these types MAY be represented differently by different event formats and in transport metadata fields. This specification defines a canonical string-encoding for each type that MUST be supported by all implementations.

And your interpretation of that is that when encoding to binary for example in HTTP, if no `datacontenttype` is set, then an integer should be encoded using its canonical string representation. That's a fair enough interpretation. But then there's also this:

> An application is free to hold the information in any in-memory representation of its choosing, but as the value is transposed into HTTP as defined in this specification, the assumption is that the data attribute value is made available as a sequence of bytes.

So, the HTTP binding (and every other binding that has a binary mode) assumes that the data is available as a sequence of bytes, while the main spec says that every type has a canonical string encoding. And you could say well a String is bytes when encoded, but can you see the confusion there? Why not just make the body bytes?

Every time I have commented in this thread, I have read the spec in detail to try and understand where you're coming from. This wasn't at all obvious to me. Either I'm stupid, or there's a problem in the spec. And that's my point, it's really confusing as it stands now. The body both is specified by the CloudEvents abstract type system and the `datacontenttype` attribute. If they disagree, that's an error, but the spec never says that. And you have to make some fairly big (and in my opinion, unintuitive) leaps to understand how certain things get encoded.

clemensv:
Let's be clear on what's a stake here if we drop `Any`:

```
{
    "id" : "12345678",
    "time": "2019-07-02T17:15:33Z",
    "type" : "com.vasters.hello",
    "specversion": "0.3"
    "data" : {
        "foo" : "bar"
    }
}
```

This is the simplest case. 

I can take this event and render that in JSON or in Avro and in Protobuf (and CBOR when we add it) without declaring `datacontenttype` and the event would be transcoded just fine if the encoding switches. 

I don't want to give up the simple case where `data` is a single value or a simple map of key/pairs and gets encoded inside a structured envelope. If the data type goes to "undefined", `datacontenttype` becomes required and then we necessarily end up with abominations like flowing JSON inside of Protobuf, because if it says "application/json" then that's what the Protobuf wrapper MUST carry. We lose transcoding of the data attribute via the abstract type system. 

I'm also against having the awful limitations of JSON be the only guide here. If you use an encoding with proper support for the most basic types (binary, datetime, integer vs. float), a lot of the confusion vanishes.  CBOR to Avro to Protobuf to AMQP can be super clean with no loss of type fidelity for what's used in common business systems and with the model we have in place, it's actually possible to ignore the existence of JSON altogether in a concrete implementation.

duglin:
> If the data type goes to "undefined", datacontenttype becomes required

Is this true? In your example w/o `datacontenttype` I would think the implied value is json since that's what the rest of the message is in.  Right?

I sort of thought our type system was really there just to provide a bit of an abstract model so that we didn't have to say how each attribute was to be serialized, rather we could just describe how the type is serialized in one spot - and then extensions can piggy-back on it too. More of a spec space-saving thing more than anything else. So, with that in mind, I didn't see much of an issue with `data` being un-typed because we have special serialization rules for it anyway. And in doing that we could talk about maps within `data` if we saw a need.  Am I missing something?

clemensv:
@duglin If `data` is un-typed, the only way to give it shape it is `datacontenttype`. There's no "implied value" outside the type model we define here. If you leave that to the chosen encoding, you instantly lose all transcoding ability. If you start with CBOR using dates and decimals in the data section, you're on your own transcoding that to JSON since there's no constraining rule-set as we have with the abstract model. The example I have above isn't JSON, it's a projection of the abstract model into JSON. 

Without the type model putting constraints on what's legitimate in the data field, we'd be bolting every event to a particular encoding for the `data` field.

duglin:
Still feel like I'm missing something. Since `data` can be lots of different types, if we see:
```
{ "data": "MjAxOS8wNy8wMiAxMjowMDowMAo=" }
```
I'm not sure how we know how to interpret `MjAxOS8wNy8wMiAxMjowMDowMAo=` just because our spec defines `data` as one of our types instead of just "undefined". I still don't know if it's a string vs base64-encoded data - the `datacontentecoding` should help here, I think. But then that doesn't tell me whether the decoded string ("2019/07/02 12:00:00") is a timestamp or just a string. If this were on `ce-time` then I'd know because we define it as a timestamp, but in `data` it could be a string or a timestamp.  Likewise, a decode "2" could be a string or an int.  Only the app knows. So, I'm back to... how does `data` being defined as "undefined" vs "one of our CE types" actually change anything?

clemensv:
@duglin The data content is a string. No further statement can be made about it. 

If you qualify it with `datacontentencoding`="Base64" then it's a binary. If you qualify it with ´datacontenttype`="application/jwt" then it's a JSON Web Token and that can even stand without the ´datacontentencoding` qualification because that's implied.

If we make the type of ´data´ undefined, the field can't be a string, because we then haven't defined for `data` what String means. You would have to declare a content-type and treat the content of data based on that - and, as said, then the data is bolted to the content-type.

You definitely couldn't use a map of key value pairs for the value of ´data´ that can be transformed in an interoperable way across encodings, because there's no translation model. Right now, the encodings all have a mapping to the abstract type model and the spec is using that type model. "Undefined" tosses all of that out for `data´.

evankanderson:
@clemensv Is the problem here that JSON does not have sufficient type information on its own to enable body translation? In particular, representations like CBOR make a distinction between String and Binary data, but JSON does not (and therefore, we end up with cases like `"Mg=="` when receiving a payload bytes("2") which we don't know whether it is JSON or binary).

I wonder whether making JSON structured the special case would help -- i.e. if we consider a CloudEvent to be:

<table>
<tr><td>Attributes</td></tr>
<tr><td>Data</td></tr>
</table>

and then JSON provides a particular way of wrapping the two together into a single JSON object. This might make it easier to talk about separate handling of the data component compared with the attributes. Looking at all the transports that define a binary mode, they all have this model in common, which suggests that it's a pretty common pattern for message transport.

evankanderson:
(Separate thread, because separate discussion)

@clemensv Your mention of transcoding the _data payload_ of the event between encodings like Avro, protobuf and CBOR is interesting -- I haven't seen this spelled out as a goal of CloudEvents before, and I would actually be surprised to discover this in my infrastructure.

Are you suggesting that the `data` attribute might be transcoded e.g. from JSON to CBOR when going into a particular eventing system, and then converted (for example) to Avro if that's the consumer's preference? I did not previously expect that type of behavior to be in scope for this group.

jroper:
I agree with @evankanderson, I had thought these types only existed to try and make the spec a little smarter about existing formats, not to introduce a new format. The spec says:

> CloudEvents is a specification for describing event data in common formats to provide interoperability across services, platforms and systems.

Now, for all the services, platforms and systems I'm aware of, eg SQS, Kinesis, EventHubs, Knative Events, Kafka, MQTT, NATS, HTTP, etc etc etc, users select a payload encoding, eg json, protobuf, avro, something else, and both the producer and consumer use that. None of the above systems, as far as I'm aware, require the use of any particular encoding, or do any transcoding, the encoding is opaque to them. So providing a universal abstract data type, while interesting, is orthogonal to the goal of making these services/platforms/systems interoperate, and so doesn't belong in CloudEvents.

I'm also not sure that anyone would use it. The simple case above might look great in a hello world example, but the ADT is so limited that I would consider it too high risk to use in a real world scenario. Let's say I did have a payload structure that did fit into the ADT. What if in future, as my application domain evolves, I need to add a property that contains an array of strings? The ADT doesn't support arrays, so in order to do that, I need to switch to a different encoding, eg encoding directly in JSON. But that change won't be backwards compatible with existing consumers. And arrays aren't the only thing it doesn't support, floating point numbers and 64 bit integers aren't supported either. So I'd consider it bad practice to use it.

And I think if a spec for a universal abstract data type that can be encoded to major formats were created, that would be a huge undertaking to do it well, to provide a minimum set of types to be useful, to ensure round tripping to/from formats is stable (right now we lose type information for many round trips), to ensure there are no edge cases (eg what happens if a Binary payload is encoded to JSON, then transcoded to HTTP? You get a base64ed HTTP payload, but if you go straight to HTTP, you get the raw binary HTTP payload. To correctly interpret it, a consumer needs to know what transcodings have happened along the way). It's too big and distracting for the CloudEvents spec. The best way to support this would be for a separate spec to create it (since such a universal ADT would have usefulness far beyond CloudEvents), and then it would get its own MIME type, and then CloudEvents could incorporate it just like any other encoding, with special casing in encodings that natively support the universal ADT to have them encode it natively.

What we've got today doesn't even meet the goals you suggested - if the above CloudEvent with a Map payload and no datacontenttype goes through a binary transport binding, it will come out on the other end as a String containing JSON, and then if that gets transcoded to protobuf, you end up with JSON in protobuf. Transcode it to JSON, and you get JSON inside a JSON String in JSON. And the user experience would be terrible, since now the user needs to be prepared for it to either be a map, or a string containing JSON that they'll have to decode - if they don't code to support that, then in future, when the structure of the system is changed, eg the organisation decides to migrate from one cloud to another, and so instead of events just going through one service, they get transcoded a few times to get from one cloud to another, the code will break. And what's the point of using CloudEvents if the moment you interoperate different clouds things break? CloudEvents will be a great spec if it can be robust enough to ensure that no matter what the user tries to do or doesn't do, things will work consistently and the same everywhere.

clemensv:
@jroper @evankanderson 

For simple, self-contained, structured events, transcoding happens if I pick up the event as JSON and route it onwards as CBOR or store it in an Avro container.

evankanderson:
This transcoding covers both the context attributes and the data attribute?

At this point I think I share James' concern that our transport encodings
are not rigorously specified enough to support this, though I'd love to be
proved wrong.

clemensv:
@evankanderson transcoding covers both attribute types. We're obviously transcoding today as we go between the various modes and transports. Every deserialization and serialization of an in-memory representation of an event into a transport frame or a wire event format transcodes.

clemensv:
JFYI - Reference to the original PR that added the type system. The intent was specifically to allow for `data` to carry structured information. https://github.com/cloudevents/spec/pull/152

duglin:
Closing in favor of #470

PR #462, opened 2019-06-19
Modify roadmap with a more clear goal of getting to v1.0

duglin:

### DIFF
--- a/roadmap.md
+++ b/roadmap.md
@@ -47,32 +47,40 @@ _0.2_ - Completed - 2018/12/06
 1. Incorporate learnings and feedback from interop demo to support wider
    adoption.
 1. Draft documentation, developer and/or user guide.
-1. Resolve all known large design issues (excluding security issues)
-1. Resolve all known "proposed required attributes" issues
+1. Resolve all known large design issues (excluding security issues).
+1. Resolve all known "proposed required attributes" issues.
 1. Interoperability demo 2
-   1. Details to be determined
-   1. Showcase demo at conferences - perhaps KubeCon NA 2018
+   1. Details to be determined.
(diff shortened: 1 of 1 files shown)


cneijenhuis:
👍 

Minor input: If I get this PR correctly, the plan for 0.9 is that we ideally change nothing, unless we find an issue during verification/testing. As we try to follow semver, I think `0.9` could also be called `1.0.0-alpha.1` or `1.0.0-RC.1` (see https://semver.org/#spec-item-9 ).

We could still make breaking changes if we have to ("A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version."), but it clearly communicates that we don't plan to do that.

duglin:
I like the idea of calling it 1.0.0-alpha.1 or 1.0.0-RC.1 - I like the signal it sends, as you said.

w.r.t "change nothing"... this is something I was a little uncertain of. During this phase I was thinking we'd continue to work on the 'try-for-v1.0' issues/PRs since they should be non-breaking changes, and those would go into the v1.0 release.  That would not adhere to the "change nothing" statement. WDYT?

The other option is to push those out to a v1.1 - but I'd prefer to not do a version bump so quickly.

Would love some feedback on this....

cneijenhuis:
Maybe we can differentiate between changes that clarify the spec, but don't have an influence on "valid" implementations (unless one particular implementation misunderstood something); and changes that are not backwards-compatible.

With "change nothing", I meant the latter type - i.e. it changes nothing for the code that I wrote.

duglin:
ok with that I took a stab at updating it - see what you think

duglin:
Approved on the 6/28 call

PR #438, opened 2019-06-07
Avro Event Format for CloudEvents

fabiojose:
This PR closes #426.

This is the first version, please comment and improve!

PR #463, opened 2019-06-20
Avro Event Format for CloudEvents

### LINKED ISSUE #426: Avro format enconding
I would like to propose an official Avro encoding schema to be supported by CloudEvents.

I've seen a lot of uses cases, in special here in my current project, with Apache Kafka and the many of them were using Avro as the format. Inspired by [this](https://gist.github.com/rbramley/c00bbd95f3f6ecc0cc1f90e5d6d6c90c) Gist (thanks @rbramley), the following schema could be the first draft:

```json
{
  "namespace":"io.cloudevents",
  "type":"record",
  "name":"CloudEvent",
  "version": "1",
  "doc": "CloudEvents Spec for Avro format encoding (the version is just for the schema itself)",
  "fields":[
    {
      "name":"specversion",
      "type":"string"
    },
    {
      "name":"type",
      "type":"string"
    },
    {
      "name":"source",
      "type":"string"
    },
    {
      "name":"id",
      "type":"string",
    },
    {
      "name":"time",
      "type":[
        "null",
        "int"
      ],
      "logicalType": "date"
    },
    {
      "name":"schemaurl",
      "type":[
        "null",
        "string"
      ]
    },
    {
      "name":"contenttype",
      "type":[
        "null",
        "string"
      ]
    },
    {
      "name":"data",
      "type":[
        "null",
        "string"
      ]
    },
    {
      "name":"extensions",
      "type":[
        "null",
        {
          "type":"map",
          "values":[
            "string",
            "int",
            "long",
            "float",
            "double",
            "boolean",
            "bytes"
          ]
        }
      ]
    },
  ]
}
```

- Official Avro spec available [here](https://avro.apache.org/docs/1.8.2/spec.html)

__Extensions__

To define a top-level extension (like below) is not allowed by Avro spec for `avsc`, because we do not have the feature for dynamic fields like JSON. Because of this, you will see the `extension` attribute to the example above.

```json
...
    "comexampleextension2" : {
        "othervalue": 5
    }
...
```

For that, we may employ Avro spec for `ìdl` that is a higher-level language for authoring Avro schemas.

- Official Avro IDL docs available [here](https://avro.apache.org/docs/1.8.2/idl.html)

__Special Types__

What I mean with _special types_, logical types in JSON Schema, are: `uri-reference`. We do not have this type in Avro.

For that, again, we may employ Avro IDL.

fabiojose:

Closes #426

Accidentally I've closed the original PR, but open this new one with the right changes.


NickLarsenNZ:
Original PR: https://github.com/cloudevents/spec/pull/438

duglin:
@fabiojose can you update the README to point to this doc?

avro-format.md
```diff
@@ -0,0 +1,177 @@
...
+
+Users of Avro MUST use a message whose binary encoding is identical
+to the one described by the [CloudEvent Avro Schema](./spec.avsc):
+
+```json
+{
+  "namespace":"io.cloudevents",
+  "type":"record",
+  "name":"CloudEvent",
+  "version":"0.3",
```
duglin: 0.4-wip

avro-format.md
```diff
@@ -0,0 +1,177 @@
...
+```
+
+## 3 Examples
+
+The following table shows exemplary mappings:
+
+| CloudEvents     | Type      | Exemplary Avro Value                           |
+| --------------- | --------- | ---------------------------------------------- |
+| type            | string    | `"com.example.someevent"`                      |
+| specversion     | string    | `"0.2"`                                        |
```
duglin: 0.4-wip

spec.avsc
```diff
@@ -0,0 +1,53 @@
+{
+  "namespace":"io.cloudevents",
+  "type":"record",
+  "name":"CloudEvent",
+  "version":"0.2",
```
duglin: 0.4-wip

fabiojose: Done!

avro-format.md
fabiojose: Done!

avro-format.md
fabiojose: Done!

README.md
```diff
@@ -38,6 +38,7 @@ The following documents are available:
 | NATS Transport Binding        | [v0.3](https://github.com/cloudevents/spec/blob/v0.3/nats-transport-binding.md) | [master](https://github.com/cloudevents/spec/blob/master/nats-transport-binding.md) |
 | Protocol Buffers Event Format | [v0.3](https://github.com/cloudevents/spec/blob/v0.3/protobuf-format.md)        | [master](https://github.com/cloudevents/spec/blob/master/protobuf-format.md)        |
 | Web hook                      |      [v0.3](https://github.com/cloudevents/spec/blob/v0.3/http-webhook.md)      |      [master](https://github.com/cloudevents/spec/blob/master/http-webhook.md)      |
+| AVRO Event Format             | - | [master](./avro-format.md) |
```
duglin: can you move this so the list if alphabetized?

duglin:
We tentatively approved this on today's call, but people asked for one more week to make sure. Please look it over so we can resolve it next week

README.md
duglin: @fabiojose can you also add this to the Makefile so it gets verified?

duglin:
@JemDay I believe you asked for one more week,last week.  You ok with this one?

README.md
fabiojose: @duglin  done!

avro-format.md
```diff
@@ -0,0 +1,177 @@
...
+map-of-maps in a recursive way, i.e., every `map` declaration REQUIRE the
+`values` with primitives types or a declared type.
+
+### 2.3 Optional Attributes
+
+CloudEvents Spec defines optional attributes. The Avro format defines that
+every optional field MUST use the `null` type and the actual type through
+the [union][avro-unions].
+
+Declaring a type which may be either a `null` or `string`, i.e., an optional
```
duglin: s/may/might/

avro-format.md
```diff
@@ -0,0 +1,177 @@
...
+attributes they define.
+
+### 2.2 Mapping Any-typed Attributes
+
+`Any`-typed CloudEvents values can either hold a `String`, or a `Binary` value,
+or a `Map`, or any of all other types. Avro type system SHOULD NOT satisfy the
+map-of-maps in a recursive way, i.e., every `map` declaration REQUIRE the
+`values` with primitives types or a declared type.
+
+### 2.3 Optional Attributes
```
duglin: s/Optional/OPTIONAL/

avro-format.md
```diff
@@ -0,0 +1,177 @@
...
+
+CloudEvents Spec defines optional attributes. The Avro format defines that
+every optional field MUST use the `null` type and the actual type through
```
duglin: s/every optional field/these fields/

avro-format.md
duglin: is the above a complete sentence? We need to do something about "optional", the RFC checker isn't happy. I would suggest we either remove the "i.e" part, if that still makes sense

fabiojose: Did not make sense! Removed!

clemensv:
I don't think the definition of a CloudEvent as a "record" satisfies our need for having extensibility, because record requires all fields to be known for encoding.

JemDay:
I think the prior comment from @clemensv may be touching slightly on what i'd been thinking about given proposal was very structured (which in a way i like) vs the approach taken with proto which way much more minimalistic. The proto schema is more aligned with the example @clemensv cited.

As it stands with the current schema proposal a mechanism to carry 'extensions' is required.

fabiojose:
Very clever @clemensv 

I tested it and got the following schema:

```json
{
  "namespace":"io.cloudevents",
  "type":"record",
  "name":"CloudEvent",
  "version":"0.4-wip",
  "doc":"Avro Event Format fo CloudEvents",
  "fields":[
    {
      "name":"metadata",
      "type":{
        "type":"map",
        "values":[
          "null",
          "int",
          "float",
          "string",
          "boolean",
          "long",
          "CloudEvent"
        ]
      }
    }
  ]
}
```

It looks more flexible than the first version.

evankanderson:
We don't have `float` or `boolean` types defined at the moment.

We may also need to support `binary` types.

fabiojose:
Good points @evankanderson!

Done!

duglin:
@clemensv @evankanderson does this look ok now?

JemDay:
Disclaimer : Not an avro expert..

For consistency reasons with protobuf...

- Could we change the name of the definition file to _cloudevent.avsc_
- Does avro allow you to follow the same model as used in cloudevent.proto

Otherwise +1

fabiojose:
Hi @JemDay 

"Could we change the name of the definition file to cloudevent.avsc": Sure!
- I had used the `spec.avsc` to follow the same naming convention used by `spec.json`.

"Does avro allow you to follow the same model as used in cloudevent.proto"
- unfortunately, not

JemDay:
Thanks ... thinking a bit more, maybe aligning around _spec.<format>_ is better

Still +1

fabiojose:
Good point! I will change the file name.

fabiojose:
I think that `cloudevent.avsc` is better.

duglin:
at this point I think consistency with the others is more important tho

duglin:
@fabiojose is this one ready to go? it was approved to but I don't want to merge if you still have a pending change

fabiojose:
Done, no pending changes!

The file will be `spec.avsc`, just like before.

catkins-miso:
It seems there would have been a performance impact removing the first-class context attributes in favor of the attribute map approach. I understand the point of adding the attribute map for extensions, but I think the schema-on-read approach could have been isolated without impacting other use cases. In particular, if I'm listening to a firehose topic (events of many types)--or implementing an intermediary--I need to quickly evaluate the type context attribute. Not to mention Avro has a pretty mechanics for [forward and backwards compatibility](https://avro.apache.org/docs/current/spec.html#Schema+Resolution) which I _think_ could have addressed concerns about introducing new context attributes.

evankanderson:
I think the attributes may have been simplified in the interim for v1 (when this was written, they were a json-style `string` -> `value or object` mapping, rather than a `string` -> `simple type` mapping.

Unfortunately, I think the v1.0 spec ship has sailed; I'm not sure what to suggest here.

catkins-miso:
Thanks for the response @evankanderson. I did a little preliminary benchmarking (Apache Avro 1.9 dotnetcore), and I am seeing no meaningful difference in the binary serde performance from the attributes and having distinct fields.

PR #465, opened 2019-06-20
Allow for a broader schemaurl

### LINKED ISSUE #311: A schema (constraints) for context and data attributes identified by schemaURL
Many of the context attributes (e.g., eventType, source, eventID, contentType) are not tightly constrained to maintain flexibility in specific implementations.  However, as noted in the intro to the spec "event publishers (producers) tend to describe events differently. The lack of a common way of describing events means developers are constantly re-learning how to receive (consume) events."    Loosely constrained attributes will not effectively address this key problem.

The spec currently describes the schemaURL attribute as "a link to the schema that the 'data' attribute adheres to".    Suggest that this be broadened to "a link to the schema defining further constraints to context and data attributes"  Thus, a single schema can ensure semantic interop of context attributes between event producers and consumers within a specific domain or ecosystem - not just data.

Further, suggest that the schemaURL and cloudEventsVersion attributes are "schema attributes" that describe the "context and data attributes" and should not be included as "context attributes" describing the occurrence (data).

duglin:
closes #311

/cc @dmigliori 

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -356,6 +356,10 @@ on the definition of OPTIONAL.
 
 - Type: `URI-reference`
 - Description: A link to the schema that the `data` attribute adheres to.
+  However, the schema MAY also include constraints for any other CloudEvents
+  attributes - as long as those constraints do not conflict with this
+  specification.
+
   Incompatible changes to the schema SHOULD be reflected by a different URL. See
   [Versioning of Attributes in the Primer](primer.md#versioning-of-attributes)
   for more information.


clemensv:
How is that going to work in practice? If I have a schema for structured data in 'data', that's a clearly scoped matter. If that schema were to constrain an attribute/property "id", how would that now be interpreted? Does that apply to the CloudEvent attribute, does that apply to a top-level property inside data? What complicates this further is that schemas are commonly bound to a particular format. Net is that I doubt the schemaUrl attribute to be the right place to also put schema constraints for context attributes.

I understand why some people might want to put a formal pattern or range constraint on particular CloudEvent attributes, either in the core or in extensions, but I really consider that a solution concern. I believe such a requirement can be very well satisfied at the application level and that it doesn't impact cross-system interop. 

An implementation of an event broker might quite well have authorization or content rules that cause particular events to be rejected, but those would be configured in the broker based on matching event types, sources, and/or subjects and not carried inside the event - because the event can't just bring its own door opener along.

The reference to the schema exists primarily to allow the receiving party to even be able to decode the payload for when the payload encoding is dependent on external schema; it's not there for the event to carry its own constraints.

duglin:
@dmigliori any thoughts on @clemensv's comment ^^^

dmigliori:
@clemensv thanks for your comments.    Putting the use of the SchemaURL attribute aside, I'd like your thoughts on how to address this issue, which dovetails @duglin's PR #441 on how to use CE as an umbrella event format to transport events produced in another format (e.g., GitHub, GS1 EPCIS).  If these other events are "adapted" to enable transport within a CE event (including extensions) and aggregated together via middleware to deliver to event consumers, then the consumers need some indication of what adapted format each of these CE events conforms to (for both data validation and processing).  This extends beyond the data attribute and is not specific to a single event producer.  See example below of 3 events using different "sub" schemas (github, microsoft, kubecon demo):

{
    "specversion" : "0.4-wip",
    "subject" : "123",
    "id" : "A234-1234-1234",
    "comexampleextension2" : {
        "othervalue": 5
    },
    "datacontenttype" : "text/xml",
}
{
    "cloudEventsVersion" : "0.1",
    "eventType" : "Microsoft.Storage.BlobCreated",
    "eventTypeVersion" : "",
    "source" : "/subscriptions/{subscription-id}/resourceGroups/{resource-group}/providers/Microsoft.Storage/storageAccounts/{storage-account}#blobServices/default/containers/{storage-container}/blobs/{new-file}",
    "eventID" : "173d9985-401e-0075-2497-de268c06ff25",
    "eventTime" : "2018-04-28T02:18:47.1281675Z",
    "data" : {
      "api": "PutBlockList",
      "clientRequestId": "6d79dbfb-0e37-4fc4-981f-442c9ca65760",
      "requestId": "831e1650-001e-001b-66ab-eeb76e000000",
      "eTag": "0x8D4BCC2E4835CD0",
      "contentType": "application/octet-stream",
      "contentLength": 524288,
      "blobType": "BlockBlob",
      "url": "https://oc2d2817345i60006.blob.core.windows.net/oc2d2817345i200097container/oc2d2817345i20002296blob",
      "sequencer": "00000000000004420000000000028963",
      "storageDiagnostics": {
        "batchId": "b68529f3-68cd-4744-baa4-3c0498ec19f0"
      }
    }
}
{
    "specversion": "0.3",
    "type": "TransferAction.ActionStatus.CompletedActionStatus",
    "source": "Carrier.IBMC",
    "subject": "e6583953-0728-4761-9966-bfb12ed3cd35",
    "id": "f4e35896-39db-45ff-a54c-11eb4a8f4160",
    "time": "2019-06-23T22:51:45Z",
    "contenttype": "application/json",
    "data": {
        "actionStatus": "CompletedActionStatus",
        "fromLocation": "Supplier.ORCLS",
        "toLocation": "Retailer.IBMR",
        "offer": "small"
    },
    "cause": "67d72269-a304-4591-9727-8e56720eddfc"
}

I think the "adaptors" being developed for PR #441 will be much more effective being constructed as schemas tied to URLs rather than text within the CE primer.

I plan to create an EPCIS to CE adaptor and am working with GS1's committee that is creating a JSON version of EPCIS (to get this format as close to CE format as possible to minimize the adaption requirements).  Below is a link to the draft EPCIS JSON schema:

https://github.com/dannyhaak/epcis2-json-schema/blob/master/json_schema.json
Note its schema URL is https://id.gs1.org/epcis-schema.json

I proposed using CE's schemaURL attribute for this to unnecessarily propose a new attribute.  As far as I can tell, the CE spec doesn't detail any examples on how the schemaURL attribute is implemented or if the URL content needs to comply to a certain interoperable format (e.g., json-schema).  Thus, I didn't see a reason why the URL content couldn't include schema for multiple CE attributes.

clemensv:
@dmigliori Here's how I might solve this:

Introduce an extension attribute "eventSchemaUrl" that refers to a schema that covers the entirety of the event and not just the data. The base schema for such schemas ought to reflect the chosen CloudEvents specVersion, and the schemas further restrict the base schema mandating particular formats for the attributes and/or specific additional extensions that must be used.

The reason I would do that as an extension and not as a core spec element/rule is twofold:
* Since this requires base schemas that also cover the normative parts of CloudEvents, we'd likely get into a discussion about which schema formats/types are preferable and that's potentially ultra-contentious.
* While I recognize validation as being important for a range of use-cases, I don't think those cases make up a large portion of the usage scenarios for CloudEvents.

I'd definitely be in favor of something like this as one of the official set of extensions, even through I'm unsure whether I would want to bake such enforcement support into an event router.

duglin:
I have mixed feeling about this. I think I understand the usecase at a conceptual level, but I wonder about the layers here. By that I mean, how many different fields will people need to check to know that this event is of interest to them? We already have `type` and `source` as key ones. But with this extension they would need to check that first to make sure they understand how `type` was constructed. I just wonder, in practice, how often people will find that they get the same `type` and `source` from multiple input streams into the same environment such that they need another field to differentiate things?

@dmigliori Maybe it'll help me if you could elaborate on what kind of validation/processing you're thinking of that would change based on the value of this extension and in what scenarios things would fall apart w/o this field?

Ultimately, an extension does no harm and allows us to test it out... but I'd like to better understand the problem space first just to make sure that we're not ignoring a fundamental issue in the core spec.

dmigliori:
@duglin @clemensv

Since the potential problem doesn't seem obvious at this point, maybe I'm missing something.  The way I would solve the problem wouldn't involve any change to the CE schema - thus the PR was just a description change.  It would just be handled through a URI value in the existing schemaURL attribute that linked to a structured/interoperable schema file (e.g., json-schema) that could apply to key-value pairs within the CE "data" attribute, as well as define what CE extensions (e.g. "cause") were required.   Then a single CE content validator could validate all CE attribute values for multiple sub-schemas (e.g. GS1 EPCIS).

Since I'm not advocating a CE schema change, let's put this aside until broader CE adoption makes it an obvious issue - or not.

dmigliori:
It would still be great to see an example in  the  primer of how the schemaURL attribute was intended to be used - including sample content of the file accessible via the schemaURL value.  It's not clear how adding additional content to that file to support attributes other than "data" would cause disruption.

duglin:
per @dmigliori's comment, closing

Issue #414, opened 2019-04-05
What happens when... for JSON Batch Format.


n3wscott:
I have few implementation questions around json batch format. 

1. Is it expected that the receiver is delivered the array? Or can an implementation not expose batching and deliver the events one at a time?
1. What happens when that receiver has an issue with one of _n_ of these events? The response code should be? 
1. If a valid response is `[]`, is that also a valid request? IE, should an SDK prevent you from sending`[]`?

cneijenhuis:
1. I'm not sure I'm getting the question correctly. Batching is optional and must only be used if both producer and consumer support it.
If you have a setup with a middleware, the middleware can add or remove the batching, or change the batch size.

Examples:
Publisher pushes single events every second (no batching). Middleware is a queue and aggregates events. Consumer polls the queue (with batching) every 5 seconds, and the Middleware responds with a batch of 5 events.
Publisher publishes a batch of 5 events. Middleware takes this batch, transforms each event from JSON into Protobuf and sends them one-by-one via grpc.

In other words, the publisher MUST expect that the batch will be broken up. It is just a temporary optimization on one hop, and probably will change on the next hop.

Does that help?

2. Probably `400`, but the webhook spec is not very prescriptive in terms of error responses. Do you think it makes sense to specify a response format, where the status is declared per event?

3. Clemens says "no": https://github.com/cloudevents/spec/pull/370#discussion_r251797790 😉

duglin:
@clemensv what was your intention w.r.t. batch responses?

IMO w.r.t. the response if there's an issue with one of the events... I think the answer might be the same as what would happen if the processing of a single event failed 1/2 thru it's business logic. Do we roll it all back? Do you try to respond with some partial success error code?  I think the same applies to batching and the options are:
- 200 : successfully processed the incoming event or events
- 202 : successfully accepted the incoming event or events
- 4xx/5xx : was not able to process the incoming event or events - how far it got is unknown. Whether it rolled back any is unknown. At least from our spec perspective, since we don't deal with processing models.

Net: I think we could add text to the primer stating this.

JemDay:
FYI .. I've been really slow adding any language to the webhook spec for this .. sorry.

cneijenhuis:
I hereby volunteer to open a small PR, but I'm fine with Scott or Jem driving this as well 😉

duglin:
@cneijenhuis thanks a ton!  I know @n3wscott is super swamped these days would appreciate any help.

[...]

PR #466, opened 2019-06-22
Add section to primer about error handling

### LINKED ISSUE #434: Should we document unhappy paths?
While chatting at KubeCon+CloudNativeCon with @noelo they asked a very good question: how does CloudEvents handle the unhappy paths? What happens if a 500 is returned by the HTTP transport? What about a 429? What about all other transports? Apparently AMQP does have some ACK/ NACK included.

Do we want to document this? Would a primer per-transport make sense for this?

duglin:
Closes #434

@n3wscott @cneijenhuis perhaps we could add some more text to this section to resolve https://github.com/cloudevents/spec/issues/414 ?

### DIFF
--- a/primer.md
+++ b/primer.md
@@ -20,6 +20,7 @@ This document is a working draft.
 - [Architecture](#architecture)
 - [Versioning of Attributes](#versioning-of-attributes)
 - [CloudEvent Attributes Extensions](#cloudevent-attribute-extensions)
+- [Creating CloudEvents](#creating-cloudevents)
 - [Qualifying Protocols and Encodings](#qualifying-protocols-and-encodings)
 - [Proprietary Protocols and Encodings](#proprietary-protocols-and-encodings)
 - [Prior Art](#prior-art)
@@ -202,6 +203,14 @@ application protocol. The [HTTP Webhook](http-webhook.md) specification is
 not specific to CloudEvents and can be used to post any kind of one-way event
 and notifications to a conformant HTTP endpoint. However, the lack of such a
 specification elsewhere makes it necessary for CloudEvents to define it.
+
+### Transport Error Handling
+
+The CloudEvents specification, for the most part, does not dictate a processing
+model associated with the creation or processing of CloudEvents. As such, if
+there are errors during the processing of a CloudEvent, the software
+encountering the error is encouraged to use the normal transport-level
+error reporting to report them.
 
 ## Versioning of Attributes
 


primer.md
```diff
@@ -187,6 +188,13 @@ not specific to CloudEvents and can be used to post any kind of one-way event
 and notifications to a conformant HTTP endpoint. However, the lack of such a
 specification elsewhere makes it necessary for CloudEvents to define it.
 
+### Transport Error Handling
+
+The CloudEvents specification, for the most part, does not dictact a processing
```
cneijenhuis:
```suggestion
The CloudEvents specification, for the most part, does not dictate a processing
```

duglin: fixed!

primer.md
```diff
@@ -187,6 +188,13 @@ not specific to CloudEvents and can be used to post any kind of one-way event
 and notifications to a conformant HTTP endpoint. However, the lack of such a
 specification elsewhere makes it necessary for CloudEvents to define it.
 
+### Transport Error Handling
+
+The CloudEvents specification, for the most part, does not dictate a processing
+model associated with the creation or processing of CloudEvents. As such, if
+there are errors during the processing of a CloudEvent it is assumed that the
+normal transport-level error reporting mechanisms will be used.
```
n3wscott: I don't think this is good enough advice. Should the producer retry? Should it propagate the error downstream? how does it propagate errors that cross transports??

duglin: IMO those are semantic decisions that are out of scope of the spec. To me, CE is about adding metadata to an event to help with routing/filtering/etc.  In the most basic form, binary, we add a few bits of metadata (e.g. 4 http headers) and that's it. How retries or errors are handled do not change just because we said "add 4 new ce-xxx headers".  The questions you're asking were there before CE and they'll be there after CE because we aren't trying to solve those problems.  

Our git repo README says:
> CloudEvents is a specification for describing event data in common formats to provide interoperability across services, platforms and systems.

"describing event data" - not "transporting events". Or put another way, we're not defining an eventing/messaging transport framework spec. Follow-on specs/projects could choose to work on those though.  I'm not dismissing there might be issues with what you're concerned about, I'm just saying it's out of scope for what we agreed to focus on.

duglin: And this PR tries to make it clear that yes we are "punting"  :-)

timbray: Could we say something to be a bit more proactive about encouraging people to report problems?  In typical eventing infrastructure there won't be a direct channel to feedback to a sender of broken messages, so good citizenship of the ecosystem requires making noise when receiving broken data. So I'd change the language to say something like "if there are errors during the processing of a CloudEvent, the software encountering the error is encouraged to use the normal transport-level error reporting to report them."

ColinSullivan1:
LGTM.

duglin:
Approved on the 7/25 call - w/ @timbray's suggested edit

Issue #460, opened 2019-06-18
Revise Map encoding for attributes in AMQP


clemensv:
Looking at [the AMQP bindings](https://github.com/cloudevents/spec/blob/v0.3/amqp-transport-binding.md#3132-amqp-application-property-values), it's not clear that ["Map" values](https://github.com/cloudevents/spec/blob/v0.3/spec.md#context-attributes) can be represented as requested in AMQP, because the [application-properties](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-application-properties) section does not allow values of the [map](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-types-v1.0-os.html#type-map) type.

> The keys of this map are restricted to be of type string (which excludes the possibility of a null key) and the values are restricted to be of simple types only, that is, excluding map, list, and array types.

_Originally posted by @evankanderson in https://github.com/cloudevents/spec/pull/456#issuecomment-502879572_

That constraint is one that the AMQP TC now regrets, but it's difficult to change now. The AMQP-native way to do this would be to store the Map as a binary-encoded AMQP map, which is a little weird, but would be close to the wire model you'd have if maps were allowed. The more straightforward way for apps would be to store it as a JSON-encoded string.

evankanderson:
I think either format would be fine, but unless the AMQP TC wants to revise the standard, we should probably suggest one of the workarounds -- I'm not sure which to recommend, but I suspect that the binary-encoded AMQP map would need some sort of indication for bridging into other formats so that it's clear that the value is an encoded Map and not an arbitrary Binary string.

duglin:
@clemensv do you want to work on a PR for this one?

clemensv:
@duglin Yes, I will work on this.

clemensv:
@duglin @evankanderson  

As the linked thread shows, @rgodfrey reminded me that the key reason for the type limitation is that message/event selectors and filters generally don't allow for traversing complex structures, also because that can become rather expensive and drives complexity (suddenly people will want expressions that search inside a collection). 

It would be possible to create a capability override in AMQP to allow for complex types (lists and maps) in application-properties even if the core spec disallows it, but the capability override might not be consistently supported on a multi-hop route. In reality, we're not sure whether implementations have a hard enforcement of the rule that the values must not be complex, but filters will certainly not be able to do anything with that content.

[...]

PR #467, opened 2019-06-24
Remove Map from allowed Context Attribute types

evankanderson:
Per discussion in #460 and https://github.com/cloudevents/spec/pull/456, it's not clear that Map and Any types in Context Attributes are worth the trouble:

(me)
> Map values are one of the more complex areas of the spec, and the only one
> which does not seem to be used in the core or extensions. Furthermore, it
> sounds like several other ecosystems (HTTP and AMQP) don't support them,
> and the AMQP decision was guided by careful consideration.

This is not quite correct, because Map and Any are used by the `data` attribute in core (and hence this PR does not remove those types entirely)

It's not clear that context attributes needed for routing or other middleware decisions benefit from this flexibility (see also [this thread](https://lists.oasis-open.org/archives/amqp/201906/msg00014.html)), and the possibility of Any types as the value in a Map adds a **lot** of complexity to the encoding.

### DIFF
--- a/http-transport-binding.md
+++ b/http-transport-binding.md
@@ -100,12 +100,13 @@ identically to [HTTP over TLS]([RFC2818][RFC2818]).
 
 ## 2. Use of CloudEvents Attributes
 
-This specification does not further define any of the [CloudEvents][ce] event
-attributes.
+This specification does not further define any of the core [CloudEvents][ce]
+event attributes.
 
 Two of the event attributes, `datacontenttype` and `data` are handled specially
 and mapped onto HTTP constructs, all other attributes are transferred as
-metadata without further interpretation.
+metadata without further interpretation (except that extensions may define
+additional header mappings).
 
 This mapping is intentionally robust against changes, including the addition and
 removal of event attributes, and also accommodates vendor extensions to the
@@ -202,7 +203,8 @@ specification, header names are case-insensitive.
 ##### 3.1.3.2 HTTP Header Values
 
 The value for each HTTP header is constructed from the respective attribute
-type's canonical string representation.
+type's canonical string representation in the [JSON][json-format] string
+encoding.
 
 Some CloudEvents metadata attributes can contain arbitrary UTF-8 string content,
 and per [RFC7230 Section 3][rfc7230-section-3], HTTP headers MUST only use
--- a/http-transport-binding.md
+++ b/http-transport-binding.md
@@ -196,11 +196,6 @@ Examples:
     * `id` maps to `ce-id`
     * `specversion` maps to `ce-specversion`
 
-`Map`-typed CloudEvents attributes MUST be flattened into a set of HTTP headers,
-where by the name of each header carries the prefix "ce-", an infix reflecting
-the map attribute followed by a dash ("-"), and the name of the map entry key,
-e.g. "ce-attrib-key".
-
 Note: per the [HTTP](https://tools.ietf.org/html/rfc7230#section-3.2)
 specification, header names are case-insensitive.
 
--- a/http-transport-binding.md
+++ b/http-transport-binding.md
@@ -105,7 +105,7 @@ event attributes.
 
 Two of the event attributes, `datacontenttype` and `data` are handled specially
 and mapped onto HTTP constructs, all other attributes are transferred as
-metadata without further interpretation (except that extensions may define
+metadata without further interpretation (except that extensions MAY define
 additional header mappings).
 
 This mapping is intentionally robust against changes, including the addition and
@@ -203,8 +203,7 @@ specification, header names are case-insensitive.
 ##### 3.1.3.2 HTTP Header Values
 
 The value for each HTTP header is constructed from the respective attribute
-type's canonical string representation in the [JSON][json-format] string
-encoding.
+type's [canonical string representation][ce-types].
 
 Some CloudEvents metadata attributes can contain arbitrary UTF-8 string content,
 and per [RFC7230 Section 3][rfc7230-section-3], HTTP headers MUST only use
@@ -461,6 +460,7 @@ Content-Length: nnnn
 - [RFC7540][rfc7540] Hypertext Transfer Protocol Version 2 (HTTP/2)
 
 [ce]: ./spec.md
+[ce-types]: ./spec.md#type-system
 [json-format]: ./json-format.md
 [json-batch-format]: ./json-format.md#4-json-batch-format
 [content-type]: https://tools.ietf.org/html/rfc7231#section-3.1.1.5
--- a/mqtt-transport-binding.md
+++ b/mqtt-transport-binding.md
@@ -176,7 +176,8 @@ in the MQTT PUBLISH message.
 ##### 3.1.3.2 User Property Values
 
 The value for each MQTT PUBLISH User Property MUST be constructed from the
-respective CloudEvents attribute type's canonical string representation.
+respective CloudEvents attribute type's canonical string representation in the
+[JSON][json-format] string encoding.
 
 #### 3.1.4 Examples
 
--- a/mqtt-transport-binding.md
+++ b/mqtt-transport-binding.md
@@ -176,8 +176,8 @@ in the MQTT PUBLISH message.
 ##### 3.1.3.2 User Property Values
 
 The value for each MQTT PUBLISH User Property MUST be constructed from the
-respective CloudEvents attribute type's canonical string representation in the
-[JSON][json-format] string encoding.
+respective CloudEvents attribute type's [canonical string
+representation][ce-types].
 
 #### 3.1.4 Examples
 
@@ -312,6 +312,7 @@ Topic Name: mytopic
   Format
 
 [ce]: ./spec.md
+[ce-types]: ./spec.md#type-system
 [json-format]: ./json-format.md
 [oasis-mqtt-3.1.1]: http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/mqtt-v3.1.1.html
 [oasis-mqtt-5]: http://docs.oasis-open.org/mqtt/mqtt/v5.0/mqtt-v5.0.html
--- a/spec.md
+++ b/spec.md
@@ -197,8 +197,8 @@ as a string.
   represent the `Any` type.
 - `Any-context` - A variant type which can take the shapes allowed in a context
   attribute, namely `Integer`, `String`, `Binary`, `URI-reference` or
-  `Timestamp`. Unknown extensions in the context attributes should be treated as
-  the `Any-context` type; transport mappings should explicitly document how to
+  `Timestamp`. Unknown extensions in the context attributes SHOULD be treated as
+  the `Any-context` type; transport mappings SHOULD explicitly document how to
   represent the `Any-context` type.
 
 ## Context Attributes
@@ -216,7 +216,7 @@ and the event data will be materialized. For example, in the case of a JSON
 serialization, the context attributes and the event data might both appear
 within the same JSON object.
 
-Context attributes must be one of the following types:
+Context attributes MUST be one of the following types:
 
 - `Integer`
 - `String`
@@ -224,7 +224,7 @@ Context attributes must be one of the following types:
 - `URI-reference`
 - `Timestamp`
 
-In particular, `Any` and `Map` values are not allowed in context attributes,
+In particular, `Any` and `Map` values MUST NOT be used in context attributes,
 including in extension attributes.
 
 ### REQUIRED Attributes
--- a/spec.md
+++ b/spec.md
@@ -211,8 +211,8 @@ and the event data will be materialized. For example, in the case of a JSON
 serialization, the context attributes and the event data might both appear
 within the same JSON object.
 
-`Any` and `Map` values MUST NOT be used in context attributes, including in
-extension attributes.
+Values of type `Map`, either directly or via the `Any` type, MUST NOT be used
+in context attributes, including in extension attributes.
 
 ### REQUIRED Attributes
 
--- a/spec.md
+++ b/spec.md
@@ -95,7 +95,7 @@ routing can be performed based on information contained in the event, but an
 event will not identify a specific routing destination. Events will contain two
 types of information: the [Data](#data) representing the Occurrence and
 [Context](#context) metadata providing contextual information about the
-Occurrence.
+Occurrence. A single occurrence MAY result in more than one event.
 
 #### Producer
 
@@ -287,7 +287,8 @@ The following attributes are REQUIRED to be present in all CloudEvents:
 #### type
 
 - Type: `String`
-- Description: Type of occurrence which has happened. Often this attribute is
+- Description: This attribute contains a value describing the type of event
+  related to the originating occurrence.  Often this attribute is
   used for routing, observability, policy enforcement, etc. The format of this
   is producer defined and might include information such as the version of the
   `type` - see
--- a/spec.md
+++ b/spec.md
@@ -182,8 +182,7 @@ as a string.
 - `Binary` - Sequence of bytes.
   - String encoding: Base64 encoding per
     [RFC4648](https://tools.ietf.org/html/rfc4648).
-- `Map` - `String`-indexed dictionary of `Any`-typed values. Only valid in the
-  `data` attribute.
+- `Map` - `String`-indexed dictionary of `Any`-typed values.
   - String encoding: JSON Object per
     [RFC 7159, Section 4](https://tools.ietf.org/html/rfc7159#section-4)
 - `URI-reference` - Uniform resource identifier reference.
@@ -195,11 +194,6 @@ as a string.
   `String`, `Binary`, `Map`, `URI-reference` or `Timestamp`. The type system is
   intentionally abstract, and therefore it is left to implementations how to
   represent the `Any` type.
-- `Any-context` - A variant type which can take the shapes allowed in a context
-  attribute, namely `Integer`, `String`, `Binary`, `URI-reference` or
-  `Timestamp`. Unknown extensions in the context attributes SHOULD be treated as
-  the `Any-context` type; transport mappings SHOULD explicitly document how to
-  represent the `Any-context` type.
 
 ## Context Attributes
 
@@ -216,16 +210,8 @@ and the event data will be materialized. For example, in the case of a JSON
 serialization, the context attributes and the event data might both appear
 within the same JSON object.
 
-Context attributes MUST be one of the following types:
-
-- `Integer`
-- `String`
-- `Binary`
-- `URI-reference`
-- `Timestamp`
-
-In particular, `Any` and `Map` values MUST NOT be used in context attributes,
-including in extension attributes.
+`Any` and `Map` values MUST NOT be used in context attributes, including in
+extension attributes.
 
 ### REQUIRED Attributes
 
--- a/spec.md
+++ b/spec.md
@@ -27,7 +27,7 @@ differently.
 
 The lack of a common way of describing events means developers are constantly
 re-learning how to consume events. This also limits the potential for libraries,
-tooling and infrastructure to aide the delivery of event data across
+tooling and infrastructure to aid the delivery of event data across
 environments, like SDKs, event routers or tracing systems. The portability and
 productivity that can be achieved from event data is hindered overall.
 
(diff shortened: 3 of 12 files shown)


http-transport-binding.md
```diff
@@ -100,12 +100,13 @@ identically to [HTTP over TLS]([RFC2818][RFC2818]).
...
 
-This specification does not further define any of the [CloudEvents][ce] event
-attributes.
+This specification does not further define any of the core [CloudEvents][ce]
+event attributes.
 
 Two of the event attributes, `datacontenttype` and `data` are handled specially
 and mapped onto HTTP constructs, all other attributes are transferred as
-metadata without further interpretation.
+metadata without further interpretation (except that extensions may define
```
duglin: s/may/MAY/

jroper:
Related to this is #457, where `data` is redefined to have no specified type, but instead is required to have a `Binary` encoding. If that change gets merged, both `Any` and `Map` as defined in this PR can be completely removed, since `data` doesn't use them. This means this PR could rename `Any-context` back to `Any`.

spec.md
```diff
@@ -194,6 +195,11 @@ as a string.
   `String`, `Binary`, `Map`, `URI-reference` or `Timestamp`. The type system is
   intentionally abstract, and therefore it is left to implementations how to
   represent the `Any` type.
+- `Any-context` - A variant type which can take the shapes allowed in a context
```
duglin: I'm not sure I understand the purpose or value of this. If it's really only for extensions, and we treat unknown extensions as strings, why do we need this?

duglin:
We discussed this on today's call. I believe there was general agreement that there's an issue to solve, but it wasn't clear how to proceed. There were 4 options mentioned:
1 - do nothing, leave it as-is
2 - restrict Maps to just one level (so no nesting of Maps)
3 - remove Maps as a type
4 - adopt this PR

My interpretation of what was said on the call : there seemed to be a lot of interest in dropping Maps entirely, but some people did worry that not allowing for structured extensions would limit things too much or end up with people inventing their own way to represent maps via encoding them as strings or other types. And then we might have an interop problem around decoding those custom things since there is no single way to do it.  Please correct me if I got this wrong.

Please chime in on this one as it's kind of a biggie to solve prior to v1.0

spec.md
```diff
@@ -210,6 +216,17 @@ and the event data will be materialized. For example, in the case of a JSON
 serialization, the context attributes and the event data might both appear
 within the same JSON object.
 
+Context attributes MUST be one of the following types:
```
clemensv: I believe the half sentence in line 227 is sufficient for this PR.

`Map` values MUST NOT be used in context attributes, including in extension attributes.

spec.md
```diff
@@ -182,7 +182,8 @@ as a string.
 - `Binary` - Sequence of bytes.
   - String encoding: Base64 encoding per
     [RFC4648](https://tools.ietf.org/html/rfc4648).
-- `Map` - `String`-indexed dictionary of `Any`-typed values.
+- `Map` - `String`-indexed dictionary of `Any`-typed values. Only valid in the
```
clemensv: It's sufficient to say this only in one place normatively.

spec.md
clemensv: I would leave `Any` untouched. Stating that the Map type can't be used for context attributes is sufficient. `Any` is not a submarine through which you can smuggle `Map` undetected.

mqtt-transport-binding.md
```diff
@@ -176,7 +176,8 @@ in the MQTT PUBLISH message.
 ##### 3.1.3.2 User Property Values
 
 The value for each MQTT PUBLISH User Property MUST be constructed from the
-respective CloudEvents attribute type's canonical string representation.
+respective CloudEvents attribute type's canonical string representation in the
+[JSON][json-format] string encoding.
```
clemensv: We define the canonical representations here. We've decoupled them from JSON now.

evankanderson: Thanks, updated the link to the type system

spec.md
evankanderson: I'm concerned about confusion between line 150 and later, but I'll remove if this seems clear to everyone.

> The following abstract data types are available for use in attributes.

spec.md
evankanderson: Done. I'm slightly uncomfortable with the formulation, but the result is clear if you read the whole spec:

> The value of a context attribute is Integer, String, Binary, Map, URI-reference, Timestamp, or Any, except that it can't be Any or Map.

If we accept some form of #457, we can completely remove Map, and the sentence simplifies to:

> The value of a context attribute is Integer, String, Binary, Map, URI-reference, Timestamp, or Any.

spec.md
```diff
@@ -211,6 +211,9 @@ and the event data will be materialized. For example, in the case of a JSON
 serialization, the context attributes and the event data might both appear
 within the same JSON object.
 
+`Any` and `Map` values MUST NOT be used in context attributes, including in
+extension attributes.
```
duglin: I like this version of the PR better - it feels less confusing to me.  One thing I think we need to do is expand a bit of this sentence. Normally I would say we should do it in the Primer because the text in the spec should be obvious and the Primer is just for a more in-depth discussion, but in this case there's no indication at all as to why we're making this rule.

Do we need to say "Any" or can "Map" alone cover it? I wonder if someone will read it as: Any == one of int, string, etc... and therefore I can't use any of those.  Or maybe: Values of type `Map`, either directly or via the `Any` type, MUST NOT be used in context attributes, including in extension attributes.

duglin: ping @evankanderson

JemDay:
General Comment :

I understand the drive for this change but would ask that we consider retaining some model that allows for _extension_s to be represented as a map in some constrained form; i believe there is value in keeping them grouped together. 

I would propose that an extension be defined as a _"string indexed dictionary of string typed values"_ (or something similar) to prevent the map-of-maps situation.

Doing this means that existing transport bindings for extensions continue to work unchanged, and the spec.json change would be very subtle:

```
"extensions": {
      "type": "object",
      "additionalProperties" : {
		"type" : "string"
      }
 },
```

duglin:
I have mixed feeling about this, but one of the things that I keep thinking about is when someone (maybe @clemensv) mentioned that just about any serialization of maps in binary format (e.g. as http headers) will result in a really non-trivial mechanism to do filtering over that data. And when I combine that with the idea that CloudEvents metadata should not really be used to carry business logic (that's what `data` is for), and people will want to do quick filtering/routing on the CloudEvent metadata, it pushes me more towards the idea that it may not be a bad thing to force people into a model of limiting what they put into CloudEvents metadata. Which means having Maps could be viewed as encouraging more complex and large pieces of data.

And I guess, it's worth reminding people that just because a piece of data isn't part of the CloudEvents properties doesn't mean that it can't still be used for filter/routing... people will just need to do something like `data.mymap["name"]="Jeff"` instead of `ce-mymap["name"]="Jeff"`.

Of course I also once suggested that everything be strings... so what do I know  :-) 
(although, that's almost where we are now)

spec.md
evankanderson: Done.

FWIW, I find this construction more complicated than defining a separate type, because now we can't say elsewhere in the spec "Unknown extension attributes should be assumed to be of type `Any`" (for example) without including a reference to this clarification.

evankanderson:
Observation from a number of transports:

* HTTP has string-string header maps, but no support for nested structure.
* AMQP has [string-type headers, but does not support nested structure](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-application-properties).
* Google PubSub has [string-string properties](https://cloud.google.com/pubsub/docs/reference/rest/v1/PubsubMessage).
* MQTT has [string-string user properties](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901027)
* SQS supports a [max of 10 string-type attributes, where type is Int, Binary, String](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-attributes.html)
* Kafka has string-byte [headers](https://kafka.apache.org/documentation/#record).

JemDay:
Apologies if this is a bit rambling, hopefully you can see what i'm driving at:

####

My understanding of how extensions that allow for mutiple attributes would be represented (currently and with my proposal)

Example using the _sequence_ and _partioning_ extension along with a company (Bobs Bait Shop - BBS) private internal extension.

##### HTTP, Structured, JSON "data"

```
Content-Type : application/cloudevents+json
Content-Length : nnnn

{
    "specversion" : "0.5",
    "type" : "org.jem.example",
    "time" : "2019-07-12T20:10:00Z",
    "id" : "aaaa-bbbb-cccc",
    "extensions" : {
        "sequence" : {
            "sequence" : "99",
            "sequencetype" : "Integer"
        },
        "bbs" : {
            "ctx" : "dnflkdsnflksdnflkdnf",
            "correlation" : "666-yyyy-aaaa"
        },
        partionkey : "P1"
    }
    "data" : {
         "quote" : "Now is the winter of our discount tents"
    }
}
```

##### HTTP, Binary, JSON data

```
ce-specversion : 0.5
ce-type : org.jem.example
ce-time : 2019-07-12T20:10:00Z
ce-id : aaaa-bbbb-cccc
ce-sequence-sequence : 99
ce-sequence-sequencetype : Integer
ce-bbs-ctx : dnflkdsnflksdnflkdnf
ce-bbs-correlation : 666-yyyy-aaaa
ce-partionkey : P1
Content-Length : nnn

{
    "quote" : "Now is the winter of our discount tents"
}
```

Having sketched this out, and assuming i'm correct, then i believe my proposal _has legs_. :-)

Without some formal namespace/grouping in structured mode everything gets flattened.

```
{
    "specversion" : "0.5",
    "type" : "org.jem.example",
    "time" : "2019-07-12T20:10:00Z",
    "id" : "aaaa-bbbb-cccc",
    "sequence" : "99",
    "sequencetype" : Integer",
    "ctx" : "dnflkdsnflksdnflkdnf",
    "correlation" : "666-yyyy-aaaa",
    "partionkey" : "P1",
    "data" : {
         "quote" : "Now is the winter of our discount tents"
    }
}
```

As an SDK provider i can no longer say to the application code "here are all the attributes of _Bobs-Bait-Shop_ custom extension".

duglin:
Actually, since we don't put extensions into a bag it would be:
**HTTP, Structured, JSON "data"**
```
Content-Type : application/cloudevents+json
Content-Length : nnnn

{
    "specversion" : "0.5",
    "type" : "org.jem.example",
    "time" : "2019-07-12T20:10:00Z",
    "id" : "aaaa-bbbb-cccc",
    "sequence" : {
        "sequence" : "99",
        "sequencetype" : "Integer"
    },
    "bbs" : {
        "ctx" : "dnflkdsnflksdnflkdnf",
        "correlation" : "666-yyyy-aaaa"
    },
    partionkey : "P1"
    "data" : {
         "quote" : "Now is the winter of our discount tents"
    }
}

```

The Binary version looks right.
However, if we did ban maps I think the structured version would be:
```
{
    "specversion" : "0.5",
    "type" : "org.jem.example",
    "time" : "2019-07-12T20:10:00Z",
    "id" : "aaaa-bbbb-cccc",
    "sequencesequence" : "99",
    "sequencesequencetype" : Integer",
    "bbsctx" : "dnflkdsnflksdnflkdnf",
    "bbscorrelation" : "666-yyyy-aaaa",
    "partionkey" : "P1",
    "data" : {
         "quote" : "Now is the winter of our discount tents"
    }
}
```
I would question the choice of using a prefix of "sequence" and a property of "sequence" instead of the first being a bit more descriptive (or just prefixing those two `sequence` ones at all), but it does work in the sense that you can just tell people to look for all properties that start with "sequence" or "bbs".

JemDay:
Hmmm .. according to the spec.json file we *do* put extensions into a bag ;-)

duglin:
LOL well that's a bug - good catch!  PR coming....

duglin:
@JemDay https://github.com/cloudevents/spec/pull/472

duglin:
@evankanderson I think this line:  https://github.com/cloudevents/spec/pull/467/files#diff-958e7270f96f5407d7d980f500805b1bR422   might need to be updated

JemDay:
@duglin - I found the sequence/sequence weird as well, we might want to consider having guidance of extension writers to use more terse naming (much like JWT's).

duglin:
@JemDay I agree, but I wasn't suggesting that it should be that way. I was just showing what it would look like if you blindly squashed things. In a real-world situation, I wouldn't name things such that I ended up with `sequencesequence` - I would be a better smarter about it :-)

JemDay:
Another attempt to clarify my concern with the propose dchnages, how do we move between structrued, mode, binary mode, and back without loss?

#### (1) Start with a structured event

```
Content-Type : application/cloudevent+json
Content-Length : nnnn

{
    "specversion" : "0.5",
    "type" : "org.jem.example",
    "time" : "2019-07-18T16:26:00Z",
    "id" : "aaaa-bbbb-cccc",
    "sequence" : {
        "sequence" : "99",
        "sequencetype" : "Integer"
    },
    "bbs" : {
        "ctx" : "ABCD",
        "correlation" : "111-222-333"
    },
    partionkey : "P1",
    "data" : {
         "message" : "The quick brown fox jumped over the lazy dog"
    }
}
```

#### (2) Now, re-encode it into Binary

```
Content-Type : application/json
Content-Length : nnnn

ce-specversio" : 0.5
ce-type : org.jem.example,
ce-time : 2019-07-18T16:26:00Z,
ce-id : aaaa-bbbb-cccc,
   
---- Fill in the blanks for the extension encoding
---- then proceed to (3)

{
    "message" : "The quick brown fox jumped over the lazy dog"
}
```

#### (3) Proof

So, only with the information above , how do you transform the binary representation into the structured one we started with using the various binding sprecifications.

If you can show a translation from (2) -> (1) with no prior knowledge of (1) then i'm happy.

I believe my proposal (further up the PR) _'restricting a extension to be a bag-of-strings (or primitives)'_, eg using an `ce-bbs-ctx` scheme allows for this to occur.

evankanderson:
> @evankanderson I think this line: https://github.com/cloudevents/spec/pull/467/files#diff-958e7270f96f5407d7d980f500805b1bR422 might need to be updated

>> This specification places no restriction on the type or semantics of the
extension attributes.

We should probably fix that sentence anyway; we don't support sequences (arrays) or many other types (including boolean).

evankanderson:
Summary of pros and cons:

**Pros of removing `Map` from context attributes**
* Adds an extra (fairly complex) concept to the type system
* Without custom mapping ([like HTTP](https://github.com/cloudevents/spec/blob/master/http-transport-binding.md#3131-http-header-names)), it is hard to filter on `Map`-valued attributes
* Unknown extensions in the SDK are of type `Or<String, Map>`, which is awkward
* Does not match the underlying header semantics on [6 surveyed transports](https://github.com/cloudevents/spec/pull/467#issuecomment-510563541)
__Minor pros__
   * HTTP has only a single non-alnum character for header key separation; by using this for `Map`, we have no separators in context attribute names
   * `Map`s are recursive, it's not clear that [HTTP](https://github.com/cloudevents/spec/blob/master/http-transport-binding.md#3131-http-header-names) handles the recursive case


**Pros of having `Map`s as context attributes**
* Allows nesting of objects for association
* Allows using a `Map` as a namespace for extension
* Maps naturally in some languages

evankanderson:
With respect to Jem's example, I would recommend the following instead:

**(1) Start with a structured event**
```
Content-Type : application/cloudevent+json
Content-Length : nnnn

{
    "specversion" : "0.5",
    "type" : "org.jem.example",
    "time" : "2019-07-18T16:26:00Z",
    "id" : "aaaa-bbbb-cccc",
    "datacontentype": "application/json",
    "sequence" : "99",
    "sequence_type" : "Integer",
    "bbs_ctx" : "ABCD",
    "bbs_correlation" : "111-222-333"
    "partion_key : "P1",
    "data" : {
         "message" : "The quick brown fox jumped over the lazy dog"
    }
}
```

**(2) Now, re-encode it into Binary**

```http
POST /... HTTP/1.1
Content-Length: nnnn
CE-Specversion: 0.5
CE-Type: org.jem.example
CE-Time: 2019-07-18T16:26:00Z
CE-Id : aaaa-bbbb-cccc
CE-Sequence: 99
CE-Sequence-Type: Integer
CE-Bbs-Ctx: ABCD
CE-Bbs-Correlation: 111-222-333
CE-Partition-Key: P1

{
    "message" : "The quick brown fox jumped over the lazy dog"
}
```


In addition to this PR, this assumes a second PR which would add "_" to the allowed set of interior characters in context attributes (i.e. not "__foo" or "foo__"). "_" would be a language- and content-encoding specific delimiter character; in HTTP, it would be mapped to "-" (and the character after a "-" would be capitalized in this implementation, though HTTP headers are case-insensitive). "_" is chosen over "-" or other characters as it is a valid separator character in most programming languages, even though [it is uncommon in HTTP header names](https://stackoverflow.com/questions/22856136/why-underscores-are-forbidden-in-http-header-names).

cneijenhuis:
> this assumes a second PR which would add _ to the allowed set of interior characters in context attributes

As I said on the call, this was already discussed in the working group. I personally did not agree to do away with any kind of separation between words, but the working group had a vote and accepted the proposal by Clemens. See https://github.com/cloudevents/spec/pull/321 , in particular

> To make up for the loss of the word separation by case, it would be great to have some kind of word separator available in addition to lower-case letters and digits, but dots and dashes and underscores clash with identifier rules in various languages/protocols. I don't think the result is all that terrible and the wire doesn't care about aesthetics.

While I personally don't like the tradeoff made in #321 (and have voted against it), I'd prefer if we can just keep it as-is and move on. Re-discussing all tradeoff decision every couple of months is not a good way to make progress towards a `1.0` 😉

duglin:
Focusing just on this one:
> Unknown extensions in the SDK currently are of type Or<String, Map>, which is awkward 

I'm curious to know how people expect SDKs/code to deal with this.

When the extension is in a format like JSON then I suspect the answer is the same as how you'd deal with any unknown JSON property since that could be a singular value or an object/map.

When we're dealing with binary mode, e.g. http binary, and we allow any arbitrary depth of maps, then it's harder. However, if we limited maps to just one level and kept the binary serialization as `ce-map-key=value` then that should be easy to reconstruct as a map - which then puts us back to the previous state of how to return a map vs singular value back to the app.

I'd like to know what people think about the compromise position that @JemDay mentioned... keep maps but just limit them to one level.

duglin:
At the end of this discussion on yesterday's call we briefly touched on whether maps could be added back in post v1.0.  At the time I had my doubts, however I now wonder if we could add them by doing this:
- add something to spec today that says something like: as a consumer if you get an attribute that starts with "ce-" but then contains characters that are not allowed to be there, then you MUST generate an error and not process the message beyond the logging or error reporting aspect of it. Meaning, no business processing at all.
- if in v1.1 we add maps back in, then the `-` char will suddenly be a valid character. Which means an existing v1.0 consumer will still continue to error on it, but a v1.1 consumer will work just fine. So, while we can't force a v1.0 consumer to understand a v1.1 message, we can guarantee that it won't let it thru and there's a way for the sender to detect that it wasn't just ignored (assuming it can check the logs).

Thoughts?

evankanderson:
Assuming that we want to remove `Map` from context attributes (but retain it as a possible expansion of `Any` in the `data` payload), I think this is ready to vote/merge.

I may be out on Thursday; I'll see whether I can make the call, but I have a meeting conflict at that time.

Secondary question -- would you prefer that I squash the commits before this is merged?

duglin:
Given the magnitude of this change, we decided to start an async vote that ends at the start of next week's call. The following have voting rights:

Adobe, ControlBEAM, Google, IBM, NATS/Synadia, PayPal, Red Hat, RX-M, SAP, VMWare, Vlad Ionescu, Erik Erikson

You can either vote via comment in this PR or at the start of the call next week - but a vote via comment (with reasons why) would be great for those who are unsure and would like to understand your reasoning.

rperelma:
I vote in favor of this proposal, i.e. removing Map from the set of supported attribute types, and the reason is because it adds complexity to the spec, as has been discussed extensively in this PR.

deissnerk:
SAP votes for approving this PR.

Vlaaaaaaad:
I'm in favor of this PR.

We'll make a decision now, release 1.0-rc1 and as per Doug's comment we'll see what the real world feedback is.

markpeek:
VMware votes for approving this PR. Agreed with simplicity for 1.0.

ColinSullivan1:
Synadia votes to approve.  Keep things simple.  Maps can be added in the future if deemed necessary  through a use case.

PR #469, opened 2019-06-27
Clarifications on batching to address #414

cneijenhuis:
Addresses 1. and 3. from #414 

I've left out 2., because I think it has been clarified with #466

### DIFF
--- a/primer.md
+++ b/primer.md
@@ -125,7 +125,9 @@ of how to serialize the event in different formats (e.g. JSON) and transports
 Batching of multiple events into a single API call is natively supported by some
 transports. To aid interoperability, it is left up to the transports if and how
 batching is implemented. Details may be found in the transport binding or in the
-transport specification.
+transport specification. A batch of CloudEvents carries no semantic meaning and
+is not ordered. An [Intermediary](spec.md#intermediary) can add or remove
+batching as well as assign events to different batches.
 
 ### Non-Goals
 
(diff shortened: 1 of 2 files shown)


primer.md
```diff
@@ -125,7 +125,9 @@ of how to serialize the event in different formats (e.g. JSON) and transports
 Batching of multiple events into a single API call is natively supported by some
 transports. To aid interoperability, it is left up to the transports if and how
 batching is implemented. Details may be found in the transport binding or in the
-transport specification.
+transport specification. A batch of CloudEvents carries no semantic meaning and
```
duglin: The "no semantic meaning" part might feel a bit off since it might be possible that someone else defined a semantics meaning to the batch. I think perhaps you meant: The CloudEvents specification does not define any semantic meaning, or ordering, to a batch of CloudEvents.   WDYT?

duglin: ping @cneijenhuis

cneijenhuis: The problem, from my point of view, is that existing transports with a batching mechanism will break/re-group batches, and future ones will probably too.

So one SHOULD NOT define a meaning/ordering for a batch, and if a producer does, the consumer SHOULD simply act as if that definition doesn't exist to guarantee interoperability if the transport between the two is changed.

Since this is the primer, I tried to keep it short and didn't use SHOULD NOT.

Your proposal is also fine by me, it doesn't change much.

duglin: I'm ok either way - no biggie

duglin:
Approved on the july 11th call

PR #470, opened 2019-07-11
Clarified data and datacontenttype

### LINKED ISSUE #261: Map and Object type problems

jroper:
Supersedes #457
Fixes #261

Clarified the relationship between `data` and `datacontenttype`, documenting the constraints between the two and their allowed values and types.

@clemensv I think this sums up what we discussed.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -336,10 +336,6 @@ on the definition of OPTIONAL.
   event format defines the relationship in
   [section 3.1](./json-format.md#31-special-handling-of-the-data-attribute).
 
-  When this attribute is omitted, the `data` attribute simply follows the event
-  format's encoding rules. For the JSON event format, the `data` attribute value
-  can therefore be a JSON object, array, or value.
-
   For the binary mode of some of the CloudEvents transport bindings, where the
   `data` content is immediately mapped into the payload of the transport frame,
   this field is directly mapped to the respective transport or application
@@ -347,8 +343,23 @@ on the definition of OPTIONAL.
   and the content-type metadata mapping can be found in the respective transport
   mapping specifications.
 
+  Implementations MAY introduce this parameter when encoding `Integer`, `Map`
+  and `String` to a binary mode of a CloudEvents transport binding. For example,
+  when encoding a `Map` in binary mode, the implementation MAY encode the value
+  as JSON if the underlying transport encoding does not support `Map` payloads,
+  in accordance to the normative rules for the `Map` type, and therefore MUST
+  set the `datacontenttype` to `application/json`. When decoding the event, the
+  new type of `data` will therefore be `Binary`, with a `datacontenttype` of
+  `application/json`.
+
 - Constraints:
-  - OPTIONAL
+  - MUST be compatible with the type of the `data` attribute, given the
+    following rules:
+    - `Binary` - REQUIRED, and can be any media type that can be used to decode
+      the value of the `data` attribute.
+    - `String` - OPTIONAL, and MUST be `text/*` if present.
+    - `Integer` - MUST NOT be present.
+    - `Map` - MUST NOT be present.
   - If present, MUST adhere to the format specified in
     [RFC 2046](https://tools.ietf.org/html/rfc2046)
 - For Media Type examples see
@@ -451,8 +462,15 @@ encapsulated within the `data` attribute.
 
 - Type: `Any`
 - Description: The event payload. The payload depends on the `type` and the
-  `schemaurl`. It is encoded into a media format which is specified by the
-  `datacontenttype` attribute (e.g. application/json).
+  `schemaurl`. If the type is `Binary` or `String`, it is encoded using a 
+  media format which is specified by the `datacontenttype` attribute 
+  (e.g. `application/json`).
+
+  The abstract type (ie, `Binary`, `Text`, `Integer` or `Map`) MAY change when
+  encoded using a binary mode transport binding. For example, a `Map` MAY
+  become `Binary`, with a `datacontenttype` attribute set to 
+  `application/json`.
+
 - Constraints:
   - OPTIONAL
 


spec.md
```diff
@@ -336,19 +336,30 @@ on the definition of OPTIONAL.
...
-  can therefore be a JSON object, array, or value.
-
   For the binary mode of some of the CloudEvents transport bindings, where the
   `data` content is immediately mapped into the payload of the transport frame,
   this field is directly mapped to the respective transport or application
   protocol's content-type metadata property. Normative rules for the binary mode
   and the content-type metadata mapping can be found in the respective transport
   mapping specifications.
 
+  Implementations MAY introduce this parameter when encoding `Integer`, `Map`
```
clemensv: I would turn that on its head as a MUST rule for Binary/String. This should also be "attribute", not "parameter":

> The attribute MUST be set for `Binary` content in the `data` field, with an accurate media-type.
> If no IANA registered media-type is available, a vendor extension or `application/octet-stream`
> MAY be used.
>
> The attribute SHOULD be set for `String` content in the `data` field. If set, the media-type MUST     
> describe a text format. If the attribute is omitted for `String` content, the default is `text/plain`.
> If the character encoding is not UTF-8 or one of its subsets (US-ASCII, ISO-8859-1) or if the 
> character encoding of the overall event diverges, the character encoding MUST be declared with
> a charset parameter and the text MUST be encoded as `Binary`.
>
> The attribute MAY be set for all other types in order to enforce encoding of the `data` attribute 
> according to the rules of a particular event format, which might differ from the event format 
> chosen for the overall event. Using "application/json" will encode the `data` content according to 
> the JSON event format rules. When the event itself is JSON encoded, the content is inlined into 
> the JSON object's `data` field. When the event is encoded in a different event format, the content 
> is carried as `String` or `Binary` under consideration of the aforementioned character encoding rules.
>
> If the attribute is omitted, the content of the `data` attribute is encoded according to the rules of
> the chosen event format, which also allows for transcoding of the content as an event is 
> forwarded through routes that use different event formats. An event might initially published via
> MQTT with Protobuf encoding for preserving bandwidth, but then routed onwards via Kafka 
> using JSON encoding since the consumers are expecting JSON content.  
>
> For using the "binary" mode of any transport binding, the `datacontenttype` MUST be declared.

spec.md
```diff
@@ -336,19 +336,30 @@ on the definition of OPTIONAL.
...
+  new type of `data` will therefore be `Binary`, with a `datacontenttype` of
+  `application/json`.
+
 - Constraints:
-  - OPTIONAL
+  - MUST be compatible with the type of the `data` attribute, given the
+    following rules:
+    - `Binary` - REQUIRED, and can be any media type that can be used to decode
+      the value of the `data` attribute.
+    - `String` - OPTIONAL, and MUST be `text/*` if present.
```
clemensv: I've covered this ground also with the proposal above, but just for completeness of the review:
* There are many "application/*" types that are also text
* Your rule for Map and Integer here (MUST NOT) contradicts the rule above (MAY).

spec.md
```diff
@@ -451,8 +462,15 @@ encapsulated within the `data` attribute.
 
 - Type: `Any`
 - Description: The event payload. The payload depends on the `type` and the
-  `schemaurl`. It is encoded into a media format which is specified by the
-  `datacontenttype` attribute (e.g. application/json).
+  `schemaurl`. If the type is `Binary` or `String`, it is encoded using a 
+  media format which is specified by the `datacontenttype` attribute 
+  (e.g. `application/json`).
+
+  The abstract type (ie, `Binary`, `Text`, `Integer` or `Map`) MAY change when
```
clemensv: We agree on the effect; I've tried to explain this differently in the proposal above ("The attribute MAY be set for all other types..."). What happens is that the in-memory structure at the publisher might contain a map, but `datacontenttype` is set to "application/json" while the event is encoded as Avro. In that case, JSON rules apply to `data` and the attribute ends up being a String holding JSON inside the Avro-encoded event and therefore pops out like that on the other end, but I don't think that ought to be a real surprise. We might write a primer paragraph that clarifies that behavior.    

When you use the normative "MAY", you're permitting something, but here you're actually just describing a potential effect, which is a "might".

spec.md
jroper: The problem with allowing non `"text/*"` types to be string is that when encoding to JSON, we have no other heuristic to know when it should be base64 unencoded or not when it's received. Let's say I am the sender, and I send a `String` with a `datacontenttype` of `application/mytexttype`, and I encode it as JSON, so it gets put as is. Then it gets received by a transcoder that transcodes to protobuf, but that transcoder has no knowledge of my mime type, so assumes it's binary. Because it's binary, it base64 decodes the string, and ends up putting junk in the protobuf binary field - the event is now corrupt. The point is, to prevent such corruptions, we need a robust way for JSON decoders to know if the data is string or binary, only allowing `text/*` is robust, but if we remove that rule, then we need a different way.

The rule for `Map` and `Integer` here doesn't contradict the rule above, because as soon as you encode the `Map` to JSON bytes, it becomes `Binary`, so is no longer a `Map`, and so this rule not applicable.

clemensv: My point is that "text/*" is not a reliable indicator for text content. "application/xml" is text. 

There's no ambiguity around text vs binary because we have a rule that if `Binary` is inside a string, you must set `datacontentencoding` to "Base64". Base64 in a string in `data` cannot exist without that attribute being present. That was what @alanconway raised and we fixed that problem with that attribute.

In binary transport mode, the payload is a sequence of bytes or text as indicated by the media type. In doubt, e.g. if you can't tell what the content type means, it's always binary and if you find a "charset" parameter, you know that you ought to run this through a character encoder that yields text. 

The Map/Integer rules are contradictory because the JSON event format retains the ability to inline JSON content as JSON into the `data` attribute. The content only turns into binary if you transcode the outer event into something else. Otherwise, `data` just carries a JSON value and with the declaration in place, the JSON may also contain bool or values that deviate from our rule set. Without the declaration, you are limited to the abstract type set.

spec.md
deissnerk: @clemensv 
If an event router receives an event in HTTP binary mode, how can it distinguish events for which the producer explicitly set datacontenttype to application/json from an event that was produced without a datacontenttype? How does it know when to transcode the content of the data attribute and when to continue treating it just as a JSON document?

duglin: In binary mode... does it matter? In binary mode, I believe that we can only really be concerned about how to get the CE specific attributes, which means the `ce-` http headers. The http body is (basically) out of scope for us (aside from the abstract notion of it mapping to our `data` attribute) but whatever normal processing the receiver has should apply, just as if CloudEvents were never used at all.  No?

deissnerk: Perhaps I am still missing something, but isn't there a difference between data adhering to the CE type system, that is only represented as JSON for a single hop (HTTP binary), and data that was intended to be a JSON document? I don't see, how it is possible to distinguish these cases after an HTTP binary hop.

duglin: Maybe I'm missing something too :-)  But to me, `data` should have zero restrictions on it regardless of binary vs structured. CE should never restrict, or change, what can be sent for business logic (ignoring serialization aspects). The fact that we are even having this discussion kind of bothers me and I think it's due to presence of our type system. I don't think it should apply to `data` since doing so implies some kind of restrictions and I hope that wasn't the intent.

Using it for our other attributes just seems unnecessary to me since we require a string mapping and unknown attribute would remain as strings. Known attributes, since they're known, can be converted from string to **whatever** by the entity that understands it's type.  And that can all be done (and specified in the spec) w/o a formal type system that seems to be introducing more questions than necessary.

To me, if we defined a new attribute called `size`, then:
```
name: size
type: int
constraints: must be between 0 and 100
```
can be expressed as:
```
name: size
constraints: must be a string representation of an integer between 0 and 100
```
and I think we'd get the exact same serializations and code written, but less questions/concerns.  And even with this, if the transport does support types metadata, the spec for that transport can specify that an `int` type be used if it wants.

duglin:
Closing in favor of #484 per the 8/15 call

PR #472, opened 2019-07-12
We no longer put extensions into an 'extensions' bag

duglin:
Jem noticed this


JemDay:
I assume this change happened before i joined; i will refrain from commenting on the connivence of this change in light of discussions in other PR's ;-)

duglin:
Approved on 7/18 call

PR #475, opened 2019-07-31
s/schemaurl/dataschema/ to be consistent

### LINKED ISSUE #474: schemaurl -> dataschemaurl ?
Anyone else think it's odd that we have:
- datacontenttype
- datacontentencoding

but then we have:
- schemaurl

when all 3 are about `data` ? Should `schemaurl` really be `datacontentschemaurl` ?
Or just `datacontentschema` since we don't put the type into other property names.
I'd actually be ok with just `dataschema[url]` since I think starting with `data` is the most important thing.

Of course we could just do:
- datatype
- dataencoding
- dataschema

Thoughts?

duglin:
Let's discuss. But I tend to think it's better to get this right
for v1.0 than to continue to be inconsistent and let a pre-v1.0
version dictate things forever.

Closes #474

### DIFF
--- a/SDK.md
+++ b/SDK.md
@@ -128,7 +128,7 @@ Here's the list of getters:
     Source() -> URL
     EventID() -> string
     EventTime() -> RFC3339
-    DataSchemaURL() -> string
+    DataSchema() -> URL
     DataContentType() -> string
     Data() -> serializable
     Extensions() -> map[string]string
--- a/SDK.md
+++ b/SDK.md
@@ -128,7 +128,7 @@ Here's the list of getters:
     Source() -> URL
     EventID() -> string
     EventTime() -> RFC3339
-    SchemaURL() -> string
+    DataSchemaURL() -> string
     DataContentType() -> string
     Data() -> serializable
     Extensions() -> map[string]string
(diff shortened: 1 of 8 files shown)


duglin:
On last week's call we briefly discussed this and "someone" mentioned that they thought the `schemaurl` pointed to the schema of the entire CloudEvent, not just `data`.  That's part of the reason I want to make this change. Many people will not know (or read) the CE spec like we do and so we need to pick names that are as clear as possible for those who don't.

If you have an opinion on this (for or against) please speak up here or on the call this week. I'd like to resolve this issue (one way or the other) if we can.

duglin:
Per today's call I:
- changed the attribute to: “dataschema”
- made it a URI
- Add URI to type section

SDK.md
```diff
@@ -128,7 +128,7 @@ Here's the list of getters:
     Source() -> URL
     EventID() -> string
     EventTime() -> RFC3339
-    SchemaURL() -> string
+    DataSchema() -> string
```
n3wscott: URL

duglin: Makes sense to me - I'll pick up this change - if someone thinks it's incorrect we can open another PR and revert it.  @clemensv @markpeek @fabiojose @deissnerk

duglin:
Approved on the 8/15 call

PR #478, opened 2019-08-05
Tweak URI-references and be smart with Strings

### LINKED ISSUE #473: How to deal with odd chars in strings
In a chat with @jroper he was asking about how to deal with new-lines in string. This is related to Even's PR https://github.com/cloudevents/spec/pull/456 .

It seems to me that we can take two approaches:
1 - add text that says something to the effect: don't be stupid :-)  Yes you can put all sorts of weird text into strings but if you want your stuff to be consumable by as many people as possible then don't do weird stuff that'll just break people. This is about interop, not pushing for edge-case testing of the spec.  We left the spec open so that people can have the freedom to leverage some of those odd cases if they really needed to, but those might only be useful in very confined environments with certain transports where interop with everyone (across multiple transports) isn't the main objective.
2 - add rules to each spec (e.g. http transport) that explains what to do in cases like this. E.g. URL-encode the string in http binary mode. 

Doing part of 2 wouldn't necessarily be a bad thing, but I'm leaning more toward 1 a the general rule.  Comments?

duglin:
I can't think of a good reason why URI-references should be allowed to
be blank since that seems about as useful as the property not being
there at all.

Closes #473

### DIFF
--- a/primer.md
+++ b/primer.md
@@ -19,7 +19,8 @@ This document is a working draft.
 - [Design Goals](#design-goals)
 - [Architecture](#architecture)
 - [Versioning of Attributes](#versioning-of-attributes)
-- [CloudEvent Attributes Extensions](#cloudevent-attribute-extensions)
+- [CloudEvent Attributes](#cloudevent-attributes)
+- [CloudEvent Attribute Extensions](#cloudevent-attribute-extensions)
 - [Creating CloudEvents](#creating-cloudevents)
 - [Qualifying Protocols and Encodings](#qualifying-protocols-and-encodings)
 - [Proprietary Protocols and Encodings](#proprietary-protocols-and-encodings)
@@ -182,7 +183,7 @@ The CloudEvents specification set defines four different kinds of protocol
 elements that form a layered architecture model.
 
 1. The [base specification](spec.md) defines an abstract information model
-   made up of attributes (key-value pairs) and associated rules for what 
+   made up of attributes (key-value pairs) and associated rules for what
    constitutes a CloudEvent.
 2. The [extensions](./spec.md#extension-context-attributes) add use-case specific
    and potentially overlapping sets of extension attributes and associated
@@ -193,7 +194,7 @@ elements that form a layered architecture model.
    an application protocol.
 4. The transport bindings, e.g. [HTTP](http-transport-binding.md), defines
    how the CloudEvent is bound to an application protocol's transport frame,
-   in the case of HTTP to the HTTP message. The transport binding does not 
+   in the case of HTTP to the HTTP message. The transport binding does not
    constrain how the transport frame is used, meaning that the HTTP binding
    can be used with any HTTP method and with request and response messages.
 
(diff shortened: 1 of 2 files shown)


duglin:
I think this help address some of https://github.com/cloudevents/spec/pull/456  too

primer.md
```diff
@@ -235,6 +236,54 @@ general, this is true for all CloudEvents attributes as well.
...
+situations. So, CloudEvent producers are encouraged to only use values that
+have the best chance of being properly processed by all possible formats
+and transports that their CloudEvents might be used with.
+
+### URI-reference Values
+
+Attributes of type URI-reference have quite a large set of valid values that
+can be used. For example, it is possible that the `source` attribute could be
+as short as `home` or could be more meaningful, such as
+`com.example.my-component`. The CloudEvents specification allows for a wide
```
cneijenhuis: `example.com/my-component` sounds more like a proper URI-ref to me (unless `my-component` is a new TLD I haven't heard of yet 😉 )

primer.md
```diff
@@ -235,6 +236,54 @@ general, this is true for all CloudEvents attributes as well.
...
+for attribute values so as to allow for the specification to be used in
+situations where those potentially problematic characters are desired.
+
+However, this should not be interpreted as encouragement for producers to
+push the boundaries of what characters should be used just for the sake of
+edge-testing implementations. CloudEvents is striving to help with
+interoperability while still being flexible enough to be used in unique
+situations. So, CloudEvent producers are encouraged to only use values that
+have the best chance of being properly processed by all possible formats
+and transports that their CloudEvents might be used with.
```
cneijenhuis: For SaaS publishers, their CloudEvents may end up being used with any format or transport. It'd be good to know the minimum one can rely on. ASCII? URL encoding?

primer.md
duglin: fixed :-)

primer.md
duglin: I thought about that too, but then I noticed that define Strings as `Sequence of printable Unicode characters`
which to me removed a fair about of the reasons for needing to URL encode stuff. Plus I didn't think we'd want to get into when, or when not, to encode Strings since then we'd need a mechanism by which consumers would have to know when they're encoded.  That's when I decided "just don't be dumb" was good enough.

But if someone has any ideas...

primer.md
```diff
@@ -235,6 +236,54 @@ general, this is true for all CloudEvents attributes as well.
...
+
+### URI-reference Values
+
+Attributes of type URI-reference have quite a large set of valid values that
+can be used. For example, it is possible that the `source` attribute could be
+as short as `home` or could be more meaningful, such as
+`http://home.example.com`. The CloudEvents specification allows for a wide
+range of values to allow for it to be used in as any use cases as possible,
+without undo burden on implementations.
```
rperelma: s/undo/undue

duglin: LOL fixed

primer.md
```diff
@@ -235,6 +236,54 @@ general, this is true for all CloudEvents attributes as well.
...
+choice when the CloudEvent is meant to be used in a very public environment,
+it may make perfect sense in a closed system where a full reverse-DNS name
+might be unnecessary. Or perhaps there is a severe size limitation on
+the size of the CloudEvent and shorter URI-references are needed.
+
+These are decisions that CloudEvent producers will need to decide for
+themselves based on their needs and expected usage of their events.
+
+Additionally, the CloudEvents specification does not define a `baseURI`
+to be used along with the URI-reference type attributes. This is because,
```
deissnerk: I wonder, if it prohibits that event producers would define a base URI in their documentation?

primer.md
```diff
@@ -235,6 +236,54 @@ general, this is true for all CloudEvents attributes as well.
...
+unless otherwise stated, the URI-references are not meant to be normalized
+or used for any other purpose other than comparing its string value against
+some other URI-reference string value for equality. The use of URI-reference,
+instead of just a String, was done to encourage the use of meaningful values
```
deissnerk: Referring to a complex specification such as [RFC3986](https://tools.ietf.org/html/rfc3986) just to encourage people to chose meaningful values while ignoring most of the specification, seems a bit strange.

primer.md
duglin: I don't think we can stop anyone from doing so... but I'd question the need for it  :-)

primer.md
duglin: maybe but I think there is something to be said for encouraging some level of consistency, but still allowing a lot of freedom.  Kind of like a best-practice.  But, having said that, this is just my memory of our previous chats and if people think I'm remembering incorrectly, or they just don't want to even mention this aspect of it, I'd be ok with removing this sentence if that would help

duglin:
up updated based on today's call.  Let me know what you think

deissnerk:
Thank you @duglin! LGTM

spec.md
```diff
@@ -363,6 +368,7 @@ on the definition of OPTIONAL.
   for more information.
 - Constraints:
   - OPTIONAL
+  - If present, MUST be a non-empty URI-reference
```
evankanderson: I'm more skeptical of the value of a URI-reference here; I sort of understand the thinking that we want to allow a subset of general strings as `source` attributes even if they don't conform to the
```URI           = scheme ":" hier-part [ "?" query ] [ "#" fragment ]```
production. Since `schemaurl` is OPTIONAL rather than required, why not require that the schema URL be an actual URI and not relative to some (unknown) base?

Comment (deissnerk) [reply spec.md]: @evankanderson I think that's part of #475.

**Comment by duglin** [comment]
moved the "RECOMMENDED" line into the constraints section and deleted the rest of the paragraph - per today's call.

Approved with the above change on the 8/15 call

Can I get an LGTM to make sure I didn't mess anything up before I merge?

File: spec.md [hunk 7]
```diff
@@ -251,6 +251,8 @@ The following attributes are REQUIRED to be present in all CloudEvents:
 
 - Constraints:
   - REQUIRED
+  - MUST be a non-empty URI-reference
+  - RECOMMENDED that an absolute URI be used
```
cneijenhuis: Disclaimer: I'm not a native English speaker. `be used` sounds strange to me here.
Either try `is used` or maybe `An absolute URI is RECOMMENDED`?

duglin: works for me!

duglin:
thanks for the review @cneijenhuis 

merging.  Approved on the 8/15 call

PR #479, opened 2019-08-06
Allow for a LOA w.r.t. voting rights

duglin:
It's been mentioned that we have no provision for a leave-of-absence
(e.g. vacations) w.r.t. voting rights. Luckily, we don't vote very often
and they're usually land-slides so I don't think it's impacted anything, but
we should probably have some kind of official allowance for it.

See what ya'll think...


duglin:
Add a comment that it's ok to notify of LOA via slack message - per today's call.

Approved on the 8/15 call with the above edit.

Can I get an LGTM before I merge?

rperelma:
LGTM

duglin:
thanks @rperelma 
merging

PR #482, opened 2019-08-11
Simplify data + datacontenttype

cneijenhuis:
Addresses https://github.com/cloudevents/spec/issues/261, and is an alternative to https://github.com/cloudevents/spec/pull/470 and https://github.com/cloudevents/spec/pull/471

As discussed on last weeks call, this is an attempt to simplify the handling of `datacontenttype` for both producers and consumers. This PR trades off some features and flexibility, but (hopefully) makes the overall process easier.

There are three major changes:
* `data` can only be a `Binary` or a `Map`. (`String` may be added later, see https://github.com/cloudevents/spec/pull/471/files#r312744230)
* `datacontenttype` is REQUIRED
* If the type is `Map`, a new media-type (`application/cloudevents-typed-data+{format}` - I'm open for different naming suggestions) MUST be used. Intermediaries CAN change `{format}`, which @clemensv seems to be fond of 😉 .

Note: The diff is pretty big because I moved `datacontenttype` into the required section, but I only added the 3rd paragraph (lines 242 - 246) and some examples (lines 255 - 261).

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -218,6 +218,48 @@ in context attributes, including in extension attributes.
 
 The following attributes are REQUIRED to be present in all CloudEvents:
 
+#### datacontenttype
+
+- Type: `String` per [RFC 2046](https://tools.ietf.org/html/rfc2046)
+- Description: Content type of the `data` attribute value. This attribute
+  enables the `data` attribute to carry any type of content, whereby format and
+  encoding might differ from that of the chosen event format. For example, an
+  event rendered using the [JSON envelope](./json-format.md#3-envelope) format
+  might carry an XML payload in its `data` attribute, and the consumer is
+  informed by this attribute being set to "application/xml". The rules for how
+  the `data` attribute content is rendered for different `datacontenttype`
+  values are defined in the event format specifications; for example, the JSON
+  event format defines the relationship in
+  [section 3.1](./json-format.md#31-special-handling-of-the-data-attribute).
+
+  For the binary mode of some of the CloudEvents transport bindings, where the
+  `data` content is immediately mapped into the payload of the transport frame,
+  this field is directly mapped to the respective transport or application
+  protocol's content-type metadata property. Normative rules for the binary mode
+  and the content-type metadata mapping can be found in the respective transport
+  mapping specifications.
+
+  If `data` is of type `Binary`, an accurate media-type MUST be set.
+  If `data` is of type `Map`, the media-type
+  `application/cloudevents-typed-data+{format}` MUST be used, where `{format}`
+  is the chosen event format. An [Intermediary](#intermediary) CAN change the
+  event format (e.g. when the format of the envelope is changed).
+
+- Constraints:
+  - REQUIRED if `data` is present
+  - MUST NOT be present if `data` is not present
+  - If present, MUST adhere to the format specified in
+    [RFC 2046](https://tools.ietf.org/html/rfc2046)
+- For Media Type examples see
+  [IANA Media Types](http://www.iana.org/assignments/media-types/media-types.xhtml)
+- Examples
+  - `data` is of type `Binary`
+    - application/json
+    - image/jpeg
+  - `data` is of type `Map`
+    - application/cloudevents-typed-data+json
+    - application/cloudevents-typed-data+avro
+
 #### id
 
 - Type: `String`
@@ -325,38 +367,6 @@ on the definition of OPTIONAL.
   - If present, MUST adhere to
     [RFC 2045 Section 6.1](https://tools.ietf.org/html/rfc2045#section-6.1)
 
-#### datacontenttype
-
-- Type: `String` per [RFC 2046](https://tools.ietf.org/html/rfc2046)
-- Description: Content type of the `data` attribute value. This attribute
-  enables the `data` attribute to carry any type of content, whereby format and
-  encoding might differ from that of the chosen event format. For example, an
-  event rendered using the [JSON envelope](./json-format.md#3-envelope) format
-  might carry an XML payload in its `data` attribute, and the consumer is
-  informed by this attribute being set to "application/xml". The rules for how
-  the `data` attribute content is rendered for different `datacontenttype`
-  values are defined in the event format specifications; for example, the JSON
-  event format defines the relationship in
-  [section 3.1](./json-format.md#31-special-handling-of-the-data-attribute).
-
-  When this attribute is omitted, the `data` attribute simply follows the event
-  format's encoding rules. For the JSON event format, the `data` attribute value
-  can therefore be a JSON object, array, or value.
-
-  For the binary mode of some of the CloudEvents transport bindings, where the
-  `data` content is immediately mapped into the payload of the transport frame,
-  this field is directly mapped to the respective transport or application
-  protocol's content-type metadata property. Normative rules for the binary mode
-  and the content-type metadata mapping can be found in the respective transport
-  mapping specifications.
-
-- Constraints:
-  - OPTIONAL
-  - If present, MUST adhere to the format specified in
-    [RFC 2046](https://tools.ietf.org/html/rfc2046)
-- For Media Type examples see
-  [IANA Media Types](http://www.iana.org/assignments/media-types/media-types.xhtml)
-
 #### schemaurl
 
 - Type: `URI-reference`
@@ -452,7 +462,7 @@ encapsulated within the `data` attribute.
 
 ### data
 
-- Type: `Any`
+- Type: `Binary` or `Map`
 - Description: The event payload. The payload depends on the `type` and the
   `schemaurl`. It is encoded into a media format which is specified by the
   `datacontenttype` attribute (e.g. application/json).


clemensv:
The PR was supposed to express the rules more concisely, not to cut features.

* This takes away the ability to carry string-typed values such as any kind of plain text or text based formats including, for instance, XML or CSV inside the "data" element without character set de/encoding, because the attribute is made required and you're not allowing it to describe strings. 
* You're referring to "application/json" as an example of a binary type. That means that you would have to encode declared JSON content as Base64 text inside the "data" element in structured mode for the JSON encoding, and you also must use the dataencontextencoding attribute with it, per the rules of that attribute. This is at odds with the [JSON event format rules](https://github.com/cloudevents/spec/blob/master/json-format.md#31-special-handling-of-the-data-attribute).
* The correlation you create between the "chosen event format" and the newly introduced media type is not clear. In binary mode, apps don't chose an event format, because the event is completely projected onto the transport frame. As a result, there is nothing to tie that new media type to in that mode. In structured mode, forcing this new media type to be consistent with the chosen event format means that the datacontenttype information is superfluous, because you'll know what the value must be once you know what the event format is. Exactly because of that, I'm allowing the attribute to be optional overall and if omitted, this is the default case. 

Here's the brief summary of the effective rule set coming out of the base PR:
* Binary data is described with a mandatory datacontenttype value and carried as binary data. If carried inside a format that only supports string (JSON structured), the data must be Base64 encoded and that must be declared via datacontentencoding.
* String data is described with an optional datacontenttype value for helping the receiver interpret the string. If the character encoding differs from the event format, the string is instead treated as a binary value.
* Maps don't have a datacontenttype value and their encoding is implied by the surrounding event format. Maps can't be carried in binary mode.

cneijenhuis:
> The PR was supposed to express the rules more concisely, not to cut features.

If this is what you understood from the call, I'm sorry, I didn't express myself properly. I don't think the rules can be expressed much more concisely.
I volunteered to open a PR that is more concise, because the underlying rules are more simple (which comes at the cost of some features).

> This takes away the ability to carry string-typed values such as [...] XML or CSV [...] JSON

As I noted in the description, `String` may be added later, but currently `String` does not allow one to carry XML, CSV or JSON because all of them may contain non-printable characters. For CSVs in particular, line breaks are an integral part of the file format.

I can pretend that `String` allows non-printable characters under the assumption that we'll change the `String` definition (I'm not sure this will be a no-brainer, given it opens another can of worms in HTTP headers...).

Either way, to keep the discussion focused, I suggest to discuss `String` independently. I'll create an extra issue.

---

> That means that you would have to encode declared JSON content as Base64 text inside the "data" element in structured mode for the JSON encoding, and you also must use the dataencontextencoding attribute with it, per the rules of that attribute. This is at odds with the JSON event format rules.

I assume you are referring to `String encoding: Base64 encoding per RFC4648.` in the definition of `Binary`?
From my understanding, the `String encoding` (that is also found for other types) is the default fallback for formats and transport if they specify no other mapping. If they specify another mapping, that takes precedence. E.g. avro represents `Binary` as `bytes`, and also doesn't use the default string encoding for `Map` but uses `map`: https://github.com/cloudevents/spec/blob/master/avro-format.md#21-type-system-mapping

See also the first paragraph of the type system section:

> Each of these types MAY be represented differently by different event formats and in transport metadata fields.

---

> The correlation you create between the "chosen event format" and the newly introduced media type is not clear. In binary mode, apps don't chose an event format, because the event is completely projected onto the transport frame. As a result, there is nothing to tie that new media type to in that mode.
> Maps can't be carried in binary mode.

My reading was that in binary mode, the transport has to map it somehow, with the default option being the previously discussed `String encoding` - aka use JSON. But I accept that I'm wrong here, because then I've actually added a feature 😉 
Maps are supported in binary mode with this PR.

rperelma:
I really want to ensure we preserve the case where the event envelope is JSON, the data is also JSON,  and therefore we don't need datacontenttype.

spec.md
```diff
@@ -218,6 +218,48 @@ in context attributes, including in extension attributes.
 
 The following attributes are REQUIRED to be present in all CloudEvents:
 
+#### datacontenttype
```
duglin: What if instead of putting it under the "required" section, we keep it down below but rename that section to something like "Conditional/Optional Attributes" ?

duglin:
@clemensv wrote a summary of #471 as:
> - Binary data is described with a mandatory datacontenttype value and carried as binary data. If carried inside a format that only supports string (JSON structured), the data must be Base64 encoded and that must be declared via datacontentencoding.
> - String data is described with an optional datacontenttype value for helping the receiver interpret the string. If the character encoding differs from the event format, the string is instead treated as a binary value.

Is there any reason that the spec text couldn't be reduced down to something like the this text and all/most of the rest of #471 be moved into the primer?

Regardless, @clemensv @cneijenhuis since I believe both of you are in Europe, perhaps you guys could ping each other via slack tomorrow and chat?

clemensv:
> As I noted in the description, String may be added later, but currently String does not allow one to carry XML, CSV or JSON because all of them may contain non-printable characters. For CSVs in particular, line breaks are an integral part of the file format.

You can of course represent CSV as a string, because you'll always encode the string either as plaintext in the body of a transport frame (binary) where newline would obviously be acceptable as-is since you're picking a media type that sets the encoding rules, or you use the string representation of an event format. JSON obviously [has escape characters] (https://tools.ietf.org/html/rfc7159#section-7) that allow for putting CSV into the `data` attribute and binary formats generally take any UTF-8. The restriction on non-printable characters is something that might make sense to relax for TAB/CR/LF to be super-correct, even though the escape characters deal with that for the cases I was worried about.

>  From my understanding, the String encoding (that is also found for other types) is the default fallback for formats and transport if they specify no other mapping

There are string representations for each type in the type system. Since your PR's rules don't allow for strings to be stored in `data`, the only way to represent such content is binary. If, as you enumerated, application/json is a binary type, the only conformant way to encode strings in the data field is then to render them into byte stream using the default text encoding (UTF-8) and then encode the byte stream as Base64 and set that as `data` with the datacontentencoding set.

>   My reading was that in binary mode, the transport has to map it somehow,

The transport bindings may not be sufficiently clear here and that appears worth fixing. For HTTP, the rule is "For the binary mode, the HTTP Content-Type value maps directly to the CloudEvents datacontenttype attribute." What that means, in effect, is that if you don't declare a content-type, there is no rule to encode the entity body as, if you read further in the HTTP specs, the implied type were application/octet-stream, i.e. you can an opaque bag of bytes. There's no path for the event format to become applicable.   

> Maps are supported in binary mode with this PR.

They are, but with a media type that requires intimate knowledge of CloudEvents, while our simplest case, where a CloudEvent is just a plain, self-contained JSON object now needs to carry a declaration whose value is already implied by the format per-se.  The simplest case therefore got more complicated.    

```
{
    "specversion" : "0.4-wip",
    "id" : "A234-1234-1234",
    "datacontenttype" : "application/cloudevents-typed-data-json",
    "data" : {
          "hello" : "world"
    }
}
``` 

vs.

```
{
    "specversion" : "0.4-wip",
    "id" : "A234-1234-1234",
    "data" : {
          "hello" : "world"
    }
}
```

evankanderson:
If the concern about printable characters applies mostly to the `data` attribute, Doug's suggested PR in https://github.com/cloudevents/spec/issues/483#issuecomment-520940556 seems to handle this nicely and avoid the need for special-casing what the "printable" character range is.

cneijenhuis:
> I really want to ensure we preserve the case where the event envelope is JSON, the data is also JSON, and therefore we don't need datacontenttype.

@rperelma I tend to agree, but #471 doesn't do that either.

With #471 you don't need datacontenttype if the event envelope is JSON and the data is a CloudEvents-Map (formatted as JSON).
You need datacontenttype if the data is a "real" JSON (e.g. contains an array, a boolean, a floating point number, a null, ...).

See https://github.com/cloudevents/spec/pull/471/files#diff-ed0940ecbc7be65c47b9414d516b3490R169-R177

cneijenhuis:
> because you'll always encode the string either as plaintext in the body of a transport frame (binary) where newline would obviously be acceptable

I've never disputed that. The problem is that you'd never get there if you actually enforce the "string-as-defined-by-CE". I.e. a CloudEvents SDK SHOULD refuse to accept a string with newline currently, even before attempting to hand it over to a transport.

Anyway, I think we all agree that this is broken, and again I'd like to keep this discussion out of this issue, and I am more than happy to include String here once that is fixed 😉 

> but with a media type that requires intimate knowledge of CloudEvents

Yes, for good reason, because (given #471) you need intimate knowledge of CloudEvents!

```
{
    [...]
    "data" : {
          "distance" : 1
    }
}

{
    [...]
    "data" : {
          "distance" : 1.5
    }
}

{
    [...]
    "data" : {
          "distance" : 1.5
    }
}
```

How is someone without intimate knowledge of CloudEvents going to understand why 1 is a valid CloudEvent, 2 isn't, but 3 is?

> The simplest case therefore got more complicated.

You've picked an example where the JSON happens to be a CloudEvents-Map. Without conscious effort on the side of the producer, this is unlikely to hold. Just checking the Azure EventGrid docs, I can quickly find multiple examples where it doesn't:
* https://docs.microsoft.com/en-us/azure/event-grid/event-schema-azure-maps#event-examples uses floats
* https://docs.microsoft.com/en-us/azure/media-services/latest/media-services-event-schemas?toc=%2fazure%2fevent-grid%2ftoc.json#liveeventingestheartbeat uses boolean
* https://docs.microsoft.com/en-us/azure/event-grid/event-schema-iot-hub#example-event emits `null` instead of dropping "primaryThumbprint"

IMO, the simplest path to adopting CloudEvents is "I have an existing event in JSON, let's make it a CloudEvent". Here, you effectively have to set `datacontenttype` to `application/json`, or you are playing with fire (see examples above).
As @rperelma said, if there is a case where `datacontenttype` should be optional, it is where both envelope and data are JSON.

cneijenhuis:
Closing in favor of #484 - dropping `Map` from `data` is obviously even simpler. I didn't expect that proposal to gain traction, but I'm happy it did 👍

Issue #481, opened 2019-08-09
datacontentencoding is not OPTIONAL


cneijenhuis:
The notational convention states:

> For clarity, when a feature is marked as "OPTIONAL" this means that it is OPTIONAL for the Producer [...] of a message to support that feature. In other words, a producer can choose to include that feature in a message if it wants [...]

The `datacontentencoding` is listed as an OPTIONAL attribute, however producers MUST or MUST NOT set the attribute - in other words, they MUST support the feature (which can mean they have to omit the attribute):

> The attribute MUST be set if the `data` attribute contains string-encoded binary data. Otherwise the attribute MUST NOT be set.

The same will be true for `datacontenttype` if we adopt either #470 or #471 .

My suggestion is to either move them to REQUIRED attributes and clarify that this includes a requirement to NOT set the attribute, or to create a new section for attributes that are producers are REQUIRED to use correctly (incl. omitting them), but consumers have to handle them not being there.

rperelma:
Good point @cneijenhuis I agree with your suggestion. I think # 1 is best, and explain more clearly what REQUIRED means.

duglin:
@jroper @clemensv do you think you could modify the PRs to add text in the "constraints" section to indicate when they are REQUIRED?

cneijenhuis:
@duglin If I'm not mistaken, their PRs already indicate when it is REQUIRED in the constraints. My point is that an OPTIONAL attribute MUST NOT state it is REQUIRED in some circumstances in the constraints section.

A "skimming" reader may skip the optional attributes - it would be quite a surprise that these are not optional, after all...

duglin:
@cneijenhuis if #484 is adopted, does the new header for the "optional attributes" section address your concern?

cneijenhuis:
Mostly, but I think the [Notational Convention](https://github.com/cloudevents/spec/blob/master/spec.md#notational-conventions) section should be changed as well, e.g. to:

```
For clarity, when a feature is marked as "OPTIONAL" this means that it is
OPTIONAL for both the [Producer](#producer) and [Consumer](#consumer) of a
message to support that feature. In other words, a producer can choose to
include that feature in a message if it wants, and a consumer can choose to
support that feature if it wants. A consumer that does not support that feature
will then silently ignore that part of the message. The producer needs to be
prepared for the situation where a consumer ignores that feature. An
[Intermediary](#intermediary) SHOULD forward OPTIONAL attributes.

[...]

PR #484, opened 2019-08-13
tweak how we talk about data and datacontentencoding

duglin:
This might address #482 #471 #470 

### DIFF
--- a/amqp-transport-binding.md
+++ b/amqp-transport-binding.md
@@ -22,7 +22,7 @@ This document is a working draft.
 2. [Use of CloudEvents Attributes](#2-use-of-cloudevents-attributes)
 
 - 2.1. [datacontenttype Attribute](#21-datacontenttype-attribute)
-- 2.2. [data Attribute](#22-data-attribute)
+- 2.2. [data](#22-data)
 
 3. [AMQP Message Mapping](#3-amqp-message-mapping)
 
@@ -70,7 +70,7 @@ In the _structured_ content mode, event metadata attributes and event data are
 placed into the AMQP message's [application data][data] section using an
 [event format](#14-event-formats).
 
-In the _binary_ content mode, the value of the event `data` attribute is placed
+In the _binary_ content mode, the value of the event `data` is placed
 into the AMQP message's [application data][data] section as-is, with the
 `datacontenttype` attribute value declaring its media type; all other event
 attributes are mapped to the AMQP [application-properties][app-properties]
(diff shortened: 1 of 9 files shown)


clemensv:
I've closed #471 since this is ultimately a cleaner solution to the set of problems we're having.

This preserves the simplest case where `data` is just another bit of JSON inside the outer envelope for a structured JSON event without having to add a declaration. 

Even though this PR abandons the abstract `Map` type altogether and we're effectively leaving structured data inside `data` without a more specific definition, we can probably take that risk until implementations turn up real issues, with a little extra tweak. 

Generally, serializers are able to (de-)serialize a generic object graph for the respective runtime. An intermediary can therefore transcode the event if it reads it in one format and writes it in another and uses the respective serialization format's mapping to the runtime that is being used. Some of that might however be lossy.

If I build an intermediary in C# and read a structure based on an event format that knows about timestamps, my resulting graph will contain DateTime. If I serialize that back out as JSON, which doesn't have a timestamp concept, the respective field will be a string, meaning that I lose the type information.

I'm okay with a date turning into a string via intermediaries if I can be sure that the string has a predictable format. I want to avoid the situation where some implementations choose RFC1123 and others choose RFC3339 for dates, specifically in the chaotic world of JSON/JavaScript where RFC1123 dates are still common. 

Therefore, we should augment the JSON event format with a statement that any date and time expressions, anywhere and even inside `data`, MUST follow RFC3339. That's the one simple type I'm really worried about with JSON because the format itself doesn't take a stance.

duglin:
For CloudEvents defined metadata (e.g. our `time` attribute) I agree with you that being very clear about the syntax of the time string, to ensure interop, is very important. Then, as you said, even if it's converted into a string people should still be able to reason about it and get interop even w/o a strongly typed protocol.

I'm less clear on the idea of mandating what's inside of `data` since that's not within our domain to control. I agree with you that if an event is transmitted across multiple transports and/or multiple formats that things can be lossy, but I view that as a problem for the implementation to deal with. I say this because anyone who wants to create one of those converters (e.g. gateways) will have this problem even without CloudEvents in the picture - meaning, we didn't introduce the issue - the converter did by coming into existence. So, it's not our problem to solve.

clemensv:
It may not be out problem to solve, but we could make interop a little easier by taking a stance. How about a SHOULD? :)

duglin:
To be clear... would you want to say that about just JSON or any format that uses structured?

rperelma:
I think this is a great PR, @duglin, thanks for putting it together!

cneijenhuis:
💯 👍 

Probably should have posted this comment here: https://github.com/cloudevents/spec/issues/481#issuecomment-521536897

I still think we should give @jroper some credit, the idea is more-or-less the same as in #457

duglin:
I think all credit goes to @jroper and @clemensv :-)  I'm just trying to repeat the words I've heard from them in a way that's acceptable to people.

spec.md
```diff
@@ -146,33 +145,23 @@ Messages can be delivered through various industry standard protocol (e.g. HTTP,
...
-data type in the encoding or in transport metadata fields.
-
-An attribute value of type `Timestamp` might indeed be routed as a string
-through multiple hops and only materialize as a native runtime/language type at
-the producer and ultimate consumer. The `Timestamp` might also be routed as a
-native transport type and might be mapped to/from the respective
-language/runtime types at the producer and consumer ends, and never materialize
-as a string.
-
+- `Boolean` - a boolean value of "true" or "false".
```
evankanderson: Did you mean to add `Boolean` to the type system in this PR?

duglin: yes I added it because someone may want to define an extension of type boolean. I pointed it out on the call and there was no objection.

amqp-transport-binding.md
```diff
@@ -237,7 +237,7 @@ content-type: application/cloudevents+json; charset=UTF-8
 #### 3.2.2. Event Data Encoding
 
 The chosen [event format](#14-event-formats) defines how all attributes,
```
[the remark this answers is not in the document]
duglin: done

duglin:
fixed comma issue per @evankanderson  - ready for another review/check

cneijenhuis:
I still think we can clarify what `conditional` attributes mean, but it isn't the main point of this PR. I'll submit a follow-on PR for #481 .

In general, LGTM 👍

duglin:
thanks @cneijenhuis merging.

Approved on the 8/15 call.

Issue #483, opened 2019-08-12
String type: "printable Unicode characters" and follow-on problems


cneijenhuis:
Previous discussion started by @evankanderson here: https://github.com/cloudevents/spec/pull/456#issuecomment-507860941 and in the following comments. Also, here: https://github.com/cloudevents/spec/pull/471/files#r312744230 and here: https://github.com/cloudevents/spec/pull/482#issuecomment-520437625

The `String` type currently only allows "printable" Unicode characters.

There are multiple problems:
- As @evankanderson has pointed out, while "printable" is defined for ASCII, it isn't clear what it means in Unicode.
- One may assume that space is not "printable" (but per ASCII, it is)
- We can't use `String` for the `data` attribute. The most obvious reason is that line breaks can't be used, which is an issue for plain text, but also if we want to model JSON, XML or CSV as `String`.

`String` is used in three places:
- As a key for `Map`. The "printable" requirement makes sense (and may be even to weak?)
- As a value in regular attributes. The "printable" requirement makes sense, or we'd have to rework our HTTP transport a bit (as headers are terminated by a line break).
- As a value in `data`. The requirement is too restrictive.

cneijenhuis:
I think there are three paths forward:

1. We keep the requirement, but specify it more clearly as Evan suggested here: https://github.com/cloudevents/spec/pull/456#issuecomment-509433468 JSON, XML, CSV and even multi-line plain text will have to be modeled as `Binary`.
2. We have a separate `String` type for use within `data`.
3. We drop the "printable" requirement completely, and make sure our transports are able to deal with control characters.

duglin:
what about:
4. Stop trying to define the type for `data`. 

If we're in structured mode then the type is implicit based on the format (e.g. json) being used.
If we're in binary mode then it's just a blob of bytes - unless `ce-dataencoding` is defined.

I think this aligns with our "data PRs" too. But, more importantly, I don't think it actually changes what goes on the wire - it's really just a doc change.

Then we don't need to worry about map keys.

I dunno, maybe I'm missing it, but it feels like we're over complicating things.

clemensv:
@duglin That is good thinking. I like that approach. In that case, we just need the datacontentencoding hint in JSON to tell whether the string is meant to be a binary, and the datacontenttype to figure out what's in the binary if it is a binary,   #471 #482

duglin:
@clemensv @cneijenhuis (et al), see what you think:

full set of changes based on my comment above:
[...]

PR #485, opened 2019-08-14
Change definition of string to remove "Printable" and discourage use …

timbray:
…of control characters.

Signed-off-by: Tim Bray <timbray@amazon.com>

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -179,7 +179,10 @@ as a string.
   `Integer` values in this range.
   - String encoding: Integer portion of the JSON Number per
     [RFC 7159, Section 6](https://tools.ietf.org/html/rfc7159#section-6)
-- `String` - Sequence of printable Unicode characters.
+- `String` - Sequence of Unicode characters. "Control" characters in
+  the range U+0000-U+001F, except for U+000A LINE FEED and U+000D 
+  CARRIAGE RETURN, SHOULD NOT be included, since they have no agreed-on
+  meaning and predictably cause interoperability problems.
 - `Binary` - Sequence of bytes.
   - String encoding: Base64 encoding per
     [RFC4648](https://tools.ietf.org/html/rfc4648).


timbray:
BTW I'd be happy to change SHOULD NOT to MUST NOT.

clemensv:
Great.

timbray:
Oh yeah, U+0009, you're right.

evankanderson:
This is definitely an improvement, thanks!

A previously-ambiguous example is: 🏃‍♀️, which seems handled pretty clearly. That glyph is:

```
U+1F3C3 : RUNNER {running, marathon, track and fields}
U+200D : ZERO WIDTH JOINER [ZWJ]
U+2640 : FEMALE SIGN {Venus; alchemical symbol for copper}
U+FE0F : VARIATION SELECTOR-16 [VS16] {emoji variation selector}
```

It's not clear that either ZWJ or VS16 is a "printable character", even though the resulting glyph is printable. So, thanks for writing this up more precisely!

Nits incoming:

* I'm wondering if we want to define this either in terms of [unicode character classes](https://unicodebook.readthedocs.io/unicode.html#categories) or some other way that will adjust to Unicode as it changes over time.
* Alternatively, we should support all the unicode characters, even ones that don't make sense in this context like \<DEL\>: "AB". It looks like JSON describes how to encode "problematic" Unicode characters in values; perhaps each of the transport encodings should explicitly do the same?

Nits:
* Why are we allowing U+000A and U+000D? At least for HTTP, it seems like these would break header parsing particularly badly (though we probably need to update the HTTP spec to describe how to encode non-ASCII String attribute values, so that may be separate work).
* I think we might want to exclude at least all characters in the `Cc` category, which includes U+0000 through U+001F, but also U+007F (\<DEL\>) through U+009F:

   https://www.compart.com/en/unicode/category/Cc

   (As further evidence, the start/end selected area and other device control characters don't seem to have a good use as event metadata...)

   Similarly, there are a set of reserved "noncharacters" which are only intended for internal use which we might want to exclude: http://www.unicode.org/faq/private_use.html#nonchar4

cneijenhuis:
Thanks for addressing #483 👍 

Assuming we adopt #484 and don't enforce these rules on `data`, I agree with @evankanderson that we probably shouldn't allow multi-line strings because of the HTTP headers.

timbray:
> * I'm wondering if we want to define this either in terms of [unicode character classes](https://unicodebook.readthedocs.io/unicode.html#categories) or some other way that will adjust to Unicode as it changes over time.

That would certainly be fairly conventional. A variety of protocols exclude certain character classes.  I'm particularly in favor of excluding C0 controls for a variety of reasons.

> * Alternatively, we should support all the unicode characters, even ones that don't make sense in this context like <DEL>: "A�B". It looks like JSON describes how to encode "problematic" Unicode characters in values; perhaps each of the transport encodings should explicitly do the same?

I think we should stay away from encoding and keep our discussion strictly in terms of Unicode characters. Every transport I know of has well-established practices for turning them into bits on the wire and we shouldn't get in the way.

> * Why are we allowing U+000A and U+000D? At least for HTTP, it seems like these would break header parsing particularly badly (though we probably need to update the HTTP spec to describe how to encode non-ASCII String attribute values, so that may be separate work).

Well, in the data payload you obviously need them.  Other than that I can't see why.

> * I think we might want to exclude at least all characters in the `Cc` category, which includes U+0000 through U+001F, but also U+007F (<DEL>) through U+009F:
>   https://www.compart.com/en/unicode/category/Cc
>   (As further evidence, the start/end selected area and other device control characters don't seem to have a good use as event metadata...)
>   Similarly, there are a set of reserved "noncharacters" which are only intended for internal use which we might want to exclude: http://www.unicode.org/faq/private_use.html#nonchar4

I'm in general in favor of excluding trash characters.

duglin:
> BTW I'd be happy to change SHOULD NOT to MUST NOT.

Just a personal preference, but I'd prefer SHOULD NOT. I'm not a fan of being too parental. In general I think people who care about interop will "do the right thing" or people will stop using them. But because there are times when people need to do things that are not best for interop, but are necessary for their case, I like to give people the option to do so.  That's why I think SHOULD NOT in this case is better - gives a strong best-practice but also an "out".

evankanderson:
HTTP headers are defined in US-ASCII, so I think the HTTP binary transport needs to define a way to take a unicode string value and map it to an ASCII string:

spec.md
```diff
@@ -179,10 +179,17 @@ as a string.
   `Integer` values in this range.
   - String encoding: Integer portion of the JSON Number per
     [RFC 7159, Section 6](https://tools.ietf.org/html/rfc7159#section-6)
-- `String` - Sequence of Unicode characters. "Control" characters in
-  the range U+0000-U+001F, except for U+000A LINE FEED and U+000D 
-  CARRIAGE RETURN, SHOULD NOT be included, since they have no agreed-on
-  meaning and predictably cause interoperability problems.
+- `String` - Sequence of allowable Unicode characters. The following characters
+  are disallowed:
+  - the "control characters" in the ranges U+0000-U+001F and U+007F-009F (both
```
cneijenhuis: Either `U+0000-U+001F and U+007F-U+009F` or `U+0000-001F and U+007F-009F` 😉

spec.md
```diff
@@ -179,10 +179,17 @@ as a string.
...
+	ranges inclusive), since most have no agreed-on meaning, and some, such as
+	U+000A (newline), are not usable in contexts such as HTTP headers.
```
cneijenhuis: Indentation (both lines should start with 4 spaces)

spec.md
timbray: I had to look at the comment for an endless time to spot the fix but I did eventually …

duglin:
Approved on the 8/22 call.

However, @timbray there's a merge conflict - can you resolve that? Then I can merge it.

PR #488, opened 2019-08-16
Clarifications on the HTTP transport binding

evankanderson:
Update based on feedback from #456 

1. Define Context-Encoding header mapping based on `datacontentencoding`

2. More clearly define percent-encoding (and particularly percent-decoding) rules.

### DIFF
--- a/http-transport-binding.md
+++ b/http-transport-binding.md
@@ -69,10 +69,9 @@ This specification defines three content modes for transferring events:
 _binary_, _structured_ and _batched_. Every compliant implementation SHOULD
 support the _structured_ and _binary_ modes.
 
-In the _binary_ content mode, the value of the event `data` is placed
-into the HTTP request or response body as-is, with the `datacontenttype`
-attribute value declaring its media type; all other event attributes are mapped
-to HTTP headers.
+In the _binary_ content mode, the value of the event `data` is placed into the
+HTTP request or response body as-is, with the `datacontenttype` attribute value
+declaring its media type; all other event attributes are mapped to HTTP headers.
 
 In the _structured_ content mode, event metadata attributes and event data are
 placed into the HTTP request or response body using an
@@ -120,17 +119,17 @@ compliant media-type expression.
 
 ### 2.2. data
 
-`data` is assumed to contain opaque application data that is
-encoded as declared by the `datacontenttype` attribute.
+`data` is assumed to contain opaque application data that is encoded as declared
+by the `datacontenttype` attribute.
 
 An application is free to hold the information in any in-memory representation
 of its choosing, but as the value is transposed into HTTP as defined in this
-specification, the assumption is that the `data` value is made
-available as a sequence of bytes.
+specification, the assumption is that the `data` value is made available as a
+sequence of bytes.
 
 For instance, if the declared `datacontenttype` is
-`application/json;charset=utf-8`, the expectation is that the `data`
-value is made available as [UTF-8][rfc3629] encoded JSON text to HTTP.
+`application/json;charset=utf-8`, the expectation is that the `data` value is
+made available as [UTF-8][rfc3629] encoded JSON text to HTTP.
 
 ## 3. HTTP Message Mapping
 
@@ -164,18 +163,11 @@ For the _binary_ mode, the HTTP `Content-Type` header value corresponds to
 (should be populated from or written to) the CloudEvents `datacontenttype`
 attribute.
 
-#### 3.1.2. HTTP Content-Encoding
+#### 3.1.2. Event Data Encoding
 
-For the _binary_ mode, the HTTP `Content-Encoding` header value corresponds to
-(should be populated from or written to) the CloudEvents `datacontentencoding`
-attribute.
+The [`data`](#22-data) byte-sequence is used as the HTTP message body.
 
-#### 3.1.3. Event Data Encoding
-
-The [`data`](#22-data) byte-sequence is used as the HTTP
-message body.
-
-#### 3.1.4. Metadata Headers
+#### 3.1.3. Metadata Headers
 
 All [CloudEvents][ce] attributes with exception of `datacontenttype`,
 `datacontentencoding`, and `data` MUST be individually mapped to and from
@@ -191,7 +183,7 @@ any revision of such a specification, MUST also define explicit mapping rules
 for all other transport bindings that are part of the CloudEvents core at the
 time of the submission or revision.
 
-##### 3.1.4.1. HTTP Header Names
+##### 3.1.3.1. HTTP Header Names
 
 Except for attributes
 [explicitly handled in this specification](#2-use-of-cloudevents-attributes),
@@ -207,7 +199,7 @@ Examples:
 Note: per the [HTTP](https://tools.ietf.org/html/rfc7230#section-3.2)
 specification, header names are case-insensitive.
 
-##### 3.1.4.2. HTTP Header Values
+##### 3.1.3.2. HTTP Header Values
 
 The value for each HTTP header is constructed from the respective attribute
 type's [canonical string representation][ce-types].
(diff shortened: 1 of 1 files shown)


http-transport-binding.md
```diff
@@ -162,12 +162,17 @@ efficient transfer and without transcoding effort.
 For the _binary_ mode, the HTTP `Content-Type` value maps directly to the
 CloudEvents `datacontenttype` attribute.
 
-#### 3.1.2. Event Data Encoding
+#### 3.1.2. HTTP Content-Encoding
+
+For the _binary_ mode, the HTTP `Content-Encoding` value maps directly to the
```
rperelma: Don't understand `maps directly`. Is it:
a) Should be identical to
b) Supersedes
c) If one is present, the other is not needed

evankanderson: This language is copied from 3.1.1.

I believe that it is trying to say "If you have a `datacontentencoding` attribute in the CloudEvent, set the `Content-Encoding` header in HTTP. If you have a `Content-Encoding` header in HTTP, set the `datacontentencoding` attribute in the CloudEvent."

This is about mapping an abstract CloudEvent to a specific transport in a bidirectional manner. This means that you either:

1. Have a CloudEvent and want an HTTP rendering
2. Have an HTTP message and want to derive a CloudEvent

In your ontology, I believe this is closest to (a), but (b) and (c) don't make sense to me as options, because you can't have both at once.

evankanderson: Fixed the exception on line 178 to include `datacontentencoding`, thanks for helping me notice.

rperelma: Your explanation is so much better than the original obtuse "maps to". Suggest adopting it!

evankanderson: Any other opinions on this?

I found "maps to" fairly intuitive and much shorter, but if this is a cause for confusion, I'm happy to spell it out further.

duglin: I'm ok either way... but rather than having to spell it out for every attribute, perhaps something like: ... the HTTP xxx header's value is populated from the CloudEvents `xxx` attribute.

Although, that sure sounds like "mapping"  :-)

evankanderson: Sold, I'll update this.

evankanderson: Ugh, it actually is a bidirectional mapping. I'll try "corresponds to".

evankanderson: Updated.

clemensv:
Based on the discussion in the last call, @duglin asked me to add a note here, @evankanderson 

a) The PR #492 obsoletes the Content-Encoding discussion because it eliminates the attribute. The attribute was never meant to map to Content-Encoding. An HTTP implementation can nevertheless still use encodings like compression, but that's orthogonal to CloudEvents.

b)  For strings, we need a hybrid of your clarification and the text it replaces. Tim's clarification for the String type in the core spec now makes it clear that we must be able to handle Unicode. That means the step "that the string MUST first be encoded according to UTF-8" from the prior text is indeed vital, but the following step in that text is unclear. The result of UTF-8 encoding is a sequence of characters of which the ASCII7 characters are unchanged and should show up as such (a pure plain ASCII string should just be plain ASCII in effect), but all further characters are percent encoded. That composes with your rule about the "token" set but we need to get both under one roof.

evankanderson:
Thanks for pointing out #490. I'll try to rework the text in 3.1.4.2 (will be 3.1.3.2 after removing `datacontentencoding`) to square the circle between what we allow in `String` and what's allowed in an HTTP header.

For example, HTTP headers dictate that leading and trailing whitespace are permitted around the header value -- [section 3.2](https://tools.ietf.org/html/rfc7230#section-3.2); similarly, multiple header values for the same field name should generally be combined into a single line using the `,` character -- [section 3.2.2](https://tools.ietf.org/html/rfc7230#section-3.2.6).

evankanderson:
I think I got what Clemens was suggesting; I suggested a two step encoding process:

1. Convert to ASCII using percent-encoding
2. Convert to HTTP header using header escaping

I also ran the prettier.io markdown formatter and rationalized some of the header formatting and end-matter links.

evankanderson:
This should be compatible with #492 now.

http-transport-binding.md
```diff
@@ -160,19 +159,19 @@ efficient transfer and without transcoding effort.
 
 #### 3.1.1. HTTP Content-Type
 
-For the _binary_ mode, the HTTP `Content-Type` value maps directly to the
-CloudEvents `datacontenttype` attribute.
+For the _binary_ mode, the HTTP `Content-Type` header value corresponds to
+(should be populated from or written to) the CloudEvents `datacontenttype`
```
duglin: s/should/MUST/

http-transport-binding.md
```diff
@@ -200,28 +199,29 @@ Examples:
...
+before applying the header encoding rules described in [RFC7230, section
+3.2.6][rfc7230-section-3-2s6].
 
-Non-printable ASCII characters and non-ASCII characters MUST first be encoded
-according to UTF-8, and then each octet of the corresponding UTF-8 sequence MUST
-be percent-encoded to be represented as HTTP header characters, in compliance
-with [RFC7230, sections 3, 3.2, 3.2.6][rfc7230-section-3]. The rules for
-encoding of the percent character ('%') apply as defined in [RFC 3986 Section
-2.4.][rfc3986-section-2-4].
+When decoding an HTTP message into a Cloud Event, these rules must be applied in
```
duglin: s/must/MUST/

http-transport-binding.md
```diff
@@ -200,28 +199,29 @@ Examples:
...
+reverse -- [RFC7230, section 3.2.6][rfc7230-section-3-2-6] decoding to an ASCII
+string, and then a **single round** of percent-decoding as described in
+[RFC3986, section 2.4][rfc3986-section-2-4] to produce a valid UTF-8 String.
+(Note that applying percent-decoding an incorrect number of times may result in
```
duglin: s/may/might/

evankanderson:
Updated, this got lost between `git commit` and `git push`.

http-transport-binding.md
```diff
@@ -200,28 +199,29 @@ Examples:
...
-and per [RFC7230 Section 3][rfc7230-section-3], HTTP headers MUST only use
+and per [RFC7230, section 3][rfc7230-section-3], HTTP headers MUST only use
 printable characters from the US-ASCII character set, and are terminated by a
-CRLF sequence.
+CRLF sequence with optional whitespace around the header value.
 
-Therefore, and analog to the encoding rules for Universal character set host
-names in URIs [RFC3986 3.2.2][rfc3986], the string value MUST be further encoded
-as follows:
+String values which contain Unicode characters outside the ASCII range MUST be
```
duglin: we agreed on the call yesterday to remove the `which contain Unicode characters outside the ASCII range` text.

http-transport-binding.md
```diff
@@ -200,28 +199,29 @@ Examples:
...
 
 The value for each HTTP header is constructed from the respective attribute
 type's [canonical string representation][ce-types].
 
 Some CloudEvents metadata attributes can contain arbitrary UTF-8 string content,
-and per [RFC7230 Section 3][rfc7230-section-3], HTTP headers MUST only use
+and per [RFC7230, section 3][rfc7230-section-3], HTTP headers MUST only use
 printable characters from the US-ASCII character set, and are terminated by a
-CRLF sequence.
+CRLF sequence with optional whitespace around the header value.
```
duglin: s/optional/OPTIONAL/

duglin:
We approved this on the 9/5 call with the agreement to remove "which contain Unicode characters outside the ASCII range" from line 212

duglin:
@evankanderson thanks!!  Merging per our approval on the call yesterday.

Issue #486, opened 2019-08-15
Should datacontentencoding list any transformations applied to the content wrapped by the data property?


willfranco:
Proposal:

The datacontentencoding should list all the transformations that were applied to the content that is wrapped by the data property.

For example: 

 In the case that we're compressing and encrypting the content wrapped by the data property, we should be listing in the datacontentencoding property such transformations in the order they were applied to the data.

{
...
"datacontentencoding" : "gzip; aesgcm"
...
}

Clients will be able to reconstitute the data by applying the same transformations in the reverse order.

Another nice change would be to change the datacontentencoding from a String to a JSON array to simplified its parsing.

"datacontentencoding" : "gzip; aesgcm" vs "datacontentencoding" : ["gzip", aesgcm"]

Thanks.

duglin:
@clemensv any thoughts on this?

evankanderson:
I think https://github.com/cloudevents/spec/pull/484 addresses this?

duglin:
Closed due to https://github.com/cloudevents/spec/pull/492 - if incorrect please let me know

Issue #477, opened 2019-08-05
Should datacontentencoding maps to Content-Transfer-Encoding ?


fabiojose:
Should `datacontentencoding` maps to `Content-Transfer-Encoding` in the HTTP Transport Binding?

There is no mention about that like we have with `datacontenttype`, which maps to `Content-Type`. It is maybe an issue specifically with the binary content mode.

- datacontentencoding: https://github.com/cloudevents/spec/blob/master/spec.md#datacontentencoding
- HTTP Transport Binding: https://github.com/cloudevents/spec/blob/master/http-transport-binding.md#13-content-modes

duglin:
I suspect this might be slightly different though because in the switch from 'binary' to 'structured' mode your choice of encoding `data` may need to change drastically based on what your structured format allows in the `data` attribute's value character set. And visa-verse.

E.g. when going from structured to binary you may actually go from base64 encoded to no-encoding at all.

I wonder if it would make sense to say:
- `datacontentencoding` and HTTP's `Content-Encoding` headers serve similar purposes
- whether they are the same values in both structured and binary modes will vary depending on the structure serialization chosen

Dunno. What do other people think? Could just be more of an FYI kind of thing to help people understand why `datacontentencoding` exists

duglin:
hmm is it Content-Encoding or Content-Transfer-Encoding?

fabiojose:
Like we have in the spec, `datacontentenconding` is based on the `Content-Transfer-Encoding` spec. 

Yes, is `Content-Transfer-Encoding` what I am asking.

Here: https://tools.ietf.org/html/rfc2045#section-6.8

clemensv:
While the fields correspond directly, I wouldn't prescribe a mapping because it's more than a little weird to carry binary as Base64 where you can carry Binary raw. For an SMTP transport, I would prescribe a mapping, but not for any of the transports we've got.

fabiojose:
Just to be clear for me, consider bellow.

If we have this event:

```json
{
  "specversion": "0.3",
  "id": "45c83279-c8a1-4db6-a703-b3768db93887",
  "type": "com.github.pull.create",
  "source": "https://github.com/cloudevents/spec/pull/123",
  "time": "2019-06-21T17:31:00.000Z",
  "datacontenttype": "application/json",
  "datacontentencoding": "base64",
  "data": "eyJtdWNoIjoid293In0="
}
```

And map that event to binary, we end up with this? `ce-datacontentencoding:base64`

[...]

PR #491, opened 2019-08-27
Changed datacontentencoding to dataencoding

clemensv:
Per discussion on the call last week.

This PR renames `datacontentencoding` to `dataencoding` and tightens the rules around it since there have been several misunderstandings about what the attribute is meant to help with, see #486 #481 #477 

The attribute is ONLY for structured mode and ONLY for when a value cannot be correctly expressed in the chosen event format's type system and needs further encoding. The specific concern is making binary fit into JSON.

The attribute does not map to transport-level content encoding concerns like compression or signatures or encryption or the like, specifically not to Content-Transfer-Encoding fields of SMTP or the Content-Encoding or Transfer-Encoding fields of HTTP. This is a CloudEvents specific construct for the aforementioned purpose. That does not at all preclude that one uses message-level compression in HTTP for transferring a CloudEvent; it's simply an orthogonal concern.

### DIFF
--- a/adapters/aws-s3.md
+++ b/adapters/aws-s3.md
@@ -16,7 +16,7 @@ same pattern as described in the following table:
 | `source`              | "eventSource" value + `.` + "awsRegion" value + `.` + "s3.buckets.name" value  |
 | `specversion`         | `0.4-wip`                                       |
 | `type`                | `com.amazonaws.s3.` + "eventName" value         |
-| `datacontentencoding` | Omit                                            |
+| `dataencoding` | Omit                                            |
 | `datacontenttype`     | S3 event type (e.g. `application/json`)         |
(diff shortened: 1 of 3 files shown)


clemensv:
@evankanderson I agree on the motivation to push that off into the JSON spec based on what we have right now. Your argument that XML and YAML might chose different hints also makes sense. 

The lack of such annotations in JSON makes it a little iffy to express the encoding mode as an "encoding-private" attribute, though.

An alternate proposal for JSON I can think of is to simply call, in JSON encoding, a field that holds regular string data or inline JSON (which is factually the same thing) ´data´ and call the alternate  field that holds base64-encoded data ´base64data´ - and kill dataencoding altogether.

PR #492, opened 2019-08-29
remove dataencoding and intro data_base64 in JSON 

clemensv:
Alternate to #491

### DIFF
--- a/json-format.md
+++ b/json-format.md
@@ -89,19 +89,9 @@ inference using the rules from the mapping table, whereby the only potentially
 ambiguous JSON data type is `string`. The value is compatible with the
 respective CloudEvents type when the mapping rules are fulfilled.
 
-### 2.3. Mapping Data
+### 2.3. Examples
 
-If an implementation determines that the actual type of `data` is a `String`,
-the value MUST be represented as [JSON string][json-string] expression and the
-member name inside the JSON object MUST be `data`. For `Binary`, the value MUST
-be represented as a [JSON string][json-string] expression containing the 
-[Base64][base64] encoded binary value, and the member name inside the JSON 
-object MUST be `data_base64`.
-
-
-### 2.4. Examples
-
-The following table shows exemplary mappings:
+The following table shows exemplary attribute mappings:
 
 | CloudEvents     | Type          | Exemplary JSON Value            |
 | --------------- | ------------- | ------------------------------- |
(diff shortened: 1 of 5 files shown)


json-format.md
```diff
@@ -92,9 +92,12 @@ respective CloudEvents type when the mapping rules are fulfilled.
...
 
 If an implementation determines that the actual type of `data` is a `String`,
-the value MUST be represented as [JSON string][json-string] expression; for
-`Binary`, the value MUST represented as [JSON string][json-string] expression
-containing the [Base64][base64] encoded binary value.
+the value MUST be represented as [JSON string][json-string] expression and the
+member name inside the JSON object MUST be `data`. For `Binary`, the value MUST
+be represented as a [JSON string][json-string] expression containing the 
+[Base64][base64] encoded binary value, and the member name inside the JSON 
+object MUST be `data_base64`.
```
duglin: While it's implied, it doesn't actually ban someone from having both `data` and `data_base64` in the JSON. Can we add something like:
```
Note, `data` and `data_base64` MUST NOT both appear at the same time within a CloudEvent serialization.
```
Someone may try to be "smart" and fill in both (when possible, like when it's a string) and that's a situation I think we need to avoid.

Also, could someone misinterpret the `Binary` usage here? I believe you mean that the data type of `data` is binary and you're not referring to the binary serialization of the CE.  Perhaps:
- - -
For cases where `data` is a `Binary` value, it MUST be represented ....
- - -

json-format.md
```diff
@@ -92,9 +92,12 @@ respective CloudEvents type when the mapping rules are fulfilled.
 ### 2.3. Mapping Data
```
evankanderson: Terrible nit, but I think this should be in section 3 now, because we've removed `data` from the attributes, and section 2 is "Attributes".

evankanderson: That would also put this text closer to the text in section 3.1 'Special Handling of "data"'.

evankanderson:
Does this also need to update [`spec.json`](https://github.com/cloudevents/spec/blob/master/spec.json)?

clemensv:
I'll address @duglin and @evankanderson concerns tomorrow. I agree with all comments made.

clemensv:
I consolidated the `data` handling into section 3 of the JSON spec and also dropped rules we no longer need because we gave up on having "Map" as an option for data. What remains is the difference between binary and everything else. Strings are just another JSON value and need no special treatment. I also stressed that `data` and `data_base64` are mutually exclusive.

duglin:
LGTM

duglin:
Approved on the Sept 5 call

PR #493, opened 2019-08-29
Cleanup: Map and add URI in formats; re-add "Type System" heading

cneijenhuis:
Some cleanup in our event formats:
* Removed `Map`
* Added `URI`

All event formats reference back to the `Type System` (e.g. https://github.com/cloudevents/spec/blob/master/json-format.md#22-type-system-mapping also includes a relative link).
Therefore I think it is worth adding a heading for it that can be linked back to.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -155,7 +155,7 @@ These attributes, while descriptive of the event, are designed such that they
 can be serialized independent of the event data. This allows for them to be
 inspected at the destination without having to deserialize the event data.
 
-#### Type System
+### Type System
 
 The following abstract data types are available for use in attributes. Each of
 these types MAY be represented differently by different event formats and in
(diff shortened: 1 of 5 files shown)


spec.md
```diff
@@ -155,6 +155,8 @@ These attributes, while descriptive of the event, are designed such that they
 can be serialized independent of the event data. This allows for them to be
 inspected at the destination without having to deserialize the event data.
 
+#### Type System
```
duglin: @cneijenhuis let me know when you've changed this to `###` and then I'll merge

cneijenhuis: @duglin Done :)

duglin:
@cneijenhuis thanks!

Approved on the 8/29 call.

PR #495, opened 2019-08-30
Reword SDK.md to be more general.

n3wscott:


clemensv:
LGTM - discuss in SDK call

duglin:
Found a couple of typos but we can fix those later I think this is a good next step in its evolution. 
LGTM

All, I’d like to see if we can approve this on the normal call tomorrow so please look it over. There haven’t been any changes for a few days.

duglin:
Approved on the 10/10 call

PR #497, opened 2019-08-30
Update AVRO mappings to match current event schema and type system.

evankanderson:
It looks like we didn't update the AVRO format to actually include a place to store `data` when we graduated the payload from a special attribute to a header + payload model.

### DIFF
--- a/avro-format.md
+++ b/avro-format.md
@@ -139,7 +139,7 @@ described by the [CloudEvent Avro Schema](./spec.avsc):
 ## 3 Data
 
 Before encoding, the AVRO serializer MUST first determine the runtime data type
-of the content. This may be determined by examining the data for invalid UTF-8
+of the content. This can be determined by examining the data for invalid UTF-8
 sequences or by consulting the `datacontenttype` attribute.
 
 If the implementation determines that the type of the data is binary, the value
--- a/avro-format.md
+++ b/avro-format.md
@@ -2,8 +2,8 @@
 
 ## Abstract
 
-The Avro Format for CloudEvents defines how events attributes are expressed
-in the [Avro 1.9.0 Specification][avro-spec].
+The Avro Format for CloudEvents defines how events attributes are expressed in
+the [Avro 1.9.0 Specification][avro-spec].
 
 ## Status of this document
 
@@ -13,7 +13,8 @@ This document is a working draft.
 
 1. [Introduction](#1-introduction)
 2. [Attributes](#2-attributes)
-3. [Examples](#3-examples)
+3. [Data](#3-data)
+4. [Examples](#4-examples)
 
 ## 1. Introduction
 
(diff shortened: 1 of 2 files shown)


avro-format.md
```diff
@@ -47,108 +48,81 @@ The CloudEvents type system MUST be mapped to Avro types as follows.
...
+[union][avro-unions].
 
 Example:
 
 ```json
-[
-  "null",
-  "string"
-]
+["null", "string"]
```
evankanderson: I can undo this JSON auto-fomatting if preferred.

avro-format.md
```diff
@@ -47,108 +48,81 @@ The CloudEvents type system MUST be mapped to Avro types as follows.
...
+      "type": "bytes"
     }
   ]
 }
 ```
 
-## 3 Examples
+## 3 Data
+
+The `data` of the CloudEvent should be encoded in a top-level field called
```
duglin: s/should/MUST/

avro-format.md
```diff
@@ -47,108 +48,81 @@ The CloudEvents type system MUST be mapped to Avro types as follows.
...
+`data` of type `bytes`. No additional encoding should be done, regardless of the
```
duglin: s/should/will/. 
Or if you want it to be normative; Additional encoding MUST NOT be done, regardless ....

avro-format.md
```diff
@@ -47,108 +48,81 @@ The CloudEvents type system MUST be mapped to Avro types as follows.
...
     {
-      "name":"attribute",
-      "type":{
-        "type":"map",
-        "values":[
-          "null",
-          "int",
-          "string",
-          "bytes",
-          "CloudEvent"
```
duglin: Not familiar with Avro... is the "CloudEvent" above replaced by the stuff on line 95? (the data/bytes thingy)

evankanderson: The "CloudEvent" was a recursive allowed value of the same object, to support `Map` values.

Without `Map`, these structures don't need to be recursive in their definition, which makes me much more comfortable.

The actual diff here is `+boolean -CloudEvent`

clemensv:
Even though we dropped 'Map' as a "portable" cross-encoding construct, JSON retains the ability to carry structured data and Avro should, too. The recursive schema (which is required for the encoder to work) enabled that. The Avro schema is not a strict schema in the sense that it accurately tracks all constraints of CloudEvents; it just makes the encoder work. 

With the change you are now limiting "data" in Avro to what "data_base64" does in JSON.  

In effect, I should be able to read a JSON event into an in-memory structure and then write that structure back out into Avro.

With limiting "data" to binary, you're now forcing me to pick an encoding for that structured data, and if that were Avro, I would have to make my own schema and if that were JSON then I would wonder why I am using Avro to begin with.

The "normal" case remains a self-contained event with data containing structured data and that needs to work in all encodings.

evankanderson:
I updated the `.avsc` file to define a union type of JSON Value and `bytes`, which is very unpretty because:

> [Unions may not immediately contain other unions](https://avro.apache.org/docs/current/spec.html#Unions)

I'm also no Avro expert, so there may be a better way. Feel free to send a pull request with corrections against this branch if there are better ways to do this: https://github.com/evankanderson/spec/tree/hadoop-doop-doop

evankanderson:
(Tested via python and:

```python
import avro.schema
avro.schema.parse(open("spec.avsc", "rb").read())
```
)

evankanderson:
Whoops, I left in my virtualenv... let me back that out.

avro-format.md
```diff
@@ -47,108 +48,125 @@ The CloudEvents type system MUST be mapped to Avro types as follows.
...
     }
   ]
 }
 ```
 
-## 3 Examples
+## 3 Data
+
+Before encoding, the AVRO serializer MUST first determine the runtime data type
+of the content. This may be determined by examining the data for invalid UTF-8
```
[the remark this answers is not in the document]
evankanderson: I changed this to "can" rather than "might".

duglin:
Approved on the 9/12 - just waited for the RFC2119 keyword fix.
Thanks @evankanderson 
Merging.

Issue #435, opened 2019-06-06
AWS Support of CloudEvents


timbray:
Hi. We're working on a bunch of event-related stuff at AWS with a view to releases in 2019. It has been internally proposed that we "Support CloudEvents" and this seems to me like a good idea. No promises, but my opinion is that if we could find a plausible way to do it, there is a high likelihood that we would.

**Important**  This is not an official communication, this is Tim looking for options I can use internally.

## Background

We already have an AWS Event envelope that is widely used across AWS chiefly through the "CloudWatch Events" service. There's no formal schema or anything but it's documented [here](https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/CloudWatchEventsandEventPatterns.html).

The required top-level fields are:

* `version`: (Corresponds to CloudEvents `specversion`)
* `detail-type`: (`type`)
* `source`: (`source`)
* `id`: (`id`)
* `time`: (`time`)
* `detail`: (`data`)

Fields in AWS Events not in CloudEvents: `region`, `account`, `resources`.

Fields in CloudEvents not in AWS Events: `datacontenttype` (So far we're all-JSON all the time.)

## Problems

1. There are already huge numbers of AWS Events per second flowing through our system and there's no API, the only documentation is of the bits on the wire. And, there are a huge number of AWS customers who are making use of this service. Therefore there is no reasonable prospect of us changing field names.
1. CloudEvents isn't finished, but we want to ship production software in 2019. In theory, the working group could do a wholesale redesign of the attributes at any time. Can any organization plausibly claim to "Support CloudEvents" at this stage in history? It's not obvious to me how.

## Appeal for input

Before I dive any deeper, I'd welcome general comments along the lines of "No, just wait till CloudEvents is finished", or "That is a really bad idea because X" for some value of X, or "You can effectively support CloudEvents by doing A, B, and C." Thanks in advance.

## Some notes about possibilities

1. CloudEvents could consider freezing the field names. I.e. adopt some sort of policy that whereas new attributes may be added, there's a commitment not to change the names of the ones that exist.
 1.1. And if you wanted to try making AWS an offer it couldn't refuse, consider changing the field names to match ours. We could add support for the extra CloudEvents fields and our extras could be extensions. There are some semantic incompatibilities (e.g. our "source" field looks like a relative URI reference) but probably there are workarounds. (OK, long shot).
[...]

PR #499, opened 2019-09-04
Amazon SNS transport binding

to-masz:
Regarding discussion in #435 *AWS Support of CloudEvents* I am proposing here Amazon SNS transport binding specification.

Information for reviewers:
* It is very similar and highly inspired by binary mode of existing Kafka transport binding.
* IMO it is very good approach for handling the CE specification with SNS, ... but 🙈 the most problematic part can be the count limit of message attributes (headers), that is 10. The CE attributes limit is 100. 
* I have some doubts about JSON format of examples. Maybe URI would be better to illustrate REST nature of SNS API, but I find JSON more readable.
* The SQS transport binding should follow and should have the same assumptions. However you can use both technologies separately and we need to figure out how to deal with SNS first.
* I see it as WIP. Feel free to contribute by both edits and discussion!

### DIFF
--- a/sns-transport-binding.md
+++ b/sns-transport-binding.md
@@ -15,7 +15,7 @@ This document is a working draft.
 1. [Introduction](#1-introduction)
 - 1.1. [Conformance](#11-conformance)
 - 1.2. [Relation to SNS](#12-relation-to-sns)
-- 1.3. [Content Modes](#13-content-modes)
+- 1.3. [Message structure](#13-message-structure)
 - 1.4. [Security](#14-security)
 2. [Use of CloudEvents Attributes](#2-use-of-cloudevents-attributes)
 - 2.1. [data](#21-data)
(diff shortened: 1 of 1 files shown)


clemensv:
AWS SNS is a proprietary product and isn't based on a standard that meets our criteria and therefore this binding should reside in an external repo and then be referenced in the https://github.com/cloudevents/spec/blob/master/proprietary-specs.md document, which we created for this purpose.

to-masz:
@clemensv makes sense 👮👍 

@timbray are you able to make it happen? I believe the right place for that would be within your https://github.com/aws organisation. And this is for sure a good step to _«we "Support CloudEvents"»_ 😉

duglin:
If people agree that this doesn't belong in our repo then we can close this PR. However, I'll hold off doing that so we can use this as a reminder for @timbray since there's an open question for him

timbray:
I think I agree that this probably belongs in AWS-land rather than CloudEvents land.

Also, it's not obvious to me that CE should use SNS message attributes at all, I'd also want to think about using the JSON encoding as a message payload.

duglin:
ok - I'm going to close this since there's general agreement that it's not within scope for us, and you guys can work off-line to find the best home for this.

Issue #443, opened 2019-06-08
Multi-tenant system support


lennygran:
This is more like a question, rather an issue. There are plenty of systems required multi-tenancy support. Usually, such systems include `tenantId` or similar attribute as part of the event envelope. The main rationale here is very often the tenant-based rules, such as routing, security, etc. may be applied outside of the event handling, but the intermediaries that have no business understanding of the event body. 

Alternatively, this type of information can be carried via an extension.  

Do you see a value of defining such attribute? If so, what would be your preference - the context attribute or extension?

duglin:
My initial reaction is that it would be an extension, if part of CE at all, because it seems like this might be better put into whatever authN/authZ scheme you're using as part of the transport layer.  Just my 2c

lennygran:
@duglin, yes, I was thinking about an extension. But I wanted to see if others have a different opinion. I would not directly relate it to the authN/Z in particular because this information is often used in routing, scheduling/processing priorities, etc., therefore falling into generic messaging concerns. It also feels that this would be useful for products that are dealing with multi-tenancy, hence would be nice to have it as part of CE for consistency. 

If no objections, I'll create a PR and put it for review.

duglin:
never any objection to PRs  ;-)

PR #502, opened 2019-09-10
Extension to support multi-tenancy

fnobilia:
Original issue: #443 

Most of modern systems support multi-tenancy and therefore it would nice having a way to handle tenancy within the [Cloudevents specification](https://github.com/cloudevents/spec).

In this PR, a new extension has been added to let publisher specifying a tenant identifier if need. The suggested attribute is called `tenantid`. Using an extension to specify the tenant associated to the event empowers to event receiver to apply the required tenant-based rules without having to handle the data field.

Signed-off-by: Francesco Nobilia f.nobilia@gmail.com


cneijenhuis:
Isn't the tenant supposed to be part of the `source`?

The spec calls out `Often this will include information such as [...] the organization publishing the event` and gives an example of `/cloudevents/spec/pull/123`.

Similar, all three adapters are for multi-tenant systems, and the tenant ends up being part of the source: https://github.com/cloudevents/spec/blob/master/adapters.md

I'd prefer to keep the tenant in the `source`. Maybe you can make a PR that clarifies the `source` section instead?

(Minor: In a lot of multi-tenant systems, you have tenants-within-tenants.
E.g. in GitHub, you sign up an org, and that org has multiple repositories. Repos may have different access rights.
E.g. in AWS, again an org signs up, but for e.g. S3, you have buckets which again have different access rights.
Therefore, tenantId may be better modeled as a `URI-reference` to support e.g. `/cloudevents/spec`. Only `cloudevents` isn't sufficient from an access rights perspective, and only `spec` is not unique...)

fnobilia:
@cneijenhuis thx for the comments!

I agree with you using a `URI-reference` would be better approach. Given the fact `URI-reference` can describe hierarchies, they can help to represents `tenants-within-tenants` and address your other comment.

When I was editing the spec, I initially put `URI` and then reverted to `String` because I considered products where tenant identifier is just a UUID and handling this case with a URI is a bit ugly. If we go for the `URI` format, how would you define UUID, maybe applying [RFC 4122](https://blogs.msdn.microsoft.com/dandris/2008/09/19/uuid-uri-schemes/)? For instance, in case of tenant identifier being `7673868d-231e-490d-9c4f-19288e7e668d`, would you suggest to use `urn:uuid:7673868d-231e-490d-9c4f-19288e7e668d` as a value for the tenant identifier attribute?

Regarding the idea of using the `source` field to specify the tenant and discard the `tenant identifier` extension, I am not convinced. Your suggestion perfectly works for situations like the one reported in [spec.md](https://github.com/cloudevents/spec/blob/master/spec.md) where the tenant is part of the source attribute (e.g. `https://example.com/storage/tenant/container`), but what if the source does not contain the tenant at all? For instance, how would you handle cases like `Microsoft event-grid` reported in the [prime.md](https://github.com/cloudevents/spec/blob/master/primer.md#microsoft---event-grid) where the tenant identifier is an attribute distinct from the source? It would be nice to have an extension in Cloudevents to model this case too.

cneijenhuis:
> but what if the source does not contain the tenant at all?

In many cases, the `source` field doesn't map back to an individual field from an existing event, but is constructed from several fields. E.g. https://github.com/cloudevents/spec/blob/master/adapters/github.md#commitcommentevent and https://github.com/cloudevents/spec/blob/master/adapters/aws-s3.md

> how would you handle cases like Microsoft event-grid [...]

I wouldn't at all, because Microsoft has defined the mapping already: https://docs.microsoft.com/en-us/azure/event-grid/cloudevents-schema 😉 (and `tenantId` isn't part of it...)

Ignoring that, I'd include the `tenantId` as the first part of the `source`, e.g. `/{tenantId}/{topic}#{subject}`.

---

This is also how we map our proprietary commercetools events to a CloudEvent.

The proprietary format includes:
```
"projectKey": "tenant",
"resource": {
  "typeId": "product",
  "id": "d1229e6f-2b79-441e-b419-180311e52754"
}
```
which will be mapped to a  CloudEvent `source` of:
`/tenant/products/d1229e6f-2b79-441e-b419-180311e52754`

(The `id` should probably end up in `subject` and not `source` by now, but that wasn't part of CloudEvents yet when I defined the mapping initially. I'll review it for 1.0 😉 The same is probably true for the EventGrid mapping above...)

fnobilia:
Thank you @cneijenhuis and @evankanderson for your comments!

Out of curiosity, how did you solve the *spoofing* issue mentioned by @evankanderson 

> seems like including a (spoofable) tenantid as a CloudEvents attribute that can be sent into the multi-tenant system might not be a desirable property.

for other fields like source?

evankanderson:
One way to solve the spoofing issue is to offer different delivery
endpoints (PubSub topics, queue names, etc) for each tenant. In that case,
rate limiting would be applied based on the queue information, rather than
based on the payload content.

Even within a single cloud, there are cases for forwarding an event from
one tenant to another and then routing that event within the second tenant.
Given that scenario, quota and costs should be accrued to the tenant which
is sending the message, rather than the tenant that originally created the
message.

cneijenhuis:
@fnobilia Sorry, I'm not sure if I get the question correctly.

Trying to make an example: A user `john@doe.com` adds the file `foo.txt` to a S3 bucket called `bar`.
In this case, the `source` would be `aws:s3.us-east-1.bar` and the `subject` `foo.txt`.

The tenant id is `bar`, and part of the `source`. The user is neither part of `source` nor `subject`.

Not sure if that already helps already?

duglin:
@fnobilia is this still needed?

fnobilia:
@duglin I'll close it! Thx to @cneijenhuis and @evankanderson for the comments and the help to better understand the `cloudevents` project.

PR #504, opened 2019-09-11
Remove protobuf format until someone has time to redo it.

evankanderson:
Alas, poor proto... needs update for #492, and I don't have time.


evankanderson:
Note that since proto binary encoding does not contain the keys, either something like the AVRO mapping (see #497), or something like the HTTP transport will be needed for the payload. I would suggest simply encoding the payload as `bytes` and not transcoding JSON structured data into proto specifically.

Ideally, a CL that re-introduced this format would also include [gRPC][https://grpc.io] bindings.

duglin:
Since @erikerikson is AFK for a bit he asked me to complete this in a new PR. See: https://github.com/cloudevents/spec/pull/507

PR #505, opened 2019-09-12
make extensions follow the same pattern as normal attrs

### LINKED ISSUE #498: HTTP Binary spec: possibility of dropping attributes
If you receive an event over HTTP Binary with an unknown extension attribute which has special header mapping rules, it will be dropped.

Currently, we have [Distributed Tracing](https://github.com/cloudevents/spec/blob/master/extensions/distributed-tracing.md) which defines a mapping from the tracing headers to CE attributes and vice-versa.

Imagine a new extension attribute which defines a similar mapping, and a client built before the extension attribute was known or defined. When receiving the attribute over HTTP Binary mode, the attribute will be dropped because the corresponding HTTP header does not begin with the magic string `CE-` (case insensitive).

Depending on the attribute, this data loss might substantially change the processing of the event. As a possibly-bad-idea, imagine a `processby` attribute which maps to the HTTP `If-Unmodified-Since`. If this header were stripped by middleware by accident, it might cause a event to be applied at the destination even if it should not be.

```
[client] --(HTTP)--> [middleware] --(AMQP)--> [processor]
```

duglin:
Fixes #498 

This PR makes CE extensions follow the same rules as all other CE attributes. Meaning:
1 - they MUST be serialized with `ce-` prefixes in transports like binary-http
2 - they MAY have a secondary serialization, but they MUST still have the `ce-` prefixed one too.
3 - if they do, and the values differ, then the `ce-` version is the CE attribute. The other one, if picked-up by the middleware, is 'some other' metadata that it's passing along.

This means that a CE receiver no longer needs to look any place except in one location for ALL CloudEvents data. This also means that ALL CloudEvents extensions will be picked up by ALL receivers.  W/o this PR, it's possible for someone to define a CE extension that maps to an existing HTTP header (like `datacontenttype` does) and because the HTTP header it maps to is a well-known HTTP header, the CE receiver might not realize it's a CE extension and not pass it along to the app. This PR avoids that problem.

In the end, this PR gets us back to our "first principles" that CloudEvents is trying to standardize the location of a well-defined set of metadata - even if it ends up duplicating the data. For example, we've talked about how some of our data would be duplicated from the "business logic"(`data`) and that's ok. So, let's do the same for extensions too. By allowing some attributes (CE defined or extensions) to ONLY have a different serialization (and possibly location), we're violating our own stated goal. But, allowing for a "secondary" serialization allows for us to have a single location for all CE data but then also allow for existing metadata defined by other specs to still be properly filled-in.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -406,8 +406,10 @@ See
 [CloudEvent Attributes Extensions](primer.md#cloudevent-attribute-extensions)
 for additional information concerning the use and definition of extensions.
 
-This specification places no restriction on the type or semantics of the
-extension attributes. Each definition of an extensions SHOULD fully define all
+This specification places no restriction on the semantics of the
+extension attributes, but they MUST use one of the types defined within the
+[type system](#type-system).
+Each definition of an extensions SHOULD fully define all
 aspects of the attribute - e.g. its name, semantic meaning and possible values
 or even to indicate that it places no restrictions on its values. New extension
 definitions SHOULD use a name that is descriptive enough to reduce the chances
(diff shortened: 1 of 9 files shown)


duglin:
ok - I updated all of the docs that I think need to be updated.
Please take a look.  If there are alternatives then please open a complete PR before EOD Tuesday so we can consider it. This is the last open v1.0 issue/PR and we'd like to get to v1.0-rc1 on next week's call!!

duglin:
ping @evankanderson @n3wscott @clemensv @deissnerk @Vlaaaaaaad @cneijenhuis

clemensv:
This is good. 

Why are you asking for the datacontenttype info to flow double for MQTT5, though? That protocol version now does have that info just like HTTP.

clemensv:
LGTM

spec.md
```diff
@@ -419,13 +421,31 @@ extensions that might be of interest.
 Each specification that defines how to serialize a CloudEvent will define how
 extension attributes will appear.
 
+Extensions attribtue MUST be serialized using the same general pattern as all
```
deissnerk: s/Extensions attribtue/Extension attributes

duglin: thanks - fixed

duglin:
Approved on the 9/19 call.

PR #506, opened 2019-09-12
Syntax fixes

duglin:
Evan noticed we have more than one #data header so I tried to fix that.

Also, we still used a complex/map in an example.


duglin:
Perhaps we can approve this one off-line, if people review and LGTM it?

duglin:
Conditionally approved on the 9/12 call. No objections since then, so merging.

PR #507, opened 2019-09-14
remove proto - see issue 504

duglin:


duglin:
This was conditionally approved on the 9/12 call per https://github.com/cloudevents/spec/pull/504

PR #509, opened 2019-09-19
Clarify interoperability of extensions (#508)

### LINKED ISSUE #508: Needs clarification: interoperability of extensions
This is not a spec bug,downgrading to a doc issue. See the explanation at https://github.com/cloudevents/spec/issues/508#issuecomment-533153659
----
Original text:

The current definition of extension attributes makes interoperability essentially impossible:

"This specification places no restriction on the type or semantics of the extension attributes. Each definition of an extensions SHOULD fully define all aspects of the attribute - e.g. its name, semantic meaning and possible values or even to indicate that it places no restrictions on its values."

There is no way for a general purpose cloudevents library to represent such an extension unless it has been customized to be aware of it. There is no way for a transport binding implementation to decode or encode such an extensions unless it has been so customized. There is no way for a generic event bridge to even pass-through such extension values unless it is aware of them.  I'm trying to fix cloudevents/sdk-go and the cloudevents AMQP transport to deal with extensions and AFAIK  it's impossible except for int, bool and string valued extensions.

I suggest that extension attributes should be constrained to use the same type system as normal context attributes.

Even that is problematic for the JSON binding, as there's no way to distinguish time or binary values from strings. For JSON I suggest those extensions be encoded as a JSON object with a type tag:
    {
      "anInt": 1,
      "aString": "foo",
      "aTime": { "time": "1969-03-21T0:00:00Z"},
      "aBinary": { "binary": "base64encodedstuff" }
    } 

I think that would be sufficient to round-trip extension value safely thru any binding that supports the cloudevents type system.

There is also the issue of URI and URI-reference, I think it's OK to encode those as strings in JSON since they are so closely allied to strings, and any endpoint that knows what the extension means will be capable of using a string. It's not 100% faithful to the CE type system but probably workable.
    .

alanconway:
Fixes #508

Signed-off-by: Alan Conway <aconway@redhat.com>

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -150,8 +150,8 @@ protocol, in particular how events are mapped to messages in that protocol.
 ## Context Attributes
 
 Every CloudEvent conforming to this specification MUST include context
-attributes designated as REQUIRED and MAY include one or more OPTIONAL context
-attributes.
+attributes designated as REQUIRED, MAY include one or more OPTIONAL context
+attributes and MAY include one or more extension attributes.
 
 These attributes, while descriptive of the event, are designed such that they
 can be serialized independent of the event data. This allows for them to be
@@ -210,9 +210,8 @@ string-encoding for each type that MUST be supported by all implementations.
 - `Timestamp` - Date and time expression using the Gregorian Calendar.
   - String encoding: [RFC 3339](https://tools.ietf.org/html/rfc3339).
 
-All context attributes MUST be of scalar type (e.g. string, integer) that have a
-string-encoding defined. They MUST NOT be of complex type (e.g. structures,
-map).
+All context attribute values MUST be of one of the types listed above.
+Attribute values MAY be presented as native types or canonical strings.
 
 A strongly-typed programming model that represents a CloudEvent or any extension
 MUST be able to convert from and to the canonical string-encoding to the
@@ -235,7 +234,7 @@ language/runtime types at the producer and consumer ends, and never materialize
 as a string.
 
 The choice of serialization mechanism will determine how the context attributes
-and the event data will be materialized. For example, in the case of a JSON
+and the event data will be serialized. For example, in the case of a JSON
 serialization, the context attributes and the event data might both appear
 within the same JSON object.
 
@@ -420,39 +419,33 @@ on the definition of OPTIONAL.
 
 ### Extension Context Attributes
 
-CloudEvents producers MAY include additional context attributes in the event
-that might be used in ancillary actions related to the processing of the event.
+A CloudEvent MAY include any number of additional context attributes with
+distinct names, known as "extension attributes". Extension attributes MUST
+follow the same [naming convention](#attribute-naming-convention) and use the
+same [type system](#type-system) as standard attributes. Extension attributes
+have no defined meaning in this specification, they allow external systems to
+attach metadata to an event, much like HTTP custom headers.
+
+Extension attributes are always serialized according to binding rules like
+standard attributes. However this specification does not prevent an extension
+from copying event attribute values to other parts of a message, in order to
+interact with non-CloudEvents systems that also process the message. Extension
+specifications that do this SHOULD specify how receivers are to interpret
+messages if the copied values differ from the cloud-event serialized values.
+
+#### Defining Extensions
+
 See
 [CloudEvent Attributes Extensions](primer.md#cloudevent-attribute-extensions)
 for additional information concerning the use and definition of extensions.
 
-This specification places no restriction on the semantics of the extension
-attributes, but they MUST use one of the types defined within the
-[type system](#type-system). Each definition of an extensions SHOULD fully
-define all aspects of the attribute - e.g. its name, semantic meaning and
-possible values or even to indicate that it places no restrictions on its
-values. New extension definitions SHOULD use a name that is descriptive enough
-to reduce the chances of name collisions with other extensions. In particular,
-extension authors SHOULD check the
-[documented extensions](documented-extensions.md) document for the set of known
-extensions - not just for possible name conflicts but for extensions that might
-be of interest.
-
-Each specification that defines how to serialize a CloudEvent will define how
-extension attributes will appear.
-
-Extension attributes MUST be serialized using the same general pattern as all
-CloudEvents context attributes. For example, in binary HTTP, that means they
-MUST appear as HTTP headers with the `ce-` prefix. The specification of an
-attribute MAY define a secondary serialization where the data is duplicated in
-some other location within the message.
-
-In cases where a secondary serialization is defined, the extension specification
-MUST also state what a receiver of the CloudEvent is to do if the data differs
-between those two serialization locations. Additionally, senders need to be
-prepared for intermediaries, and receivers, to not know about their extension
-and therefore the specialized serialization version will most likely not be
-processed as a CloudEvent extension attribute.
+The definition of an extension SHOULD fully define all aspects of the
+mattribute - e.g. its name, type, semantic meaning and possible values. New
+extension definitions SHOULD use a name that is descriptive enough to reduce the
+chances of name collisions with other extensions. In particular, extension
+authors SHOULD check the [documented extensions](documented-extensions.md)
+document for the set of known extensions - not just for possible name conflicts
+but for extensions that might be of interest.
 
 Many protocols support the ability for senders to include additional metadata,
 for example as HTTP headers. While a CloudEvents receiver is not mandated to


duglin:
@alanconway rebase needed

spec.md
```diff
@@ -406,18 +404,28 @@ See
...
+The definition of an extension attribute SHOULD define its [type](#type-system),
+name, semantic meaning and constraints on legal values.
+
+New extension definitions SHOULD use a name that is descriptive enough to reduce
+the chances of name collisions with other extensions. In particular, extension
+authors SHOULD check the [documented extensions](documented-extensions.md)
+document for the set of known extensions - not just for possible name conflicts
+but for extensions that might be of interest.
+
+The consumer of an extension attribute MUST know the extension definition to use
```
duglin: I don't think we can make this a MUST since 1) we can't enforce or check it, and 2) a "consumer" doesn't actually need to know about it if all it's doing is passing it along. I tried to reword it to keep the gist of what you're trying to say, but I can't seem to find a wording that doesn't seem obvious.....e.g. "if a consumer wants to successfully process an extension then it needs to know about it" feels silly  :-)

What is the requirement you're trying to make/enforce here?

spec.md
```diff
@@ -406,18 +404,28 @@ See
...
+it meaningfully, including being able to interpret the [canonical
+string-encoding](#type-system) for the attributes type.
+
+Intermediaries that need to transcode events between different formats or
+transport bindings do not need to understand the extension semantics. They can
+forward extension values either by mapping between the native types of each
+format, or by converting to the [canonical string-encoding](#type-system). The
+final consumer is responsible for determining the semantics of the attribute.
```
duglin: Seems ok except the last sentence since technically an intermediary could understand and do something with an extension too, if it wanted.  Could we remove it?  Or perhaps say.... "In those cases, the final consumer....".  Then it's clear that it's just about situations where the intermediaries are unaware of the extension.

alanconway:
Dropping this PR, @duglin beat me to it by a day - this change is sufficient for me:

f41061e 2019-09-19 make extensions follow the same pattern as normal attrs (#505)

alanconway:
Had another think, here's my new take. Less is more! .

spec.md
```diff
@@ -419,25 +421,7 @@ the set of known extensions - not just for possible name conflicts but for
 extensions that might be of interest.
 
 Each specification that defines how to serialize a CloudEvent will define how
-extension attributes will appear.
-
-Extension attribtue MUST be serialized using the same general pattern as all
```
duglin: why did you drop this section? It's kind of the most important pieces from #508

alanconway: It's from #505 not #508.

Extensions should follow the same patterns in the sense of the abstract event model - i.e. use the same type system.

However event encodings can't  be required to follow "the same patterns" in how they encode extensions, since this spec doesn't have any notion of any event encoding,  so it can't say what "patterns" might be best.

The AMQP, HTTP et. al. specs do say what those encodings look like, and of course it makes sense to follow similar patterns when designing a new binding or format, but it doesn't make sense to mandate that in *abstract* spec. 

E.g. the idea of "alternate locations" is used in HTTP and AMQP but not JSON. The requirement that extensions MUST be encoded using "similar patterns" to standard attributes isn't even followed by existing HTTP and AMQP binary encodings - they use a special "alternate location" pattern for datacontenttype, which can't possibly be followed for extension attributes. The original JSON format put extensions in a separate map - som future binding might need to do the same.

In short, a binding MUST be able to faithfully encode and decode an abstract event to/from its internal representation - but how it does so is for the binding spec.

duglin: I guess it depends on whether there is value in the abstract spec providing the overall guidance that all binding specs need to follow.  If the group decides that it's ok to leave it up to each binding/transport spec, then we'd need to define "secondary" in each spec because right now those other specs point to the main spec for that definition.

What do others think?  Leave it up to the binding specs or have a central statement about the consistency?

alanconway: It's fine to have examples or suggestions, but not in normative language. The spec needs to be clear on what is absolutely required of a binding, and not confuse that with suggestions about what might be a good idea for some (or even most) bindings.

duglin: It's not an example though, it says that extensions are serialized the same way as normal CE attributes. Yes each transport might define its own way of doing that serialization but I think its helpful to say (at the spec level) that all attributes (core and extensions) are equal and treated as such.  Then each transport can define what those generic rules are for that specific transport.

alanconway: CE attributes are *not* all serialized the same way, not even core attributes in our existing bindings. datacontenttype is treated differently in AMQP and HTTP because they happen to have a spot that clearly should carry the content-type. We don't know what oddities future protocol bindings might have, so I don't see any value in MUSTy language on the subject. As long as a binding faithfully represents the abstract CE event, there's no need for the core spec to say how it should be done.

alanconway: Concrete example: the AMQP CE spec chose to map datacontentype to AMQP message property content-type. It chose *not* to map CE "id" and "time" attributes to AMQP message properties message-id and creation-time. I agree with those choices, but it would be a mistake to force similar choices on future bindings from the core spec; they should be informed by the context of the protocol we are binding to.

duglin:
rebase needed

spec.md
```diff
@@ -419,25 +421,7 @@ the set of known extensions - not just for possible name conflicts but for
...
-CloudEvents context attributes. For example, in binary HTTP, that means they
-MUST appear as HTTP headers with the `ce-` prefix. The specification of an
-attribute MAY define a secondary serialization where the data is duplicated
-in some other location within the message.
-
-In cases where a secondary serialization is defined, the extension
-specification MUST also state what a receiver of the CloudEvent is to do
```
duglin: Are we losing something by dropping this MUST?  Granted it's more of a statement for spec authors instead of people writing the code, but isn't the guidance/requirement needed to ensure interop ?

duglin:
rebase needed

spec.md
alanconway: IMO no. This MUST belongs in binding specs that introduce such notions as "alternative locations" - e.g. it's there in the  AMQP spec.  In the core spec it's confusing and unhelpful speculation about the possible detailed inner workings of some bindings, and entirely irrelevant and meaningless for others (e.g. JSON format, MQTT 3.1 etc.)

duglin: Not sure I agree since then each protocol spec would need to duplicate the same type of text when it's generic text that isn't protocol specific.  However, even so, this PR doesn't add the text to the protocol docs so wouldn't more work be needed before we dropped the text here?

alanconway: Nope. If you want a MUST in the core it should say "binding specs MUST be clear and unambiguous enough that different implementations of the binding spec will inter-operate" - it should not say how particular bindings might or might not accomplish that. However I think that's already implied by the definition of a binding spec.

spec.md
alanconway: Agreed the MUST is wrong. I'm trying to say - extension attributes can't be relied on to carry type information so the consumer has to Just Know from somewhere else. They can't write code like `if attr[foo].isa timestamp...` they have to know to say `ts := AsTimestamp(attr[foo])`

alanconway: Oops, already fixed.

spec.md
alanconway: Put back a modified version of this now I get your concern.

spec.md
alanconway: See latest

spec.md
```diff
@@ -438,27 +440,31 @@ extension authors SHOULD check the
...
-and therefore the specialized serialization version will most likely not be
-processed as a CloudEvent extension attribute.
+A binding MAY serialize the same attribute value to multiple locations in a
+message. For example HTTP tracing repeats some attribute values in "ce-" headers
+and in other headers examined by tracing agents. Such bindings MUST specify how
+receivers should handle conflicting values for duplicated attributes, and how
+senders should format messages to ensure that the desired attribute value is
+transmitted.
 
 Many protocols support the ability for senders to include additonal metadata,
```
alanconway: Many protocols support the ability for senders to include additonal metadata,
for example as HTTP headers. While a CloudEvents receiver is not mandated to
process and pass them along, it is RECOMMENDED that they do so via some
mechanism that makes it clear they are non-CloudEvents metadata.

spec.md
```diff
@@ -438,27 +440,30 @@ extension authors SHOULD check the
...
+`datacontentype` attribute is stored in the standard HTTP `Content-Type` header
+so that it is visible to other HTTP processes.
 
-In cases where a secondary serialization is defined, the extension specification
-MUST also state what a receiver of the CloudEvent is to do if the data differs
-between those two serialization locations. Additionally, senders need to be
-prepared for intermediaries, and receivers, to not know about their extension
-and therefore the specialized serialization version will most likely not be
-processed as a CloudEvent extension attribute.
+A binding MAY serialize the same attribute value to multiple locations in a
```
duglin: This focuses on a "binding" defining the duality when I tend to think of it as the spec that defines the extension. We don't define any of these attributes in our specs (core or protocol) so it feels misleading to hint that a protocol spec would know anything about these special attributes.

Would it be bad to say: An extension attribute specification MAY define an additional location to which it would be serialized (in additional to the REQUIRED location defined by the protocol binding specifications). ... Such specifications MUST specify how....

spec.md
duglin: Basically yes and I think it was driven by those duplicate serialized extensions.  So, in the tracing case we were thinking that it would appear as `ce-traceparent` and HTTP header `traceparent`, and if they had different values we wanted to encourage implementations to expose both to the receiving app so they could decide what to do about the diff.  If they only ever saw the ce- one (and couldn't even get to the other one) then we're kind of setting people up for a situation where they're forced to only deal with one of the two.

spec.md
alanconway: Aha, we reach the heart of our misunderstanding. We are talking about different things with the same words :)
- *extension attributes* are name-value pairs defined by the core and serialized by a binding.
- *extension spec* describes an arbitrary external system that defines and uses named  extension attributes to accomplish some task.

I tripped over "extension MAY *serialize* ..." - it made me think the extension spec was somehow changing the serialization rules of the binding. Now I see that's not the case: e.g. the trace extension is *using* extension attributes to annotate a HTTP message for its own purposes. Those annotations are not part of the binding's serialization of the event and will be ignored by the receiving binding (unless it also implements the extension.)

Of course the *extension spec* must say how to deal with those annotations, which is what  you were trying to advise about from the beginning. It's not part of the core spec, but I'm ok with including it as advisory if we clarify the distinction. Maybe say "annotate" rather than "serialize to an alternate location" or some such. I'm specced out for today, do you want to have a go?

alanconway: OK, not specced out - check the latest. Moved back towards your original text, but with more clarity about what is *binding* spec and what is *extension* spec. I think it makes sense now.

duglin: `extension attributes are name-value pairs defined by the core and serialized by a binding.`
Well, name-value pairs yes, but not defined by the core. They're defined by someone who isn't us (ignoring the fact that we provide a convenient hosting location for people who don't want to do it themselves).

Yes to your 2nd para.

Not sure what your 3rd (last) para is saying though.  An extension spec will, at a minimum, define a new extension attribute. If it does nothing else then that attribute will be serialized like all other CE attributes for whatever format/binding is used (ignore any special cases like 'content-type').  That extension spec MAY also define alternate locations for the data to appear but the data MUST also be serialized using those default serialization rules. And, if there is an alternate location then the spec MUST say what to do if those multiple locations have different values when it is received.

Does that help?

alanconway: > `extension attributes are name-value pairs defined by the core and serialized by a binding.`
> Well, name-value pairs yes, but not defined by the core. 

I want to make it clear that there are 3 distinct layers:
1. core spec defines event to include open set of named attributes
2. binding spec defines how to serialize attributes
3. extension spec choses attribute names and says what to use them for.

I did a minor re-order to make this clearer.
  
> Yes to your 2nd para.
> 
> Not sure what your 3rd (last) para is saying though. 

Me neither. Maybe you can have another go at it. Here's where I'm having trouble:

>An extension spec will, at a minimum, define a new extension attribute. If it does nothing else then that attribute will be serialized like all other CE attributes for whatever format/binding is used (ignore any special cases like 'content-type'). 

IMO that goes without saying and saying it is confusing. The extension spec has nothing to do with how the attribute is serialized - that's the binding spec. The extension spec is simply picking a name and using the facilities provided by core and binding.

>That extension spec MAY also define alternate locations for the data to appear 

I'm struggling with this: it's not an "alternate location" for the attribute, it's just one thing an extension might do with attribute values - like storing them in a database, logging them or making routing decisions. I accept this is an important case you want to highlight, but I feel this language sounds like you're mandating something about how events are serialized.

> but the data MUST also be serialized using those default serialization rules. 

This sounds wrong - the only reason to copy data to different parts of the message is to communicate with some external system that doesn't know about the CE spec. Our encodings might fit, or might not. If an extension is setting timestamps to be read by a system that expects milliseconds-since-epoch, there's no point using RFC3999 strings.

I'll let you take another shot at what you're trying to say there. I think I agree with the substance but I found the language confusing so others might too. Hopefully you see what was confusing me...

duglin: > > An extension spec will, at a minimum, define a new extension attribute. If it does nothing else then that attribute will be serialized like all other CE attributes for whatever format/binding is used (ignore any special cases like 'content-type').
> 
> IMO that goes without saying and saying it is confusing. The extension spec has nothing to do with how the attribute is serialized - that's the binding spec. The extension spec is simply picking a name and using the facilities provided by core and binding.

The reason I think it needs to be said is because of all of the discussions we've had in the past around how extensions are serialized. E.g. are they just like normal CE attributes or do they get special "extensions" bucket? You missed a lot of the fun, but it's been a challenge to get everyone on the same page here. Some folks really wanted extensions to be serialized different from spec-defined attributes and as such I think explicitly saying that they're not special helps.  At worst, you can say it's redundant and I can live with that if it stops people from making poor choices and breaking our extension model/plans.

> > That extension spec MAY also define alternate locations for the data to appear
> 
> I'm struggling with this: it's not an "alternate location" for the attribute, it's just one thing an extension might do with attribute values - like storing them in a database, logging them or making routing decisions. I accept this is an important case you want to highlight, but I feel this language sounds like you're mandating something about how events are serialized.

What do you think it's mandating? I'm not sure I would put this into the same category as "one thing an extension might do with an attribute value - like storing them in a DB" because those actions are not visible on the wire. I suspect most people will (rightfully) assume that any attribute will only appear once in a CE message because for the majority of the cases that's true (ignoring the fact that an attribute value might also appear in the `data` section). So, explicitly saying that it's ok for these edge cases (where a value can appear in more than one location in a message) is something I think is valuable to say. Or put another way, I think someone could claim that putting a value in 2 locations is violating the spec because one of those locations isn't prefixed with `ce-` when clearly the rules for that protocol says it should be. And I think it's important to make it clear that it is ok to do so, as long as the "normal" serialization version is there.

> > but the data MUST also be serialized using those default serialization rules.
> 
> This sounds wrong - the only reason to copy data to different parts of the message is to communicate with some external system that doesn't know about the CE spec. Our encodings might fit, or might not. If an extension is setting timestamps to be read by a system that expects milliseconds-since-epoch, there's no point using RFC3999 strings.

Yes, the reason for the 2nd serialization is to work with another system that doesn't know about the ce- version.  Why is that wrong? What needs to happen though is that the extension spec needs to say what a CE receiver should do about conflicts between the two locations.

> > I'll let you take another shot at what you're trying to say there. I think I agree with the substance but I found the language confusing so others might too. Hopefully you see what was confusing me...

I'd like to turn this around :-)  and ask you what's wrong with the original text that's in the spec given I think your understanding of the reasons behind it may have shifted?

duglin:
LGTM

duglin:
Approved on the 10/17 call

PR #516, opened 2019-09-26
Update example for 'application/avro' content

### LINKED ISSUE #515: Avro formatted data using application/avro
When considering an Avro formatted event, following https://github.com/cloudevents/spec/blob/master/avro-format.md, I was wondering why the second example in section 4 (where it is using `application/avro` as the contenttype), shows the `data` mapping as a `string` with an example that appears to be base64 encoded?

Would this not be binary instead following the `bytes` Avro type? By determining that `datacontenttype` is set to `application/avro` and given that Avro is itself binary formatted, why would you need to encode it as a base64 encoded string? We treat the JSON example as bytes, why not this?

timjroberts:
This PR updates the example for 'application/avro' content data where it should be written using the Avro `bytes` type.

Closes #515

### DIFF
--- a/avro-format.md
+++ b/avro-format.md
@@ -167,7 +167,7 @@ The following table shows exemplary mappings:
 |             |        |                                                |
 | dataschema  | string | `"http://registry.com/subjects/ce/versions/1"` |
 | contenttype | string | `"application/avro"`                           |
-| data        | string | `"Q2xvdWRFdmVudHM="`                           |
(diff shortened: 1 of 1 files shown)


duglin:
@fabiojose @clemensv wanna review this one?

fabiojose:
LGTM

clemensv:
LGTM

duglin:
Approved on the 10/10 call

Issue #519, opened 2019-10-02
Relation between AMQP Format and Transport Binding is unclear


deissnerk:
I've been reviewing the AMQP related parts of the spec, and I'm having trouble figuring out how the [data section](https://github.com/cloudevents/spec/blob/master/amqp-format.md#3-data) of the AMQP format relates to the AMQP transport binding. I found [this](https://github.com/cloudevents/spec/blob/master/amqp-transport-binding.md#14-event-formats):

> All implementations of this specification MUST support the JSON event format as well as the AMQP event format for the application-properties section

This restricts the usage of the AMQP event format to the application-properties section, which excludes the handling of `data`.

In [section 2.2](https://github.com/cloudevents/spec/blob/master/amqp-transport-binding.md#22-data) I found:
>An application is free to hold the information in any in-memory representation of its choosing, but as the value is transposed into AMQP as defined in this specification, the assumption is that the data value is made available as a sequence of bytes.

Finally [section 3.1.2](https://github.com/cloudevents/spec/blob/master/amqp-transport-binding.md#312-event-data-encoding) simply states:
>The data byte-sequence is used as the AMQP application-data section.

So the handling of `data` as described in the AMQP format with its usage of AMQP values would not take place. As I don't think the AMQP format would make sense for the structured encoding, either, this pretty much excludes that this rule from the AMQP format is applied:
>For other types (non-binary data without a datacontenttype attribute), the implementation MUST translate the data value into an AMQP type system value and the value MUST be stored in an AMQP value section.

All this really confuses me. In the examples section of the JSON format there is this example:

```
{
    "specversion" : "1.0-rc1",
    "type" : "com.example.someevent",
    "source" : "/mycontext",
    "id" : "C234-1234-1234",
    "time" : "2018-04-05T17:31:00Z",
    "comexampleextension1" : "value",
    "comexampleothervalue" : 5,
    "datacontenttype" : "application/json",
    "data" : {
        "appinfoA" : "abc",
        "appinfoB" : 123,
        "appinfoC" : true
    }
}
```
How would an SDK transform this into AMQP format/binary encoding? What would happen, if the `datacontenttype` was omitted? I suppose, according to the AMQP format some kind of AMQP value would be created. Would it be an AMQP Map?

alanconway:
Here's my understanding:

Structured format (e.g. JSON):
    AMQP { content-type: "application/cloudevents+json"; data-section=<json event bytes>}

[...]

PR #521, opened 2019-10-04
Clarify meaning of missing datacontentype.

### LINKED ISSUE #520: How to determine if data is binary?
Several event formats now contain a section about the handling of data, which specifies that implementations need to determine if data is binary. In #371 @alanconway [pointed out](https://github.com/cloudevents/spec/issues/371#issue-400799222) that MIME types do not provide this information.
If an event bridge receives events in HTTP binary mode and is supposed to convert them to JSON format, how does it determine in a reproducible way, if it needs to base64 encode the data?
So what would implementations do?

alanconway:
Clarify the meaning of missing datacontentype, when it is allowed, and what that means when translating events between formats/protocols.

Also replaced "transport" with "protocol" everywhere, I believe that is now the preferred term. "Transport" is not defined in the terminology.

Fixes #520

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -119,15 +119,33 @@ include information about the occurrence, details about the data that was
 changed, or more. See the [Event Data](#event-data) section for more
 information.
 
+#### Event Format
+
+An Event Format specifies how to serialize a CloudEvent as a sequence of bytes.
+Stand-alone event formats, such as the [JSON format](json-format.md), specify
+serialization independent of any protocol or storage medium.  Protocol Bindings
+MAY define formats that are dependent on the protocol.
+
 #### Message
 
 Events are transported from a source to a destination via messages.
+
+A "structured-mode message" is one where the event is fully encoded using
+a stand-alone event format and stored in the message body.
+
+A "binary-mode message" is one where the event data is stored in the message
+body, and event attributes are stored as part of message meta-data.
 
 #### Protocol
 
 Messages can be delivered through various industry standard protocol (e.g. HTTP,
 AMQP, MQTT, SMTP), open-source protocols (e.g. Kafka, NATS), or platform/vendor
 specific protocols (AWS Kinesis, Azure Event Grid).
+
+#### Protocol Binding
+
+A protocol binding describes how events are sent and received over a given
+protocol, in particular how events are mapped to messages in that protocol.
 
 ## Context Attributes
 
@@ -158,7 +176,7 @@ descriptive and terse and SHOULD NOT exceed 20 characters in length.
 
 The following abstract data types are available for use in attributes. Each of
 these types MAY be represented differently by different event formats and in
-transport metadata fields. This specification defines a canonical
+protocol metadata fields. This specification defines a canonical
 string-encoding for each type that MUST be supported by all implementations.
 
 - `Boolean` - a boolean value of "true" or "false".
@@ -205,14 +223,14 @@ _datetime_ type in a given implementation, but it MUST be settable providing an
 RFC3339 string, and it MUST be convertible to an RFC3339 string when mapped to a
 header of an HTTP message.
 
-A CloudEvents transport binding or event format implementation MUST likewise be
+A CloudEvents protocol binding or event format implementation MUST likewise be
 able to convert from and to the canonical string-encoding to the corresponding
-data type in the encoding or in transport metadata fields.
+data type in the encoding or in protocol metadata fields.
 
 An attribute value of type `Timestamp` might indeed be routed as a string
 through multiple hops and only materialize as a native runtime/language type at
 the producer and ultimate consumer. The `Timestamp` might also be routed as a
-native transport type and might be mapped to/from the respective
+native protocol type and might be mapped to/from the respective
 language/runtime types at the producer and consumer ends, and never materialize
 as a string.
 
@@ -323,16 +341,20 @@ on the definition of OPTIONAL.
   example, the JSON event format defines the relationship in
   [section 3.1](./json-format.md#31-handling-of-data).
 
-  When this attribute is omitted, `data` simply follows the event format's
-  encoding rules. For the JSON event format, the `data` value can therefore be a
-  JSON object, array, or value.
+  For some binary mode protocol bindings, this field is directly mapped to the
+  respective protocol's content-type metadata property. Normative rules for the
+  binary mode and the content-type metadata mapping can be found in the
+  respective protocol
 
-  For the binary mode of some of the CloudEvents transport bindings, where the
-  `data` content is immediately mapped into the payload of the transport frame,
-  this field is directly mapped to the respective transport or application
-  protocol's content-type metadata property. Normative rules for the binary mode
-  and the content-type metadata mapping can be found in the respective transport
-  mapping specifications.
+  In some event formats the `datacontenttype` attribute MAY be omitted. For
+  example, if a JSON format event has no `datacontenttype` attribute, then it is
+  implied that the `data` is a JSON value conforming to the "application/json"
+  media type. In other words: a JSON-format event with no `datacontenttype` is
+  exactly equivalent to one with `datacontenttype="application/json"`.
+
+  When translating an event message with no `datacontenttype` attribute to a
+  different format or protocol binding, the target `datacontenttype` SHOULD be
+  set explicitly to the implied `datacontenttype` of the source.
 
 - Constraints:
   - OPTIONAL
@@ -432,7 +454,7 @@ prepared for intermediaries, and receivers, to not know about their extension
 and therefore the specialized serialization version will most likely not be
 processed as a CloudEvent extension attribute.
 
-Many transports support the ability for senders to include additonal metadata,
+Many protocols support the ability for senders to include additonal metadata,
 for example as HTTP headers. While a CloudEvents receiver is not mandated to
 process and pass them along, it is RECOMMENDED that they do so via some
 mechanism that makes it clear they are non-CloudEvents metadata.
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -325,19 +330,28 @@ on the definition of OPTIONAL.
...
+  `datacontenttype="application/json"` and its `data` can be any JSON object,
+  array, or value.
+
+  Formats and protocols that do not have an implied `datacontenttype` MUST NOT
+  omit this attribute. Agents that translate messages between formats or
+  protocols MUST ensure that an implied `datacontenttype` is made explicit in
+  the new format. For example when translating a JSON event with no
+  `datacontenttype` to AMQP, the AMQP `content-type` message property MUST be
+  set to "application/json".
+
```
deissnerk: +1

spec.md
```diff
@@ -325,19 +330,28 @@ on the definition of OPTIONAL.
...
-  this field is directly mapped to the respective transport or application
+  For the binary mode of some of the CloudEvents protocol bindings, where the
+  `data` content is immediately mapped into the payload of the protocol frame,
+  this field is directly mapped to the respective protocol or application
   protocol's content-type metadata property. Normative rules for the binary mode
-  and the content-type metadata mapping can be found in the respective transport
+  and the content-type metadata mapping can be found in the respective protocol
   mapping specifications.
 
+  This attribute MAY be omitted if the event format has an implied
```
duglin: Use of the word "implied" here can be confusing. If I'm understanding things, this is really talking about the case where it's a "structured" message and therefore w/o a datacontenttype attribute we can assume that `data` just follows the format of the payload itself, right?  If so, would this wording be ok:

```
In structured mode, where the CloudEvents attributes are siblings to `data` in the message
payload, this attribute MAY be omitted when the format of `data` is the same as the format
of the message itself. For example, when the payload is of type JSON, `datacontenttype` can
be omitted when `data` is also of type JSON.
```

spec.md
```diff
@@ -325,19 +330,28 @@ on the definition of OPTIONAL.
...
+  `datacontenttype`. For example, the JSON format implies "application/json". A
+  JSON-format event with no `datacontenttype` is exactly equivalent to one with
+  `datacontenttype="application/json"` and its `data` can be any JSON object,
+  array, or value.
+
+  Formats and protocols that do not have an implied `datacontenttype` MUST NOT
```
duglin: I believe this is really just talking about binary mode, right?
If so, I have mixed feeling about this. While I do agree that it's a good idea, best practice, to include a content-type on all messages, I tend to think of binary mode being used in cases where CE is being added to an existing message (while not always true, I do think that's going to be a common case). And so in those cases we have an existing message flow that works w/o a content-type attribute, and now the addition of CE will mandate that senders include an attribute that previously wasn't needed. That seems like an unnecessary burden and out of scope for CE to mandate since we're not trying to control how the payload of the message is processed - just how to find common metadata.

Now.... on the flip side, as I said, I do think it's a good idea to include it and I can certainly understand someone thinking that CE isn't just about CE attributes but also about promoting interop and so nudging people to include a content-type attribute is just an overall good thing.

Would it be horrible if instead of making it a MUST for binary mode, we make it a SHOULD or STRONGLY RECOMMENDED ?

duglin:
@alanconway you'll need to sign your commits

spec.md
alanconway: I think it's consistent with what the spec already says, and it rules out loopy interpretations like my previous "it must be a CE type" nonsense.

It does mean that you cannot use an AMQP value-section message as an event without additional clarifications in the AMQP binding spec and/or proprietary transformations - but that can be solved in the AMQP binding spec, and that's the right place to solve it

alanconway: @clemensv thoughts?

spec.md
alanconway: Agreed it's confusing and again the use of terms "binary" and "structured" is causing us problems. Structured mode *messages* are not events an do not have a datacontentype attribute. They *contain* an event in some in some "application/cloudevents+" encoding (or multiple events now we have batch formats). 

Only when you decode an individual event  do you see a dataconenttype attribute - and at that point it is irrelevant whether the event arrived wrapped in a structured-mode message, as part of a batch, or directly (e.g. JSON events read from stdin.)

So this issue applies only to *directly-encoded* events. We use the term "binary" for HTTP and AMQP directly-encoded events, we don't for JSON since that would exacerbate the base64 confusion. "binary" and "structured" are really just the wrong terms. I'll take another shot avoiding these terms and "implied'

spec.md
alanconway: Lemme take another shot.

alanconway:
Took a shot at clarifying - whaddya think now?

I went with SHOULD, but I really don't think it makes sense to translate an event with a well-known, if not explicitly stated datacontentype (JSON) into say an AMQP or HTTP binary message with a blob of bytes and no clues what to do with them.

spec.md
```diff
@@ -119,16 +119,34 @@ include information about the occurrence, details about the data that was
...
+An Event Format specifies how to serialize a CloudEvent as a sequence of bytes.
+Stand-alone event formats, such as the [JSON format](json-format.md), specify
+serialization independent of any protocol or storage medium.  Protocol Bindings
+MAY define formats that are dependent on the protocol.
+
 #### Message
 
 Events are transported from a source to a destination via messages.
 
+A "structured-mode message" is one where the event is fully encoded using
```
duglin: regardless of where we land on this PR - I like adding these definitions to the spec itself.

spec.md
```diff
@@ -325,16 +343,24 @@ on the definition of OPTIONAL.
   event format defines the relationship in
   [section 3.1](./json-format.md#31-handling-of-data).
 
-  When this attribute is omitted, `data` simply follows the event
-  format's encoding rules. For the JSON event format, the `data` value
-  can therefore be a JSON object, array, or value.
+  Some event formats have a default value for `datacontenttype`. Events in such
```
duglin: It feels odd to talk about a "default content type" because it comes across like something specific to the format - almost like you're saying JSON defines a default content type and that's not really what's going on. All we're doing is saying that `datacontenttype` can be omitted if it's the same value as the event format itself - just an optimization. Can we just say that instead?

spec.md
```diff
@@ -325,16 +343,24 @@ on the definition of OPTIONAL.
...
+  SHOULD propagate the `datacontenttype` from source to target as follows:
 
-  For the binary mode of some of the CloudEvents transport bindings, where the
-  `data` content is immediately mapped into the payload of the transport frame,
-  this field is directly mapped to the respective transport or application
-  protocol's content-type metadata property. Normative rules for the binary mode
-  and the content-type metadata mapping can be found in the respective transport
-  mapping specifications.
+  * Source has a value: copy it to target unmodified.
+  * Source has no value, source format has a default: set the target to the
```
duglin: I think I can live with this if we can tweak the wording to not use the term "default" per my comment above

duglin:
Does this address #519 as well?

spec.md
alanconway: It's not the same. The media type for a JSON-formatted event is "application/cloudevents+json". The media  type for an arbitrary JSON value is "application/json". It is obvious, but it needs to be explicit *as a media type* for use in non-JSON contexts. Other formats/bindings may not have a less obvious default, or none at all. 

This thread started with confusion about the AMQP type system not being expressible in a cloud event. I'm trying to say that is because AMQP doesn't *have* a default media type, therefore any CE data ever carried over AMQP must be described by some other media type. Formats that do have a default media type (JSON) will get tagged with it as they enter the AMQP world and can be properly parsed when they exit.

spec.md
alanconway: Pick a word. We want to say "the datacontentype value to be assumed when none is explicitly present".

spec.md
alanconway: Yep, I find them very counter-intuitive so we should write them down somewhere obvious.

alanconway:
> Does this address #519 as well?

I think it does. 
@deissnerk do you agree?

alanconway:
CI failure is an unresponsive external link: https://goextend.io
It's still down 503.

duglin:
Yeah, I noticed that and was hoping it was a temporary thing but if it lasts a day or two more I'm gonna just remove that link from our docs

spec.md
clemensv: I agree with the intent of this. I would prefer 'implied' over 'default':

Some event formats have an implied value for datacontenttype. For example, in the JSON format `datacontenttype` has an implicit value of "application/json", which permits for `data` to be any JSON value. If overriden with a different than the implied value, `data` must follow the respectuve content-type's rules.

alanconway: Reworded, back to "implied".

alanconway:
> @alanconway I think it is a prerequisite to address #519. If this PR with its definition for _formats_ gets accepted, the AMQP format could even be removed and the type mapping for attributes be moved over to the AMQP protocol binding.

That would be my preference.

> As long as the AMQP format remains protocol specific and does not define a content-type on its own, the only AMQP values I could think of, would be String and Binary. I don't see any advantage in this compared to binary mode where CE data is always put into AMQP data.

The current spec suggests (and I agree) that you set content-type and use a single data section for the body. That means that AMQP types (even binary and string) are never used as data - there are no value-section messages. I would be forgiving on incoming messages and accept string or binary value-sections as if they had been data sections - some AMQP APIs hide those distinctions from the user.

> Btw the term _transport_ occurs in file names, too. Do we need follow-up PRs to adjust terminology in all the protocol bindings?

That's above my pay-grade guvnor.

spec.md
alanconway: I think this is discussed more below & hopefully addressed by latest update.

spec.md
alanconway: see latest update

clemensv:
LGTM

duglin:
Approved on the 10/10 call

PR #523, opened 2019-10-04
Remove unrestricted extensions. Lint.

n3wscott:
- removed:

```
The specification places no restrictions on the type of the extension	
attributes. Meaning, they may be simple types (e.g. strings, integers), complex	
(e.g. structured) or undefined collection of attributes.
```

- ran lint.

### DIFF
--- a/primer.md
+++ b/primer.md
@@ -115,10 +115,10 @@ solely for the purpose of proper delivery, and processing, of the message. Data
 that is not intended for that purpose should instead be placed within the event
 (`data`) itself.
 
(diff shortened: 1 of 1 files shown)


duglin:
LGTM thanks @n3wscott 

Just to move things along, since I consider this a typo ( meaning we just forgot to remove this text when we changed it so that extensions are required to be of CE type) then if I could get one more LGTM we can merge w/o waiting for the call.

duglin:
@Vlaaaaaaad thanks
merging

Issue #328, opened 2018-10-10
Web Hooks: How to distinguish denied delivery and missing support of validation?


deissnerk:
The [web hook spec](https://github.com/cloudevents/spec/blob/master/http-webhook.md#4-abuse-protection) states:
> If the handshake outcome is that the delivery is denied, the sender MUST NOT deliver events to the target.

A few lines later it says:
> Delivery targets SHOULD support the abuse protection feature. If a target does not support the feature, the sender MAY choose not to send to the target, at all, or send only at a very low request rate.

I wonder, if there is a way for the sender to determine, if a target just does not support http OPTIONS, or if the event delivery is denied. In both cases I would expect a response with status 405 and no `WebHook-Allowed-Origin:` header.

duglin:
would it be a 405 or 401 Unauthorized?  wouldn't each sender have its own auth/creds and a rogue sender would fail the auth check?

clemensv:
If the target doesn't support OPTIONS it doesn't support validation. You get a 405 and no header. If it supports OPTIONS but no validation, you get a success code response, but the header will be missing. The WebHook-Allowed-Origin header really is the success indicator. 

Without validation, it's really up to the target for whether it wants to allow deliveries. Mind that the protection mechanism is primarily for the delivering party and not for the target. We want to give the delivering party a way to protect itself from becoming a DDoS engine.

If the target doesn't allow deliveries without prior validation, it would just respond with 401.

deissnerk:
Thank you, @clemensv. If I understand that correctly, the sending party MUST NOT send events, if the target reacts with 401. In all other cases it is up to the sender, if it wants to take the risk of being abused. Correct?

duglin:
Do we need a PR to clarify things?

fohoffmann:
During implementation of chapter 4 “Abuse Protection” we had a discussion why the 401 approach isn’t sufficient for the sender to protect itself being abused: Assuming the spec adds in chapter 3 “Authorization” a mandatory recommendation to stop sending events to a target in case it returns a 401 due to negative authorization check. I.e. to treat a 401 as an implicit "unsubscribe". This would solve the “Abuse Protection” problem and adds the advantage to react on later changes:

Case 1: An attacker subscribes a “foreign” Webhook at sender. Sender sends first event, target does not expect the event and reacts with 401. Sender takes this as unsubscribe, spam is stopped with the first event.

[...]

PR #528, opened 2019-10-08
Removed statement about denial

deissnerk:
I removed a requirement that a sender MUST NOT send notifications in case of denial by the target, as there is no way for the sender to distinguish denial from a target just not supporting the handshake. As the validation request is mostly about protecting the sender, the statement was unnecessary.

If people come to the conclusion that the statement should not be removed, we need some clear way to distinguish the validation response of a target that does not support validation and a target that denies event delivery.

This PR solves #328

### DIFF
--- a/http-webhook.md
+++ b/http-webhook.md
@@ -188,9 +188,7 @@ it.
 
 Reaching the delivery agreement is realized using the following validation
 handshake. The handshake can either be executed immediately at registration time
-or as a "pre-flight" request immediately preceding a delivery. If the handshake
-outcome is that the delivery is denied, the sender MUST NOT deliver events to
-the target.
+or as a "pre-flight" request immediately preceding a delivery. 
 
 It is important to understand is that the handshake does not aim to establish an
 authentication or authorization context. It only serves to protect the sender


duglin:
@clemensv what are your thoughts on this?

Would it help if there was a well defined "no, go away" response? Then we can still have the "MUST NOT send" wording if the response is used, and anything else means the producer just doesn't know and therefore it might be ok for them to send events.

clemensv:
If the handshake is not supported, the sender should not deliver. That's the point of having it. The receiver MUST opt in. 

You are not required to support that mutual handshake for delivering requests per HTTP, but if a sender wants the protection and thus supports this spec, it must insist on receiver opt-in .

duglin:
If that's the intent, @deissnerk would it help if we added text along the lines of "If a sender wants to guarantee the receiver is prepared to accept event, then the following applies...." or something like that. To make it clear that that section is only true if both sides support the handshake.  Otherwise, all bets are off.

deissnerk:
There is already the following text:
>To protect the sender from being abused in such a way, a legitimate delivery target needs to indicate that it agrees with notifications being delivered to it.

A bit further down in the same section you find:
>Delivery targets SHOULD support the abuse protection feature. If a target does not support the feature, the sender MAY choose not to send to the target, at all, or send only at a very low request rate.

I think that is sufficient, as it is about protecting the sender. A MUST NOT on denial is not possible, as the handshake does not clearly indicate denial to the sender, and in case of an actual abuse I wouldn't expect the receiver to send a denial we could specify here.

clemensv:
If @deissnerk believes that clarifies the intent, I'm okay with the change.

duglin:
Conditionally approved on the 10/10 call.

Since this was submitted a bit after the 2-day deadline we're going to give people until end of day tomorrow (Friday) to raise any concerns - then if we can get 2 LGTMs we'll merge.

duglin:
LGTM
can I get one more?

markpeek:
LGTM

duglin:
thanks @markpeek 
Merging

PR #529, opened 2019-10-09
Adjust AMQP spec to changes in #521

### LINKED ISSUE #519: Relation between AMQP Format and Transport Binding is unclear

deissnerk:
This PR assumes the changes of #521 to be accepted and fixes #519. It renames the AMQP transport binding into protocol mapping.
With the clear definition of event format in the main spec, a separate AMQP event format became obsolete. I copied the type mapping for the attributes over to the AMQP protocol binding, where it is only relevant for the binary mode, and removed the AMQP format.


amqp-protocol-binding.md
```diff
@@ -68,7 +65,7 @@ modes.
 
 In the _structured_ content mode, event metadata attributes and event data are
 placed into the AMQP message's [application data][data] section using an
-[event format](#14-event-formats).
+event format as defined in the CloudEvents [spec][ce].
 
 In the _binary_ content mode, the value of the event `data` is placed
 into the AMQP message's [application data][data] section as-is, with the
```
alanconway:
```
- declaring its media type;
+ declaring its media type mapped to the AMQP `content-type` message property;
```

amqp-protocol-binding.md
```diff
@@ -94,33 +89,14 @@ mandate specific existing features to be used.
 This specification does not further define any of the [CloudEvents][ce] event
 attributes.
 
-Two of the event attributes, `datacontenttype` and `data` are handled specially
-and mapped onto AMQP constructs, all other attributes are transferred as
+One event attribute, `datacontenttype` is handled specially in *binary* content
+mode and mapped onto AMQP constructs. All other attributes are transferred as
```
alanconway: "mapped on to the AMQP content-type message property" - "AMQP constructs" is needlessly vague, it's clearer to be explicit.

amqp-protocol-binding.md
```diff
@@ -151,13 +127,25 @@ directly to the CloudEvents `datacontenttype` attribute.
...
+expectation is that the event data is made available as [UTF-8][rfc3629] encoded
+JSON text for use in AMQP.
+
 #### 3.1.3. Metadata Headers
 
-All [CloudEvents][ce] attributes with exception of `datacontenttype` and `data`
-MUST be individually mapped to and from the AMQP
+All [CloudEvents][ce] attributes with exception of `datacontenttype` MUST be
+individually mapped to and from the AMQP
 [application-properties][app-properties] section, with exceptions noted below.
```
alanconway: Strike ", with exceptions noted below", we already mentioned datacontentype and there aren't any more exceptions.

amqp-protocol-binding.md
```diff
@@ -186,8 +174,32 @@ Examples:
...
+| ------------- | --------------------------- |
+| Boolean       | [boolean][amqp-boolean]     |
+| Integer       | [long][amqp-long]           |
+| String        | [string][amqp-string]       |
+| Binary        | [binary][amqp-binary]       |
+| URI           | [string][amqp-string]       |
+| URI-reference | [string][amqp-string]       |
+| Timestamp     | [timestamp][amqp-timestamp] |
+
+A CloudEvents AMQP *binary* mode implementation MUST allow for attribute values
```
alanconway: That types be "convertible" is a property of an SDK,  not the protocol binding - the core spec already requires it. 
I'd say something like "all attribute values in an AMQP binary message MAY be represented using the native AMQP types above, or MAY be in canonical string form. An implementation MUST be able to interpret both forms on an incoming AMQP message. An implementation SHOULD use the native AMQP form on outgoing AMQP messages when it is efficient to do so, but MAY forward values as canonical strings. A robust implementation MAY further relax the requirements for incoming messages (for example accepting numeric types other than AMQP long) but MUST use only the prescribed AMQP types or canonical strings on outgoing messages.

duglin:
```
./documented-extensions.md: Can't find: ./amqp-transport-binding.md
```
bad link now

README.md
```diff
@@ -31,7 +31,7 @@ The following documents are available:
 |                               |
 | **Optional Specifications:**  |
 | AMQP Event Format             |      [v1.0-rc1](https://github.com/cloudevents/spec/blob/v1.0-rc1/amqp-format.md)       |      [master](https://github.com/cloudevents/spec/blob/master/amqp-format.md)       |
-| AMQP Transport Binding        | [v1.0-rc1](https://github.com/cloudevents/spec/blob/v1.0-rc1/amqp-transport-binding.md) | [master](https://github.com/cloudevents/spec/blob/master/amqp-transport-binding.md) |
+| AMQP Transport Binding        | [v1.0-rc1](https://github.com/cloudevents/spec/blob/v1.0-rc1/amqp-protocol-binding.md) | [master](https://github.com/cloudevents/spec/blob/master/amqp-protocol-binding.md) |
```
duglin: first one needs to remain with the old name since it's already out there

amqp-protocol-binding.md
deissnerk: I like the proposed text and updated my proposal accordingly. Some people might argue with some of the MUST/SHOULD/MAY statements. I like how it is leaning towards the AMQP types without enforcing it.

README.md
deissnerk: fixed

amqp-protocol-binding.md
```diff
@@ -186,8 +174,29 @@ Examples:
...
+| URI-reference | [string][amqp-string]       |
+| Timestamp     | [timestamp][amqp-timestamp] |
+
+All attribute values in an AMQP binary message MAY be represented using the
+native AMQP types above, or MAY be in canonical string form. An implementation
+MUST be able to interpret both forms on an incoming AMQP message. An
+implementation SHOULD use the native AMQP form on outgoing AMQP messages when it
+is efficient to do so, but MAY forward values as canonical strings. A robust
+implementation MAY further relax the requirements for incoming messages (for
+example accepting numeric types other than AMQP long) but MUST use only the
```
duglin: is the MUST at the end in conflict with the MAY in the first line?  "MUST use only AMQP types" vs "MAY use AMQP types or MAY use strings"

clemensv:
LGTM

duglin:
Conditionally approved on the 10/10 call with a minor wording change that @deissnerk will make today.

While this missed the 2-day deadline the group decided to approve it on the call and give people until end of day tomorrow (friday) to raise any concerns, then when we get 2 LGTMs we'll merge.

duglin:
LGTM
can I get one more?

markpeek:
LGTM

duglin:
thanks @markpeek 
merging

PR #532, opened 2019-10-10
remove bad link for now

duglin:


duglin:
executive decision.... we need our CI to pass so I'm merging this

Issue #530, opened 2019-10-10
Clarify Kafka transport mapping of "content-type"


gunnarmorling:
Structured mode is described as a way for conveying messages completely self-contained, without relying on any Kafka headers. The example in [3.3.4](https://github.com/cloudevents/spec/blob/master/kafka-transport-binding.md#334-example) shows this as a header, though. I'll be happy to send a PR for doing the needed adjustments, but it's not quite clear to me yet what it should be.

duglin:
@bluemonk3y @clemensv 
Is the example ok because `content-type` is special like it is in http?

I'm actually more wondering about this line:
```
"datacontenttype" : "com.example.someevent",
```
shouldn't that be a valid type per RFC 2046  ?

gunnarmorling:
Yes, so there's both in this example, `datacontenttype` and `content-type`. The relationship between the two is not quite clear to me, I think this warrants some clarification. 

Also 3.3.1 speaks of "The [Kafka content-type] property field"; apart from the odd formatting with the brackets I'm not quite clear what a "property field" in the context of Kafka actually is. If it's indeed meant to be used as a message header (as indicated example, but a bit against the spirit of structured type), then I think it should be named like that.

bluemonk3y:
My interpretation is:
* `type`: represents the class or data type itself; the deserialized format of the class. For example if you decode a payload then a `com.company.Payment` is the data type.

* `datacontenttype` represents the serialization format for example, `application/avro` json etc

The AMQP spec uses type and  content-type respectively
https://github.com/cloudevents/spec/blob/master/amqp-transport-binding.md

However, it also seems like the 2 might be confused with each other (I had to look at other docs to confirm the above points which is a fail IMO ;)  Would `contentformat` or similar make more sense? It seems like type being used in the 2 attributes makes it confusing.

gunnarmorling:
@bluemonk3y, ok, so it actually is the case then that in _structured_ mode a Kafka message-level header should be used to convey the serialization format in RFC 2046 format?

If so, the followin sentence is confusing to me:

> The binary mode only applies to Kafka 0.11.0.0 and above, because Kafka 0.10.x.x and below lack support for message level headers.

The same applies to structured mode then, too, as otherwise the fact it's only given for _binary_ mode to me implies that structured would work with earlier versions. So I think this sentence should be removed then and be replaced by something stating that 0.11.0.0 is needed in both modes due the reliance on message headers.

gunnarmorling:
[...]

PR #534, opened 2019-10-11
#530 Clarification around datacontenttype

gunnarmorling:
Related to: https://github.com/cloudevents/spec/issues/530

Fixes: #530

### DIFF
--- a/kafka-transport-binding.md
+++ b/kafka-transport-binding.md
@@ -62,8 +62,6 @@ In the *structured* content mode, event metadata attributes and event data are
 placed into the Kafka message value section
 using an [event format](#14-event-formats).
 
-
-
 ### 1.4. Event Formats
 
 Event formats, used with the *structured* content mode, define how an event is
@@ -102,10 +100,10 @@ here.
 
 The receiver of the event can distinguish between the two content modes by
 inspecting the `ce_datacontenttype` [Header][Kafka-Message-Header] of the 
-Kafka message. If the value is prefixed with the CloudEvents media type
-`application/cloudevents`, indicating the use of a known
-[event format](#14-event-formats), the receiver uses *structured* mode, otherwise
-it defaults to *binary* mode.
+Kafka message. If no value is given, or if the value is prefixed with the
+CloudEvents media type `application/cloudevents`, indicating the use of a known
+[event format](#14-event-formats), the receiver uses *structured* mode,
+otherwise it uses *binary* mode.
 
 If a receiver finds a CloudEvents media type as per the above rule, but with an
 event format that it cannot handle, for instance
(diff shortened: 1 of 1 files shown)


duglin:
@gunnarmorling can you add a comment to this one? If nothing else at least a `Fixes #...` so we know the issue it's related to?

duglin:
related to: https://github.com/cloudevents/spec/issues/530

duglin:
@gunnarmorling @bluemonk3y where are we on this?

Also - rebase needed

bluemonk3y:
@duglin  - I left a comment on #530

kafka-transport-binding.md
```diff
@@ -102,10 +100,10 @@ here.
 
 The receiver of the event can distinguish between the two content modes by
 inspecting the `ce_datacontenttype` [Header][Kafka-Message-Header] of the
```
duglin: shouldn't this be `content-type` ?
@clemensv

kafka-transport-binding.md
```diff
@@ -102,10 +100,10 @@ here.
 
 The receiver of the event can distinguish between the two content modes by
 inspecting the `ce_datacontenttype` [Header][Kafka-Message-Header] of the 
-Kafka message. If the value is prefixed with the CloudEvents media type
-`application/cloudevents`, indicating the use of a known
-[event format](#14-event-formats), the receiver uses *structured* mode, otherwise
-it defaults to *binary* mode.
+Kafka message. If no value is given, or if the value is prefixed with the
```
duglin: If I'm correct, and it should be `content-type` above, then I think "no value" should default to binary not structured, and then the original text would make more sense.

kafka-transport-binding.md
duglin: @gunnarmorling if we fix this line would that address your confusion mentioned in #530 ?
I think the rest of the text in the spec is correct and doesn't need to be updated (except for the typo for `type`).

gunnarmorling: Yes, using `content-type` consistently as the name of the header property helps 👍

kafka-transport-binding.md
gunnarmorling: How will this work with Apache Kafka < 0.11 then though (also commented about this on your new PR)? Headers cannot be used there, so all the other `ce_...` attributes couldn't be given. Am I missing something?

gunnarmorling:
Superseded by #537; leaving it open until the pending discussions have been resolved.

gunnarmorling:
Closing, #537 addresses these concerns.

kafka-transport-binding.md
alanconway: MQTT <= 3.1 >= 4 has a similar issue. The approach there was MUST use JSON format and not support anything else on older MQTT versions.

PR #537, opened 2019-10-15
another try at clarifying kafka

### LINKED ISSUE #530: Clarify Kafka transport mapping of "content-type"

duglin:
Closes #530 
Closes #534 

### DIFF
--- a/kafka-protocol-binding.md
+++ b/kafka-protocol-binding.md
@@ -49,20 +49,19 @@ in the Kafka protocol as [Kafka messages][Kafka-Message-Format].
 The specification defines two content modes for transferring events:
 *structured* and *binary*.
 
-The *binary* mode *only* applies to Kafka 0.11.0.0 and above, because Kafka
-0.10.x.x and below lack support for message level headers.
-
-In the *binary* content mode, the value of the event `data` MUST be
-placed into the Kafka message's value section as-is, with the
-`ce_datacontenttype` header value declaring its media type; all other
-event attributes MUST be mapped to the Kafka message's 
-[header section][Kafka-Message-Header].
-
 In the *structured* content mode, event metadata attributes and event data are
-placed into the Kafka message value section
-using an [event format](#14-event-formats).
+placed into the Kafka message value section using an
+[event format](#14-event-formats).
 
+In the *binary* content mode, the value of the event `data` MUST be placed into
+the Kafka message's value section as-is, with the `content-type` header value
+declaring its media type; all other event attributes MUST be mapped to the
+Kafka message's [header section][Kafka-Message-Header].
 
+Implementations that use Kafka 0.11.0.0 and above MAY use either *binary*
+or *structured* modes. Implementations that use Kafka 0.10.x.x and below
+MUST use only use *structured* mode and encode the event in JSON. This is
+because older versions of Kafka lacked support for message level headers.
 
 ### 1.4. Event Formats
 
@@ -95,28 +94,32 @@ value is made available as [UTF-8][RFC3629] encoded JSON text.
 
 ## 3. Kafka Message Mapping
 
-With Kafka 0.11.0.0 and above, the content mode is chosen by the sender of the 
+With Kafka 0.11.0.0 and above, the content mode is chosen by the sender of the
 event. Protocol usage patterns that might allow solicitation of events using a
 particular content mode might be defined by an application, but are not defined
 here.
 
 The receiver of the event can distinguish between the two content modes by
-inspecting the `ce_datacontenttype` [Header][Kafka-Message-Header] of the 
-Kafka message. If the value is prefixed with the CloudEvents media type
-`application/cloudevents`, indicating the use of a known
-[event format](#14-event-formats), the receiver uses *structured* mode, otherwise
-it defaults to *binary* mode.
+inspecting the `content-type` [Header][Kafka-Message-Header] of the
+Kafka message. If the header is present and its value is prefixed with the
+CloudEvents media type `application/cloudevents`, indicating the use of a known
+[event format](#14-event-formats), the receiver uses *structured* mode,
+otherwise it defaults to *binary* mode.
 
 If a receiver finds a CloudEvents media type as per the above rule, but with an
 event format that it cannot handle, for instance
 `application/cloudevents+avro`, it MAY still treat the event as binary and
 forward it to another party as-is.
 
+If the `content-type` header is not present then the receiver uses
+*structured* mode with the JSON event format.  
+
 ### 3.1. Key Attribute
-The 'key' attribute is populated by a partitionKeyExtractor function. The 
-partitionKeyExtractor is a protocol specific function that contains bespoke logic 
-to extract and populate the value. A default implementation of the extractor will 
-use the [Partitioning](extensions/partitioning.md) extension value. 
+The 'key' attribute is populated by a partitionKeyExtractor function. The
+partitionKeyExtractor is a protocol specific function that contains bespoke
+logic to extract and populate the value. A default implementation of the
+extractor will use the [Partitioning](extensions/partitioning.md) extension
+value.
 
 ### 3.2. Binary Content Mode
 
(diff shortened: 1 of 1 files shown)


duglin:
ping @bluemonk3y @gunnarmorling    does this help?

kafka-protocol-binding.md
```diff
@@ -95,28 +94,29 @@ value is made available as [UTF-8][RFC3629] encoded JSON text.
...
 here.
 
 The receiver of the event can distinguish between the two content modes by
-inspecting the `ce_datacontenttype` [Header][Kafka-Message-Header] of the 
+inspecting the `content-type` [Header][Kafka-Message-Header] of the
 Kafka message. If the value is prefixed with the CloudEvents media type
 `application/cloudevents`, indicating the use of a known
-[event format](#14-event-formats), the receiver uses *structured* mode, otherwise
-it defaults to *binary* mode.
+[event format](#14-event-formats), the receiver uses *structured* mode,
```
gunnarmorling: What's the intended behavior if there is *no* `content-type` header at all? We cannot default to binary in that case, as then things wouldn't work with AK < 0.11 at all (no header support as you described above). I tried to address that in my original PR by saying if the header property isn't provided, then the receiver uses structured mode with the JSON event format.

So how about adding this bit to make this clear:

> If no value is present, the receiver uses *structured* mode with the JSON event format. If the value is prefixed with the CloudEvents media type...

kafka-protocol-binding.md
```diff
@@ -202,8 +202,8 @@ hops, and across multiple protocols.
 
 #### 3.3.1. Kafka Content-Type
 
-The [Kafka `content-type`] property field MUST be set to the media
-type of an [event format](#14-event-formats).
+The [Kafka `content-type`] property field MUST be set to the media type of an
```
gunnarmorling:
```suggestion
The Kafka message header property `content-type` MUST be set to the media type of an
```

kafka-protocol-binding.md
```diff
@@ -246,12 +246,16 @@ content-type: application/cloudevents+json; charset=UTF-8
 
 {
     "specversion" : "1.0-rc1",
-    "datacontenttype" : "com.example.someevent",
+    "type" : "com.example.someevent",
+	"source" : "/mycontext/subcontext",
+	"id" : "1234-1234-1234",
+	"time" : "2018-04-05T03:56:24Z",
```
gunnarmorling: The indentation is a bit off due to tabs instead of whitespace.

kafka-protocol-binding.md
```diff
@@ -246,12 +246,16 @@ content-type: application/cloudevents+json; charset=UTF-8
...
+    "datacontenttype" : "application/json",
```
gunnarmorling: Perhaps use application/xml to make apparent that the data content type can differ from the content type of the message itself (similar to how it's shown in the JSON event format spec)?

kafka-protocol-binding.md
duglin: What do people normally do when an Kafka message comes in w/o a content-type? Is it totally undefined and they're on their own? For AK < 0.11 we require things to use structured+JSON - would it be horrible to do the same when content-type is there for AK > 0.11 (meaning require JSON)?

kafka-protocol-binding.md
```diff
@@ -184,12 +184,12 @@ ce_type: "com.example.someevent"
...
 ce_id: "1234-1234-1234"
 ce_time: "2018-04-05T03:56:24Z"
-ce_datacontenttype: application/avro
+content-type: application/avro
        .... further attributes ...
 
 ------------------- value --------------------
 
-            ... application data ...
+            ... application data encoded in AVRO ...
```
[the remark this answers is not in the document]
duglin: done

kafka-protocol-binding.md
duglin: done

kafka-protocol-binding.md
duglin: done

kafka-protocol-binding.md
duglin: Done. What's funny about you focusing on this is that I was tempted to add a sentence in there about how datacontenttype is optional in this case since the data payload is encoded the same way as the envelope - but decided not to at the last minute.  Using "XML" doesn't give me that option :-)

kafka-protocol-binding.md
```diff
@@ -202,8 +202,8 @@ hops, and across multiple protocols.
 
 #### 3.3.1. Kafka Content-Type
 
-The [Kafka `content-type`] property field MUST be set to the media
-type of an [event format](#14-event-formats).
+The Kafka message header proeprty `content-type` MUST be set to the media type
```
gunnarmorling:
```suggestion
The Kafka message header property `content-type` MUST be set to the media type
```

duglin: sorry -fixed

kafka-protocol-binding.md
gunnarmorling: So actually this is what made stumble upon all this when reading it initally. AFAIK, it's not that `content-type` is a well-defined or commonly used header property in Kafka (of course I might be wrong and just not have heard about it). Really message key and value are binary data in Kafka and producers and consumers must agree on what format is used for a specific topic. So yes, if consumers don't have the knowledge about the content type either by means of such header property or out-of-bands exchange and agreement of the content type, they are on their own.

JSON seems like a reasonable default to me in the context of CloudEvents. The expectation is (AFAICS) that most people by now are on Kafka versions with header support, making the problem disappear.

kafka-protocol-binding.md
gunnarmorling: Yes, I was thinking the same :) Either get rid of it or make actually use of it to define a different content type for data.

kafka-protocol-binding.md
duglin: ok see what you think of my latest edits

gunnarmorling: Yes, all looking good to me now. Thanks!

Maybe just on 3.3.1 though:

> The Kafka message header property content-type MUST be set to the media type of an event format.

It MUST be set to such value when it actually is given, but it's not that it MUST be given at all. I assume that's what you mean, but it currently may be understood as if that header property is mandatory, which it isn't.

Perhaps this: "If given, the Kafka message header property..."?

Also fine for me if you think that the current wording actually doesn't preclude it being not given.

duglin: done!

gunnarmorling:
@duglin, on a related note: the JSON event format spec also speaks of a `data_base64` attribute as an alternative to `data`. This one is never referenced here. I suppose that if a message with `data_base64` is sent to Kafka in binary mode, its value should go into the Kafka message value. Should this be reflected here, or is `data` in the context here a bit more abstract and actually refers to either one of `data` or `data_base64` in the JSON event format sense?

kafka-protocol-binding.md
gunnarmorling: Cool, this PR LGTM then. Thanks!

duglin:
>  I suppose that if a message with data_base64 is sent to Kafka in binary mode, its value should go into the Kafka message value

Right. The attribute named `data_base64` should only ever appear in the JSON serialization and in structured mode.  In binary no attribute name is needed, we just stick the raw data directly into the payload/message-value.

> Should this be reflected here, or is data in the context here a bit more abstract and actually refers to either one of data or data_base64 in the JSON event format sense?

yup exactly. `data` is the abstract attribute and how it appears (under "data", or "data_base64" or raw in the payload) will vary based on how the event is serialized.

gunnarmorling:
> data is the abstract attribute...

Gotcha. All LGTM then, thanks for the PR and going through the review iterations!

clemensv:
LGTM. Good clarification.

gunnarmorling:
@duglin, @bluemonk3y, I won't be able to attend the call today unfortunately, but as stated above the PR LGTM. Thanks again!

@bluemonk3y, one thing I'm kinda curious about is whether the `content-type` header name should somehow be more "well-defined" on the Kafka side of things (unless it already is), e.g. presented in the docs and/or via constants in the client APIs. If so, it'd have to allow for distinguishing between key and value content types, too. But that's a discussion to have somewhere else I suppose.

bluemonk3y:
@gunnarmorling  - regarding content-type. There are 2 aspects to consider here.
1. CloudEvents are self-describing (or stand alone). They carry all required information to decode, trust, process, route etc. In that sense it changes how we think about configuration. Traditionally, a KafkaSerDes defines the type, schema etc using config and properties. This semantic changes that and the SerDes needs to (at runtime) load and process the CE properly. 
2. Passing a `content-type` in the header makes sense. Kafka headers are arbitrary and without official standards. I believe we can presume (or inherit) existing standards as part of this. SpringCloudStream Kafka already does this and so do other Kafka wrappers. You raise an excellent point about them relating to Key or Value. So far, its been traditionally applied to the value. I presume we can discuss "key-content-type" later on, in that sense we start getting transport specific.

PR #539, opened 2019-10-16
adding contrib entry

bluemonk3y:
Signed-off-by: Neil Avery <neil@confluent.io>


duglin:
exec decision - merging

PR #567, opened 2020-02-06
Paragraph about nested events to the primer

### LINKED ISSUE #72: Nested events?
Events occur on different layers of a system. This might lead to nesting of events. E.g. an API gateway may receive an HTTP POST from a web hook, which already contains event data from a SaaS application. In addition the API gateway may be configured to raise an event to a function, when a POST request to a certain URL is received. 
How is this situation handled?

## Alternative 1: Nesting
The API gateway wraps the SaaS application event into an API gateway event and sends it to the function.

## Alternative 2: Forwarding
The API gateway recognizes the event context in the HTTP header and forwards the application event to the function using the protocol of choice. No API gateway event would be triggered. 

## Alternative 3: ?

deissnerk:
To finally resolve #72 , I added a paragraph to the primer.


primer.md
```diff
@@ -399,6 +399,31 @@ original event source. And as such, it is expected that CloudEvents attributes
 related to the event producer (such as 'source`and`id`) would be changed from
 the incoming CloudEvent.
 
+There might exist special cases in which it is necessary to create a CloudEvent
```
duglin: Minor thing, and I don't have a strong opinion but since it popped in my head, should we add a `### Nested CloudEvents` header here so that this stands out a bit?

duglin:
Approved on the 2/13 call

PR #577, opened 2020-03-01
How to determine binary CE vs random non-CE message

duglin:
Add some guidance for how to determine when an incoming message is a binary
CE vs some random non-CE message. The same receiver might accept CE and non-CE
messages so we should have an interoperable way for them to know what they're
getting.

If people are ok with this general direction then I can make the same change to the other transport specs.

### DIFF
--- a/http-protocol-binding.md
+++ b/http-protocol-binding.md
@@ -147,6 +147,16 @@ If a receiver detects the CloudEvents media type, but with an event format that
 it cannot handle, for instance `application/cloudevents+avro`, it MAY still
 treat the event as binary and forward it to another party as-is.
 
+When the `Content-Type` header is not prefixed with the CloudEvents media type,
+being able to know when the message ought to be attempted to be parsed as a
+CloudEvent can be a challenge. While this specification can not mandate that
+senders do not include any of the CloudEvents HTTP headers when the message
+is not a CloudEvent, it would be reasonable for a receiver to assume that if
+the message has all of the mandatory CloudEvents attributes as HTTP headers
+then it's probably a CloudEvent. However, as with all CloudEvent messages, if
+it does not adhere to all of the normative language of this specification
+then it is not a valid CloudEvent.
+
 ### 3.1. Binary Content Mode
 
 The _binary_ content mode accommodates any shape of event data, and allows for
--- a/amqp-protocol-binding.md
+++ b/amqp-protocol-binding.md
@@ -115,6 +115,16 @@ If a receiver detects the CloudEvents media type, but with an event format that
 it cannot handle, for instance `application/cloudevents+avro`, it MAY still
 treat the event as binary and forward it to another party as-is.
 
+When the `content-type` message property is not prefixed with the CloudEvents
+media type, being able to know when the message ought to be attempted to be
+parsed as a CloudEvent can be a challenge. While this specification can not
+mandate that senders do not include any of the CloudEvents message properties
+when the message is not a CloudEvent, it would be reasonable for a receiver
+to assume that if the message has all of the mandatory CloudEvents attributes
+as message properties then it's probably a CloudEvent. However, as with all
+CloudEvent messages, if it does not adhere to all of the normative language of
+this specification then it is not a valid CloudEvent.
+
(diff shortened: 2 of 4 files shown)


http-protocol-binding.md
```diff
@@ -147,6 +147,14 @@ If a receiver detects the CloudEvents media type, but with an event format that
 it cannot handle, for instance `application/cloudevents+avro`, it MAY still
 treat the event as binary and forward it to another party as-is.
 
+When the `Content-Type` header is not prefixed with the CloudEvents media type,
+being able to distinguish between a _binary_ CloudEvent and a non-CloudEvent
+message can be a challenge. While this specification can not mandate that
+senders do not include any of the CloudEvents HTTP headers when the message
+is not a CloudEvent, it would be a reasonable for a receiver to assume that if
+the message has all of the mandatory CloudEvents attributes as HTTP headers
```
slinkydeveloper: What if it doesn't have all of the mandatory fields? Should be considered non-event or not valid event? Maybe is it more convenient to state: if a request has an id attribute, then it should be handled as event (even not valid one)

http-protocol-binding.md
```diff
@@ -147,6 +147,14 @@ If a receiver detects the CloudEvents media type, but with an event format that
 it cannot handle, for instance `application/cloudevents+avro`, it MAY still
 treat the event as binary and forward it to another party as-is.
 
+When the `Content-Type` header is not prefixed with the CloudEvents media type,
+being able to distinguish between a _binary_ CloudEvent and a non-CloudEvent
```
slinkydeveloper: So maybe the word "distinguish", at least to me, is unclear. With distinguish do you mean "start the parsing, even of a non valid event" or "start the parsing of a valid event?".

If I understood the goal here, you want to give an hint to the spec implementer on how to implement the parsing process of the HTTP envelope. If that's the case, I think it would be useful to specify that if it has the specversion, then it's an event, potentially non valid

http-protocol-binding.md
duglin: If the incoming message doesn't have all (4) required attributes then it is not a valid CE. While SDKs might choose to treat it as a CE but flag it as "incomplete", that is there choice. But from a spec perspective I don't think we can say anything other than all 4 are required to be a CE, anything else isn't a CE.

http-protocol-binding.md
duglin: See comment below, from a spec perspective, if an incoming CE doesn't adhere to all of the MUSTs in the spec then it is not a valid CE. What a receiver chooses to do with it at that point is an impl choice - the spec is silent on that because we purposely decided not to get into the processing model side of things (for the most part).

So, I guess "distinguish" can be interpreted as "whether someone should try to parse it as a CE" yes, but even then there's no guarantee that it'll pass all of the MUST checks. I've added a little of of rewording -  see what you think.

slinkydeveloper: I like the rewording, thanks!

http-protocol-binding.md
```diff
@@ -147,6 +147,16 @@ If a receiver detects the CloudEvents media type, but with an event format that
 it cannot handle, for instance `application/cloudevents+avro`, it MAY still
 treat the event as binary and forward it to another party as-is.
 
+When the `Content-Type` header is not prefixed with the CloudEvents media type,
+being able to know when the message ought to be attempted to be parsed as a
+CloudEvent can be a challenge. While this specification can not mandate that
+senders do not include any of the CloudEvents HTTP headers when the message
+is not a CloudEvent, it would be a reasonable for a receiver to assume that if
+the message has all of the mandatory CloudEvents attributes as HTTP headers
+then it probably a CloudEvent. However, as with all CloudEvent messages, if
```
JemDay: it -> it's

duglin:
on 3/12 call, there was agreement with the overall direction - so next step is to modify the other transport specs appropriately.

http-protocol-binding.md
duglin: fixed - thanks

duglin:
Updated all transport specs except nats. PTAL

duglin:
Approved on the 3/26 call.

PR #580, opened 2020-03-02
[WIP] Drafting the clouddiscovery API specification

mikehelmick:


discovery.md
```diff
@@ -0,0 +1,378 @@
...
+`producer` and `type`. This takes a denormalized view where the composite key
+of documents is the `producer` and `type`. Producer is defined as an arbitrary
+human readable string and type is the CloudEvent type attribute.
+
+// TODO
+// 1) need to rationalize fanout, broker aggregator scenario
+// 2) Does discovery need to be 2 levels? - producers -> types -> sources
+
+### Producer entity Attributes
+
```
deissnerk: It's interesting that producer becomes the first-class citizen for discovery, while we intentionally left it out of the event context for CloudEvents. I understand that you need something to enumerate on in addition to type, and that source is not suitable for this. Maybe I am really thinking too much about terminology, but could this be a _Provider_?

discovery.md
```diff
@@ -0,0 +1,378 @@
...
+- Constraints:
+  - OPTIONAL
+  - If present, must be non-empty
+- Examples:
+  - "Github"
+  - "Awesome Cloud Storage"
+  - "Awesome Cloud\*"
+
+##### source
+
```
deissnerk: Does this only apply to `source`, or is it matched against `sourcestructure`, too?

discovery.md
```diff
@@ -0,0 +1,378 @@
...
+- [Notations and Terminology](#notations-and-terminology)
+- [API Specification](#api-specification)
+- [Protocol Bindings](#protocol-bindings)
+- [Privacy & Security](#privacy-and-security)
+
+## Overview
+
+The goal of the CloudDiscovery API is to enable connection between consumers of
+events and producers / aggregators / brokers of events and to facilitate the
+creation of CloudSubscriptions. The discovery and subscription APIs are
```
duglin: I'd prefer if we didn't talk about control-plane vs data-plane aspects since I think that'll be an implementation detail. Our job is to just define the externals surface (APIs).

discovery.md
```diff
@@ -0,0 +1,378 @@
...
+control-plane APIs to enable events to be delivered over the eventing
+data-plane.
+
+Discovery must allow for an event producer or middleware component to advertise
```
duglin: we'll need to add the RFC2119 boilerplate text and go thru and verify each use of the keywords - but not needed for the first PR. Just FYI

discovery.md
```diff
@@ -0,0 +1,378 @@
...
+cloud provider where all producers produce events, there may be dozens of
+sources (producer systems) and hundreds of event types. The discovery funnel of
+producer first, then event types helps users navigate without having to see large
+lists of event types. Both of these cases require that filters can be used
+in the discovery API to narrow down the selection of available events.
+
+The CloudEvent source attribute is a potential source of high fanout. For
+example consider a blob storage system where each directory constitutes a
+distinct `source` attribute value. For this reason, source is not the starting
+point for discovery, but an attribute that can be used in creating subscriptions.
```
duglin: It might be ok to say that `source` isn't the starting point, I would guess the producer is, so are you saying that people won't be able to traverse from "producer" to "source" to "events" via the discovery API?

Asking for events from just one bucket/dir seems like a fairly popular use case, I think it be natural to then ask for the list of events that a bucket/dir can generate? This way they don't see events that don't apply to other resources.

duglin: Are you envisioning "source" being a "resource type" or a specific instance of a "resource type"? I'm guessing "resource type", which would reduce the fan-out, no?

discovery.md
```diff
@@ -0,0 +1,378 @@
...
+producers. These take two forms:
+
+1. Middleware
+2. Aggregator (or *Events Broker*)
+
+In the Middleware case, discovery is a combined set of producers for where the
+discovery is passed through largely unmodified and subscriber URIs are generally
+absolute (back to the producer).
+
+In the aggregator use case, the aggregator is expected to modify the output of
```
duglin: I suspect this section will probably need more text and as such would probably be really good info for a primer since from a spec perspective we shouldn't care what the specific values are for the URIs - we just need to make sure we can support multiple implementation models.

Net: I think we can mention that the spec will support many implementation choices (such as the ones mentioned here), but then leave the guidance bits for the primer.

discovery.md
```diff
@@ -0,0 +1,378 @@
...
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+### Terminology
+
+This specification defines the following terms:
+
+### Producer
```
duglin: these probably need 4 #'s so they're one more level down

discovery.md
```diff
@@ -0,0 +1,378 @@
...
+
+### Intermediary
+
+An "intermediary" receives a message containing an event for the purpose of
+forwarding it to the next receiver, which might be another intermediary or a
+consumer. A typical task for an intermediary is to route the event to receivers
+based on the information in the event context.
+
+### Middleware
+
```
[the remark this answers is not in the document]
duglin: Did a quick search of the term "middleware" in the spec and it's really not used very much at all - and definitely not in any normative way. So, I'm wondering if we need to define it at all. I'd prefer to just talk about the two actors involved:
1 - entity trying to discover what's available - typically so they can subscribe
2 - the discovery endpoint. Whether it's the actual producer, middleware, broker... from a spec perspective we don't care what it's called or how people think of it, it's the same API.

discovery.md
```diff
@@ -0,0 +1,378 @@
...
+Specific to discovery and subscription capabilities. Events middleware systems
+provide discovery services that may be directly passing through discovery from
+producers, or may aggregate or otherwise alter the discovery. This includes
+events brokers that may also be relaying discovery AND subscription information
+further upstream to event producers. Multiple middleware systems may exist
+between the producer and the consumer.
+
+### Consumer
+
+The entity receiving the event - the destination for the events related to a
```
duglin: Could we just use the CE definition here so that we don't get all tangled-up in the push vs pull aspects?

discovery.md
```diff
@@ -0,0 +1,378 @@
...
+
+The discovery API takes the declarative API design approach. This document
+structure focuses on discovery where each document has a two part key of
+`producer` and `type`. This takes a denormalized view where the composite key
+of documents is the `producer` and `type`. Producer is defined as an arbitrary
+human readable string and type is the CloudEvent type attribute.
+
+// TODO
+// 1) need to rationalize fanout, broker aggregator scenario
+// 2) Does discovery need to be 2 levels? - producers -> types -> sources
```
duglin: I wonder if it wouldn't be `producers -> source types -> types`

discovery.md
```diff
@@ -0,0 +1,378 @@
...
+
+### Producer entity Attributes
```
duglin: I'm struggling with how this API works, and I admit, it could just be I need to be educated.

Let's take the scenario where the Provider is static w.r.t. the types of resources that can produce events, and the list of event types. For example, let's take an object store, with buckets and objects and CRUD events for each. That's a well defined list.

I kind of feel like there should be two flavors of query APIs:
1 - one that lists the types of resources (buckets and objects) and the types of events for each (CRUD). And in this respect the results from a query to this side of things will generally always return the same data because it's static.
2 - one that lists the specific instances of each resources - e.g. the list of buckets or objects I can see and subscribe to. And this one, almost, feels out of scope for us because I would imagine that each Provider already offers a way for people to "list their resources". But, in the interest of allowing full interop perhaps we could suggest simple List() type of APIs.

This popped in my head because below it lists things like 'source' as an attribute and (ignoring ACLs that might change which source+event each user can see), I would think the same set of events should be visible on all buckets.  So, if I think about this as a table I would prefer to see:

| provider | source resource | event |
| --- | --- | --- |
| store | bucket | created |
| store | bucket | deleted |
| store | object | created |
| store | object | deleted |

instead of:

| provider | source resource | event |
| --- | --- | --- |
| store | amys-bucket | created |
| store | amys-bucket | deleted |
| store | johns-bucket | created |
| store | johns-bucket | deleted |
... repeat for all buckets ...
| store | amys-bucket/object1 | created |
| store | amys-bucket/object1 | deleted |
... repeat for all objects in all buckets ...

In other words, I would expect one query to give me a (sort of) schema, and then other queries to give me the specific values I can put into that schema info when I create the subscription.

Am I explaining this well enough? :-)

discovery.md
duglin: Kind of related, one thing that's missing from the spec right now is how the data is returned? Is it just list where each entry is a flat list of attributes, or were people expecting some kind of object structure to it?  While not fancy, I'm partial to something that looks more like a table being returned because it's easy to process and also very easy to implement the search query - becomes a trivial SQL-like thing w/o concern for nested objects and how to express that (on input or output).

discovery.md
deissnerk: I like `source types`. I've been thinking a lot about this event catalog topic over the past months, and I often get to the question, how to categorize and identify (groups of) sources.

discovery.md
deissnerk: Yes, I think I get your point, and I agree. 😄

discovery.md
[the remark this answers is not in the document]
mikehelmick: just removed this section.

discovery.md
```diff
@@ -0,0 +1,378 @@
...
+- Description: Identifies the producer of the event. The producer string SHOULD
+  be human readable and can identify a top level producer system.
+- Constraints:
+  - REQUIRED
+  - MUST be a non-empty string
+- Examples:
+  - "GitHub.com"
+  - "Awesome Cloud Storage"
+  - "com.example.microservices.userlogin"
+
```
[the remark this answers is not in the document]
mikehelmick: I don't think they're counter to the definition. The intent here is that these could be human readable for building a discovery/subscription UX.

discovery.md
mikehelmick: My thought process here - going back to the discovery funnel [in the doc](https://docs.google.com/document/d/10S2MnYiy8wNBO8fXdtAviX0GvJXeoFqNOHG6Qh1-EAs/edit?ts=5e1767a6#heading=h.ihz1tqidrou8l)

As someone wanting to consume events, I see a list of "producers"  - systems that I can rationalize and understand.

Then I would select an event type from that producers

Then, if necessary, select a source from available sources for that producer + event type.

discovery.md
mikehelmick: just to source - `sourcestructure` is meant to be informational. 

I don't see a use case for including it in the query, but I could be missing something.

discovery.md
mikehelmick: ack. I did remove all uses of "middleware"

discovery.md
mikehelmick: I think the most likely discovery flow is 

producer -> type -> source

discovery.md
mikehelmick: removed this todo, wasn't meant to be left in

discovery.md
mikehelmick: added examples in the doc for my thinking on this

discovery.md
mikehelmick: I did rework this section

discovery.md
```diff
@@ -0,0 +1,436 @@
...
+- [Overview](#overview)
+- [Notations and Terminology](#notations-and-terminology)
+- [API Specification](#api-specification)
+- [Protocol Bindings](#protocol-bindings)
+- [Privacy & Security](#privacy-and-security)
+
+## Overview
+
+The goal of the CloudDiscovery API is to enable connection between consumers of
```
duglin: s/connection/connections/

discovery.md
```diff
@@ -0,0 +1,436 @@
...
+events and producers / aggregators / brokers of events and to facilitate the
+creation of CloudSubscriptions.
+
+Discovery must allow for an event producer or intermediary component to
```
duglin: s/must allow/allows/

discovery.md
```diff
@@ -0,0 +1,436 @@
...
+advertise the event types that are available, provide the necessary information
+to consume that event (schema / delivery protocol options), and the necessary
+information to create a subscription. The output of the API must be such that
```
duglin: s/must be/ought to be/

discovery.md
```diff
@@ -0,0 +1,436 @@
...
+
+The second case becomes relevant if multiple providers support the same event
+types. Use case 1 is likely the dominant use case. Given the example of a public
+cloud provider where all producers produce events, there may be dozens of
+sources (producer systems) and hundreds of event types. The discovery funnel of
+producer first, then event types helps users navigate without having to see large
+lists of event types. Both of these cases require that filters can be used
+in the discovery API to narrow down the selection of available events.
+
+The CloudEvent source attribute is a potential source of high fanout. For
```
duglin: s/source/`source`/ just on the first use

discovery.md
```diff
@@ -0,0 +1,436 @@
...
+example consider a blob storage system where each directory constitutes a
+distinct `source` attribute value. For this reason, source is not the starting
```
duglin: s/source/`source`/ - 2nd usage

duglin: 2 lines down too

discovery.md
```diff
@@ -0,0 +1,436 @@
...
+
+1. Aggregator
+2. Events Broker
+
+In the aggregator case, discovery is a combined set of producers for where the
+discovery is passed through largely unmodified and subscriber URIs are generally
+absolute (back to the producer).
+
+In the events broker use case, it is allowable to modify the output of
+discovery and subscription URIs will be relative to the broker itself.
```
duglin: Do you think the consumer will need to know when each of these two cases is actually in use? I was kind of assuming they wouldn't, and if that's the case then it might be good to add a sentence at the end mentioning that so people understand that these two cases are just mentioned to help explain the use cases for the spec and not because it'll impact consumer implementation details.

discovery.md
```diff
@@ -0,0 +1,436 @@
...
+The entity managing the lifecycle of a Subscription on behalf of an Event
+Consumer. In some instances this might be the same entity as the Event Consumer.
+
+## API Specification
+
+The discovery API takes the declarative API design approach. This document
+structure focuses on discovery where each document has a two part key of
+`producer` and `type`. This takes a denormalized view where the composite key
+of documents is the `producer` and `type`. Producer is defined as an arbitrary
+human readable string and type is the CloudEvent type attribute.
```
duglin: s/type/`type`/g

discovery.md
```diff
@@ -0,0 +1,436 @@
...
+  - If present, MUST be a non-empty URI
+- Examples:
+  - "http://cloud.example.com/docs/blobstorage"
+
+#### source
+
+- Type: `List` of `String`
+- Description: CloudEvents [`source`](https://github.com/cloudevents/spec/blob/master/spec.md#source-1)
+  attribute values available from this producer for this event type. Values here
+  may be used to create subscriptions from this producer.
```
duglin: s/may/can/

discovery.md
```diff
@@ -0,0 +1,436 @@
...
+- Examples:
+  - Single Internet-wide unique URI with a DNS authority.
+   - ["https://github.com/cloudevents"]
+  - Application-specific identifiers
+   - ["/sensors/tn-1234567/alerts", "/sensors/tr-1234567/alerts"]
+
+#### sourcestructure
+
+- Type: `String`
+- Description: Documentation of the format of the source attribute for this
```
duglin: Is the intent that we define the template language used?

discovery.md
```diff
@@ -0,0 +1,436 @@
...
+- Examples:
+  - "com.github.pull.create"
+  - "com.example.object.delete.v2"
+
+#### specversion
+
+- Type: `String` per [RFC 2046](https://tools.ietf.org/html/rfc2046)
+- Description: CloudEvents [`specversion`](https://github.com/cloudevents/spec/blob/master/spec.md#specversion)
+  that will be used for events published for this producer, event type
+  combination. Compliant event producers MUST use a value of `1.0` when
```
duglin: Is there any reason to lock this spec down to just v1.0 of CE?

discovery.md
```diff
@@ -0,0 +1,436 @@
...
+  the `data` attribute of subscribed events.
+- Constraints:
+  - OPTIONAL
+  - If present, MUST be a non-empty URI
+
+#### dataschemacontent
+
+- Type: `String`
+- Description: An inline representation of the schema of the `data` attribute
+  encoding mechanism. This is an alternative to using the `dataschema`
```
duglin: We may need to define the language/spec used

discovery.md
```diff
@@ -0,0 +1,436 @@
...
+- Examples:
+  - ["HTTP"]
+  - ["HTTP", "AMQP", "KAFKA"]
+
+#### extensions
+
+- Type: `Map` of `String` to `String`
+- Description: Associative map of CloudEvents [Extension Content Attributes](https://github.com/cloudevents/spec/blob/master/spec.md#extension-context-attributes)
+  that are used for this event `type`. Keys MUST be confirm to the extension
+  content attributes naming rules and value are the type of the extension
```
duglin: s/content/context/ I think

discovery.md
```diff
@@ -0,0 +1,436 @@
...
+  attribute, confirming to the CloudEvents [type system](https://github.com/cloudevents/spec/blob/master/spec.md#type-system).
```
duglin: s/confirming/conforming/

discovery.md
```diff
@@ -0,0 +1,436 @@
...
+- Constraints:
+  - OPTIONAL
+
+#### subscriptionendpoint
+
+- Type: `URI-reference`
+- Description: URI indicating where CloudSubscriptions `subscribe` API calls
+  should be sent to.
```
duglin: s/should/MUST/

discovery.md
```diff
@@ -0,0 +1,436 @@
...
+  confirming to the CloudEvents [type system](https://github.com/cloudevents/spec/blob/master/spec.md#type-system).
+  TODO: Needs resolution with CloudSubscriptions API
+- Constraints:
+  - OPTIONAL
+- Examples:
+
+#### subscriptionttl
+
+- Type: `Integer`
+- Description: Time (in seconds) after which subscriptions of this will will
```
duglin: s/of this will will/to this producer will/   perhaps?

discovery.md
```diff
@@ -0,0 +1,436 @@
...
+  for each producer entity that matches the query.
+- Constraints:
+  - OPTIONAL
+  - If absent, a default value of false is used.
+
+## Example Usage
+
+For these examples, we'll assume discovery is being served from an events broker
+that has access to events from multiple producer systems. Specifically, we will
+show these producers, each with these event types.
```
duglin: s/with these/with the specified/

discovery.md
```diff
@@ -0,0 +1,436 @@
...
+ * `cloud.storage.object.update`
+* "Cloud RDBMS"
+ * `cloud.rdbms.row.insert`
+ * `cloud.rdbms.row.delete`
+ * `cloud.rdbms.table.create`
+ * `cloud.rdbms.table.drop`
+
+### Query for available producers.
+
+In order to configure a subscription, a user wants to brows available events
```
duglin: s/brows/browse/

discovery.md
```diff
@@ -0,0 +1,436 @@
...
+  - "com.github.pull.\*"
+
+##### expandsources
+
+- Type: boolean
+- Description: If true, the `source` response attribute will be expanded
+  for each producer entity that matches the query.
+- Constraints:
+  - OPTIONAL
+  - If absent, a default value of false is used.
```
duglin: s/false/`false`/

discovery.md
```diff
@@ -0,0 +1,436 @@
...
+The request for events from an Event Producer system.
+
+#### Event Subscriber
+
+The entity managing the lifecycle of a Subscription on behalf of an Event
+Consumer. In some instances this might be the same entity as the Event Consumer.
+
+## API Specification
+
+The discovery API takes the declarative API design approach. This document
```
duglin: It's not clear to me what the sentence about "declarative API design approach" means in this context since I think all we're really designing is an API with a set of inputs and outputs - not the processing model behind it. Right?

discovery.md
```diff
@@ -0,0 +1,436 @@
...
+advertise the event types that are available, provide the necessary information
+to consume that event (schema / delivery protocol options), and the necessary
+information to create a subscription. The output of the API must be such that
+tooling can be built where all possible event producers and types aren’t known
+in advance.
+
+There are several discovery use cases to consider from the viewpoint of event
+consumers.
+
+1. What sources are available, and what event types do they produce, and
```
akashrv: It seems like "sources" and the CE-source attribute are different in this sentence. Is source == producer? Will help to provide links to producer and source definition if they are present somewhere else.

discovery.md
mikehelmick: changed second use to "cause"

discovery.md
mikehelmick: Good question for the group.

That wasn't my intention as the items in there can't necessarily be interpreted by tooling.

discovery.md
mikehelmick: should we make it 0.3 and 1.0? Since that's what SDKs will support?

discovery.md
mikehelmick: added "dataschematype"

discovery.md
duglin: Ah, in that case we may want to say that this is for human consumption

discovery.md
duglin: I was more worried about v1.1, v1.2... coming out and this spec not claiming support for them. Do we need to put any constraints on it at all?

discovery.md
```diff
@@ -0,0 +1,448 @@
...
+
+### Query for available producers.
+
+In order to configure a subscription, a user wants to browse available events
+on this broker. We start by listing the available producers, an empty discovery
+query and `expandsources` set to `false`
+
+`/discovery?expandsources=false`
+
+The returned data would correspond to this table. Some attributes are omitted.
```
akashrv: Is it fair to assume that the response will have all the attributes of the producer entity?

discovery.md
```diff
@@ -0,0 +1,448 @@
...
+
+The discovery API takes the declarative API design approach. This document
+structure focuses on discovery where each document has a two part key of
+`producer` and `type`. This takes a denormalized view where the composite key
+of documents is the `producer` and `type`. Producer is defined as an arbitrary
+human readable string and `type` is the CloudEvent `type` attribute.
+
+### Producer entity Attributes
+
+The `Producer` entity is the main component of the discovery API. This section
```
akashrv: Is this producer or source? Referring to the definitions that are listed in the document above.
Producer=The "producer" is a specific instance, process or device that creates the data structure describing the CloudEvent.
Source=The "source" is the context in which the occurrence happened. In a distributed system it might consist of multiple Producers. If a source is not aware of CloudEvents, an external producer creates the CloudEvent on behalf of the source.

discovery.md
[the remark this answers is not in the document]
mikehelmick: This was renamed to provider to not conflict with the CE producer terminology. This is really meant to be part of the discovery journey and the way that a user things of the grouping.

discovery.md
mikehelmick: Removing source for now, based on feedback from others.

discovery.md
mikehelmick: removing source from current RC of discovery

discovery.md
mikehelmick: In trying to add wording or rewrite this, I decided it is best removed.

discovery.md
mikehelmick: ok, based on some feedback I got - I removed this for this cut.

discovery.md
mikehelmick: actually - going to remove this for now.

discovery.md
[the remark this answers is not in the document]
mikehelmick: thanks - clarified that this is about providers and has nothing to do with ce-source

discovery.md
mikehelmick: removed sources

discovery.md
mikehelmick: renamed this throughout to provider.

references to source are gone.

duglin:
Approved as initial draft on the 3/26 call.
Thanks @mikehelmick and sub-team!!

PR #582, opened 2020-03-03
Fix ordering of elements for id attribute. Make consistent with rest …

mikehelmick:
…of spec.

Signed-off-by: Mike Helmick <helmick@google.com>

All of the other attributes have example after constraints, ID had examples before constraints.


duglin:
LGTM
merging since it's just typo-like fix

PR #587, opened 2020-03-11
PubSub Binding v1.0

nachocano:
- Adding the PubSub binding spec for CloudEvents.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ The following documents are available:
 | Kafka Protocol Binding        | [v1.0](https://github.com/cloudevents/spec/blob/v1.0/kafka-protocol-binding.md) | [master](https://github.com/cloudevents/spec/blob/master/kafka-protocol-binding.md) |
 | MQTT Protocol Binding         | [v1.0](https://github.com/cloudevents/spec/blob/v1.0/mqtt-protocol-binding.md) | [master](https://github.com/cloudevents/spec/blob/master/mqtt-protocol-binding.md)  |
 | NATS Protocol Binding         | [v1.0](https://github.com/cloudevents/spec/blob/v1.0/nats-protocol-binding.md) | [master](https://github.com/cloudevents/spec/blob/master/nats-protocol-binding.md)  |
+| Pub/Sub Protocol Binding      | [v1.0](https://github.com/cloudevents/spec/blob/v1.0/pubsub-protocol-binding.md) | [master](https://github.com/cloudevents/spec/blob/master/pubsub-protocol-binding.md)  |
 | Web hook                      |     [v1.0](https://github.com/cloudevents/spec/blob/v1.0/http-webhook.md)      |      [master](https://github.com/cloudevents/spec/blob/master/http-webhook.md)      |
 |                               |
 | **Additional Documentation:** |
--- a/README.md
+++ b/README.md
@@ -37,7 +37,6 @@ The following documents are available:
 | Kafka Protocol Binding        | [v1.0](https://github.com/cloudevents/spec/blob/v1.0/kafka-protocol-binding.md) | [master](https://github.com/cloudevents/spec/blob/master/kafka-protocol-binding.md) |
 | MQTT Protocol Binding         | [v1.0](https://github.com/cloudevents/spec/blob/v1.0/mqtt-protocol-binding.md) | [master](https://github.com/cloudevents/spec/blob/master/mqtt-protocol-binding.md)  |
 | NATS Protocol Binding         | [v1.0](https://github.com/cloudevents/spec/blob/v1.0/nats-protocol-binding.md) | [master](https://github.com/cloudevents/spec/blob/master/nats-protocol-binding.md)  |
-| Pub/Sub Protocol Binding      | [v1.0](https://github.com/cloudevents/spec/blob/v1.0/pubsub-protocol-binding.md) | [master](https://github.com/cloudevents/spec/blob/master/pubsub-protocol-binding.md)  |
 | Web hook                      |     [v1.0](https://github.com/cloudevents/spec/blob/v1.0/http-webhook.md)      |      [master](https://github.com/cloudevents/spec/blob/master/http-webhook.md)      |
 |                               |
 | **Additional Documentation:** |
--- a/pubsub-protocol-binding.md
+++ b/pubsub-protocol-binding.md
@@ -85,15 +85,15 @@ event metadata.
 
 ### 2.1. datacontenttype Attribute
 
-The assumption is that the `datacontenttype` attribute contains a [RFC2046][rfc2046]
+The `datacontenttype` attribute SHOULD contain a [RFC2046][rfc2046]
 compliant media-type expression.
 
 ### 2.2. data Attribute
 
-The assumption is that the `data` attribute contains opaque application data that is
+The `data` attribute SHOULD contain opaque application data that is
 encoded as declared by the `datacontenttype` attribute.
 
-An application is free to hold the information in any in-memory representation
+An application MAY hold the information in any in-memory representation
 of its choosing, but as the value is transposed into a Pub/Sub message as defined 
 in this specification, the assumption is that the `data` attribute value is made
 available as a sequence of bytes.
@@ -113,8 +113,8 @@ The receiver of the event can distinguish between the two content modes by
 inspecting the [Pub/Sub Attribute `Content-Type`][content-type] of the
 [Pub/Sub message][pubsub-message]. If the attribute is present and its value is prefixed 
 with the CloudEvents media type `application/cloudevents`, indicating the use of a known
-[event format](#14-event-formats), the receiver is expected to use _structured_ mode,
-otherwise it SHOULD default to _binary_ mode.
+[event format](#14-event-formats), the receiver SHOULD use _structured_ mode,
+otherwise it MUST default to _binary_ mode.
 
 If a receiver finds a CloudEvents media type as per the above rule, but with an
 event format that it cannot handle, for instance
--- a/pubsub-protocol-binding.md
+++ b/pubsub-protocol-binding.md
@@ -85,12 +85,12 @@ event metadata.
 
 ### 2.1. datacontenttype Attribute
 
-The `datacontenttype` attribute is assumed to contain a [RFC2046][rfc2046]
+The assumption is that the `datacontenttype` attribute contains a [RFC2046][rfc2046]
 compliant media-type expression.
 
 ### 2.2. data Attribute
 
-The `data` attribute is assumed to contain opaque application data that is
+The assumption is that the `data` attribute contains opaque application data that is
 encoded as declared by the `datacontenttype` attribute.
 
 An application is free to hold the information in any in-memory representation
@@ -113,8 +113,8 @@ The receiver of the event can distinguish between the two content modes by
 inspecting the [Pub/Sub Attribute `Content-Type`][content-type] of the
 [Pub/Sub message][pubsub-message]. If the attribute is present and its value is prefixed 
 with the CloudEvents media type `application/cloudevents`, indicating the use of a known
-[event format](#14-event-formats), the receiver uses _structured_ mode,
-otherwise it defaults to _binary_ mode.
+[event format](#14-event-formats), the receiver is expected to use _structured_ mode,
+otherwise it SHOULD default to _binary_ mode.
 
 If a receiver finds a CloudEvents media type as per the above rule, but with an
 event format that it cannot handle, for instance
@@ -128,9 +128,8 @@ efficient transfer and without transcoding effort.
 
 #### 3.1.1. Content-Type
 
-For the _binary_ mode, the [Pub/Sub Attribute `Content-Type`][content-type] of the 
-[Pub/Sub message][pubsub-message] property, if present, MUST be mapped
-directly to the CloudEvents `datacontenttype` attribute.
+For the _binary_ mode, the [Pub/Sub Attribute `Content-Type`][content-type], if present, MUST 
+be used as the CloudEvents `datacontenttype` attribute.
 
 #### 3.1.2. Event Data Encoding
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -37,6 +37,7 @@ The following documents are available:
 | Kafka Protocol Binding        | [v1.0](https://github.com/cloudevents/spec/blob/v1.0/kafka-protocol-binding.md) | [master](https://github.com/cloudevents/spec/blob/master/kafka-protocol-binding.md) |
 | MQTT Protocol Binding         | [v1.0](https://github.com/cloudevents/spec/blob/v1.0/mqtt-protocol-binding.md) | [master](https://github.com/cloudevents/spec/blob/master/mqtt-protocol-binding.md)  |
 | NATS Protocol Binding         | [v1.0](https://github.com/cloudevents/spec/blob/v1.0/nats-protocol-binding.md) | [master](https://github.com/cloudevents/spec/blob/master/nats-protocol-binding.md)  |
+| Pub/Sub Protocol Binding      | [v1.0](https://github.com/cloudevents/spec/blob/v1.0/pubsub-protocol-binding.md) | [master](https://github.com/cloudevents/spec/blob/master/pubsub-protocol-binding.md)  |
```
ericlem: Google Cloud Pub/Sub?   As-is this naming seems too generic.

pubsub-protocol-binding.md
```diff
@@ -0,0 +1,276 @@
+# Pub/Sub Protocol Binding for CloudEvents - Version 1.0
```
ericlem: Same comment on name

nachocano:
fyi @n3wscott

pubsub-protocol-binding.md
nachocano: Done, thanks!

README.md
nachocano: removing these links for now as the markup linter complaints... 
I guess we should add the links in a follow up PR

pubsub-protocol-binding.md
```diff
@@ -0,0 +1,276 @@
...
+This specification does not further define any of the [CloudEvents][ce] event
+attributes.
+
+This mapping is intentionally robust against changes, including the addition and
+removal of event attributes, and also accommodates vendor extensions to the
+event metadata.
+
+### 2.1. datacontenttype Attribute
+
+The `datacontenttype` attribute is assumed to contain a [RFC2046][rfc2046]
```
grantr: "is assumed to" doesn't have a clear meaning. Can we replace that phrase and others with RFC2119 words?

- "is free to"
- "the assumption is"
- "the expectation is"

pubsub-protocol-binding.md
```diff
@@ -0,0 +1,276 @@
...
+event. Protocol usage patterns that might allow solicitation of events using a
+particular content mode might be defined by an application, but are not defined
+here.
+
+The receiver of the event can distinguish between the two content modes by
+inspecting the [Pub/Sub Attribute `Content-Type`][content-type] of the
+[Pub/Sub message][pubsub-message]. If the attribute is present and its value is prefixed 
+with the CloudEvents media type `application/cloudevents`, indicating the use of a known
+[event format](#14-event-formats), the receiver uses _structured_ mode,
+otherwise it defaults to _binary_ mode.
```
grantr: "the receiver uses": Is this guidance to the implementor or a statement of fact? This should use RFC2119 words.

pubsub-protocol-binding.md
```diff
@@ -0,0 +1,276 @@
...
+### 3.1. Binary Content Mode
+
+The _binary_ content mode accommodates any shape of event data, and allows for
+efficient transfer and without transcoding effort.
+
+#### 3.1.1. Content-Type
+
+For the _binary_ mode, the [Pub/Sub Attribute `Content-Type`][content-type] of the 
+[Pub/Sub message][pubsub-message] property, if present, MUST be mapped
+directly to the CloudEvents `datacontenttype` attribute.
```
grantr: > "The Pub/Sub Attribute `Content-Type` of the Pub/Sub message property

What is the message property? That term isn't defined or used anywhere else. I think this can just be "The Pub/Sub Attribute `Content-Type`", dropping the last 5 words referring to the property.

> MUST be mapped directly to

What does it mean to be "mapped directly to"? Must the values be identical? Must they have a deterministic transformation between them? This should be clearer.

pubsub-protocol-binding.md
```diff
@@ -0,0 +1,276 @@
...
+The [`data`](#22-data-attribute) byte-sequence MUST be used as the value of the Pub/Sub
+message data payload.
+
+#### 3.1.3. Metadata Attributes
+
+All [CloudEvents][ce] attributes 
+MUST be individually mapped to and from distinct Pub/Sub message attributes, with
+exceptions noted below.
+
+CloudEvents extensions that define their own attributes MAY define a diverging
```
grantr: nit: "divergent". "Diverging" implies continuing change. Or just use "different" which is clearer and also apt.

pubsub-protocol-binding.md
```diff
@@ -0,0 +1,276 @@
...
+mapping to Pub/Sub attributes for those attributes, especially if specific attributes
+need to align with Pub/Sub features or with other specifications that have explicit
+Pub/Sub attribute bindings.
+
+An extension specification that defines a diverging mapping rule for Pub/Sub, and
```
grantr: I'm not sure what the purpose of this requirement is. It seems like an unusual burden and one that will be hard to police. If this is the desired state of binding specifications, it should really be part of the general CloudEvents spec. I favor removing it.

pubsub-protocol-binding.md
```diff
@@ -0,0 +1,276 @@
...
+```text
+"Content-Type": "application/cloudevents+json; charset=UTF-8"
+```
+
+#### 3.2.2. Event Data Encoding
+
+The chosen [event format](#14-event-formats) defines how all attributes,
+including the `data` attribute, are represented.
+
+The event metadata and data is then rendered in accordance with the event format
```
grantr: nit: "are then rendered"

pubsub-protocol-binding.md
nachocano: Makes sense, other specs used this terminology, but let's make it clearer here..

pubsub-protocol-binding.md
nachocano: removed...

I ended up putting MUST be used... we shouldn't be transforming it...

pubsub-protocol-binding.md
nachocano: done, not sure if you meant to remove all of it. Just double check I didn't misunderstood

pubsub-protocol-binding.md
nachocano: done! thanks!

pubsub-protocol-binding.md
nachocano: rephrased to: `the receiver is expected to use _structured_ mode,
otherwise it SHOULD default to _binary_ mode.` WDYT?

pubsub-protocol-binding.md
nachocano: I think this was introduced at the time that extensions could be nested maps... So I ended up removing it... not entirely sure though. WDYT

pubsub-protocol-binding.md
```diff
@@ -85,12 +85,12 @@ event metadata.
 
 ### 2.1. datacontenttype Attribute
 
-The `datacontenttype` attribute is assumed to contain a [RFC2046][rfc2046]
+The assumption is that the `datacontenttype` attribute contains a [RFC2046][rfc2046]
```
grantr:
```suggestion
The `datacontenttype` attribute SHOULD contain a [RFC2046][rfc2046]
```

pubsub-protocol-binding.md
```diff
@@ -85,12 +85,12 @@ event metadata.
...
 compliant media-type expression.
 
 ### 2.2. data Attribute
 
-The `data` attribute is assumed to contain opaque application data that is
+The assumption is that the `data` attribute contains opaque application data that is
```
grantr:
```suggestion
The `data` attribute SHOULD contain opaque application data that is
```

pubsub-protocol-binding.md
```diff
@@ -85,12 +85,12 @@ event metadata.
...
 encoded as declared by the `datacontenttype` attribute.
 
 An application is free to hold the information in any in-memory representation
```
grantr:
```suggestion
An application MAY hold the information in any in-memory representation
```

pubsub-protocol-binding.md
```diff
@@ -113,8 +113,8 @@ The receiver of the event can distinguish between the two content modes by
 inspecting the [Pub/Sub Attribute `Content-Type`][content-type] of the
 [Pub/Sub message][pubsub-message]. If the attribute is present and its value is prefixed 
 with the CloudEvents media type `application/cloudevents`, indicating the use of a known
-[event format](#14-event-formats), the receiver uses _structured_ mode,
-otherwise it defaults to _binary_ mode.
+[event format](#14-event-formats), the receiver is expected to use _structured_ mode,
```
grantr:
```suggestion
[event format](#14-event-formats), the receiver SHOULD use _structured_ mode,
```

pubsub-protocol-binding.md
```diff
@@ -113,8 +113,8 @@ The receiver of the event can distinguish between the two content modes by
...
+otherwise it SHOULD default to _binary_ mode.
```
grantr:
```suggestion
otherwise it MUST default to _binary_ mode.
```

grantr: My reading of the spec implies this is a MUST, but feel free to make this SHOULD if you disagree.

nachocano: is a MUST you are right...

pubsub-protocol-binding.md
grantr: Now that extensions must have untyped string attribute values only, I can't think of a case where such a mapping would be required. I think it's fine to remove it. We can add this section back later if needed.

nachocano:
> IMO "assume" or "expectation" translates to SHOULD. "is free to" translates to MAY. I suggested some translations.

Sounds good! Let me add this changes, I don't want to commit the suggestion because I may have a signoff issue :)

pubsub-protocol-binding.md
nachocano: Done! thanks!

pubsub-protocol-binding.md
nachocano: done! thanks!

pubsub-protocol-binding.md
nachocano: done, thanks!

pubsub-protocol-binding.md
nachocano: done

pubsub-protocol-binding.md
nachocano: done

pubsub-protocol-binding.md
nachocano: sounds good!

clemensv:
Google PubSub is a proprietary product and the specification should therefore be hosted in a Google repository and then linked to from the proprietary protocol list per our governance rules:

See [Qualifying Protocols and Encodings](https://github.com/cloudevents/spec/blob/master/primer.md#qualifying-protocols-and-encodings)

The Google PubSub protocol meets none of the qualification criteria for a binding to be included in the CloudEvents repository directly.

nachocano:
Oops, thanks for pointing that out Clemens. Will put it in our repo and
update this PR to just add the link to it.

pubsub-protocol-binding.md
```diff
@@ -0,0 +1,263 @@
...
+
+#### 3.1.3. Metadata Attributes
+
+All [CloudEvents][ce] attributes 
+MUST be individually mapped to and from distinct Pub/Sub message attributes.
+
+##### 3.1.3.1 Pub/Sub Attribute Names
+
+Except for attributes
+[explicitly handled in this specification](#2-use-of-cloudevents-attributes),
```
cneijenhuis: I find this confusing - there are no attributes explicitly handled there, right? I don't know what I'm supposed to take away from this.

pubsub-protocol-binding.md
```diff
@@ -0,0 +1,263 @@
...
+
+If a receiver finds a CloudEvents media type as per the above rule, but with an
+event format that it cannot handle, for instance
+`application/cloudevents+avro`, it MAY still treat the event as binary and
+forward it to another party as-is.
+
+### 3.1. Binary Content Mode
+
+The _binary_ content mode accommodates any shape of event data, and allows for
+efficient transfer and without transcoding effort.
```
cneijenhuis: I think it's worth pointing out that the Binary Content Mode can only be used if the metadata attributes fit within the limits of Pub/Sub.

```
Attributes per message: 100
Attribute key size: 256 bytes
Attribute value size: 1024 bytes
```

The value size could potentially be an issue...

pubsub-protocol-binding.md
```diff
@@ -0,0 +1,263 @@
...
+
+### 3.2. Structured Content Mode
+
+The _structured_ content mode keeps event metadata and data together in the
+payload, allowing simple forwarding of the same event across multiple routing
+hops, and across multiple transports.
+
+#### 3.2.1. Content-Type
+
+The [Pub/Sub Attribute `Content-Type`][content-type] MUST be set to the media type of
```
cneijenhuis: Is the casing intentionally?

I'd find `content-type` more consistent with the binary mode. `contentType` is more consistent with the other PubSub attributes (e.g. `messageId`).

pubsub-protocol-binding.md
```diff
@@ -0,0 +1,263 @@
...
+the naming convention for the Pub/Sub attribute mapping of well-known CloudEvents
+attributes is that each attribute name MUST be prefixed with "ce-".
+
+Examples:
+
+    * `time` maps to `ce-time`
+    * `id` maps to `ce-id`
+    * `specversion` maps to `ce-specversion`
+
+Attribute names are case-insensitive.
```
cneijenhuis: Is this a good idea? Let's say that one day Pub/Sub will allow filtering on attributes (FINALLY 😉 ), are you sure it's going to be case-insensitive?

I'd vote for lower-casing.

nachocano:
Closing this guy and moving it to our repo, it was my bad...

Thanks @cneijenhuis for the great comments, will address them in our repo, and also @clemensv for pointing out that guidelines doc!

PR #590, opened 2020-03-13
Draft for subscription API

clemensv:
This will be updated with the bindings through next week

### DIFF
--- a/subscriptions-api-openapi.yaml
+++ b/subscriptions-api-openapi.yaml
@@ -1,0 +1,206 @@
+openapi: "3.0.2"
+info:
+  title: CloudEvents Subscriptions API
+  version: "0.1 WIP"
+servers:
+  - url: https://server.example.com/v0.1
+paths:
+  /subscriptions:
+    get:
+      description: Retrieve one or multiple subscriptions
+      parameters:
+        - in: "query"
+          name: "id"
+          description: identifier of a specific subscription
+          required: false
+          schema:
+            type: string
+      responses:
+        "200":
+          description: OK
+          content:
+            application/json:
+              schema:
+                oneOf:
+                  - $ref: "#/components/schemas/Subscription"
+                    description: the subscription that was requested with teh 'id' parameter
+                  - type: array
+                    description: a list of subscriptions
+                    items:
+                       $ref: "#/components/schemas/Subscription"
+    put:
+      description: Update a subscription
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/Subscription"
+      responses:
+        "200":
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Subscription"
+    post:
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/Subscription"
+      responses:
+        "200":
+          description: Updated successfully
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Subscription"
+        "201":
+          description: Created successfully
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Subscription"
+          headers:
+            Location:
+              schema:
+                type: string
+                format: url
+        "400":
+          description: Invalid or malformed request
+    delete:
+      description: Delete a subscription
+      parameters:
+        - in: "query"
+          name: "id"
+          required: true
+          schema:
+            type: string
+          description: The id of an existing subscription
+      responses:
+        "200":
+          description: Successfully deleted
+        "400":
+          description: Invalid or malformed request
+        "404":
+          description: Subscription not found
+    options:
+      description: Discover supported features and methods for this endpoint
+      responses:
+        "200":
+          description: OK
+          headers:
+            Allow:
+              schema:
+                type: string
+                default: "GET,PUT,POST,DELETE,OPTIONS"
+components:
+  schemas:
+    Subscription:
+      properties:
+        id:
+          type: string
+          description: REQUIRED. The unique identifier of the subscription in the scope of the subscription manager.
+          example: 1119920371
+        protocol:
+          type: string
+          enum: ["HTTP", "MQTT3", "MQTT5", "AMQP", "NATS", "KAFKA"]
+          description: REQUIRED. Identifier of a delivery protocol.
+          example: "HTTP"
+        protocolsettings:
+          type: object
+          anyOf:
+            - $ref: "#/components/schemas/HTTPSettings"
+            - $ref: "#/components/schemas/MQTTSettings"
+            - $ref: "#/components/schemas/AMQPSettings"
+            - $ref: "#/components/schemas/NATSSettings"
+            - $ref: "#/components/schemas/ApacheKafkaSettings"
+          description: OPTIONAL. A set of settings specific to the selected delivery protocol provider.
+          example: null
+        sink:
+          type: string
+          format: url
+          description: REQUIRED. The address to which events shall be delivered using the selected protocol.
+          example: "https://endpoint.example.com/webhook"
+        filter:
+          $ref: "#/components/schemas/Filter"
+      required:
+        - id
+        - protocol
+    Filter:
+      properties:
+        dialect:
+          type: string
+          description: Filter dialect
+          example: "simple"
+          default: "simple"
+        conditions:
+          type: array
+          description: Filter conditions
+          items:
+            type: object
+    HTTPSettings:
+      properties:
+        proxyurl:
+          type: string
+          format: url
+        proxycredentials:
+          $ref: "#/components/schemas/Credentials"
+        headers:
+          type: object
+        method:
+          type: string
+    MQTTSettings:
+      properties:
+        topicname:
+          type: string
+        qos:
+          type: integer
+          format: int32
+        retain:
+          type: boolean
+        expiry:
+          type: integer
+          format: int32
+        userproperties:
+          type: object
+      required:
+        - topicname
+    AMQPSettings:
+      properties:
+        address:
+          type: string
+        linkname:
+          type: string
+        sendersettlementmode:
+          type: string
+          enum: ["settled", "unsettled"]
+        linkproperties:
+          type: object
+    ApacheKafkaSettings:
+      properties:
+        topicname:
+          type: string
+        partitionkeyextractor:
+          type: string
+        clientid:
+          type: string
+        ackmode:
+          type: integer
+      required:
+        - topicname
+    NATSSettings:
+      properties:
+        subject:
+          type: string
+      required:
+        - subject
+    Credentials:
+      properties:
+        username:
+          type: string
+        passkey:
+          type: string
+          format: password
+        accesstoken:
+          type: string
--- a/subscriptions-api.md
+++ b/subscriptions-api.md
@@ -1,4 +1,4 @@
-# CNCF CloudEvents – Subscriptions API
+# CNCF CloudEvents – Subscriptions API - Version 0.1-rc01s
 
 ## Abstract
 
@@ -24,13 +24,14 @@ This document also does not formalize the relationship between a specific set of
 events and the subscription manager. An event producer that produces multiple
 different types of events can offer one subscription manager for all events or a
 subscription manager for each type of event, or any other combination. An event
-producer may also offer up the same set of events for subscription at multiple
-concurrent subscription managers. Such subscription offers are published in a
-CloudEvents Discovery service.
+producer can also offer up the same set of events for subscription at multiple
+concurrent subscription managers. Regardless of relationship between the event
+producer(s) and the subscription manager(s), the advertisement of the
+subscription offers are published in a CloudEvents Discovery service.
 
 As with the core CloudEvents specification, the goal of this specification is to
 reuse mechanisms based on existing standards and conventions where such exist
-and only introduce new mechanisms where required.
+and only introduce new mechanisms where needed.
 
 Therefore, this specification not only defines a new subscription management API
 for certain use-cases, but also refers to existing mechanisms available in the
@@ -74,7 +75,7 @@ event to receivers based on the information in the event context.
 
 #### Consumer
 
-An entity receiving and processing events. Consumers may receive events from
+An entity receiving and processing events. Consumers might receive events from
 producers directly or via intermediaries. A consumer might listen and wait for
 events to be delivered to it or it might actively solicit them from the producer
 or intermediary.
@@ -127,14 +128,14 @@ the event delivery channel to the consumer is initiated:
   manager to select a transport protocol, establish the desired communication
   channel, and deliver the event(s).
 
-An application using CloudEvents might use only one of these styles or a
+An end-to-end solution using CloudEvents might use only one of these styles or a
 combination of those.
 
-For instance, the application might use an MQTT broker to handle delivery of
+For instance, the solution might use an MQTT broker to handle delivery of
 events into connected devices that connect into the broker and subscribe on a
 particular MQTT topic, which acts as the subscription manager in the sense of
 this specification ("pull"-style). The events that ought to be shared with the
-connected devices might however originate elsewhere in the overall application,
+connected devices might however originate elsewhere in the overall solution,
 and some event router middleware's subscription manager might therefore be
 configured to initiate delivery of events ("push"-style) into the given MQTT
 broker topic whenever such events are available.
@@ -144,13 +145,13 @@ broker topic whenever such events are available.
 MQTT, AMQP, NATS, and Apache Kafka are protocols with a formal CloudEvents
 binding that have native mechanisms filling the role of a subscription manager.
 
-Using any of these subscription management mechanisms should allow an
+Using any of these subscription management mechanisms ought to allow an
 application to claim conformance with this specification. More specifically, an
-application that uses an MQTT or AMQP message broker as its middleware
-component should be able to use the native capabilities of those protocols to
-subscribe to CloudEvents event streams without requiring CloudEvents-specific
-mechanisms or extensions. The conformance section in this document formally
-enumerates the requirements.
+application that uses an MQTT or AMQP message broker as its middleware component
+ought be able to use the native capabilities of those protocols to subscribe to
+CloudEvents event streams without requiring CloudEvents Subscription
+API-specific mechanisms or extensions. The conformance section in this document
+formally enumerates the requirements.
 
 _The descriptions of the protocol subscription mechanisms in this section are
 non-normative. Please refer to the protocol specifications or documentation for
@@ -250,7 +251,7 @@ also have a separate CloudEvents subscription manager API endpoint
 implementation. If such a separate API endpoint exists and creating
 subscriptions is not feasible on such a separate API endpoint, the respective
 operation might not be available. For instance, creating an MQTT subscription is
-not feasible from outside an MQTT connection, but an existing subscription may
+not feasible from outside an MQTT connection, but an existing subscription might
 be enumerable through a subscription collection and deleting the subscription
 from the collection might terminate it.
 
@@ -271,21 +272,22 @@ Each subscription is represented by an object that has the following properties:
   protocols.
 - **protocolsettings** (map) - OPTIONAL. A set of settings specific to the
   selected delivery protocol provider. Options for those settings are listed
-  in the following subsection. An implementation can offer more options.
+  in the following subsection. An implementation MAY offer more options.
   Examples for such settings are credentials, which generally vary by
   transport, rate limits and retry policies, or the QoS mode for MQTT.
   See the [Protocol Settings](#322-protocol-settings) section for further details.
-- **sink** (URI) - REQUIRED. The address to which events shall be delivered using
+- **sink** (URI) - REQUIRED. The address to which events SHALL be delivered using
   the selected protocol. The format of the address MUST be valid for the
   selected protocol or one of the protocol's own transport bindings (e.g. AMQP
   over WebSockets).
 - **filter** - OPTIONAL. A filter is an expression of a particular filter
   dialect that evaluates to true or false and that determines whether an
-  instance of a cloud event should be delivered to the subscription's sink. If
-  a filter expression evaluates to false, the event MUST NOT be sent to the
-  sink. Support for particular filter dialects might vary across different
-  subscription managers. If a filter dialect is specified in a subscription
-  that is unsupported by the subscription manager, creation or updates of the
+  instance of a CloudEvent will be delivered to the subscription's sink. If a
+  filter expression evaluates to false, the event MUST NOT be sent to the sink.
+  If the expression evaluates to true, the event MUST be attempted to be
+  delivered. Support for particular filter dialects might vary across different
+  subscription managers. If a filter dialect is specified in a subscription that
+  is unsupported by the subscription manager, creation or updates of the
   subscription MUST be rejected with an error. See the [Filter
   Dialects](#323-filter-dialects) section for further details.
 
@@ -299,10 +301,6 @@ protocol-settings map, including default values where necessary.
 For HTTP, the following settings properties SHOULD be supported by all
 implementations.
 
-- **proxyurl** (URI) – OPTIONAL. The URI of the proxy server to use for sending
-  the event(s).
-- **proxycredentials** (credentials) – OPTIONAL. A _credentials_ object (see
-  "common objects" below) holding information for gaining access to the proxy.
 - **headers** (map) – OPTIONAL. A set of key/value pairs that is copied into the
   HTTP request as custom headers.
 - **method** (string) – OPTIONAL. The HTTP method to use for sending the message.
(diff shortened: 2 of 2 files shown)


subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
+# CNCF CloudEvents – Subscriptions API
```
duglin: just for consistency with the other spec perhaps we should add ` - Version 0.1-rc01`

subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
...
+immediately at or near the event producer or it can reside in some middleware
+infrastructure. This document does not formalize the relationship between the
+original event producer and the subscription manager, and how the subscription
+manager obtains the events it distributes to subscribers.
+
+This document also does not formalize the relationship between a specific set of
+events and the subscription manager. An event producer that produces multiple
+different types of events can offer one subscription manager for all events or a
+subscription manager for each type of event, or any other combination. An event
+producer may also offer up the same set of events for subscription at multiple
```
duglin: s/may/can/  unless you want to make the entire paragraph normative in which case lots of MAYs would need to be added, but I suspect the intent wasn't to be normative yet.

subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
...
+concurrent subscription managers. Such subscription offers are published in a
```
duglin: Does the "Such" refer to all of the cases listed in this paragraph or just to the last case mentioned? I think the intent was to cover all of them, if so, perhaps: Regardless of relationship between the event producer(s) and the subscription manager(s), the advertisement of the subscription offers are published in a CloudEvents Discovery service

Maybe?

subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
...
+CloudEvents Discovery service.
+
+As with the core CloudEvents specification, the goal of this specification is to
+reuse mechanisms based on existing standards and conventions where such exist
+and only introduce new mechanisms where required.
```
duglin: s/required/needed/ or "necessary"

subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
...
+
+Therefore, this specification not only defines a new subscription management API
+for certain use-cases, but also refers to existing mechanisms available in the
+specifications of transport protocols for which CloudEvents bindings exist.
+
+For instance, introducing and mandating a CloudEvents-specific subscription
```
duglin: I really like this goal of not reinventing the wheel!

subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
...
+#### Consumer
+
+An entity receiving and processing events. Consumers may receive events from
+producers directly or via intermediaries. A consumer might listen and wait for
+events to be delivered to it or it might actively solicit them from the producer
+or intermediary.
+
+#### Subscription
+
+A relationship established between a consumer and a producer or an intermediary.
```
duglin: In the case of an intermediary, does the consumer actually know it's an intermediary? If not, I wonder if the introduction of the indirection might cause some confusion for some people. Would it be safer to just stick to a 2 party system?

subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
...
+
+- Subscriptions with subscription-manager-initiated delivery ("push"-style) are
+  configured on the subscription manager for delivering events through a
+  communication channel that the subscription manager initiates when events
+  are available for delivery on the subscription. The configuration of such a
+  subscription MUST contain all information required for the subscription
+  manager to select a transport protocol, establish the desired communication
+  channel, and deliver the event(s).
+
+An application using CloudEvents might use only one of these styles or a
```
duglin: the use of the word 'application' here might throw some people - would it be incorrect to switch it to "A consumer" ?

duglin: And below...

duglin: or use "An application acting as an event consumer..."

duglin: hmmm having read the next paragraph, I think the use of the word "application" was very different than what I original thought. I think you mean something more akin to "the entire flow of an event from the source to the final consumer", right?

subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
...
+
+### 3.1 Native Subscription Management Mechanisms
+
+MQTT, AMQP, NATS, and Apache Kafka are protocols with a formal CloudEvents
+binding that have native mechanisms filling the role of a subscription manager.
+
+Using any of these subscription management mechanisms should allow an
+application to claim conformance with this specification. More specifically, an
+application that uses an MQTT or AMQP message brokers as its middleware
+component should be able to use the native capabilities of those protocols to
```
duglin: s/should/ought/

subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
...
+subscribe to CloudEvents event streams without requiring CloudEvents-specific
```
duglin: is it "CloudEvents-specific" or "CloudEvents Subscription API-specific" ?

subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
...
+The protocol used to configure the subscription manager is decoupled from the
+delivery protocol, meaning that an application can configure a push delivery
+into an AMQP destination using an API call initiated over HTTP.
+
+A subscription manager that natively exists in a middleware implementation might
+also have a separate CloudEvents subscription manager API endpoint
+implementation. If such a separate API endpoint exists and creating
+subscriptions is not feasible on such a separate API endpoint, the respective
+operation might not be available. For instance, creating an MQTT subscription is
+not feasible from outside an MQTT connection, but an existing subscription may
```
duglin: s/may/might/

subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
...
+  scope of the subscription manager.
+- **protocol** (string) - REQUIRED. Identifier of a delivery protocol. Because of
+  WebSocket tunneling options for AMQP, MQTT and other protocols, the URI
+  scheme is not sufficient to identify the protocol. The protocols with
+  existing CloudEvents bindings are identified as "AMQP", "MQTT3", "MQTT5",
+  "HTTP", "KAFKA", and "NATS". An implementation MAY add support for further
+  protocols.
+- **protocolsettings** (map) - OPTIONAL. A set of settings specific to the
+  selected delivery protocol provider. Options for those settings are listed
+  in the following subsection. An implementation can offer more options.
```
duglin: s/can/MAY/

subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
...
+  Examples for such settings are credentials, which generally vary by
+  transport, rate limits and retry policies, or the QoS mode for MQTT.
+  See the [Protocol Settings](#322-protocol-settings) section for further details.
+- **sink** (URI) - REQUIRED. The address to which events shall be delivered using
```
duglin: s/shall/SHALL/

subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
...
+  the selected protocol. The format of the address MUST be valid for the
+  selected protocol or one of the protocol's own transport bindings (e.g. AMQP
+  over WebSockets).
+- **filter** - OPTIONAL. A filter is an expression of a particular filter
+  dialect that evaluates to true or false and that determines whether an
+  instance of a cloud event should be delivered to the subscription's sink. If
```
duglin: s/cloud event/CloudEvent/
s/should/will/

subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
...
+  a filter expression evaluates to false, the event MUST NOT be sent to the
+  sink. Support for particular filter dialects might vary across different
```
duglin: do we want to mandate what happens in the case of it evaluating to 'true'? While we can't say it MUST be delivered, we can probably say that it "MUST be attempted to be delivered", or something like that.

subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
...
+
+This section enumerates protocol-specific delivery options for the
+protocol-settings map, including default values where necessary.
+
+##### 3.2.2.1. HTTP
+
+For HTTP, the following settings properties SHOULD be supported by all
+implementations.
+
+- **proxyurl** (URI) – OPTIONAL. The URI of the proxy server to use for sending
```
duglin: We might need to say that if a proxyurl is specified then the subscription manager MUST use it when trying to send events - IOW, ignoring it isn't an option.

duglin: actually, do we need to make similar statements for all of these? Do we need to say what happens if a subscription manage doesn't support it? Or is it truly optional and the consumer has no idea which ones will be adhered to?

subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
...
+
+##### 3.2.2.5. NATS
+
+- **subject** (string) - REQUIRED. The name of the NATS subject to publish to.
+
+#### 3.2.3 Filter Dialects
+
+The filter expression language supported by an event producer is indicated by
+its dialect. This is intended to allow for flexibility, extensibility and to
+allow for a variety of filter dialects without enumerating them all in this spec or
```
duglin: s/spec/specification/

duglin: and next line

subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
...
+- "Suffix match" where the filter condition is a suffix of the attribute value.
+
+This filter dialect is intentionally constrained to these filter types, since
+filtering has a potentially significant impact on the baseline performance of
+all implementations. In cases where more filter types or different expression
+languages are desired, further dialects can be introduced as extensions.
+
+Extension dialects will have varying support across event producers. It is up to
+the subscriber and producer to negotiate which filter dialects can be used
+within a given subscription.
```
duglin: When ready we might want to add a reference to the Discovery spec here since that should help with the negotiations.

subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
...
+
+The filter conditions specified in the simple dialect will be defined by
+specifying a "conditions" property holding an array of simple filter conditions,
+which are logically combined with an implicit "AND" operator. This means the
+filter criteria MUST evaluate to true for every filter in the array in order for
+a cloud event instance to be delivered to the target sink.
```
duglin: s/cloud event/CloudEvent/

subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
...
+
+Each simple filter is defined with the following properties:
+
+- type (string) - REQUIRED. Value MUST be one of the following: prefix_match,
```
duglin: can we drop the "_match" just to save typing?  not a biggie if not.

subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
...
+  suffix_match, exact_match.
+- property (string) - REQUIRED. The CloudEvents attribute to match the value
+  indicated by the "value" property against.
```
duglin: I assume this could include CE extensions, right? If so, we should probably say so to be clear.

subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
...
+- value (Any) - REQUIRED, The value to match the CloudEvents attribute against.
```
duglin: I sense the use of "Any" here is going to ask for some more text around how to do comparisons :-)

subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
...
+- subscription (subscription) - REQUIRED. Subscription object.
+
+Errors:
+
+- **ok** - the operation succeeded
+- **notfound** - a subscription with the given _id_ already exists
+
+### 3.3. HTTP Binding for the Subscription API
+
+(TBD) This will be a straightforward mapping of the described API to a basic
```
duglin: Just looking forward to the discussion about imperative vs declarative APIs... would it make sense to support a `202` + Location Header being returned for those implementations that can't commit to fully creating the subscription immediately?

subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
...
+predicting what filtering needs every system will have in the future. This spec
+will specify a single "simple" dialect, which all implementations MUST support.
+
+The dialect for a particular filter is indicated by specifying the "dialect"
+property at the root of the JSON object. The value of this is a string encoded
+unique identifier for the filter dialect. Subscriptions specifying the "filter"
+property MUST specify a dialect. All other properties are dependent on the
+dialect being used.
+
+##### 3.2.3.1. "simple" filter dialect
```
n3wscott: simple does not describe what this dialect does, I would recommend renaming "simple" to "exact"

subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
...
+
+- **ok** - the operation succeeded and returned results
+- **nocontent** - the operation succeeded and returned no results
+
+Protocol bindings and implementations of such bindings MAY add custom filter
+constraints and pagination arguments as parameters. A request without filtering
+constraints SHOULD return all available subscriptions associated with or
+otherwise visible to the party making the request.
+
+#### 3.2.4.4. Updating a Subscription
```
sroze: If a subscription’s filter gets updated, it might mean that the consumer have now previously received a subset of the messages described by the subscription (especially when the subscription manager stores the history of events). Is it worth taking on stance on the fact that we don’t expect consumers to receive historical events that have not been sent to it do to a previously “smaller” filter?

subscriptions-api-openapi.yaml
```diff
@@ -0,0 +1,206 @@
...
+  version: "0.1 WIP"
+servers:
+  - url: https://server.example.com/v0.1
+paths:
+  /subscriptions:
+    get:
+      description: Retrieve one or multiple subscriptions
+      parameters:
+        - in: "query"
+          name: "id"
```
ryanhorn: If we care about being RESTful, I'd prefer that the "fetch by id" be its own resource, with the id being part of the URL path, such as `/subscriptions/{id}`

subscriptions-api-openapi.yaml
```diff
@@ -0,0 +1,206 @@
...
+                    description: a list of subscriptions
+                    items:
+                       $ref: "#/components/schemas/Subscription"
+    put:
+      description: Update a subscription
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/Subscription"
```
ryanhorn: Same here, should the subscription id be part of the URL?

subscriptions-api.md
ryanhorn: "exact" isn't entirely accurate either though, as we allow for prefix and suffix matching. I don't really have a better suggestion than "simple", but I'm also not too particular here.

subscriptions-api.md
ryanhorn: +1

subscriptions-api.md
ryanhorn: My 2 cents is that I think this is going to be technology and implementation dependent and should be defined in the binding for the protocol.

subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
...
+  on the subscription manager for delivering events through a communication
+  channel (like an AMQP link or a MQTT connection) initiated by the consumer
+  and connecting to the subscription manager. These kinds of subscriptions are
+  typically offered by messaging or eventing middleware and their lifetime
+  might be bounded by the lifetime of the communication channel.
+
+- Subscriptions with subscription-manager-initiated delivery ("push"-style) are
+  configured on the subscription manager for delivering events through a
+  communication channel that the subscription manager initiates when events
+  are available for delivery on the subscription. The configuration of such a
```
manuelstein: IMHO the accessory "when events are available for delivery on the subscription", although vague, could be misunderstood as if the communication channel has to be initiated at the time the subscription manager got an event. But the channel could have been setup before an event becomes available or the subscription manager could employ batching and delay the setup. Maybe the temporal clause, although illustrative, is not necessary here?

subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
...
+broker topic whenever such events are available.
+
+### 3.1 Native Subscription Management Mechanisms
+
+MQTT, AMQP, NATS, and Apache Kafka are protocols with a formal CloudEvents
+binding that have native mechanisms filling the role of a subscription manager.
+
+Using any of these subscription management mechanisms should allow an
+application to claim conformance with this specification. More specifically, an
+application that uses an MQTT or AMQP message brokers as its middleware
```
manuelstein:
```suggestion
application that uses an MQTT or AMQP message broker as its middleware
```

subscriptions-api.md
```diff
@@ -0,0 +1,613 @@
...
+applications of HTTP rather than inherent features of HTTP per-se.
+
+### 3.2. Subscription Manager API
+
+Subscriptions where the delivery is initiated by the subscription manager
+("push"-style) require a mechanism to configure such subscriptions on the
+subscription manager.
+
+The subscription manager API defines a subscription object and an API for
+creating, updating and deleting subscriptions on a subscription manager.
```
manuelstein: IIUC, the motivational sentence here states that the mechanism (i.e. the protocol) was ultimately necessary in case of push-style delivery because the subscription embodies the consumer's interest in events and that lies initially with the consumer, so the SM needs to be notified of that interest. It's true, that especially "push"-style delivery requires an out-of-band mechanism to communicate a consumer's interest in events to the subscription manager.

```suggestion
The subscription manager API defines a subscription object and a protocol for 
creating, updating and deleting subscriptions on a subscription manager.

The subscription object describes the consumer's interest in events and 
defines the delivery method. The protocol allows to configure subscriptions
on the subscription manager. Especially in case of subscription manager 
initiated ("push"-style) delivery, the protocol is required to express the 
consumer's interest to the subscription manager before the subscription 
manager can initiate delivery.
```

But, wouldn't "pull"-style also require a mechanism to configure subscriptions at the subscription manager? IIUC a CloudEvent subscription manager deals in CloudEvents, i.e. routes them based on their event context. AFAICS no (mentioned) "pull"-style message delivery would **natively** support such CloudEvent context, e.g. where the consumer can initiate delivery of a specific CloudEvent type. The pulling consumer also requires a mechanism to have the subscription manager provide an endpoint/topic/queue that caters for the subscription.

The subscription manager is two things:
- the API endpoint/service that manages CRUD of declared CloudEvent Subscriptions
- the broker/router/producer/source/entity that delivers events push-style or serves interfaces for pull-style delivery

Native subscription management mechanisms may reuse the delivery transport channel to also manage e.g. topic subscriptions (in-band), but none of them talk _CloudEvent subscriptions_. So, some sort of out-of-band _subscription API_ is a necessity to communicate _CloudEvent subscription_ objects.

In general, isn't it that any CloudEvent subscription (push or pull), if expressed using this specification, always requires out-of-band communication to be setup? A consumer may use a native messaging "subscription" mechanism to _initiate delivery_, but still requires other means, e.g. to setup a topic and route messages to that topic according to the CloudEvent subscription it wants to realize.

subscriptions-api.md
deissnerk: While the consumer may not know, if it is a producer or intermediary at the other end, it should not assume one or the other case. A consumer assuming to talk directly to the producer, may create less specific subscriptions and be surprised by all the events it receives upon that subscription.

subscriptions-api.md
deissnerk: Good points.

subscriptions-api.md
deissnerk: @manuelstein I've been thinking about these questions, too. I tried to describe some of my thoughts in the _Eventing domains_ sections in the original Google doc. 
The examples you mentioned seem to go beyond subscriptions as such. Isn't it more about setting up resources for routing or consumption? Is the configuration of a topic needed for subscribing or for producing? If you made this a part of subscribing, wouldn't you still require the consumer to do an additional subscription step based on the messaging protocol? 
A similar case is the creation of a queue that gets subscribed to the according topics. I think all this can be a layer on top of the raw subscription API we are defining here. 
If you like more discussions on this topic, just ping me on Slack.

duglin:
Conditionally approved on the 3/26 call.
"Conditional" because it would be good to address the comments in the doc first so that we don't lose them. Any comments that are not easily resolved and require more discussion can have issues/PRs opened for them later.

@clemensv and sub-team - thanks!!!  Just let me know when it's ready

subscriptions-api-openapi.yaml
```diff
@@ -0,0 +1,206 @@
...
+          example: 1119920371
+        protocol:
+          type: string
+          enum: ["HTTP", "MQTT3", "MQTT5", "AMQP", "NATS", "KAFKA"]
+          description: REQUIRED. Identifier of a delivery protocol.
+          example: "HTTP"
+        protocolsettings:
+          type: object
+          anyOf:
+            - $ref: "#/components/schemas/HTTPSettings"
```
mikehelmick: same concern - we can't possibly list all configuration setting for all protocols.

in discovery this was left as a map to specify necessary subscription parameters

subscriptions-api-openapi.yaml
```diff
@@ -0,0 +1,206 @@
...
+  schemas:
+    Subscription:
+      properties:
+        id:
+          type: string
+          description: REQUIRED. The unique identifier of the subscription in the scope of the subscription manager.
+          example: 1119920371
+        protocol:
+          type: string
+          enum: ["HTTP", "MQTT3", "MQTT5", "AMQP", "NATS", "KAFKA"]
```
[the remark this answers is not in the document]
clemensv: Everyone who adds further protocols in their implementation can extend the contract. This is a baseline. I think the enum is better than a free-for-all.

subscriptions-api-openapi.yaml
clemensv: We have a set of protocols that we have a clearly considered set of settings for as a baseline. It's better to have a baseline than leave it completely open.

subscriptions-api.md
clemensv: Batching and delay or connection pooling are all optimizations, IMO. The subscription manager needs to initiate the channel when there are events to deliver. That could also just be once day, but if there are no events to deliver at all, it is likely not bothering.

subscriptions-api-openapi.yaml
clemensv: When the POST creation returns a 201 with the Location header having ".../subscriptions?id=1" or that is a URL in the list you obtain with an unqualified GET, that is not more or less RESTful than having the same value in the path. RESTful is when the information from a prior request drives the next request. Knowing where to put data into a URL ahead of time is specifically not RESTful.

TED talk aside:

We can add an option where you can also have the id in the path, but due to OpenAPI's path bias, I would need to factor the operation out, so that either option works.

subscriptions-api.md
clemensv: @sroze Do you know a broker that lets you go back in time in this fashion?

subscriptions-api.md
clemensv: "Declarative API" is not a wire concept. You can certainly choose to have a runtime that is all driven by config files, but in the end, that runtime needs to go out and deal with actual resources. 

There 202 plus Location is weird. If you can already tell where you will create a resource and you're confident enough to return 2xx, you can just as well create it. 

I get the feeling that "Declarative API" is often code for "sorry, this is very slow".

subscriptions-api.md
clemensv: Yes. I changed to this end-to-end solution.

subscriptions-api.md
clemensv: I am removing the proxy info since that doesn't belong here, after all. This is not a per-subscription config.

subscriptions-api.md
clemensv: I renamed it to "basic". That takes the simplicity connotation out.

subscriptions-api.md
clemensv: just did

subscriptions-api.md
duglin: > I get the feeling that "Declarative API" is often code for "sorry, this is very slow".

LOL  I was just looking for a way to support both models. I agree that what appears on the wire should not change based on the server's impl choice except to say that in some cases an HTTP response might contain the created resource (or an indication that it completed the processing - e.g. a 201) while in other cases it might not due to async processing. And it's that async processing model that I consider the declarative impl sitting.

> There 202 plus Location is weird. If you can already tell where you will create a resource and you're confident enough to return 2xx, you can just as well create it.

if the processing of the request is done asynchronously, then the Location URL might not actually be the final URL, it might be a URL that sends a redirect to the real resource once it's created.  In HTTP terms, I view this more as the URL to the status of the create - which the http spec does talk about.

subscriptions-api.md
clemensv: good point. since this is a string matching function, I am clarifying this to be a string and all attributes having to use their string representation.

subscriptions-api.md
clemensv: That's an implementation choice. If you're confident that you're going to get the work done, it's ok to give a 201 + Location and then keep returning 307 + Retry-After for that URI until you're actually done. 

202 is a "got it, go away".

duglin: > That's an implementation choice.

Right that's my point. If our spec allows both paths then we should be good. I was just thinking that if our spec says "just do it via http" but all of our examples, text, etc... only really talk about a "200" then I suspect a lot of client devs will only support that and will blow up on a 202+Location.

duglin:
@clemensv can you sign your commits?

clemensv:
@duglin per log, that's a debatable issue.

I have incorporated two edits that were given here, and those are signed with "Co-Authored-By: Manuel Stein <manuelstein@users.noreply.github.com>". @manuelstein 

I think those are tripped up the DCO script.

duglin:
I'd just squash them down to one commit and manually include another "Authored by" line with Manuel's name/email

duglin:
Approved (again) on the 4/2 call

Issue #574, opened 2020-02-27
Efficient alternative to batch in HTTP binding 


slinkydeveloper:
The HTTP binding has an encoding to send multiple events in the same envelope called [batch](https://github.com/cloudevents/spec/blob/master/http-protocol-binding.md#33-batched-content-mode).
This content mode is inefficient to implement in SDKs, because it requires the buffering of the whole HTTP body and the unmarshalling of the full json array.

I would love to search for more efficient encodings to send multiple events in the same envelope. Two examples that comes to my mind are:

* multipart bodies
* json streams (which are nothing more than newline separated json objects)

What do you think about it?

aslom:
Even if it is not streaming it should still provide performance benefits by sending one batch of messages instead of multiple HTTP requests?

aslom:
For multipart then CloudEvents could be send HTTP streaming with possibly adding it to next version of HTTP binding when we have implementation(s) supporting it?

cneijenhuis:
My thought is that HTTP + JSON isn't the most efficient to start with, but probably the easiest. I'd lean towards easy/widely used concepts over efficient, but less used ones.

Also, I feel like... buffering the whole HTTP body and unmarshalling the JSON aren't big issues to start with. We recommend to keep the events small: https://github.com/cloudevents/spec/blob/master/spec.md#size-limits 
Assuming batches are not gigantic in size, I don't see a big improvement. Even if you have a batch of 100 events of 64 KB each, I don't think intermediaries (which choose to support batching) will have trouble with that.

slinkydeveloper:
> My thought is that HTTP + JSON isn't the most efficient to start with, but probably the easiest. I'd lean towards easy/widely used concepts over efficient, but less used ones.

Agree, that's why I'm not saying we should replace batch. But having somehow a "batch" version of binary mode could allow to perform optimizations of several types.

> Assuming batches are not gigantic in size, I don't see a big improvement. Even if you have a batch of 100 events of 64 KB each, I don't think intermediaries (which choose to support batching) will have trouble with that.

[...]

PR #592, opened 2020-03-23
[DRAFT][WIP] Multipart Structured proposal

slinkydeveloper:
This PR is a draft to explain some ideas from #574 

The goal of this proposal is to allow to efficiently send events inside an HTTP envelope and identify them without the need to parse a huge JSON (batched mode)

This proposal sends events encoded using one or more of the valid data formats inside the various parts of a multipart/form-data envelope.

See #593, #594 for the alternatives

Pros:

* Trivial to implement because most HTTP client/server implementations out there supports the RFC7578
* Allows identification of single events using the `Content-Disposition` header (useful for function invocation use cases)
* Allowing several event formats opens the door to several optimizations from a middleware POV

Cons:

* How do I identify if the multipart/form-data contains cloudevents?
* What should the client/server do when a part is not a valid cloudevent part?

### DIFF
--- a/http-protocol-binding.md
+++ b/http-protocol-binding.md
@@ -429,6 +429,83 @@ Content-Length: nnnn
 
 ```
 
+### 3.4. Multipart Structured Content Mode
+
+In the _multipart structured_ content mode several events are sent into a single HTTP multipart
+request or response body and they can be named. A multipart structured mode contains events serialized 
+using one or several [event formats](#14-event-formats) in the same envelope. The _multipart structured_ content mode is based
+on [RFC7578 "Returning Values from Forms: multipart/form-data"](https://tools.ietf.org/html/rfc7578).
+
+#### 3.4.1. HTTP Content-Type
+
+The [HTTP `Content-Type`][content-type] header MUST be set to the media type of `multipart/form-data` 
+and must include the [boundary parameter](https://tools.ietf.org/html/rfc7578#section-4.1)
+
+```text
+Content-Type: multipart/form-data; boundary=12345
+```
+
+#### 3.4.2. Event Data Encoding
+
+Every part of the `multipart/form-data` can use its own [event format](#14-event-formats).
+
+The specific encoding of the part MUST be set using the `Content-Type` header inside the part:
+
+```text
+Content-Type: application/cloudevents+json; charset=UTF-8
+```
+
+#### 3.4.3. Naming the events
+
+Every part of the multipart envelope containing an event MAY be named, 
+using the `Content-Disposition` header:
+
+```text
+Content-Disposition: form-data; name="event_a"
+```
+
+Names MUST be unique across one single multipart envelope.
+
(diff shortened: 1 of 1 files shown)


http-protocol-binding.md
```diff
@@ -429,6 +429,83 @@ Content-Length: nnnn
 
 ```
 
+### 3.4. Multipart Structured Content Mode
+
+In the _multipart structured_ content mode several events are sent into a single HTTP multipart
+request or response body and they can be named. A multipart structured mode contains events serialized 
+using one or several [event formats](#14-event-formats) in the same envelope. The _multipart structured_ content mode is based
```
bsideup: "Examples" could use an example of the mix of two or more

http-protocol-binding.md
```diff
@@ -429,6 +429,83 @@ Content-Length: nnnn
 
 ```
 
+### 3.4. Multipart Structured Content Mode
```
bsideup: WDYT about adding support for implicit values? Something like Elasticsearch's Bulk API, where some fields can be omitted if they share common value:
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#docs-bulk-api-desc

this way it can help reducing the traffic when big amounts of events (e.g. with the same specversion) need to be transmitted.

http-protocol-binding.md
```diff
@@ -429,6 +429,83 @@ Content-Length: nnnn
...
+
+Every part of the `multipart/form-data` can use its own [event format](#14-event-formats).
+
+The specific encoding of the part MUST be set using the `Content-Type` header inside the part:
+
+```text
+Content-Type: application/cloudevents+json; charset=UTF-8
+```
+
+#### 3.4.3. Naming the events
```
bsideup: Since `name` is optional in RFC7578, I am not sure whether name is _really_ needed here.
Could you please clarify the use of it?

http-protocol-binding.md
```diff
@@ -429,6 +429,83 @@ Content-Length: nnnn
...
+Every part of the multipart envelope containing an event MAY be named, 
+using the `Content-Disposition` header:
+
+```text
+Content-Disposition: form-data; name="event_a"
+```
+
+Names MUST be unique across one single multipart envelope.
+
+#### 3.4.3 Examples
```
bsideup:
```suggestion
#### 3.4.4 Examples
```

If "3.4.3. Naming the events" is preserved

http-protocol-binding.md
slinkydeveloper: This looks like it fits better the `binary` mode

http-protocol-binding.md
slinkydeveloper: the reason for `name` is to allow to quickly identify an event without parsing it. An example use case is for invoking a "function service" with several events, and map the function parameters with the events in the envelope

slinkydeveloper:
Closing in favour of #602

PR #593, opened 2020-03-23
[DRAFT][WIP] Multipart Binary proposal

slinkydeveloper:
This PR is a draft to explain some ideas from #574 

The goal of this proposal is to allow to efficiently send events inside an HTTP envelope and identify them without the need to parse a huge JSON (batched mode)

This proposal defines a new `multipart` content type, compliant to [RFC2046](https://tools.ietf.org/html/rfc2046#section-5.1). Events are sent inside each part in a similar fashion to the binary mode

See #592, #594 for the alternatives

Pros:

* Allows reading the attributes of the event without parsing a full JSON
* Allows identification of the events reusing attributes/extensions
* Allows unbuffered dispatch of the events and other optimizations

Cons:

* Harder to implement than https://github.com/cloudevents/spec/pull/592 because most HTTP server/clients, even if they claim to support multipart bodies, supports only the RFC7578. That means that they strip out all headers from the parts, except the content type and content disposition. Even so, implementing a multipart parser is not hard and most of tools out there can be reused. (this requires further investigation) 

### DIFF
--- a/http-protocol-binding.md
+++ b/http-protocol-binding.md
@@ -429,6 +429,57 @@ Content-Length: nnnn
 
 ```
 
+### 3.4. Multipart Binary Content Mode
+
+In the _multipart binary_ content mode several events are sent into a single HTTP multipart
+request or response body and they can be named. A multipart binary mode contains one event per part, 
+encoded in a similar fashion to [binary content mode]().
+The _multipart binary_ content mode is based on [RFC2046][rfc2046].
+
+#### 3.4.1. HTTP Content-Type
+
+The [HTTP `Content-Type`][content-type] header MUST be set to the media type of `multipart/cloudevents` 
+and must include the [boundary parameter](https://tools.ietf.org/html/rfc2046#section-5.1.1)
+
+```text
+Content-Type: multipart/cloudevents; boundary=12345
+```
+
+#### 3.4.2 Examples
+
+```text
+
+POST /myresource HTTP/1.1
+Host: webhook.example.com
+Content-Type: multipart/cloudevents; boundary=12345
+
+--12345
+ce-specversion: 1.0
+ce-type: com.example.someevent
+ce-time: 2018-04-05T03:56:24Z
+ce-id: 1234-1234-1234
+ce-source: /mycontext/subcontext
+    .... further attributes ...
+Content-Type: application/json; charset=utf-8
+
+{
+    ... application data ...
+}
+--12345
+ce-specversion: 1.0
+ce-type: com.example.anotherevent
+ce-time: 2018-04-05T03:56:24Z
+ce-id: 1234-1234-1234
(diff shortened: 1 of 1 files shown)


http-protocol-binding.md
```diff
@@ -429,6 +429,57 @@ Content-Length: nnnn
 
 ```
 
+### 3.4. Multipart Binary Content Mode
```
bsideup: same comment as in https://github.com/cloudevents/spec/pull/592/files#r396477050

http-protocol-binding.md
```diff
@@ -429,6 +429,57 @@ Content-Length: nnnn
...
+#### 3.4.2 Examples
+
+```text
+
+POST /myresource HTTP/1.1
+Host: webhook.example.com
+Content-Type: multipart/cloudevents; boundary=12345
+
+--12345
+ce-specversion: 1.0
```
bsideup: since we're talking about a binary content mode, it would be nice to reduce the overhead from headers.

Otherwise, header prefixes may take more than 50% of the part size and the format may become ineffective.

bsideup: Maybe the top header can include some sort of a table that assigns indexes to (common) headers. Something like:
```http
POST /myresource HTTP/1.1
Host: webhook.example.com
Content-Type: multipart/cloudevents; boundary=12345
ce-hpack: specversion=0&type=1&time=2&id=3&source=4 /* etc etc */

--12345
0: 1.0
1: com.example.someevent
2: 2018-04-05T03:56:24Z
3: 1234-1234-1234
4: /mycontext/subcontext
```

(not a final proposal, just to demonstrate the idea)

slinkydeveloper: Probably we could define the `ce-hpack` thing directly from the spec

mikehelmick:
Did you consider HTTP/2 multiplexing to solve this issue?

What would be the advantages of this over up-leveling to multiple streams over a single connection?

http-protocol-binding.md
slinkydeveloper: Now that I look again at this, I'm not sure this should be done by this spec directly, because HTTP/2 already has the header compression.

bsideup: does it do it for multipart body tho?

slinkydeveloper:
> Did you consider HTTP/2 multiplexing to solve this issue?

The semantic of HTTP/2 multiplexing is equal to having multiple requests/responses. My goal here is to have multiple events in the context of the same request, so then who uses this encoding can establish to give to a group of events a "semantic" meaning or just treat them as a batch of events unrelated to each other.

http-protocol-binding.md
slinkydeveloper: I'm not sure, but I don't think so...

slinkydeveloper:
Closing in favour of #602

PR #594, opened 2020-03-23
HTTP Json Streaming proposal

slinkydeveloper:
This PR proposes a new HTTP content mode to allow to efficiently send events inside an HTTP envelope and identify them without the need to parse a huge JSON (batched mode).

Look at #574 for more info.

This proposal leverage the [RFC7464](https://www.rfc-editor.org/rfc/rfc7464.html) to send several events, encoded as JSON, delimited by `<RS><LF>`.

See #602 for the alternatives

## Experimental implementation

Based on the implementation done for #602, I propose an experimental implementation in sdk-go: https://github.com/slinkydeveloper/sdk-go/commit/371e0a84f76e1af9da33b0b541d60061e89ec8fc

## Features

Pros:

* Trivial to implement compared to #602
* Allows potentially infinite ordered stream of events

Cons:

* Still requires going back and forth to JSON
* Supports only JSON event format

### DIFF
--- a/http-protocol-binding.md
+++ b/http-protocol-binding.md
@@ -83,7 +83,7 @@ In the _batched_ content mode several events are batched into a single HTTP
 request or response body using an [event format](#14-event-formats) that
 supports batching.
 
-In the _json streaming_ content mode several events are streamed into a single HTTP
+In the _json streaming_ content mode several events are streamed in a single HTTP
 request or response body using the [JSON event format][json-format]. This stream is 
 potentially infinite.
 
--- a/http-protocol-binding.md
+++ b/http-protocol-binding.md
@@ -29,6 +29,7 @@ This document is a working draft.
 - 3.1. [Binary Content Mode](#31-binary-content-mode)
 - 3.2. [Structured Content Mode](#32-structured-content-mode)
 - 3.3. [Batched Content Mode](#33-batched-content-mode)
+- 3.4. [Json Streaming Content Mode](#34-json-streaming-content-mode)
 
 4. [References](#4-references)
 
@@ -65,8 +66,8 @@ which is compatible with HTTP 1.1 semantics.
 
 ### 1.3. Content Modes
 
-This specification defines three content modes for transferring events:
-_binary_, _structured_ and _batched_. Every compliant implementation SHOULD
+This specification defines four content modes for transferring events:
+_binary_, _structured_, _batched_ and _json streaming_. Every compliant implementation SHOULD
 support the _structured_ and _binary_ modes.
 
 In the _binary_ content mode, the value of the event `data` is placed into the
@@ -81,6 +82,10 @@ placed into the HTTP request or response body using an
 In the _batched_ content mode several events are batched into a single HTTP
 request or response body using an [event format](#14-event-formats) that
 supports batching.
+
+In the _json streaming_ content mode several events are streamed into a single HTTP
+request or response body using the [JSON event format][json-format]. This stream is 
+potentially infinite.
 
 ### 1.4. Event Formats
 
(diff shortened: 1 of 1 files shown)


http-protocol-binding.md
```diff
@@ -429,6 +429,48 @@ Content-Length: nnnn
...
+```text
+Content-Type: application/cloudevents-stream; charset=UTF-8
+```
+
+#### 3.4.2. Examples
+
+```text
+PUT /myresource HTTP/1.1
+Host: webhook.example.com
+Content-Type: application/cloudevents-stream; charset=UTF-8
```
bsideup: Maybe a newline delimited JSON (NDJSON) would be a better format here.

Elasticsearch could serve as an example of it used in a wild:
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#docs-bulk-api-desc

slinkydeveloper: I've chosen the RFC7464 because it's a IETF spec, while the NDJSON is not a spec http://jsonlines.org/

http-protocol-binding.md
```diff
@@ -439,6 +446,50 @@ Content-Length: nnnn
 
 ```
 
+### 3.4. Json Streaming Content Mode
+
+In the _json streaming_ content mode several events are streamed into a single HTTP
+request or response body encoding each CloudEvent using the [JSON format][json-format].
+In order to send several events, each event MUST be delimited by `<RS>` and `<LF>`
```
duglin, line 453: Any reason why it's not just a LF?

http-protocol-binding.md
duglin: My concern is more basic... as someone who might write code to produce this stream, I'm much more likely to just do the equivalent of `printf("%s\n", eventJsonString)` and adding a RS is something I would either forget or be annoyed at having to figure out how to do.

I guess I just wondering, why it's not just any random whitespace since with at least one LF between the ending of one and the start of another - wouldn't that be the easiest for people to produce and parse?

duglin:
Aside from the RS question... this seems reasonable but I'd like to hear from someone who has written a streaming service to know if this makes sense to them.

http-protocol-binding.md
slinkydeveloper: So there are a couple of things there to consider:

* As explained in the RFC7464: 

> In prose: any number of JSON texts, each encoded in UTF-8 [RFC3629],
   each preceded by one ASCII RS character, and each followed by a line
   feed (LF).  Since RS is an ASCII control character, it may only
   appear in JSON strings in escaped form (see [RFC7159]), and since RS
   may not appear in JSON texts in any other form, RS unambiguously
   delimits the start of any element in the sequence

The problem with newlines is that you could have it in the middle of a json (for example, if you're sending a pretty printed json), which makes the parsing a whole more complex. Using <RS><LF> is more simpler because you just scan your byte array and when you find those two consecutive chars you split the byte stream and parse the previous part as json (look at the implementation, and you'll get how simple it is and doesn't require any special json parser). With newline, the parser must be more complex because you need to parse progressively and keep track of the "parsing state", unless you assume that the user will never send a pretty printed json...

* On the other side there is the [SSE spec by W3C](https://www.w3.org/TR/eventsource/) sends events as newline delimited, but they assume that the payload never contains a new line (keep in mind that the sse spec is not specific to json, you could send a json but it must be on a single line)

http-protocol-binding.md
```diff
@@ -82,6 +83,10 @@ In the _batched_ content mode several events are batched into a single HTTP
 request or response body using an [event format](#14-event-formats) that
 supports batching.
 
+In the _json streaming_ content mode several events are streamed into a single HTTP
```
cneijenhuis, line 86:
```suggestion
In the _json streaming_ content mode several events are streamed in a single HTTP
```

duglin:
Agreed to close this on the 8/13 call in favor of a websocket mapping spec - Slinky/Scott/Clemens/Lance all agreed to work on that

PR #597, opened 2020-03-26
update discovery API with full REST paths and types.

mikehelmick:
Signed-off-by: Mike Helmick <helmick@google.com>

### DIFF
--- a/discovery.md
+++ b/discovery.md
@@ -101,22 +101,138 @@ be considered an event subscriber and a target user of this API.
 
 ## API Specification
 
-This document structure focuses on discovery where each document has a two part
-key of `provider` and `type`. This takes a denormalized view where the composite
-key of documents is the `provider` and `type`. Provider is defined as an
-arbitrary human readable string and `type` is the CloudEvent `type` attribute.
+This API is specified as a REST API will well defined entities and relationships
+between those entities.
 
-### Provider entity Attributes
+All responses will have a content-type of `application/json` and are either
+a single JSON encoded object or an array of objects.
 
-The `Provider` entity is the main component of the discovery API. This section
-covers the structure of that entity and the next section describes the
-requirements for the query API.
+### Entities in the API
 
-#### provider
+1. `Provider`
+2. `Type`
+3. `Producer`
+
+The `Provider` and `Type` entities form the basis of the directory and can be
+used to build user experiences around discovery. The `Producer` entity is
+keyed off of `{Provider.Name, Type.Name}` and provides the necessary details to
+create a subscription for events of that `ce-type` from the selected `provider`.
+
+### REST Paths
+
+Each path in the REST API represents either represents a list (or search) over
+and entity class, or the retrieval of an individual entity by name. Each of
+these operations MUST be supported by compliant discovery providers.
+
+#### `/types?matching=[search term]`
+
+Returns a list of all types in the discovery system, optionally filtering on
+a provided search term (`matching`).
+
+##### matching
+
+* Type: `string`
+* Description: Search term that provides case insensitive match against `type`
+  names. The parameter can match any portion of the type name.
+* Constraints:
+  * OPTIONAL
+  * If present, must be non-empty
+* Examples:
+  * `"com.storage.object"`
+    * matches:
+      * `"com.storage.object.create"`
+      * `"com.storage.object.delete"`
+      * `"com.storage.object.update"`
+
+##### Returns
+
+A JSON array of `type` entities.
+
+#### `/types/{name}`
+
+Retrieves an individual type entity by exact match on the `type` name.
+
+##### Returns
+
+A `type` entity as a JSON object.
+
+#### `/types/{name}/providers`
+
+Retrieves the details about providers that offer the specified `type`.
+
+##### Returns
+
+A JSON array of `provider` entities.
+
+#### `/providers?matching=[search term]`
+
+Returns a list of all providers int he discovery system, optionally filtering on
+a provided search term (`matching`).
+
+##### matching
+
+* Type: `string`
+* Description: Search term that provides case insensitive match against
+  `provider` names. The parameter can match any portion of the provider name.
+* Constraints:
+  * OPTIONAL
+  * If present, must be non-empty
+* Examples:
+  * `"storage"`
+    * matches:
+      * `"Awesome Cloud Storage"`
+      * `"File storage system"`
+      * `"storage Storage STORAGE"`
+
+##### Returns
+
+A JSON array of `provider` entities.
+
+#### `/providers/{name}`
+
+Retrieves an individual provider entity by exact match on the `provider` name.
+
+##### Returns
+
+A `provider` entity as a JSON object.
+
+#### `/providers/{name}/types`
+
+Retrieves the details about the types that are offered by the specified
+`provider`.
+
+##### Returns
+
+A JSON array of `type` entities
+
+#### `/producer/{provider.name}/{type.name}`
+
+Retrieves the `producer` entity that specifies the information necessary to
+create subscriptions and to consume the events. The `provider.name` and
+`type.name` items in the request path correspond to the information that can
+be obtained via the `/provider` and `/type` API calls.
+
+##### Returns
+
+A `producer` entity as a JSON object.
+
+### Entity Specifications
+
+This section details the fields that make up each of the entities referenced
+earlier in this document.
+
+#### `provider` entity
+
+Used in discovery for enumerating the different providers represented in this
+discovery system.
+
+##### `provider` entity attributes
+
+###### provider
 
 - Type: `String`
-- Description: Identifies the provider of the event. The producer string SHOULD
-  be human readable and can identify a top level producer system.
+- Description: Identifies the provider of the event by nane. The producer string
+  SHOULD be human readable and can identify a top level producer system.
 - Constraints:
   - REQUIRED
   - MUST be a non-empty string
@@ -125,7 +241,15 @@ requirements for the query API.
   - "Awesome Cloud Storage"
   - "com.example.microservices.userlogin"
 
-#### produceruri
+###### description
+
+- Type: `String`
+- Description: Human readable description .
+- Constraints:
+  - OPTIONAL
+  - If present, MUST be a non-empty string
+
+###### provider_uri
 
 - Type: `URI`
 - Description: Absolute URI that provides a link back to the producer, or
@@ -136,7 +260,44 @@ requirements for the query API.
 - Examples:
   - "http://cloud.example.com/docs/blobstorage"
 
-#### type
+
+##### `provider` entity examples
+
+```json
+{
+  "provider": "Cloud Storage Provider",
+  "description": "Blob storage in the cloud",
+  "provider_uri": "https://cloud.example.com/docs/storage"
+}
+```
+
+And a list of valid `provider` entities.
+
+```json
+[
+  {
+    "provider": "Cloud Storage Provider",
+    "description": "Blob storage in the cloud",
+    "provider_uri": "https://cloud.example.com/docs/storage"
+  },
+  {
+    "provider": "Cloud MySQL"
+  },
+  {
+    "provider": "Cloud OtherSQL",
+    "description": "Highly scalable SQL service"
+  }
+]
+```
+
+#### `type` entity
+
+Used in discovery for enumerating the different CloudEvent `type`s represented
+in this discovery system.
+
+##### `type` entity attributes
+
+###### type
 
 - Type: `String`
 - Description: CloudEvents [`type`](https://github.com/cloudevents/spec/blob/master/spec.md#type)
@@ -149,7 +310,7 @@ requirements for the query API.
   - "com.github.pull.create"
   - "com.example.object.delete.v2"
 
-#### specversion
+###### specversion
 
 - Type: `String` per [RFC 2046](https://tools.ietf.org/html/rfc2046)
 - Description: CloudEvents [`specversion`](https://github.com/cloudevents/spec/blob/master/spec.md#specversion)
@@ -159,7 +320,7 @@ requirements for the query API.
   - REQUIRED
   - MUST be a non-empty string
 
-#### datacontenttype
+###### datacontenttype
 
 - Type: `String`
 - Description: CloudEevnts [`datacontenttype`](https://github.com/cloudevents/spec/blob/master/spec.md#datacontenttype)
@@ -169,7 +330,7 @@ requirements for the query API.
   - OPTIONAL
   - If present, MUST adhere to the format specified in [RFC 2046](https://tools.ietf.org/html/rfc2046)
 
-#### dataschema
+###### dataschema
 
 - Type: `URI`
 - Description: CloudEevnts [`datacontenttype`](https://github.com/cloudevents/spec/blob/master/spec.md#dataschema)
@@ -179,7 +340,7 @@ requirements for the query API.
   - OPTIONAL
   - If present, MUST be a non-empty URI
 
-#### dataschematype
+###### dataschematype
 
 - Type: `String` per [RFC 2046](https://tools.ietf.org/html/rfc2046)
 - Description: If using `dataschemacontent` for inline schema storage, the
@@ -191,7 +352,7 @@ requirements for the query API.
   - "application/json"
   -
 
-#### dataschemacontent
+###### dataschemacontent
 
 - Type: `String`
 - Description: An inline representation of the schema of the `data` attribute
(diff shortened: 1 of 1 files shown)


discovery.md
```diff
@@ -101,18 +101,295 @@ be considered an event subscriber and a target user of this API.
 
 ## API Specification
 
-This document structure focuses on discovery where each document has a two part
-key of `provider` and `type`. This takes a denormalized view where the composite
-key of documents is the `provider` and `type`. Provider is defined as an
-arbitrary human readable string and `type` is the CloudEvent `type` attribute.
+This API is specified as a REST API will well defined entities and relationships
```
duglin: s/will/with/   - perhaps?

discovery.md
```diff
@@ -101,18 +101,295 @@ be considered an event subscriber and a target user of this API.
...
 
-This document structure focuses on discovery where each document has a two part
-key of `provider` and `type`. This takes a denormalized view where the composite
-key of documents is the `provider` and `type`. Provider is defined as an
-arbitrary human readable string and `type` is the CloudEvent `type` attribute.
+This API is specified as a REST API with well defined entities and relationships
+between those entities.
 
-### Provider entity Attributes
+All responses will have a content-type of `application/json` and are either
```
duglin, line 107: I wonder if it would read better to put the definition of the resource model first, and then talk about the shape of the API.  This way the people get familiar with the entities and then this API section becomes more of a syntax definitional thing.

discovery.md
```diff
@@ -101,18 +101,295 @@ be considered an event subscriber and a target user of this API.
...
+### Entities in the API
+
+1. `Provider`
+2. `Type`
+3. `Producer`
+
+The `Provider` and `Type` entities form the basis of the directory and can be
+used to build user experiences around discovery. The `Producer` entity is
+keyed off of `{Provider.Name, Type.Name}` and provides the necessary details to
+create a subscription for events of that `ce-type` from the selected `provider`.
```
duglin, line 119: s/provider/Provider/

discovery.md
```diff
@@ -101,18 +101,295 @@ be considered an event subscriber and a target user of this API.
...
+a provided search term (`matching`).
+
+##### matching
+
+* Type: `string`
+* Description: Search term that provides case insensitive match against `type`
+  names. The parameter can match any portion of the type name.
+* Constraints:
+  * OPTIONAL
+  * If present, must be non-empty
```
duglin, line 139: s/must/MUST/

discovery.md
```diff
@@ -101,18 +101,295 @@ be considered an event subscriber and a target user of this API.
...
+keyed off of `{Provider.Name, Type.Name}` and provides the necessary details to
+create a subscription for events of that `ce-type` from the selected `provider`.
+
+### REST Paths
+
+Each path in the REST API represents either represents a list (or search) over
+and entity class, or the retrieval of an individual entity by name. Each of
+these operations MUST be supported by compliant discovery providers.
+
+#### `/types?matching=[search term]`
```
duglin, line 127: Can you put these in the same order as the list above? Meaning Provider, Type and then Producer?

discovery.md
```diff
@@ -101,18 +101,295 @@ be considered an event subscriber and a target user of this API.
...
+* Examples:
+  * `"com.storage.object"`
+    * matches:
+      * `"com.storage.object.create"`
+      * `"com.storage.object.delete"`
+      * `"com.storage.object.update"`
+
+##### Returns
+
+A JSON array of `type` entities.
```
duglin, line 149: Will need to be more formal - e.g.:
  Upon successful processing, the response MUST be a JSON array of `type` entities.
Or something like that. Same for the rest of the doc.

discovery.md
```diff
@@ -101,18 +101,295 @@ be considered an event subscriber and a target user of this API.
...
+
+#### `/types/{name}`
+
+Retrieves an individual type entity by exact match on the `type` name.
```
duglin, line 153: since I can't add a comment down below where I really want to, I'll add it here... I think we'll probably need to be precise on the format of the name for these entities. For example, the list of allowable chars, can it start with a numeric, etc....

discovery.md
duglin: I think we'll need to include a normative definition of what the JSON looks like - e.g. exact spelling and format of the field names, any nesting (hope not), etc.

discovery.md
```diff
@@ -101,18 +101,295 @@ be considered an event subscriber and a target user of this API.
...
-This document structure focuses on discovery where each document has a two part
-key of `provider` and `type`. This takes a denormalized view where the composite
-key of documents is the `provider` and `type`. Provider is defined as an
-arbitrary human readable string and `type` is the CloudEvent `type` attribute.
+This API is specified as a REST API with well defined entities and relationships
+between those entities.
 
-### Provider entity Attributes
+All responses will have a content-type of `application/json` and are either
+a single JSON encoded object or an array of objects.
```
duglin, line 108: For interop I agree with the idea of requiring JSON but I wonder if we should say that other formats are allowed via the normal HTTP `accept` negotiation stuff.  And absence of an `accept` defaults to JSON

discovery.md
```diff
@@ -101,18 +101,295 @@ be considered an event subscriber and a target user of this API.
...
+
+Retrieves the details about providers that offer the specified `type`.
+
+##### Returns
+
+A JSON array of `provider` entities.
+
+#### `/providers?matching=[search term]`
+
+Returns a list of all providers int he discovery system, optionally filtering on
```
duglin, line 169: s/int he/in the/

discovery.md
```diff
@@ -101,18 +101,295 @@ be considered an event subscriber and a target user of this API.
...
+
+### REST Paths
+
+Each path in the REST API represents either represents a list (or search) over
+and entity class, or the retrieval of an individual entity by name. Each of
+these operations MUST be supported by compliant discovery providers.
+
+#### `/types?matching=[search term]`
+
+Returns a list of all types in the discovery system, optionally filtering on
```
duglin, line 129: Might be better to word it as:
MUST return a list of all Types in the discovery system. If the `matching` query parameter is specified then the returned list MUST only include Types that match the `search term` value.

discovery.md
duglin: Is `matching` an optional feature or are all servers required to support it? As written it's required, IMO. If you meant for it to be optional then we'll need to state that someplace.  I'm leaning towards it being required. We just need to discuss whether it should be that any part of the string contains the search-string, or just prefix or postfix - since some people asked about this on a previous call.

discovery.md
```diff
@@ -101,18 +101,295 @@ be considered an event subscriber and a target user of this API.
...
+Retrieves the details about the types that are offered by the specified
+`provider`.
+
+##### Returns
+
+A JSON array of `type` entities
+
+#### `/producer/{provider.name}/{type.name}`
+
+Retrieves the `producer` entity that specifies the information necessary to
```
duglin, line 210: can there be more than one Producer for a given Provider and Type?

discovery.md
```diff
@@ -101,18 +101,295 @@ be considered an event subscriber and a target user of this API.
...
+
+Used in discovery for enumerating the different providers represented in this
+discovery system.
+
+##### `provider` entity attributes
+
+###### provider
+
+- Type: `String`
+- Description: Identifies the provider of the event by nane. The producer string
```
duglin, line 234: I'd avoid the word "Identifies" since that might imply it needs to be unique in some way, which I don't think is your intent. Perhaps something like: A human readable name of the Provider. This attribute does not need to be unique.

Also, you listed it under `provider` but shouldn't be called `name` ?

I wonder if we need to mention which fields can change over time and which MUST remain static. For the purposes of doing updates of a downstream catalog (or just for the consumer to know what's new) we'll need to identify which property is the unique one that MUST NOT change and can be used to uniquely identify it.

discovery.md
```diff
@@ -101,18 +101,295 @@ be considered an event subscriber and a target user of this API.
...
+- Description: Human readable description .
+- Constraints:
+  - OPTIONAL
+  - If present, MUST be a non-empty string
+
+###### provider_uri
+
+- Type: `URI`
+- Description: Absolute URI that provides a link back to the producer, or
+  documentation about the producer. This is for human consumption.
```
duglin, line 256: Not sure what you mean by "human consumption" in this case

discovery.md
```diff
@@ -101,18 +101,295 @@ be considered an event subscriber and a target user of this API.
...
+- Constraints:
+  - OPTIONAL
+  - If present, MUST be a non-empty URI
```
duglin, line 259: you probably mean: If present, MUST be a non-empty absolute URI

discovery.md
```diff
@@ -101,18 +101,295 @@ be considered an event subscriber and a target user of this API.
...
+
+```json
+[
+  {
+    "provider": "Cloud Storage Provider",
+    "description": "Blob storage in the cloud",
+    "provider_uri": "https://cloud.example.com/docs/storage"
+  },
+  {
+    "provider": "Cloud MySQL"
```
duglin, line 284: I was going to suggest we use the provider_uri as the unique key, but since it's optional we may need to add an `id` to each one for that purpose.

discovery.md
```diff
@@ -101,18 +101,295 @@ be considered an event subscriber and a target user of this API.
...
+
+##### `type` entity examples
+
+A single `type` entity.
+
+```json
+{
+  "type": "com.example.storage.object.create",
+  "specversion": "1.0",
+  "datacontenttype", "application/json",
```
duglin, line 375: s/,/:/

discovery.md
```diff
@@ -101,18 +101,295 @@ be considered an event subscriber and a target user of this API.
...
+  attribute.
+- Constraints:
+  - REQUIRED
+  - MUST be a non-empty string, following the constraints as defined in the
+    CloudEvents spec.
+- Examples:
+  - "com.github.pull.create"
+  - "com.example.object.delete.v2"
+
+###### specversion
```
duglin, line 313: Need to think about this more, but it kind of feels like specversion should be more of a global kind of thing and not something that would be Type specific - or at least linked more with the Provider than with the type of the event.

discovery.md
```diff
@@ -101,18 +101,295 @@ be considered an event subscriber and a target user of this API.
...
+}
+```
+
+#### `producer` entity
+
+Once discovery narrows down to a specific event `type` that we want to subscribe
+to from a specific `provider`, the appropriate `producer` entity can be
+retrieved by key. The key is a composite key of the `provider` and `type` names.
+
+##### `producer` entity Attributes
```
duglin, line 386: Not sure if it's just me or not, but I keep mixing up Provider and Producer as I'm reading the spec. So I'm worried other people might do the same thing.  Since Producer is a term we defined in CE, I'm wondering if we should find a different term for Provider in this spec - one that doesn't start with "P".  Actually do we formally define "Provider" in here? I can't seem to find it - I thought it would be in here: https://github.com/cloudevents/spec/blob/a55fcb5874cf76448d3441d2dc7fbb9e700c7470/discovery.md#terminology

deissnerk: No, I don't think we did. I suppose it was me proposing this in some comments and on Slack. _Provider_ is supposed to be a classification/grouping for source. We should definitely discuss how exactly our domain model looks like and then define the new terms.

Thoemmeli: I believe it is even mixed up in the description @duglin was referring to. https://github.com/cloudevents/spec/blob/a55fcb5874cf76448d3441d2dc7fbb9e700c7470/discovery.md#producer-entity-attributes:

> producer entity Attributes
> The Provider entity is the main component of the discovery API. This section covers the structure of that entity and the next section describes the requirements for the query API.

The title is about `producer` but the description talks about `Provider`. It is indeed very confusing. A schematic model diagram might help.

discovery.md
```diff
@@ -101,18 +101,295 @@ be considered an event subscriber and a target user of this API.
...
+
+* Type: `string`
+* Description: Search term that provides case insensitive match against `type`
+  names. The parameter can match any portion of the type name.
+* Constraints:
+  * OPTIONAL
+  * If present, must be non-empty
+* Examples:
+  * `"com.storage.object"`
+    * matches:
```
ryanhorn, line 142: Might help to provide suffix match/exact match examples to be complete. Such as

`"prefix.com.storage.object"` and `"com.storage.object"`

discovery.md
ryanhorn: Agreed, though wouldn't this just follow the format defined in the cloud events spec? https://github.com/cloudevents/spec/blob/v1.0/spec.md#type

discovery.md
```diff
@@ -101,18 +101,295 @@ be considered an event subscriber and a target user of this API.
...
+Retrieves the `producer` entity that specifies the information necessary to
+create subscriptions and to consume the events. The `provider.name` and
+`type.name` items in the request path correspond to the information that can
+be obtained via the `/provider` and `/type` API calls.
+
+##### Returns
+
+A `producer` entity as a JSON object.
+
+### Entity Specifications
```
ryanhorn, line 219: I feel like this should come before the API endpoint definitions, as I found myself reading through those first and not fully grokking the semantics until I got to this section, then going back and re-reading.

discovery.md
ryanhorn, line 375: Something I've been thinking about is whether implementations might want to support more than 1 content type/encoding for a single event type. For example, a subscriber might have tooling and services already oriented around a specific format. Additionally, subscribers might want to request a more compact representation if bandwidth/space is a concern. 

The same could possibly be said about the schema format, though that gets more complex to specify generally, as features across different schema technologies will vary.

discovery.md
ryanhorn: +1, I've been a bit confused as well. The current "provider" definition needs to be more concrete about what a provider represents, imo. Right now it's `Used in discovery for enumerating the different providers represented in this discovery system.`, which uses the term to define itself.

discovery.md
ryanhorn, line 219: I know we're a bit early in converging on a model we all like here, but regardless, it seems like it is going to introduce concepts which aren't part of the current cloud events spec, but might be of interest to event subscribers. Might be worth thinking about exposing new cloud event attributes which enclose a reference to them (or extension attributes if discovery is optional). For example, a link to the type URL, provider URL, etc.

Some of these entities aren't part of the current cloud events spec. Thoughts on whether they might be useful to link in a cloud event instance? For example, we have a "type" identifier, but we're introducing

discovery.md
ryanhorn: I'd argue it's a valid use case. One example I can think of: exposing an audit trail as a particular event type, which could span many different sources, producers, etc.

discovery.md
```diff
@@ -101,18 +101,295 @@ be considered an event subscriber and a target user of this API.
...
+earlier in this document.
+
+#### `provider` entity
+
+Used in discovery for enumerating the different providers represented in this
+discovery system.
+
+##### `provider` entity attributes
+
+###### provider
```
ryanhorn, line 231: The entity itself is already called provider. And in the API endpoints, it seems we reference it by "name", should we use that instead here?

discovery.md
ryanhorn: +1

discovery.md
duglin: yup - so even just a pointer to that spec would be ok.

discovery.md
deissnerk: +1 from me, as well.
I think the provider (do we find a more appropriate term?) will play an important role for discovery.

discovery.md
mikehelmick: done.

discovery.md
mikehelmick: interesting about things remaining static. If a provider name changes - what happens to the associated subscriptions?

discovery.md
mikehelmick: I was using `name` as the unique key. Any reason not to do that?

discovery.md
mikehelmick: For some of those things, separation does make sense. There are items that are important the discovery / subscription configuration state that aren't needed on actual events. I think provider URL is sone of those.

discovery.md
mikehelmick: makes sense - dropped it from here. `specversion` is still available in the producer entity

discovery.md
mikehelmick: I think it's unlikely that systems would offer a variety of encodings of the data attribute on a per-subscriber basis. It would complicate their systems quite a bit.

discovery.md
mikehelmick: Following up from the call - I updated this to `EventProvider` and make the change throughout. I also added a definition.

duglin:
Approved on 4/16 call

PR #600, opened 2020-04-02
Minor updates to Cloud Events Primer

vinayvenkat:
As a recent entrant into the SIG I thought it would be a good opportunity to read through all the documentation and to suggest 
improvements where appropriate.

### DIFF
--- a/primer.md
+++ b/primer.md
@@ -196,7 +196,7 @@ elements that form a layered architecture model.
    with any HTTP method and with request and response messages.
 
 If required to assure broader interoperability, the CloudEvents specification
-set, (may or should) provide specific constraints for event delivery using a particular
+set will provide specific constraints for event delivery using a particular
 application protocol. The [HTTP Webhook](http-webhook.md) specification is not
 specific to CloudEvents and can be used to post any kind of one-way event and
 notifications to a conformant HTTP endpoint. However, the lack of such a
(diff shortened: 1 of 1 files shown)


duglin:
@vinayvenkat can you sign your commit?

vinayvenkat:
@duglin Done. Thanks

primer.md
```diff
@@ -196,7 +196,7 @@ elements that form a layered architecture model.
    with any HTTP method and with request and response messages.
 
 If required to assure broader interoperability, the CloudEvents specification
-set provides specific constraints for event delivery using a particular
+set, (may or should) provide specific constraints for event delivery using a particular
```
duglin, line 199: If we change it, I'd prefer `...the CloudEvents specification set will provide specific...`
I'm not sure how to interpret the `(may or should)`

primer.md
```diff
@@ -253,7 +253,10 @@ The `id` attribute is meant to be a value that is unique across all events
 related to one event source (where each event source is uniquely identified by
 its CloudEvents `source` attribute value). While the exact value used is
 producer defined, receivers of CloudEvents from a single event source can be
-assured that no two events will share the same `id` value. The only exception to
+assured that no two events will share the same `id` value. (We are implicitly
```
duglin, line 256: I'd remove the parens and replace the `? Since` with `, since`

primer.md
```diff
@@ -677,7 +682,9 @@ came from and where it might be going. This prevents tooling to facilitate
 successful event delivery and consumers from knowing what to do with event data.
 
 CloudEvents offers useful metadata which middleware and consumers can rely upon
-to facilitate event routing, logging, delivery and receipt.
+to facilitate event routing, logging, delivery and receipt. (Not sure if we
```
duglin, line 685: You should remove the commentary since we can't merge with it in there.
It's not meant to "impact" it, but it is meant to help "facilitate" it.

primer.md
vinayvenkat: Makes sense. I put in parenthesis as I was not sure if we should state that in the primer. Although it makes it more clear with your suggestion.

primer.md
vinayvenkat: makes sense.

duglin:
Approved on 4/9 call

PR #602, opened 2020-04-09
HTTP Multipart content mode

slinkydeveloper:
This PR merges the work done in #592 and #593 into a single content mode.

The goal of this proposal is to allow mapping efficiently more events inside an HTTP envelope without the need to parse a huge JSON (batched mode). For more details on the rationale also look at #574. 

This proposal defines a new `multipart` content type called `multipart/cloudevents`, compliant to [RFC2046](https://tools.ietf.org/html/rfc2046#section-5.1). 
Because each part has an Header and a Body section, it can encode the event as structured or binary mode similar to how the spec already defines these two content modes.

This content mode, similar to batching, doesn't define any semantic meaning between the different events sent in the same envelope. This means that is up to the user to give them a meaning or just treat them as independent between each other.

## Features

* Allows unbuffered dispatch of the events and other optimizations
* Because it already reuses the same concepts of binary and structured mode, the implementation is relatively easy for the SDKs (look below)
* The multipart envelopes are potentially infinite, allowing to use this content mode as a sort of "streaming" of events

## Experimental implementation

A sample implementation for sdk-go lives here: https://github.com/slinkydeveloper/sdk-go/commits/multipart_experiment

For reading multipart request/responses, the client directly leverage the already existing parsing logic for structured and binary mode: https://github.com/slinkydeveloper/sdk-go/commit/355a3cec5f624b732fc7f3ff67c09900e20d07e4#diff-584b2042f74d673322bd7a717e3d66b3R58

For writing the multipart request/response the client loops among the several messages of the `MultiMessage` interface and writes the single parts reusing the `binding.Write` function, used in the sdk to write all kind of messages: https://github.com/slinkydeveloper/sdk-go/commit/355a3cec5f624b732fc7f3ff67c09900e20d07e4#diff-e9340c4352f2c852878f7b66bba5967cR29 

In particular, there are three examples to test it:

* [Send/Receive](https://github.com/slinkydeveloper/sdk-go/blob/355a3cec5f624b732fc7f3ff67c09900e20d07e4/v2/cmd/samples/http_multipart/random_multi_event_send_receive/main.go): Demo of a server and a client which sends several events and, without any event parsing, applies some transformations, replying with transformed events
* [Send](https://github.com/slinkydeveloper/sdk-go/blob/355a3cec5f624b732fc7f3ff67c09900e20d07e4/v2/cmd/samples/http_multipart/random_multi_event_send/main.go): HTTP client that sends 5 random events
* [Reply with N events](https://github.com/slinkydeveloper/sdk-go/blob/355a3cec5f624b732fc7f3ff67c09900e20d07e4/v2/cmd/samples/http_multipart/random_multi_event_response/main.go): HTTP server (with upgrade to HTTP/2) that replies with N events

A sample curl response with multipart/cloudevents (including HTTP/2 upgrade):

```
% curl localhost:8080\?num=3 --http2 -v                                                                                                                                                                                            
> GET /?num=3 HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.69.1
> Accept: */*
> Connection: Upgrade, HTTP2-Settings
> Upgrade: h2c
> HTTP2-Settings: AAMAAABkAAQCAAAAAAIAAAAA
> 
< HTTP/1.1 101 Switching Protocols
< Connection: Upgrade
< Upgrade: h2c
< HTTP/2 200 
< content-type: multipart/cloudevents; boundary=63a561a72502161512131752ba061292d5948dbbceba58953f12c6ebc938
< content-length: 1153
< date: Thu, 09 Apr 2020 09:21:22 GMT
< 
--63a561a72502161512131752ba061292d5948dbbceba58953f12c6ebc938
Content-Type: application/cloudevents+json

{"data":{"id":0,"message":"hello world"},"datacontenttype":"application/json","id":"06dda122-2430-4199-b3de-980b9891632f","source":"https://github.com/cloudevents/sdk-go/v2/cmd/samples/http_multipart/random_multi_event_response","specversion":"1.0","type":"com.cloudevents.multipart.sample"}
--63a561a72502161512131752ba061292d5948dbbceba58953f12c6ebc938
Ce-Id: 4f647dbe-7dfd-4e8f-a8fe-09dafbb45238
Ce-Source: https://github.com/cloudevents/sdk-go/v2/cmd/samples/http_multipart/random_multi_event_response
Ce-Specversion: 1.0
Ce-Type: com.cloudevents.multipart.sample

{"id":1,"message":"hello world"}
--63a561a72502161512131752ba061292d5948dbbceba58953f12c6ebc938
Content-Type: application/cloudevents+json

{"data":{"id":2,"message":"hello world"},"datacontenttype":"application/json","id":"de8ca4b6-72a0-4f37-afbc-0da7beb9914c","source":"https://github.com/cloudevents* Connection #0 to host localhost left intact
/sdk-go/v2/cmd/samples/http_multipart/random_multi_event_response","specversion":"1.0","type":"com.cloudevents.multipart.sample"}%  
```

## TODO

* [ ] Should this content mode (maybe together with batching) live in another subspec, like _HTTP multi event binding spec_?
* [ ] Investigate HTTP Headers packing (https://github.com/cloudevents/spec/pull/593#discussion_r396489323)

### DIFF
--- a/http-protocol-binding.md
+++ b/http-protocol-binding.md
@@ -83,8 +83,8 @@ In the _batched_ content mode several events are batched into a single HTTP
 request or response body using an [event format](#14-event-formats) that
 supports batching.
 
-In the _multipart_ content mode several events are sent into a single HTTP
-request or response body allowing to represent, on each part, an event or in 
+In the _multipart_ content mode several events are sent in a single HTTP
+request or response body allowing to represent, on each part, an event in 
 binary mode or in structured mode.
 
 ### 1.4. Event Formats
@@ -447,9 +447,9 @@ Content-Length: nnnn
 
 ### 3.4. Multipart Content Mode
 
-In the _multipart_ content mode several events are sent into a single HTTP multipart
+In the _multipart_ content mode several events are sent in a single HTTP multipart
 request or response as described in [RFC2046, section 5.1][rfc2046-section-5.1]. 
-The _multipart_  content mode maps one event per part. Because each part consists in an 
+The _multipart_  content mode maps one event per part. Because each part consists in a 
 group of headers and a body, an event could be encoded following the same rules that applies
 to [Structured Content Mode](#32-structured-content-mode) and [Binary Content Mode](#31-binary-content-mode)
 within the part, respectively named _structured part_ and _binary part_.
@@ -459,7 +459,7 @@ Every part of the HTTP envelope MUST be or a _structured part_ or a _binary part
 #### 3.4.1. HTTP Content-Type
 
 The [HTTP `Content-Type`][content-type] header MUST be set to the media type of `multipart/cloudevents` 
-and must include the [boundary parameter](https://tools.ietf.org/html/rfc2046#section-5.1.1)
+and MUST include the [boundary parameter](https://tools.ietf.org/html/rfc2046#section-5.1.1)
 
 ```text
 Content-Type: multipart/cloudevents; boundary=12345
@@ -467,15 +467,15 @@ Content-Type: multipart/cloudevents; boundary=12345
 
 #### 3.4.2. Structured Part
 
-All rules defined in [Structured Content Mode](#32-structured-content-mode) applies from the beginning of the 
-part up to the end of the part, outlined by the boundary delimiter.
+A _structured part_ MUST conform to all rules defined by [Structured Content Mode](#32-structured-content-mode)
+from the beginning up to to the end of the part, outlined by the boundary delimiter.
 
 More [event formats](#14-event-formats) for several parts in the same envelope are allowed.
 
 #### 3.4.3. Binary Part
 
-All rules defined in [Binary Content Mode](#31-binary-content-mode) applies from the beginning of the 
-part up to the end of the part, outlined by the boundary delimiter.
+A _binary part_ MUST conform to all rules defined by [Binary Content Mode](#31-binary-content-mode)
+from the beginning up to to the end of the part, outlined by the boundary delimiter.
 
 #### 3.4.4. Distinguish between structured and binary part
 
--- a/http-protocol-binding.md
+++ b/http-protocol-binding.md
@@ -439,6 +439,57 @@ Content-Length: nnnn
 
 ```
 
+### 3.4. Multipart Binary Content Mode
+
+In the _multipart binary_ content mode several events are sent into a single HTTP multipart
+request or response body and they can be named. A multipart binary mode contains one event per part, 
+encoded in a similar fashion to [binary content mode]().
+The _multipart binary_ content mode is based on [RFC2046][rfc2046].
+
+#### 3.4.1. HTTP Content-Type
+
+The [HTTP `Content-Type`][content-type] header MUST be set to the media type of `multipart/cloudevents` 
+and must include the [boundary parameter](https://tools.ietf.org/html/rfc2046#section-5.1.1)
+
+```text
+Content-Type: multipart/cloudevents; boundary=12345
+```
+
+#### 3.4.2 Examples
+
+```text
+
+POST /myresource HTTP/1.1
+Host: webhook.example.com
+Content-Type: multipart/cloudevents; boundary=12345
+
+--12345
+ce-specversion: 1.0
+ce-type: com.example.someevent
+ce-time: 2018-04-05T03:56:24Z
+ce-id: 1234-1234-1234
+ce-source: /mycontext/subcontext
+    .... further attributes ...
+Content-Type: application/json; charset=utf-8
+
+{
+    ... application data ...
+}
+--12345
+ce-specversion: 1.0
+ce-type: com.example.anotherevent
+ce-time: 2018-04-05T03:56:24Z
+ce-id: 1234-1234-1234
+ce-source: /mycontext/subcontext
+    .... further attributes ...
+Content-Type: application/json; charset=utf-8
+
+{
+    ... application data ...
+}
+--12345
+```
+
 ## 4. References
 
 - [RFC2046][rfc2046] Multipurpose Internet Mail Extensions (MIME) Part Two:
--- a/http-protocol-binding.md
+++ b/http-protocol-binding.md
@@ -29,6 +29,7 @@ This document is a working draft.
 - 3.1. [Binary Content Mode](#31-binary-content-mode)
 - 3.2. [Structured Content Mode](#32-structured-content-mode)
 - 3.3. [Batched Content Mode](#33-batched-content-mode)
+- 3.4. [Multipart Content Mode](#34-multipart-content-mode)
 
 4. [References](#4-references)
 
@@ -65,8 +66,8 @@ which is compatible with HTTP 1.1 semantics.
 
 ### 1.3. Content Modes
 
-This specification defines three content modes for transferring events:
-_binary_, _structured_ and _batched_. Every compliant implementation SHOULD
+This specification defines four content modes for transferring events:
+_binary_, _structured_, _batched_ and _multipart_. Every compliant implementation SHOULD
 support the _structured_ and _binary_ modes.
 
 In the _binary_ content mode, the value of the event `data` is placed into the
@@ -81,6 +82,10 @@ placed into the HTTP request or response body using an
 In the _batched_ content mode several events are batched into a single HTTP
 request or response body using an [event format](#14-event-formats) that
 supports batching.
+
+In the _multipart_ content mode several events are sent into a single HTTP
+request or response body allowing to represent, on each part, an event or in 
+binary mode or in structured mode.
 
 ### 1.4. Event Formats
 
@@ -136,12 +141,13 @@ events using a particular mode might be defined by an application, but are not
 defined here. The _batched_ mode MUST NOT be used unless solicited, and the
 gesture SHOULD allow the receiver to choose the maximum size of a batch.
 
-The receiver of the event can distinguish between the three modes by inspecting
+The receiver of the event can distinguish between the four modes by inspecting
 the `Content-Type` header value. If the value is prefixed with the CloudEvents
 media type `application/cloudevents`, indicating the use of a known
 [event format](#14-event-formats), the receiver uses _structured_ mode. If the
 value is prefixed with `application/cloudevents-batch`, the receiver uses the
-_batched_ mode. Otherwise it defaults to _binary_ mode.
+_batched_ mode. If the value is prefixed with `multipart/cloudevents`, the 
+receiver uses the _multipart_ mode. Otherwise it defaults to _binary_ mode.
 
 If a receiver detects the CloudEvents media type, but with an event format that
 it cannot handle, for instance `application/cloudevents+avro`, it MAY still
@@ -439,12 +445,16 @@ Content-Length: nnnn
 
 ```
 
-### 3.4. Multipart Binary Content Mode
+### 3.4. Multipart Content Mode
 
-In the _multipart binary_ content mode several events are sent into a single HTTP multipart
-request or response body and they can be named. A multipart binary mode contains one event per part, 
-encoded in a similar fashion to [binary content mode]().
-The _multipart binary_ content mode is based on [RFC2046][rfc2046].
+In the _multipart_ content mode several events are sent into a single HTTP multipart
+request or response as described in [RFC2046, section 5.1][rfc2046-section-5.1]. 
+The _multipart_  content mode maps one event per part. Because each part consists in an 
+group of headers and a body, an event could be encoded following the same rules that applies
+to [Structured Content Mode](#32-structured-content-mode) and [Binary Content Mode](#31-binary-content-mode)
+within the part, respectively named _structured part_ and _binary part_.
+
+Every part of the HTTP envelope MUST be or a _structured part_ or a _binary part_.
 
 #### 3.4.1. HTTP Content-Type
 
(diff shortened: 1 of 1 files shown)


duglin:
I assume the structured example above is hand written, right? There are typos in the property names.. they shouldn’t be prefixes with CE-

http-protocol-binding.md
```diff
@@ -82,6 +83,10 @@ In the _batched_ content mode several events are batched into a single HTTP
 request or response body using an [event format](#14-event-formats) that
 supports batching.
 
+In the _multipart_ content mode several events are sent into a single HTTP
```
duglin, line 86: s/into/in/

http-protocol-binding.md
```diff
@@ -82,6 +83,10 @@ In the _batched_ content mode several events are batched into a single HTTP
...
+request or response body allowing to represent, on each part, an event or in
```
duglin, line 87: s/or in/in/

http-protocol-binding.md
```diff
@@ -439,6 +445,119 @@ Content-Length: nnnn
 
 ```
 
+### 3.4. Multipart Content Mode
+
+In the _multipart_ content mode several events are sent into a single HTTP multipart
```
duglin, line 450: s/in to/in/

http-protocol-binding.md
```diff
@@ -439,6 +445,119 @@ Content-Length: nnnn
...
+request or response as described in [RFC2046, section 5.1][rfc2046-section-5.1]. 
+The _multipart_  content mode maps one event per part. Because each part consists in an
```
duglin, line 452: s/in an/in a/

http-protocol-binding.md
```diff
@@ -439,6 +445,119 @@ Content-Length: nnnn
...
+### 3.4. Multipart Content Mode
+
+In the _multipart_ content mode several events are sent into a single HTTP multipart
+request or response as described in [RFC2046, section 5.1][rfc2046-section-5.1]. 
+The _multipart_  content mode maps one event per part. Because each part consists in an 
+group of headers and a body, an event could be encoded following the same rules that applies
+to [Structured Content Mode](#32-structured-content-mode) and [Binary Content Mode](#31-binary-content-mode)
+within the part, respectively named _structured part_ and _binary part_.
+
+Every part of the HTTP envelope MUST be or a _structured part_ or a _binary part_.
```
duglin, line 457: Interesting. While we’re obviously mostly concerned about CE, what is the harm in allowing any random http message to be part of the stream?

http-protocol-binding.md
```diff
@@ -439,6 +445,119 @@ Content-Length: nnnn
...
+
+#### 3.4.1. HTTP Content-Type
+
+The [HTTP `Content-Type`][content-type] header MUST be set to the media type of `multipart/cloudevents`
```
duglin, line 461: I guess if we require this media type then my previous comment doesn’t make as much sense, however what if we just used multipart/mixed? Or do we really need to know we’re dealing with CEs in advance?

http-protocol-binding.md
```diff
@@ -439,6 +445,119 @@ Content-Length: nnnn
...
+group of headers and a body, an event could be encoded following the same rules that applies
+to [Structured Content Mode](#32-structured-content-mode) and [Binary Content Mode](#31-binary-content-mode)
+within the part, respectively named _structured part_ and _binary part_.
+
+Every part of the HTTP envelope MUST be or a _structured part_ or a _binary part_.
+
+#### 3.4.1. HTTP Content-Type
+
+The [HTTP `Content-Type`][content-type] header MUST be set to the media type of `multipart/cloudevents` 
+and must include the [boundary parameter](https://tools.ietf.org/html/rfc2046#section-5.1.1)
```
duglin, line 462: s/must/MUST/

http-protocol-binding.md
```diff
@@ -439,6 +445,119 @@ Content-Length: nnnn
...
+
+```text
+Content-Type: multipart/cloudevents; boundary=12345
+```
+
+#### 3.4.2. Structured Part
+
+All rules defined in [Structured Content Mode](#32-structured-content-mode) applies from the beginning of the
```
duglin, line 470: I think there needs to be a MUST in here someplace.

http-protocol-binding.md
```diff
@@ -439,6 +445,119 @@ Content-Length: nnnn
...
+#### 3.4.2. Structured Part
+
+All rules defined in [Structured Content Mode](#32-structured-content-mode) applies from the beginning of the 
+part up to the end of the part, outlined by the boundary delimiter.
+
+More [event formats](#14-event-formats) for several parts in the same envelope are allowed.
+
+#### 3.4.3. Binary Part
+
+All rules defined in [Binary Content Mode](#31-binary-content-mode) applies from the beginning of the
```
duglin, line 477: And here

http-protocol-binding.md
slinkydeveloper: * Why a user should be interested in doing it? There are any use cases for this? 
* How a sender should handle it? Should our SDKs allow to send a part non cloudevent?
* How a receiver should handle it? It should discard the message? It should put it in a queue of non accepted messages?

http-protocol-binding.md
slinkydeveloper: How a receiver could distinguish between a multipart containing "generic stuff + cloudevents" and a multipart containing only "cloudevents". I think the same questions that i did in the comment above applies here.

http-protocol-binding.md
slinkydeveloper: reworked the sentence, check it now

JemDay:
This may well have been discussed in a meeting i missed but handling for these types of batched events needs to be addressed in the WebHook specification.

If i'm an endpoint, and i receive a multi-part message where one of the entries is mal-formed in some way do i reject the whole batch with a 4XX error ?

As i think about establishing an event gateway my thought was to return a 200 if the event made it synchronously to the handler or a 202 if we dropped it on a queue for deferred handling.

If i receive a batch of 5 events (at the gateway) with different types that i'm going to dispatch onto  different transports internally i now need to be atomic across those resources (an all-or-none-receipt) which is a big red-flag to me.

I'm not saying this approach isn't worthwhile, but it will require very clear guidance for receivers and intermediaries.

http-protocol-binding.md
duglin: It seems to me that if the sender looks at this open connection as a way to send any stream of http messages, then why should it be required to open one for CEs and one for non-CEs? If the receiver treats each section as an independent HTTP message, then I'm not sure I see the issue. Isn't that kind of the point of multi-part messages to begin with? And why each one looks like an independent message with it's own set of HTTP headers.

http-protocol-binding.md
duglin: Why does a receiver need to know if there are CEs in there at all? Why can't it treat each part as an http message and if it happens to be a CE then it hands it to the CE processor.

slinkydeveloper: how sdks should handle a non CE part? should they discard it?

http-protocol-binding.md
slinkydeveloper: The problem that i see without the "exclusivity" is that the spec then becomes inherently more complex to implement, because at the same time you should allow to send "http pure parts" and you should be able to receive them from the other side. While SDKs now relies on the abstraction of event (or message) to send/receive every kind of thing, adding the ability to have non event parts brings the problem to send/receive some protocol specific payloads.

slinkydeveloper:
> I'm not saying this approach isn't worthwhile, but it will require very clear guidance for receivers and intermediaries.

I'm not sure this guidance should come from this http multipart spec: some users could decide to do always async processing, enqueing the results and sending back a 202, some could decide to synchronously process the events and eventually fail fast if there was an error with one of them, some more elaborated could send back a response with id of "acked" messages. But all these strategies IMO should be up to the user.

slinkydeveloper:
For people reviewing this PR, do not forget that I also elaborated this one https://github.com/cloudevents/spec/pull/594 with similar concepts but using JSON Streaming, which is simpler to implement and widely supported

http-protocol-binding.md
duglin: While I do like the idea of leveraging existing technology, I think what concerns me the most is that we're creating a one-off/specialized multi-part envelope. I don't know how many other content-types do that, but it seems odd to me to suggest that each should.

What would an SDK do today if it was receiving an http message that wasn't a CE? Drop it? It seems the same answer would apply for all cases (multi-part batched or not).

We also need to look at this from a spec perspective, not just from an SDK perspective. How the SDKs choose to support this is up to them. Today the SDKs might be written with the assumption that they are the HTTP server getting the events from the internet. However, who is to say that is correct? It is just as valid to have a layer on top of the CE SDK that looks at each message coming in (or each part in the multi-part) and based on whether it's a CE or not send it to a different processor. In that case the CE SDK at that processor can be assured it is only getting CEs. So, putting something in the spec that implies a multi-part message MUST only have CEs might be too limiting.

Unless you're saying this is ONLY for cases where we're only ever going to be sending CEs, but then I'm back to.... should each content type create their own multi-part content type when the only thing they send are messages of that type?

http-protocol-binding.md
duglin: what does it do today?

slinkydeveloper: An error, but i think the comparison with a single message doesn't hold here because the semantic is pretty clear: or i get an error or i get an event. What happens here?

http-protocol-binding.md
slinkydeveloper: So let me understand your proposal:

* Do you think we should allow non-event parts?
* Do you think we should use the standard `multipart/mixed` more than `multipart/cloudevents`?
* Both of the above points?

http-protocol-binding.md
```diff
@@ -439,6 +445,119 @@ Content-Length: nnnn
...
+### 3.4. Multipart Content Mode
+
+In the _multipart_ content mode several events are sent in a single HTTP multipart
+request or response as described in [RFC2046, section 5.1][rfc2046-section-5.1]. 
+The _multipart_  content mode maps one event per part. Because each part consists in a 
+group of headers and a body, an event could be encoded following the same rules that applies
+to [Structured Content Mode](#32-structured-content-mode) and [Binary Content Mode](#31-binary-content-mode)
+within the part, respectively named _structured part_ and _binary part_.
+
+Every part of the HTTP envelope MUST be or a _structured part_ or a _binary part_.
```
deissnerk, line 457: > or a _structured part_ or a _binary part_

either a _structured part_  or a _binary part_

http-protocol-binding.md
deissnerk: Isn't it just the same as with a single event? I would say that it depends on what the receiver expects. If you set up an endpoint that accepts only CloudEvents, you return an error. If your endpoint deals with mixed messages, it has to differentiate. 
Even if we define `multipart/cloudevents`, wouldn't an SDK have to be prepared for `multipart/mixed`, too? Do we exclude putting CloudEvents into a mixed multipart message?

clemensv:
I have a hard time understanding why you are not just using RFC2046 5.1.3 "mixed" https://tools.ietf.org/html/rfc2046#section-5.1.3 if you want to send multiple entities with Multipart MIME. 

This is special casing a CloudEvents scenario where the general case works pretty well for mixed payloads already.

JemDay:
> I'm not sure this guidance should come from this http multipart spec:

I would agree, but i do think the http-webhook.md needs to be changed at the same time to set expectations as to how bulk messages should be handled.

clemensv:
> > I'm not sure this guidance should come from this http multipart spec:
> 
> I would agree, but i do think the http-webhook.md needs to be changed at the same time to set expectations as to how bulk messages should be handled.

The webhook spec is completely payload agnostic and only talks about the HTTP request mechanics. [RFC2046 5.1](https://tools.ietf.org/html/rfc2046#section-5.1) covers multipart MIME.

http-protocol-binding.md
clemensv: > * Why a user should be interested in doing it? There are any use cases for this?
> * How a sender should handle it? Should our SDKs allow to send a part non cloudevent?
> * How a receiver should handle it? It should discard the message? It should put it in a queue of non accepted messages?

A user might want to bundle a CloudEvent with an attached image, e.g. a picture that accompanies an "vehicle ran red light" event raised by a traffic camera where the payload of the event references the picture in the same bundle using a URI. That is how Email attachments work.

Multipart MIME with "mixed" covers the scenarios of packaging multiple entities far better than the special-casing you propose here. 

The problem that is being tried to solve here is already solved, and better solved, by RFC2046 5.1

http-protocol-binding.md
slinkydeveloper: > Do we exclude putting CloudEvents into a mixed multipart message?

We could avoid excluding such use case, saying that you could eventually use this spec to send events in your own mixed multipart.

> wouldn't an SDK have to be prepared for multipart/mixed, too?

I don't think so. To avoid leaking/exposing the internals of the communication/http protocol, the SDKs needs somehow to distinguish between event and non-event. If we allow non cloudevents in a multi-cloudevent request/response, how such situation should be "modeled" by the sdk without exposing the internals? Does it just returns an error on the single part?

slinkydeveloper:
> This is special casing a CloudEvents scenario where the general case works pretty well for mixed payloads already.

@clemensv I think this is similar to what @dug already mentioned: should the spec allow non cloudevents inside the payload? If yes, how the SDKs (or any other implementer) should handle such use cases? https://github.com/cloudevents/spec/pull/602#discussion_r409255682

slinkydeveloper:
BTW I've reorganized the https://github.com/cloudevents/spec/pull/594 to propose another solution to implement multi events in the same request/response. The #594 is far simpler to implement, but you still need to parse a JSON.

duglin:
Agreed to close this on the 8/13 call in favor of a websocket mapping spec - Slinky/Scott/Clemens/Lance all agreed to work on that

Issue #575, opened 2020-02-28
Mandatory fields in Avro representation of CloudEvents.


arvindkasale08:
Hi Team,

We are analyzing the use of cloudevents-avro format to send Kafka events. The native/core cloud event representation has the **id**, **source**, **specversion** and **type** as the mandatory fields in a cloud event [here](https://github.com/cloudevents/spec/blob/v1.0/spec.md) 
Contrary to this the cloud-event avro representation has a generic **attributes** map and a provision for **data** [here](https://github.com/cloudevents/spec/blob/v1.0/avro-format.md) 
We had the following questions.

1. What is the specific rationale behind losing out on the strong typing and validations in the core cloud event in the avro representation. Is there a benefit in keeping a attribute map inplace of specific strongly typed id, source, specversion and type fields with other optional metadata as attributes.

2. How does this map to when working with knative eventing. How does a system with a KafkaSource route the messages based on a cloudevent avro representation.

3. In knative eventing with a KafkaSource using avro is there a predisposition toward using binary or structured mode?

4. How would the schema validation for such a cloudevent take place if the data fields are optional?

duglin:
ping @fabiojose

BigJerBD:
+1 ,  

I also have issues understanding the existence of the attribute field, when the structured cloudevent put it all on the same level. 

the [example](https://github.com/cloudevents/spec/blob/v1.0/avro-format.md#4-examples) shown put the `data` and all the other attribute` on the same level. However the schema specifies them in two different fields,

fabiojose:
@arvindkasale08 Lets start a revision of this current schema definition. What do you think?

fabiojose:
May something like that, but maybe necessary to take into account the historical steps to reach the current schema.

Instead of generic attribute, we may have something like this for required attributes:
```json
 {
      "name": "id",
      "type": "string"
}
```

And this, for the optional one:
```json
 {
      "name": "dataschema",
      "type": ["null", "string"]
}
```

BigJerBD:
Our organisation which is currently designing a custom avro-cloudevent, with extra fields based on our requirements, currently prefer having required attributes in that fashion.

we also put 1.0 as the default spec version : 
```
{"name": "specversion",
 "type": "string",
 "default": "1.0",}
```
the use of builtin logicalTypes is also considered. Avro support `uuid` and `timestamp-millis`  for those for the `id` and `time` fields. I wonder if it could be something to be appropriate for the official cloudevent avro-format

[...]

PR #613, opened 2020-04-30
[WIP] Cloudevents avro improvement proposal

BigJerBD:
Proposal to change the Avro CloudEvent format for better field validation and better use with SDKs

#### Changing namespace
There is potential namespace collision because of the currently used namespace is and can be used in SDKs for an other non Avro CloudEvent  class

For example the Java SDK (`io.cloudevents:cloudevents-api:1.3.0`)  has a CloudEvent on the same location than the avro CloudEvent class, meaning that using avro code generation with this SDK is difficult, 

#### Using mandatory fields instead of a mapping
as mentioned in this [issue](https://github.com/cloudevents/spec/issues/575),  the mapping of value was selected to support forward and backward compatibility, however it causes problem since mandatory fields could be absent and no field validation is present.

using mandatory fields with defaults allow to have forward and backward compatibility and type verification.  

#### Use of builtin logical types
Avro support some builtin logical type `timestamp-millis` and `uuid`  for some extra validation. Since the resulting serialized message does not change, it does not cause any forward/backward compatibility problems 

Thanks for the consideration :)

### DIFF
--- a/spec.avsc
+++ b/spec.avsc
@@ -1,22 +1,56 @@
 {
-  "namespace":"io.cloudevents",
+  "namespace": "io.cloudevents.avro",
   "type":"record",
   "name":"CloudEvent",
   "version":"1.0",
   "doc":"Avro Event Format for CloudEvents",
   "fields":[
     {
-      "name":"attribute",
-      "type":{
-        "type":"map",
-        "values":[
-          "null",
-          "boolean",
-          "int",
-          "string",
-          "bytes"
-        ]
-      }
+      "name": "specversion",
+      "type": "string",
+      "default": "1.0"
+    },
+    {
+      "name": "type",
+      "type": "string"
+    },
+    {
+      "name": "source",
+      "type": "string"
+    },
+    {
+      "name": "id",
+      "type": "string",
+      "logicalType": "uuid"
+    },
+    {
+      "name": "time",
+      "type": "string",
+      "logicalType": "timestamp-millis"
+    },
+    {
+      "name": "dataschema",
+      "type": [
+        "null",
+        "string"
+      ],
+      "default": null
+    },
+    {
+      "name": "datacontenttype",
+      "type": [
+        "null",
+        "string"
+      ],
+      "default": null
+    },
+    {
+      "name": "subject",
+      "type": [
+        "null",
+        "string"
+      ],
+      "default": null
     },
     {
       "name": "data",


duglin:
@BigJerBD can you sign your commit?

duglin:
@fabiojose any comments?
Is this similar to the Java SDK conversation going on?  @slinkydeveloper

spec.avsc
```diff
@@ -1,22 +1,56 @@
 {
-  "namespace":"io.cloudevents",
+  "namespace": "io.cloudevents.avro",
```
slinkydeveloper, line 2: Really nice!

fabiojose:
According to the goals of Avro schema defined for this spec, it was created to be able to accommodate future changes without changing the schema itself.

Take a look at this good [explanation by @duglin](https://github.com/cloudevents/spec/issues/575#issuecomment-621171183)

But, without named fields, all the work must be done by SDKs. Comparing to JSON, we may use the SDK, but we could easily understand the event without using any of it. With the current Avro schema, we can't.

I think this is a question to raise in the call because now we have experienced the current Avro schema and I guess we need to __choose__ between __schema flexibility__ or __ease to use__.

spec.avsc
BigJerBD: we would need for sure probably a schema for 0.3 , the `schemaurl` would need to have a default of `null` like the 1.0 for `dataschema`.  this will ensure forward and backward compatiblity. Changing the namespace to `io.cloudevents.v03.avro` also would be needed i think

The most confusing part is indeed the implementation with SDK's, since an avro schema can be used to generate code in multiple language, the need of the java-sdk is not that useful, because  you could use the Generated CloudEvent for validation with avro packages for serialization. 

adding format/avro  could however  be one of these: 
- implementing a serialization that take sdk's cloudevent, converting it to avro cloudevent ( without validation ) then serializing it with avro libraries.  This conversion is necessary since we cannot put directly the avroCloudEvent in the EventFormat class because it takes sdk's cloudevent as an argument. The serializer would also need to do version handling like the `formats/json` do 

However as a developer It is tempting to just use the Avro CloudEvents directly since it provides all the necessary features without the EventFormat overhead

here's an example kafka Serializer for avro  for guidance 
```
import org.apache.avro.generic.GenericDatumWriter;
import org.apache.avro.generic.GenericRecord;
import org.apache.avro.io.BinaryEncoder;
import org.apache.avro.io.DatumWriter;
import org.apache.avro.io.EncoderFactory;
import org.apache.avro.specific.SpecificRecordBase;
import org.apache.kafka.common.errors.SerializationException;
import org.apache.kafka.common.serialization.Serializer;


import java.io.ByteArrayOutputStream;
import java.io.IOException;
import java.util.Map;

public class AvroSerializer<T extends SpecificRecordBase> implements Serializer<T> {

    @Override
    public void close() {
        // No-op
    }

    @Override
    public void configure(Map<String, ?> arg0, boolean arg1) {
        // No-op
    }

    @Override
    public byte[] serialize(String topic, T data) {
        try {
            byte[] result = null;

            if (data != null) {
                // here would be the place where we could convert `data` from 
                // an sdk-cloudevent to an avro-cloudevent
                ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
                BinaryEncoder binaryEncoder =
                        EncoderFactory.get().binaryEncoder(byteArrayOutputStream, null);

                DatumWriter<GenericRecord> datumWriter = new GenericDatumWriter<>(data.getSchema());
                datumWriter.write(data, binaryEncoder);

                binaryEncoder.flush();
                byteArrayOutputStream.close();
                result = byteArrayOutputStream.toByteArray();
            }
            return result;
        } catch (IOException ex) {
            throw new SerializationException(
                    "Can't serialize data='" + data + "' for topic='" + topic + "'", ex);
        }
    }
}

```

duglin:
@BigJerBD @fabiojose as with #616, is this a breaking change that would require a v2.0 of the spec?

BigJerBD:
@duglin yes it is a breaking change because : 
- it removes/adds fields that does not have default
- it changes to class location of the Avro CloudEvent somewhere else

Actually #616 is an other way to fix the namespace issue mentioned above. doing one but not both of these modification would be enough

fabiojose:
Yes @duglin, this breaks the current version of Avro schema.

Because this introduce new fields and remove others, then, the resulting event on the wire is different.

Theses changes must be do with new major version: 2.0

duglin:
@BigJerBD are you saying that if we do #616 then we don't need to do this? If so, since #616 doesn't require a major-version bump, would that be an ok resolution?

If so, perhaps we could open an issue so that if/when we do a v2.0 then we can revisit this?

BigJerBD:
For me  both requires a major bump version, but maybe I misunderstand what
deserve to be majorly bumped

Both changes move/rename  the generated class thus using this version would
give class not found exception to anyone referencing the old class
name/location

I think it would be good to revisit that for v2.0. feel free to close it
and keep that in mind :).

duglin:
@BigJerBD it is our understanding that #616 would not change what appears on the wire, would you agree with that?  If so, we were hoping to either treat this more like an "SDK" issue, or just a bug that was missed.  What are your thoughts?

BigJerBD:
Sounds good for me, the other changes can also be revisited when version
2.0 will be considered.

Feel free to keep or close the pr :)

spec.avsc
```diff
@@ -1,22 +1,56 @@
...
-      "type":{
-        "type":"map",
-        "values":[
-          "null",
-          "boolean",
-          "int",
-          "string",
-          "bytes"
-        ]
-      }
```
subanasif, line 19: If you take out "attribute", where do you suggest one puts "extensions". As per my understanding, extensions,  mandatory and optional attributes all go under "attribute" as per 1.0 spec version.

BigJerBD: (sorry for late reply I though this PR was closed and forgotten) 

Indeed I totally forgot extensions when i wrote this a while ago. The Avro specs does not provide a way to do dynamic name when using complex types , A solution could be to actually keep attributes and rename it to "extensions". 

This would actually be clean in the avro's  generated source code  since it would consist to do something like `cloudEvent.getExtensions().get("extensionName")` to get an extension,

alizard0: Can't you add extensions as you have attributes and keep the fields?

PR #616, opened 2020-05-04
Closes #615

### LINKED ISSUE #615: Rename CloudEvent in avro spec to AvroCloudEvent
Some users of Avro Event Format may use some kind of tool for source code gen, then the resulting source may collide with packages and classes defined in the SDKs.

This happens because of name definition within `spec.avsc`:

```json
{
    "namespace":"io.cloudevents",
    "type":"record",
    "name":"CloudEvent"
}
```

The suggestion is to change the name to `AvroCloudEvent`, that I think is better for the context.

This is a result of experience using Avro with Java.

fabiojose:
Change CloudEvent to AvroCloudEvent

### DIFF
--- a/spec.avsc
+++ b/spec.avsc
@@ -1,7 +1,7 @@
 {
   "namespace":"io.cloudevents",
   "type":"record",
-  "name":"CloudEvent",
+  "name":"AvroCloudEvent",
   "version":"1.0",
   "doc":"Avro Event Format for CloudEvents",
   "fields":[
@@ -31,7 +31,7 @@
             "boolean",
             {
               "type": "record",
-              "name": "CloudEventData",
+              "name": "AvroCloudEventData",
               "doc": "Representation of a JSON Value",
               "fields": [
                 {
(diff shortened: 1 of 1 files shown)


duglin:
@fabiojose are these breaking changes? On the call this week there was concern that this would break existing users and that these changes are not simply changes to things like "docs" or "description" attribtues.

fabiojose:
@duglin yes! This is a break change.

duglin:
Should we make it then? Technically that would require a v2.0 of the Avro spec and is it really worth that just for a name change?

fabiojose:
@dug this PR does not change the attributes names, types or remove them.

This breaks the way someone uses with code generation tools, but at the end, the resulting event on the wire is the same. 

BTW, should be prudent creating a version 1.1 of Avro schema.

duglin:
@fabiojose To be clear, you're saying that this does NOT require a major-version bump in the Avro spec?

Will an existing Avro client, or server, continue to work if the other side of the communication used this new file?  Based on you saying it doesn't change what appears on the wire, I'm assuming the answer is 'yes'.

duglin:
Approved on the 5/14 call as a non-on-the-wire change and more akin to a bug that needs to be fixed.

PR #624, opened 2020-05-13
Changes to data model

duglin:
I don't think of this as a radical change to what's in there today, it's more of a naming changing for "producer". This is just a first step and a rough idea to get some discussions/brainstorming to happen.

### DIFF
--- a/discovery.md
+++ b/discovery.md
@@ -140,7 +140,7 @@ its basic conceptual structure:
       dataschema: [ce-dataschema URI] ?
       dataschematype: [string per RFC 2046] ?
       dataschemacontent: [schema] ?
-      dataschemacontent: 
+      dataschemacontent:
       extensions: *
       - name: [CE context attribute name]
         type: [CE type string]
@@ -162,30 +162,6 @@ An example:
 Note: the above is just a sample and implementations are free to use any
 internal model they wish to store the data as long as they are compliant
 with the wire format/API defined by this specification.
-
-And some sample queries:
-
-- `http://.../?service=example.com`
-  - Shows entire data model for `example.com`
-
-- `http://.../?type=com.example.widget.create`
-  - Shows all Services that have `com.example.widget.create` as one of
-    its `types`.
-  - Does this show entire data model for all Services that have this
-    `type` (meaning all of its types), or does it just show the type referenced
-    in the query? I'm leaning towards all.
-
-- `http://.../?protocol=HTTP&type=com.example.widget.create
-  - Same as previous except only shows Services that also supports the
-    `HTTP` subscription protocol.
-
-
-Questions:
-- does specifying the same attribute multiple times imply OR or AND? OR
-- does specifying diffrent attributes imply OR or AND?  AND
-
--- current end of propsal --
-
 
 ### Entities in the API
 
--- a/discovery.md
+++ b/discovery.md
@@ -3,8 +3,8 @@
 ## Abstract
 
 CloudSubscriptions Discovery API is a vendor-neutral API specification for
-determining what events an Event Producer or Aggregator / Broker
-has available, as well as how to subscribe to those events.
+determining what events an Event Producer has available, as well as how to
+subscribe to those events.
 
 ## Status of this document
 
@@ -22,10 +22,13 @@ version.
 ## Overview
 
 The goal of the CloudDiscovery API is to enable connections between consumers of
-events and Event providers / producers / aggregators / brokers of events and to
-facilitate the creation of CloudSubscriptions.
+events and Event producers to facilitate the creation of
+CloudSubscriptions. Note that whether an Event producer is a single entity,
+an aggregator/broker of multiple event producers, or an intermediary is an
+implementation detail of the producer and does not change how an event
+consumer interacts with a specification compliant producer.
 
-Discovery allows for an event producer or intermediary component to
+Discovery allows for an event producer to
 advertise the event types that are available, provide the necessary information
 to consume that event (schema / delivery protocol options), and the necessary
 information to create a subscription. The output of the API ought to be such
@@ -35,10 +38,10 @@ known in advance.
 There are several discovery use cases to consider from the viewpoint of event
 consumers.
 
-1. What providers are available, and what event types do they produce?
-2. What event types are available, and from which providers?
+1. What event sources are available, and what event types do they produce?
+2. What event types are available, and from which event sources?
 
-The second case becomes relevant if multiple providers support the same event
+The second case becomes relevant if multiple sources support the same event
 types. Use case 1 is likely the dominant use case. Given the example of a public
 cloud provider where all producers produce events, there might be dozens of
 sources (producer systems) and hundreds of event types. The discovery funnel of
@@ -47,9 +50,13 @@ lists of event types. Both of these cases show the importance of using filters
 in the discovery API to narrow down the selection of available events.
 
 The CloudEvent `source` attribute is a potential cause of high fanout. For
-example consider a blob storage system where each directory constitutes a
-distinct `source` attribute value. For this reason, `source` is not part of the
-discovery system.
+example, consider a blob storage system where each directory constitutes a
+distinct `source` attribute value. For this reason, the exact CloudEvents
+`source` attribute value that might appear in a CloudEvent will not appear in
+the Discovery API query result. Instead, a higher level classifier
+(`sourceclass`) will be used to represent the abstract notion of the generic
+event producer of those events - in the example case, the blob storage service
+itself.
 
 ## Notations and Terminology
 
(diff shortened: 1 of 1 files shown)


deissnerk:
LGTM

discovery.md
```diff
@@ -22,10 +22,13 @@ version.
 ## Overview
 
 The goal of the CloudDiscovery API is to enable connections between consumers of
-events and Event providers / producers / aggregators / brokers of events and to
-facilitate the creation of CloudSubscriptions.
+events and Event producers to facilitate the creation of
```
JemDay, line 25: + a

_".. facilitate the creation of a CloudEventSubscription. .."_

discovery.md
```diff
@@ -110,6 +111,76 @@ be considered an event subscriber and a target user of this API.
...
 between those entities.
 
+To help explain the data model, the following non-normative pseudo yaml shows
+its basic conceptual structure:
+
+( `*` means zero or more, `+` means one or more, `?` means optional)
+
+```
+- sources: *
+  - sourceclass: [unique identifier for producer of the specified event types]
```
mikehelmick, line 121: I really don't like 'sourceclass' - I think the word class makes it sound general/abstract

How about 'sourcesystem' ?

discovery.md
```diff
@@ -110,6 +111,76 @@ be considered an event subscriber and a target user of this API.
 This API is specified as a REST API with well defined entities and relationships
 between those entities.
 
+To help explain the data model, the following non-normative pseudo yaml shows
+its basic conceptual structure:
+
+( `*` means zero or more, `+` means one or more, `?` means optional)
+
+```
+- sources: *
```
mikehelmick, line 120: sourcesystems ?

discovery.md
deissnerk: Imho it's not a system. For many of us it will be something like a product or service. _System_ actually suggests that it is the same as the producer. To me it really is a lot like a class. I think of the source as an instance of it. A sourceclass could be repository.github.com, and a specific source would be of form github.com/{org}/{repo}. 
Perhaps we still don't have the same understanding of the semantic of this new concept. I see at least three different possible meanings:

1. Related to physical entity: system, producer, host
2. Describe common properties and meaning: class, category, type, prototype 
3. No semantics, just multiple sources bundled together: group, family

discovery.md
duglin: Done

discovery.md
duglin: what about "service"? 

I want to subscribe to the "object store service".
github, gitlab, ... are services, right? And each has a set of "sources" about which you can ask for events.
Maybe?

duglin:
@mikehelmick I look at the REST PATH stuff in the spec and I'm not sure it presents a consistent solution.  It seems like most people would expect a single data model with a single URL path. For example, with the data model in this PR, I think people would expect a URL path like:
```
https://example.com/services/blobservice/types/com.example.widget.create
```
even though I'm not sure people would specify that many layers.
But I think the point is that it might seem a bit odd to then offer two different URL paths:
```
- https://example.com/services/???
- https://example.com/types/???
```
because I think people would expect each to present a different data model - one with "service" as the parent, and one with "type" as the parent. While that's ok if you view the query in isolation because you're focused on asking about a "service" or a "type" as the main query parameter, but then some things come to mind:
1 - each returns a different schema for the results
2 - if we want to search on some other "key" do we need to then create a new URL and schema for each one?

I know you show the same schema in both cases, but that seems inconsistent to me because the URL implies "types" is the root entity, yet the response shows "service" as the root.

But, aside from that, it's not consistent if we allowed people to query over multiple properties - say "protocol" and "type". Wouldn't it be odd if one of those entities were in the path and one in the query parameter? e.g.
```
example.com/types?matching=foo&protocol=http
```
would be more natural as:
```
example.com?types=foo&protocol=http
```
then we're not treating "types" as anything special. It also allows us to specify the same property multiple time, if we wanted to allow that.

I think the fact the data model is pretty flat means that asking people to specify things as query parameters isn't much of a burden. IOW, if "type" were 5 levels deep or shows up in multiple locations in the schema then making a special URL query path (to help the UX) could make sense, but I don't think we need it for our shallow model.

mikehelmick:
@duglin 

This is why I believe GraphQL is a better choice for the discovery API.

There are multiple entry points into the discovery flow and a flexible query language lets the client select their entry points, filter requirements, and the subset of the fields they want to receive in the reply.

discovery.md
deissnerk: I could live with "service". I suppose `source` will be many different things. @clemensv often comes up with IoT examples. Sometimes you will have one source service/class/system per source. Sometimes it will be many. As we give this thing an id, I wonder what exactly it identifies. 
Btw. what about `sourcespec`? 😄

discovery.md
```diff
@@ -110,6 +117,76 @@ be considered an event subscriber and a target user of this API.
...
+
+```
+- services: *
+  - service: [unique identifier for producer of the specified event types]
+    description: [human string] ?
+    uri: [URI reference for human documentation] ?
+    specversion: [ce-speciversion value] ? (required?)
+    subscriptionuri: [URI to which the Subscribe request will be sent]
+    subscriptionconfig: *
+    authscope: [string] ?  (explain)
```
deissnerk, line 133: I would like to see something like `sourcepattern` in here. @mikehelmick had this in an earlier proposal.

discovery.md
```diff
@@ -110,6 +117,76 @@ be considered an event subscriber and a target user of this API.
...
+    protocols: +
+    - protocol: [string]
+    types: *
+    - type: [ce-type value]
```
deissnerk, line 137: I still wonder, if the de-normalized form is the right one here. I certainly understand the simplicity when it's supposed to support UIs. It's also nice and simple, if you look for a type emitted by a single service. In scenarios where an event type is emitted by multiple services, this will make the development of tools more complex. In case you want to provide tooling to generate some event handling or creation code, you would then have to assume that an event type is always defined the same way, or do complex structure comparisons.

discovery.md
```diff
@@ -110,6 +117,76 @@ be considered an event subscriber and a target user of this API.
...
 between those entities.
 
+To help explain the data model, the following non-normative pseudo yaml shows
+its basic conceptual structure:
+
+( `*` means zero or more, `+` means one or more, `?` means optional)
+
+```
+- services: *
+  - service: [unique identifier for producer of the specified event types]
```
deissnerk, line 127: If I recall correctly, one of the proposals before we agreed on source was something like "service"+"resource". 😄

discovery.md
duglin: I assume if we added that we'd then need to find a way to define all of the various substitution values, right? 
Do you have a link to his old proposal?

discovery.md
duglin: can you provide an example of what you mean by "de-normalized" in this context?

deissnerk: In the proposal the "type" is a child of the "service". In reality the CloudEvents model defines "types" as an independent entity. You can have multiple "sources" or "services" using the type. If you define type as a child of service, it can be defined differently for every service. 
If you allow the definition of types independently from sources/services, it is a tricky question which context attributes are described in type and which are defined as part of the source/service. I could imagine a model where a service references a type and makes definitions for some attributes more specific. E.g. for a type it could be defined that it works with dataschema 1 or 2, and a service referring to this type could then restrict it to dataschema 1. I know this sounds really complex. I guess we need to agree on what kind of scenarios we want to support.

discovery.md
deissnerk: What I have in mind is similar to what AsyncAPI offers for channels:
[https://www.asyncapi.com/docs/specifications/2.0.0/#a-name-parametersobject-a-parameters-object](https://www.asyncapi.com/docs/specifications/2.0.0/#a-name-parametersobject-a-parameters-object
)

In the original CloudSubscriptions Google doc, you can find:
> “sourceStructure”
Documentation on the format of the source, for example
“github.com/[organization]/[repository]/pull/[id]

duglin: @deissnerk want to open an issue so we can track your idea?

discovery.md
duglin: perhaps open an issue so we can discuss it?

It kind of sounds like you're talking about enabling something like a type-ref thing, where the type is defined outside of a service but then referenced from within a service.  It might be useful to reduce the size of the response payload, but I kind of wonder if a user will need to know about that relationship.

duglin:
ok I removed the query stuff per the call today.
Please look it over to make sure I didn't miss anything and I'll hold off merging until EOD Friday.

Approved conditionally (w/query stuff) on the 5/21 call.

duglin:
Query stuff was removed so merging per the 5/21 call agreement to wait until EOD Friday.

Issue #610, opened 2020-04-29
Proposal to form a sub-working group for a schema registry API


clemensv:
Microsoft proposes the formation of a sub-working group to starts working on a common schema registry API interface in parallel with the ongoing efforts.

The schema registry API should, in our view, fulfill the following minimal requirements:
- Provide a management and retrieval interface over HTTP (AMQP services can leverage the [HTTP-over-AMQP](https://www.oasis-open.org/committees/document.php?document_id=65571&wg_abbrev=amqp) binding that is pending in the AMQP TC)
-  Allow for storing schema file streams for any data serialization format for both schema-driven serialization and data validation, e.g. Avro or Protobuf or XML or JSON schemas. 
- Allow for storing for multiple versions for each schema.
- Allow for storing of logical groups of such schemas. 

Our concrete initial proposal is to create a very simple REST API for this purpose, that manages a resource graph like this:

https://example.com/**schemagroups**/{schemagroup}/**schemas**/{schema}/**versions**/{version}

- "schemagroups" collection: the top level collection of schema groups. Each group is a logically related set of schemas, typically belonging to a particular application or subject to shared access control rules.
- "schemas" collection: the collection of schemas belonging to a schema group. 
- "versions": the versions of schema documents belonging to a particular schema.

The version identifiers, eg. `.../versions/4` are assigned and managed by the service and do not necessarily reflect the developer-assigned version number. A service implementation MAY enforce certain dependencies between versions. For instance, a service implementation that specifically understand Avro schemas might be configured such that a new schema version can only be added if it satisfies certain backwards compatibility criteria. If the added schema does not satisfy those criteria, the service might then report a conflict (409).  

In the "versions" collection, clients can add new schemas with POST against the collection root. The response is either 201 (Created) with a `Location` header that indicates the newly created schema version location, 400 (Bad Request) for a malformed request, 409 (Conflict) for a consistency violation as just discussed, or some other 5xx or 4xx response appropriate for the error condition.

Schema version can be deleted but not replaced, meaning there is no update.

The "versions" collection allows enumerating all versions of a schema with GET. The result is a JSON array holding URI-references to to the schema version documents. Schema version documents can be deleted with DELET on their schema version URI.

In the "schemas" collection clients can add new schemas by name using the PUT method with the schema name in the URI segment. This creates a new version collection for the schema if required and then appends the schema as the latest version. A GET against ".../schemas/{schema}" will return the latest version of the schema document stored in the schema collection. 

[...]

PR #625, opened 2020-05-20
Schema Registry Proposal

clemensv:

For initial review. I'm still updating both documents including changing some names, but the combination of OpenAPI doc and the spec doc should already tell a fairly complete story. 

Microsoft proposal for #610

### DIFF
--- a/schemaregistry/schemaregistry.md
+++ b/schemaregistry/schemaregistry.md
@@ -44,7 +44,10 @@ This schema registry deals with only three top-level elements:
 
 ## 2. Schema Registry Elements
 
-This section further describes the elements enumerated in the introduction.
+This section further describes the elements enumerated in the introduction. All
+data types used in this section are defined in the core CloudEvents
+specification and MUST follow the respective formatting and syntax rules unless
+specified otherwise.
 
 ### 2.1. Schema Group
 
@@ -73,7 +76,7 @@ rules.
 
 ### 2.1.1. Schema group attributes
 
-The data model for a schema group consists of three attributes:
+The data model for a schema group consists of these attributes:
 
 #### id
 
@@ -128,7 +131,7 @@ The data model for a schema group consists of three attributes:
 
 #### updatedtimeutc
 
-- Type: `Timestamp`
+- Type: `Timestamp` 
 - Description: Instant when the schema group was last updated
 - Constraints:
   - OPTIONAL
@@ -174,7 +177,7 @@ of schema documents. An implementation MAY add further attributes.
   - REQUIRED
   - MUST be a non-empty string
   - MUST conform with RFC3986/3.3 `segment-nz-nc` syntax
-  - MUST be unique within the scope of the schema registry
+  - MUST be unique within the scope of the schema group
 - Examples:
   - myschema
   - my-schema
@@ -204,7 +207,7 @@ of schema documents. An implementation MAY add further attributes.
 - Examples:
   - "This schema describes the toppings for a Pizza."
 
-  #### createdtimeutc
+#### createdtimeutc
 
 - Type: `Timestamp`
 - Description: Instant when the schema was added to the registry.
@@ -227,10 +230,8 @@ document or binary stream. An implementation SHOULD validate whether a
 schema version is valid according to the rules of its format, for instance
 whether it is a valid Avro schema document when the format is Apache Avro.
 
-Within the scope of the schema set, the version is identified by the combination
-of a version number and an optional format identifier. The schema version MAY
-also have an additional, optional unique identifier within the scope of the
-registry.
+The schema version MAY also have an additional, optional unique identifier
+within the scope of the registry.
 
 #### version
 
@@ -279,6 +280,8 @@ This section informally describes the HTTP API binding of this schema registry.
 The formal definition is the [OpenAPI document](./schemaregistry.yaml) that is
 part of this specification.
 
+This section is therefore non-normative.
+
 ### 3.1. Path hierarchy
 
 Schema groups contain schemas and those contain schema versions, which are the
@@ -286,7 +289,7 @@ documents required for serialization or validation.
 
 These dependencies are reflected in the path structure:
 
-`[/schemagroups]/{group-id}/schemas/{schema-id}/versions/{version}`
+`/schemagroups/{group-id}/schemas/{schema-id}/versions/{version}`
 
 The name of the first segment of the path ("/schemagroups") is an illustrative
 suggestion and MAY differ between implementations and the registry does not need
@@ -300,19 +303,105 @@ be used.
 ### 3.2. Operations at the groups level
 
 #### 3.2.1. List schema groups
+
+Schema groups are enumerated with a GET on the root of the registry hierarchy,
+in the exemplary structure above identified as `/schemagroups`.
+
+The result is a JSON encoded array of strings enumerating the `id` values of the
+schema groups.
+
 #### 3.2.2. Get schema group
+
+The details of a schema group are retrieved with a GET on the group's path in
+the schema groups collection, for instance `/schemagroups/mygroup`. 
+
+The result is a JSON object that contains the attributes of the schema group.
+
 #### 3.2.3. Create schema group
+
+A schema group is created with a PUT on the desired group's path in
+the schema groups collection, for instance `/schemagroups/mygroup`.
+
+The payload is a JSON object that contains the attributes of the schema group.
+
+The operation returns the effective attributes as a JSON object when the schwema
+group has been created or updated.
+
 #### 3.2.4. Delete schema group
+
+A schema group is deleted with a DELETE on the desired group's path in
+the schema groups collection, for instance `/schemagroups/mygroup`.
 
 ### 3.3 Operations at the schemas level 
 
 #### 3.3.1. List schemas for the group
+
+Schemas within a schema group are enumerated with a GET on the `schemas`
+collection of the group, for instance `/schemagroups/mygroup/schemas`.
+
+The result is a JSON encoded array of strings enumerating the `id` values of the
+schemas within the group.
+
 #### 3.3.2. Delete all schemas from the group
+
+All schemas of a schema group can be deleted DELETE on the `schemas`
+collection of the group, for instance `/schemagroups/mygroup/schemas`.
+
 #### 3.3.3. Add a new schema version
+
+A new schema or a new version of a schema is added to the version collection
+with a POST on the desired schema's path in the schemas collection, for instance
+`/schemagroups/mygroup/schemas/myschema`.
+
+This operation will either create a new schema and store the document under the
+first version identifier assigned by the server or will update the schema by
+assigning a new version to the given document and storing it.
+
+The payload of the request is the schema document. All further attributes such
+as the `description` or the `format` indicator are passed either via the query
+string or as HTTP headers.  
+
+The ´Content-Type´ for the payload MUST be preserved by the registry and
+returned when the schema is requested, independent of the format identifier. 
+
 #### 3.3.4. Get the latest version of a schema
+
+The latest version of a schema is retrieved via a GET on schema's path in the
+schemas collection, for instance `/schemagroups/mygroup/schemas/myschema`. 
+
+The returned payload is the schema document. Further attributes such as the
+`description` and the `format` indicator are returned as HTTP headers.
+
+The returned ´Content-Type´ is the same that was passed when the schema version
+was registered.
+
+The HEAD method SHOULD also be implemented.
+
 #### 3.3.4. Delete a schema
+
+A schema including all its versions is deleted with a DELETE on schema's path in the
+schemas collection, for instance `/schemagroups/mygroup/schemas/myschema`
 
 ### 3.4 Operations at the versions level
 
 #### 3.4.1. List all versions of the schema
+
+Versions of a schema within a schema group are enumerated with a GET on the
+`versions` collection of the schema, for instance
+`/schemagroups/mygroup/schemas/myschemas/versions`.
+
+The result is a JSON encoded array of integers enumerating the `version` values of the
+schemas within the group.
+
 #### 3.4.2. Get a specific schema version
+
+A specific version of a schema is retrieved via a GET on schema version's path in the
+`versions` collection, for instance `/schemagroups/mygroup/schemas/myschema`.
+
+The returned payload is the schema document. Further attributes such as the
+`description` and the `format` indicator are returned as HTTP headers.
+
+The returned ´Content-Type´ is the same that was passed when the schema version
+was registered.
+
+The HEAD method SHOULD also be implemented.
--- a/schemaregistry/schemaregistry.md
+++ b/schemaregistry/schemaregistry.md
@@ -1,0 +1,257 @@
+# CNCF Schema Registry API Version 0.1-rc01s
+
+## Abstract
+
+This specification defines a simple API for storing, organizing, and accessing
+schema documents for use with serialization and data validation frameworks such
+as Apache Avro and JSON Schema.
+
+## Status of this document
+
+This document is a working draft.
+
+## 1. Introduction
+
+Schema-dependent serialization formats and related frameworks such as Apache
+Avro are popular in eventing and messaging scenarios. With most of the metadata
+information such as names and types of fields being held externally in schema
+documents, these frameworks are able to produce very compact on-wire
+representations of structured data that is carries as the payload of events and
+messages. While the wire-footprint savings are often very significant, the added
+cost is that of schema handling. All communicating parties who need to
+deserialize the event or message payload need access to the schema used to
+publish it.
+
+For formats that do not depend on external schemas for serialization, schemas
+might still be required in some scenarios to allow for consumers or inspecting
+intermediaries to validate the data structure's compliance with a set of rules.
+
+This specification defines a very simple API focused on storing, organizing, and
+accessing such documents. While there are very sophisticated metadata management
+solutions available as open source und under neutral governance already, the focus
+of this specification is not great sophistication but straightforward simplicity.
+
+This schema registry deals with only three top-level elements:
+
+- Schema Group: A named collection of schemas. Each group holds a logically
+  related set of schemas, typically belonging to a particular application or
+  subject to shared access control rules.
+- Schema: A document describing the structure, names, and types of some
+  structured data payload.
+- Version: A specific version of a schema document. Even though not prescribed
+  in this specification, an implementation might choose to impose compatibility
+  constraints on versions following the initial version of a schema.
+
+> NOTE: There are two fundamentally different options for how we can organize
+> multiple schema formats and this specification draft reflects both: 
+>
+> 1) In an "ideal" REST service, any data entity can be represented ("R" in
+>    REST) in multiple different data formats, which a client can request in
+>    prioritized order during content-type negotiation. If a service offers up
+>    the same entity in JSON, Protobuf, and Avro formats, there would therefore
+>    be three distinct schemas, each defining a representation of the data
+>    entity for the respective format. From the same "REST" mindset perspective,
+>    the ideal representation of this set of schemas, which differ with regards
+>    to the format they describe, but ought to be semantically identical, is to
+>    also differentiate them only by their content type and maintain them under
+>    the exact same name as if they were one schema.
+> 2) Since the above strategy is truly RESTful, but quite esoteric if you've not
+>    grown up as a RESTafarian, the alternative strategy for concurrently
+>    handling multiple schema formats is much simpler: Constrain each schema
+>    group to a single format.
+
+## 2. Schema Registry Elements
+
+This section further describes the elements enumerated in the introduction.
+
+### 2.1. Schema Group
+
+A schema group is a named collection of schemas. Since a schema registry is
+often a resource with a scope greater than a single application and might even
+span multiple organizations, it is very useful to put a grouping construct
+around sets of schemas that are related either by ownership or by a shared
+subject matter context.
+
+Implementations of this specification MAY associate access control rules with
+schema groups. For instance, a user or a group of users might be given write
+access only to a particular schema group that their organization owns. If trade
+secret protection is required for an application or parts of it and the schema
+structure would give some of those away, read access to a group of schemas might
+likewise be restricted. 
+
+Access control rules at the schema group level MUST NOT limit visibility of the
+schema groups themselves to users authorized to enumerate schema groups at the
+registry level or to create new schema groups in order to create unambiguous
+transparency about which `id` have already been taken. An implementation MAY
+withhold all details other than the `id` attribute from a user not authorized
+to access the schema group.
+
+This specification does not define management constructs for such access control
+rules.
+
+### 2.1.1. Schema group attributes
+
+The data model for a schema group consists of three attributes:
+
+#### id
+
+- Type: `String`
+- Description: Identifies the schema group.
+- Constraints:
+  - REQUIRED
+  - MUST be a non-empty string
+  - MUST conform with RFC3986/3.3 `segment-nz-nc` syntax
+  - MUST be unique within the scope of the schema registry
+- Examples:
+  - mygroup
+  - my-group
+
+#### format
+
+- Type: `String`
+- Description: Defines the schema format managed by this schema group. The
+  formats supported by an implementation are specific to the implementation.
+  This specification does not mandate support for particular formats.
+- Constraints:
+  - OPTIONAL. If not set, each schema in the group MAY have multiple, concurrent
+    format representations. See XXXX
+  - MUST be a non-empty string, if present.
+- Examples:
+  - "avro"
+  - "json-schema"
+  - "xsd11"
+
+#### description
+
+- Type: `String`
+- Description: Explains the purpose of the schema group.
+- Constraints:
+  - OPTIONAL
+- Examples:
+  - "This group holds schemas for the fabulous example app."
+
+## 2.2. Schema and Schema Version
+
+Conceptually, a schema is a description of a data structure. Since data
+structures evolve over time, the schema describing them will also evolve over
+time. Therefore, a schema often has multiple versions.
+
+For simple scenarios, the API allows for version management to be automatic and
+transparent. Whenever a schema is updated, a new version number is assigned and
+prior schema versions are retained. The latest available schema is always the
+default version that is retrieved when the API is given just the schema `id`.
+
+A newer schema version might introduce breaking changes or it might only
+introduce careful changes that preserve compatibility. These strategies are not
+subject of this specification, but the API provides a conflict handling
+mechanism that allows an implementation to reject updates that do not comply
+with a compatibility policy, if one has been implemented.
+
+### 2.2.1. Multi-format schema sets
+
+If the schema format has not been restricted at the schema group level, each
+schema may describe the same data structure for encoding in different formats
+concurrently. That implies that the schema managed by this registry is not
+a single document, but MAY be a set of documents that all describe the exact
+same data structure, each in their own particular syntax and using their own 
+type model.
+
+The schema version therefore relates to this abstract notion of schema and the
+described data structure. All documents coexisting within the same version
+SHOULD describe the exact same data structure.
+
+### 2.2.2. Schema attributes
+
+As per the definition above, the schema object is a collection of versions
+of schema documents. It only has an identifier attribute.
+
+#### id
+
+- Type: `String`
+- Description: Identifies the schema.
+- Constraints:
+  - REQUIRED
+  - MUST be a non-empty string
+  - MUST conform with RFC3986/3.3 `segment-nz-nc` syntax
+  - MUST be unique within the scope of the schema registry
+- Examples:
+  - myschema
+  - my-schema
+
+### 2.2.3 Schema version attributes
+
+A schema version is a document. The "body" of a schema version MAY be a text
+document or binary stream. An implementation SHOULD validate whether a
+schema version is valid according to the rules of its format, for instance
+whether it is a valid Avro schema document when the format is Apache Avro.
+
+Within the scope of the schema set, the version is identified by the combination
+of a version number and an optional format identifier. The schema version MAY
+also have an additional, optional unique identifier within the scope of the
+registry.
+
+#### version
+
+- Type: `Integer`
+- Description: The version of the schema. This is a simple counter and tracks
+  the version in the scope of this schema within the schema group. The schema
+  document MAY indicate a schema that follows a different versioning scheme.
+- Constraints:
+  - REQUIRED
+  - Assigned by server.
+- Examples:
+  - 1
+  - 2
+
+#### format
+
+- Type: `String`
+- Description: 
+- Constraints:
+  - OPTIONAL. Can be used if and only if not format has been set for the schema
+    group hosting the schema.
+  - MUST be a non-empty string, if present.
+- Examples:
+  - "avro"
+  - "json-schema"
+  - "xsd11"
+
+#### id
+
+- Type: `String`
+- Description: Identifies the schema document uniquely without requiring
+  other qualifiers.
+- Constraints:
+  - OPTIONAL
+  - Assigned by the server.
+- Examples:
+  - { ... guid ... }  
+
+## 3. HTTP ("REST") API 
+
+This section informally describes the HTTP API binding of this schema registry.
+The formal definition is the [OpenAPI document](./schemaregistry.yaml) that is
+part of this specification.
+
+### 3.1. Path hierarchy
+
+Schema groups contain schemas and those contain schema versions, which are the
+documents required for serialization or validation.
+
+These dependencies are reflected in the path structure:
+
+`/schemagroups/{group-id}/schemas/{schema-id}/versions/{version}`
+
+The name of the first segment of the path is a suggestion and MAY differ between
+implementations and the registry does not need to be anchored at the site root.
+The segment names `schemas` and `versions` MUST be used.
+
+- `{group-id}` corresponds to the schema group's [`id` attribute](#id)
+- `{schema-id}` corresponds to the schema's [`id` attribute](#id-1)
+- `{version}` corresponds to the schema version's [`version` attribute](#version)
+
+### 3.2. Operations at the groups level
+
+### 3.3 Operations at the schemas level 
+
+### 3.4 Operations at the versions level
--- a/schemaregistry/schemaregistry.md
+++ b/schemaregistry/schemaregistry.md
@@ -1,4 +1,4 @@
-# CNCF Schema Registry API Version 0.1-rc01s
+# CNCF Schema Registry API Version 0.1-wip
 
 ## Abstract
 
@@ -16,7 +16,7 @@ Schema-dependent serialization formats and related frameworks such as Apache
 Avro are popular in eventing and messaging scenarios. With most of the metadata
 information such as names and types of fields being held externally in schema
 documents, these frameworks are able to produce very compact on-wire
-representations of structured data that is carries as the payload of events and
+representations of structured data that is carried as the payload of events and
 messages. While the wire-footprint savings are often very significant, the added
 cost is that of schema handling. All communicating parties who need to
 deserialize the event or message payload need access to the schema used to
@@ -28,37 +28,19 @@ intermediaries to validate the data structure's compliance with a set of rules.
 
 This specification defines a very simple API focused on storing, organizing, and
 accessing such documents. While there are very sophisticated metadata management
-solutions available as open source und under neutral governance already, the focus
+solutions available as open source and under neutral governance already, the focus
 of this specification is not great sophistication but straightforward simplicity.
 
 This schema registry deals with only three top-level elements:
 
 - Schema Group: A named collection of schemas. Each group holds a logically
-  related set of schemas, typically belonging to a particular application or
-  subject to shared access control rules.
+  related set of schemas, typically managed by a single entity, belonging to a
+  particular application and/or having a shared access control management scope.
 - Schema: A document describing the structure, names, and types of some
   structured data payload.
 - Version: A specific version of a schema document. Even though not prescribed
   in this specification, an implementation might choose to impose compatibility
   constraints on versions following the initial version of a schema.
-
-> NOTE: There are two fundamentally different options for how we can organize
-> multiple schema formats and this specification draft reflects both: 
->
-> 1) In an "ideal" REST service, any data entity can be represented ("R" in
->    REST) in multiple different data formats, which a client can request in
->    prioritized order during content-type negotiation. If a service offers up
->    the same entity in JSON, Protobuf, and Avro formats, there would therefore
->    be three distinct schemas, each defining a representation of the data
->    entity for the respective format. From the same "REST" mindset perspective,
->    the ideal representation of this set of schemas, which differ with regards
->    to the format they describe, but ought to be semantically identical, is to
->    also differentiate them only by their content type and maintain them under
->    the exact same name as if they were one schema.
-> 2) Since the above strategy is truly RESTful, but quite esoteric if you've not
->    grown up as a RESTafarian, the alternative strategy for concurrently
->    handling multiple schema formats is much simpler: Constrain each schema
->    group to a single format.
 
 ## 2. Schema Registry Elements
 
@@ -100,22 +82,28 @@ The data model for a schema group consists of three attributes:
 - Constraints:
   - REQUIRED
   - MUST be a non-empty string
-  - MUST conform with RFC3986/3.3 `segment-nz-nc` syntax
+  - MUST conform with RFC3986/3.3 `segment-nz-nc` syntax. This allows for "dot
+    notation", e.g. `org.example.myapp.module` for logical organization of
+    schema groups, even though the namespace is flat.
   - MUST be unique within the scope of the schema registry
 - Examples:
   - mygroup
   - my-group
+  - org.example.myapp.module
+  - events@example.com
 
 #### format
 
 - Type: `String`
 - Description: Defines the schema format managed by this schema group. The
-  formats supported by an implementation are specific to the implementation.
-  This specification does not mandate support for particular formats.
+  formats supported by an implementation are specific to the implementation. If
+  the format is omitted for the schema group, the format MUST be specified at
+  the schema-level. This specification does not mandate support for particular
+  formats.
 - Constraints:
-  - OPTIONAL. If not set, each schema in the group MAY have multiple, concurrent
-    format representations. See XXXX
+  - OPTIONAL.
   - MUST be a non-empty string, if present.
+  - MUST NOT be modified if at least one schema exists in the group.
 - Examples:
   - "avro"
   - "json-schema"
@@ -130,6 +118,22 @@ The data model for a schema group consists of three attributes:
 - Examples:
   - "This group holds schemas for the fabulous example app."
 
+#### createdtimeutc
+
+- Type: `Timestamp`
+- Description: Instant when the schema group was added to the registry.
+- Constraints:
+  - OPTIONAL
+  - Assigned by the server.
+
+#### updatedtimeutc
+
+- Type: `Timestamp`
+- Description: Instant when the schema group was last updated
+- Constraints:
+  - OPTIONAL
+  - Assigned by the server.
+
 ## 2.2. Schema and Schema Version
 
 Conceptually, a schema is a description of a data structure. Since data
@@ -138,8 +142,12 @@ time. Therefore, a schema often has multiple versions.
 
 For simple scenarios, the API allows for version management to be automatic and
 transparent. Whenever a schema is updated, a new version number is assigned and
-prior schema versions are retained. The latest available schema is always the
-default version that is retrieved when the API is given just the schema `id`.
+prior schema versions are retained. This specification does not mandate a
+retention policy, but implementations MAY retire and remove outdated schema
+versions.
+
+The latest available schema is always the default version that is retrieved when
+the URL to the schema is given without the version specified.
 
 A newer schema version might introduce breaking changes or it might only
 introduce careful changes that preserve compatibility. These strategies are not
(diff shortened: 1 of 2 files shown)


schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+>    REST) in multiple different data formats, which a client can request in
+>    prioritized order during content-type negotiation. If a service offers up
+>    the same entity in JSON, Protobuf, and Avro formats, there would therefore
+>    be three distinct schemas, each defining a representation of the data
+>    entity for the respective format. From the same "REST" mindset perspective,
+>    the ideal representation of this set of schemas, which differ with regards
+>    to the format they describe, but ought to be semantically identical, is to
+>    also differentiate them only by their content type and maintain them under
+>    the exact same name as if they were one schema.
+> 2) Since the above strategy is truly RESTful, but quite esoteric if you've not
```
ryanhorn, line 58: Can we have both? If a client does not specify a content type in the request, the implementation could choose a default. Or implementations could only support just one schema format.

That being said, I'm curious to hear if there's a real world example where there is value in offering multiple schema formats for the same logical schema. I can imagine in some cases that a more compact representation might be desirable (such as Avro's binary encoding over JSON), especially if bandwidth is a concern. But it does introduce some complexity, some of which will emerge in the implementations.

For example, if an implementation supports multiple formats for the same logical schema, how are they created atomically across the different representations? If the creation is not atomic, there's a chance the following scenario could technically happen with a schema registry that supports Avro and JSONSchema representations:

1. Consumer subscribes to event with Schema A
2. Schema registry registers JSONSchema version 2 of Schema A
3. Consumer receives event and requests Avro schema version 2
4. Consumer receives a 404
5. Schema registry registers Avro version 2 of Schema A

Looking at the proposed openapi spec, schema creation isn't specific about how this works, so I suppose an implementation could require sending all representations in the same request. But I wanted to call it out as complexity to consider.

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+  - mygroup
+  - my-group
+
+#### format
+
+- Type: `String`
+- Description: Defines the schema format managed by this schema group. The
+  formats supported by an implementation are specific to the implementation.
+  This specification does not mandate support for particular formats.
+- Constraints:
```
ryanhorn, line 115: Do we want to prescribe that this SHOULD be the media type for the schema as defined by the schema spec? For example, JSONSchema proposes "application/json+schema": https://json-schema.org/draft/2019-09/json-schema-core.html

Additionally, the schema format specs themselves are versioned, I wonder if this is something we need to account for here or whether we should leave that up to implementations.

schemaregistry/schemaregistry.yaml
```diff
@@ -0,0 +1,313 @@
...
+          description: OK
+          content:
+            application/json;format=avro:
+              schema:
+                type: array
+                items: 
+                  type: integer
+      tags:
+        - 'schemas'
+  /schemagroups/{group-name}/schemas/{schema-name}/versions/{version-number}:
```
ryanhorn, line 198: There might be a case for metadata to live at the version level of a schema. For example, when the schema was created, who created it, etc. I wonder if we need separate endpoints for the metadata vs fetching the raw schema itself.

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+of a version number and an optional format identifier. The schema version MAY
+also have an additional, optional unique identifier within the scope of the
+registry.
+
+#### version
+
+- Type: `Integer`
+- Description: The version of the schema. This is a simple counter and tracks
+  the version in the scope of this schema within the schema group. The schema
+  document MAY indicate a schema that follows a different versioning scheme.
```
ryanhorn, line 198: How do we envision this working exactly? This wording feels like we're opening it up for implementations to diverge from the spec without being specific about how they should do that.

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+
+This document is a working draft.
+
+## 1. Introduction
+
+Schema-dependent serialization formats and related frameworks such as Apache
+Avro are popular in eventing and messaging scenarios. With most of the metadata
+information such as names and types of fields being held externally in schema
+documents, these frameworks are able to produce very compact on-wire
+representations of structured data that is carries as the payload of events and
```
duglin, line 19: s/carries/carried/

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+deserialize the event or message payload need access to the schema used to
+publish it.
+
+For formats that do not depend on external schemas for serialization, schemas
+might still be required in some scenarios to allow for consumers or inspecting
+intermediaries to validate the data structure's compliance with a set of rules.
+
+This specification defines a very simple API focused on storing, organizing, and
+accessing such documents. While there are very sophisticated metadata management
+solutions available as open source und under neutral governance already, the focus
```
duglin, line 31: s/und/and/

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+of this specification is not great sophistication but straightforward simplicity.
+
+This schema registry deals with only three top-level elements:
+
+- Schema Group: A named collection of schemas. Each group holds a logically
+  related set of schemas, typically belonging to a particular application or
+  subject to shared access control rules.
```
duglin, line 38: "to shared access control rules"... having trouble parsing this. Perhaps just stop the sentence after "subject" ?

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+
+### 2.1. Schema Group
+
+A schema group is a named collection of schemas. Since a schema registry is
+often a resource with a scope greater than a single application and might even
+span multiple organizations, it is very useful to put a grouping construct
+around sets of schemas that are related either by ownership or by a shared
+subject matter context.
+
+Implementations of this specification MAY associate access control rules with
```
duglin, line 75: Do you see the spec talking about access control at levels below Schema Group too? If so, I wonder if this MAY/sentence should be in a more global spot to talk about access control at any layer in the hierarchy

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+secret protection is required for an application or parts of it and the schema
+structure would give some of those away, read access to a group of schemas might
+likewise be restricted. 
+
+Access control rules at the schema group level MUST NOT limit visibility of the
+schema groups themselves to users authorized to enumerate schema groups at the
+registry level or to create new schema groups in order to create unambiguous
+transparency about which `id` have already been taken. An implementation MAY
+withhold all details other than the `id` attribute from a user not authorized
+to access the schema group.
```
duglin, line 87: Section 3.1 says that this `id` will be used as part of the URL to access a particular schema. Does the MUST NOT in this paragraph imply some kind of security leak by requiring everyone who can at least see schema groups in general to also be able to see the IDs of schema groups they don't have access to?

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+## 2.2. Schema and Schema Version
+
+Conceptually, a schema is a description of a data structure. Since data
+structures evolve over time, the schema describing them will also evolve over
+time. Therefore, a schema often has multiple versions.
+
+For simple scenarios, the API allows for version management to be automatic and
+transparent. Whenever a schema is updated, a new version number is assigned and
+prior schema versions are retained. The latest available schema is always the
+default version that is retrieved when the API is given just the schema `id`.
```
duglin, line 142: s/when the API..../when the URL to the schema is given without the version specified/

Not a biggie, but I didn't want people to think that the schema ID could be used w/o the group ID too.

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+
+## 2.2. Schema and Schema Version
+
+Conceptually, a schema is a description of a data structure. Since data
+structures evolve over time, the schema describing them will also evolve over
+time. Therefore, a schema often has multiple versions.
+
+For simple scenarios, the API allows for version management to be automatic and
+transparent. Whenever a schema is updated, a new version number is assigned and
+prior schema versions are retained. The latest available schema is always the
```
duglin, line 141: seems there should be a MUST in here about "no version == latet"

duglin, line 141: Do we need to say something about the retention policy? Even just to say "the spec does not mandate one" ?

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+A newer schema version might introduce breaking changes or it might only
+introduce careful changes that preserve compatibility. These strategies are not
+subject of this specification, but the API provides a conflict handling
+mechanism that allows an implementation to reject updates that do not comply
+with a compatibility policy, if one has been implemented.
+
+### 2.2.1. Multi-format schema sets
+
+If the schema format has not been restricted at the schema group level, each
+schema may describe the same data structure for encoding in different formats
```
duglin, line 153: s/may/MAY/

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+concurrently. That implies that the schema managed by this registry is not
+a single document, but MAY be a set of documents that all describe the exact
+same data structure, each in their own particular syntax and using their own 
+type model.
```
duglin, line 157: I understand "same data structure, each in their own particular syntax". But the "using their own type model" confused me - I would think that model would be the same. Or are you implying that a particular field in one format might be an int but in another format it might be a string due to the restrictions of that particular format?

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+
+#### version
+
+- Type: `Integer`
+- Description: The version of the schema. This is a simple counter and tracks
+  the version in the scope of this schema within the schema group. The schema
+  document MAY indicate a schema that follows a different versioning scheme.
+- Constraints:
+  - REQUIRED
+  - Assigned by server.
```
duglin, line 201: I think I'm on the same track as Ryan, I'm wondering if we need to allow for the author to decide the version string so they can choose a simple int or a semvar pattern.

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+part of this specification.
+
+### 3.1. Path hierarchy
+
+Schema groups contain schemas and those contain schema versions, which are the
+documents required for serialization or validation.
+
+These dependencies are reflected in the path structure:
+
+`/schemagroups/{group-id}/schemas/{schema-id}/versions/{version}`
```
duglin, line 243: on last week's call there was a bit of a discussion around whether each group in the path is singular vs plural (e.g. `schemas` vs `schema`).  Should be a good bike-shedding discussion.

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+
+The name of the first segment of the path is a suggestion and MAY differ between
```
duglin, line 245: How do you see the query for all group getting interop w/o agreement on the first one?

schemaregistry/schemaregistry.md
clemensv: If we differentiate by media type, we would need to backfill definitions that are are not provided by the respective standards; best as vendor extensions such that we don't get in the way of them changing their minds, e.g. "application/vnd.cloudnative.schema-registry.avro"

schemaregistry/schemaregistry.yaml
clemensv: I think those extra metadata items can be exposed through metadata annotations or through an OPTIONS call if they are not part of the schema itself.

schemaregistry/schemaregistry.md
clemensv: This means that the schema document itself can use semver or some other "embedded" versioning notion while the API goes strictly by order or when changes have been added.

schemaregistry/schemaregistry.md
clemensv: We can surely open that up for also having access control levers at the other levels.

schemaregistry/schemaregistry.md
clemensv: "schemagroups" is really part of the path to the registry itself; everything defined here sits under that path. the segment could even be empty or could be multiple segments. I don't see there being interop issues, because you would always have it being at the root of the URI.

schemaregistry/schemaregistry.md
clemensv: GET /schemagroups gives you a list. 
GET /schemagroups/{group-id} gives you an object.

plural/singular/plural/singular/plural/singular

schemaregistry/schemaregistry.md
clemensv: The goal of the server-assigned integer is for versioning not to complicate the API model. With automatic numbers, you can make all updates a plain POST on the schema URI and you can enforce the compatibility rules you want using service-side logic as the update happens.  

Introducing breaking changes should really require the pain of a wholly new schema with its own backcompat versioning sequence.

Semver 1.x, 2.x, 3.x is really better captured by 

/schemagroups/myapp/schemas/foo.1/versions/{n}
/schemagroups/myapp/schemas/foo.2/versions/{n}
/schemagroups/myapp/schemas/foo.3/versions/{n}

than by 

/schemagroups/myapp/schemas/foo/versions/1.1
/schemagroups/myapp/schemas/foo/versions/2.2
/schemagroups/myapp/schemas/foo/versions/3.0

Those things under foo are not the same if they don't describe structurally compatible data.

schemaregistry/schemaregistry.md
clemensv: Yes, that is what I mean. Some formats have richer and some formats have simpler type models and the schemas might all reflect the respectively best approximation of the data types of an app that's using a type model based on the CLR or JVM that's yet different.

schemaregistry/schemaregistry.md
clemensv: If you think we need to, yes.

schemaregistry/schemaregistry.md
clemensv: Unauthorized parties can find out by probing the registry for conflicts, so there is no significant security gain in hiding the group names. This mandates that the API doesn't try to obscure the reason for potential conflicts.

schemaregistry/schemaregistry.md
clemensv: I'll try to rephrase that

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+  - "This group holds schemas for the fabulous example app."
+
+## 2.2. Schema and Schema Version
+
+Conceptually, a schema is a description of a data structure. Since data
+structures evolve over time, the schema describing them will also evolve over
+time. Therefore, a schema often has multiple versions.
+
+For simple scenarios, the API allows for version management to be automatic and
+transparent. Whenever a schema is updated, a new version number is assigned and
```
deissnerk, line 140: How does this work in combination with schema versions in multiple formats?

schemaregistry/schemaregistry.md
ryanhorn: Understood, but wouldn't semver (or other versioning schemes) imply a different data type?

schemaregistry/schemaregistry.md
ryanhorn: Thinking about my example some more, it's actually not valid in the context of cloud events, as the schema is determined by the producer of the event and a reference to the specific schema is enclosed in the cloud event metadata.

However, this did bring up another thought specifically with cloud events. If the schema format is indicated by the Content-Type, how does a cloud event consumer know which one to request? The dataschema URI wouldn't be enough. There's the datacontenttype, but that describes the encoding, not the schema format.

schemaregistry/schemaregistry.md
duglin: By doing this you're basically asking the impl to be a document management system. Would it be so bad if we left that up to the author of the schema files and they could pretty much pick any URL pattern they wanted (within their permissions/scope)? Meaning, if they PUT over an existing one then it updates it. If they PUT to a new URL them they're creating a new one. Then the impl doesn't need any versioning at all, no saving of history, etc.. that's left up to systems that are built for that kind of thing and the results are pushed here for sharing/viewing.

schemaregistry/schemaregistry.md
duglin: I know what you mean.... but wait for it.... :-)

schemaregistry/schemaregistry.md
duglin: ah, I didn't realize you meant for "schemagroups" to be an impl choice - we might want to make clearer throughout then entire doc

schemaregistry/schemaregistry.md
clemensv: It's okay not to do updates atomically, in my view. To use a schema, you'll need to have it registered. The consumer can't be getting an Avro encoded event if the schema has not been registered by the publisher, if the publisher play by the rules. I also don't want to be prescriptive about the compatibility enforcements; conventions are okay, too.

schemaregistry/schemaregistry.md
clemensv: Assignment is per format. So you would store JSON and Avro and Protobuf one after another and they would all get the same version in the simplest case. For more control, you should also be able to PUT on the version resource per-se.

deissnerk: I wonder, if this could lead to confusing situations:
```
PUT JSON, PUT Avro -> JSON V1, Avro V1
PUT JSON -> JSON V2
PUT JSON, PUT Avro -> JSON V3, Avro V2
```
Even if a client always PUTs both formats, clients can crash. You might also run into race conditions.
We could of course state that people who don't have a simple use case should refrain from this kind of usage. Another option might be to allow the automatic version creation only for schema groups that are restricted to a single format.

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+- Schema: A document describing the structure, names, and types of some
+  structured data payload.
+- Version: A specific version of a schema document. Even though not prescribed
+  in this specification, an implementation might choose to impose compatibility
+  constraints on versions following the initial version of a schema.
+
+> NOTE: There are two fundamentally different options for how we can organize
+> multiple schema formats and this specification draft reflects both: 
+>
+> 1) In an "ideal" REST service, any data entity can be represented ("R" in
```
gunnarmorling, line 48: Does a registry have to be REST-ful to begin with? Seems worth spending some cycles on challenging this, e.g. it could be interesting to have a registry which is events-based itself, i.e. it'd publish streams of schema versions. That'd put more emphasis on enabling interested consumers to keep copies of schemas they're interested themselves in a push-based manner, instead of relying on a pull-based model.

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+>    REST) in multiple different data formats, which a client can request in
+>    prioritized order during content-type negotiation. If a service offers up
```
gunnarmorling, line 50: This also would imply equal expressiveness of all supported schema languages, which is not necessarily the case (type systems might be different etc.) Also from what I see from Debezium usage (used a lot with Kafka schema registries like Red Hat's and Confluent's), people seem fine with one format or another for a given event type.

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+> 2) Since the above strategy is truly RESTful, but quite esoteric if you've not
+>    grown up as a RESTafarian, the alternative strategy for concurrently
+>    handling multiple schema formats is much simpler: Constrain each schema
+>    group to a single format.
+
+## 2. Schema Registry Elements
+
+This section further describes the elements enumerated in the introduction.
+
+### 2.1. Schema Group
```
gunnarmorling, line 67: Can there be groups of groups?

schemaregistry/schemaregistry.md
gunnarmorling: I was wondering the same.

schemaregistry/schemaregistry.yaml
```diff
@@ -0,0 +1,313 @@
...
+          description: Created
+          headers:
+            Location:
+              schema:
+                type: string
+        '409':
+          description: Schema group already exists
+      tags:
+      -  'groups'
+    delete:
```
gunnarmorling, line 67: Should DELETE actually be allowed? Did you consider a "decomission" option or "soft delete" alternatively? That'd e.g. prohibit to produce new events referencing such schema taken out of business, but existing events could still be decoded.

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+- Examples:
+  - myschema
+  - my-schema
+
+### 2.2.3 Schema version attributes
+
+A schema version is a document. The "body" of a schema version MAY be a text
+document or binary stream. An implementation SHOULD validate whether a
+schema version is valid according to the rules of its format, for instance
+whether it is a valid Avro schema document when the format is Apache Avro.
```
evankanderson, line 186: How do recipients know what type of content to expect in the body? Is this based on the `datacontenttype` in the received message (plus some sort of lookup table to map `datacontenttype` to `format`)?

If so, making `format` the same value as `datacontenttype` would simplify the API.

schemaregistry/schemaregistry.yaml
```diff
@@ -0,0 +1,313 @@
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/SchemaGroup'
+        '404':
+          description: Specified group not found
+      tags:
+      - 'groups'
+    put:
+      summary: Create schema group
+      description: Create schema group with specified format format in registry namespace.
```
evankanderson, line 47:
```suggestion
      description: Create schema group with specified format in registry namespace.
```

schemaregistry/schemaregistry.yaml
```diff
@@ -0,0 +1,313 @@
...
+          description: Invalid request
+      tags:
+        - 'runtime'
+    get:
+      summary: Get latest version of schema
+      description: Get latest version of schema. 
+      operationId: getLatestSchema
+      responses: 
+        '200':
+          $ref: '#/components/responses/SchemaBytePayloadResponse'
```
evankanderson, line 156: Do you want to include the ID that was served in this response, so that clients can retrieve the value again on subsequent calls?

schemaregistry/schemaregistry.md
evankanderson: Given the ability and suggestion to semver the `dataschema` or schema `id`, complex versioning schemes within a schema seem less necessary. A server-side check to prevent footguns should be possible even without versions, and this API doesn't actually prevent replacing a specified version with an incompatible document at the same URL (delete the existing version and/or schema, then re-create a schema with the same name and re-insert schemas until the version lines up, but the content doesn't).

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
+# CNCF Schema Registry API Version 0.1-rc01s
```
duglin, line 1: maybe "CNCF Schema Registry API - wip" since it's not an 'rc' yet.

schemaregistry/schemaregistry.md
ryanhorn: Yup, I agree with that.

Regarding my second comment, in the context of cloud events, do we think that we need to provide more information in the cloud event about how the client requests the schema referenced by dataschemauri? If a registry hosts multiple schema types for the same logical schema at the same location and the client is required to indicate which type they want by sending Content-Type, then dataschemauri is no longer enough information for a client to request the correct schema that was used to structure/encode the event.

A cloud event does enclose the datacontenttype, but that describes the encoding of the event, not the type of the schema itself.

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+>    be three distinct schemas, each defining a representation of the data
+>    entity for the respective format. From the same "REST" mindset perspective,
+>    the ideal representation of this set of schemas, which differ with regards
+>    to the format they describe, but ought to be semantically identical, is to
+>    also differentiate them only by their content type and maintain them under
+>    the exact same name as if they were one schema.
+> 2) Since the above strategy is truly RESTful, but quite esoteric if you've not
+>    grown up as a RESTafarian, the alternative strategy for concurrently
+>    handling multiple schema formats is much simpler: Constrain each schema
+>    group to a single format.
```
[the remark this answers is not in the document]
EricWittmann: What about constraining on a per-schema basis (giving each schema a "format" or "type").  The reason here is that I can imagine a logical group of schemas that aren't all the same technology.  Especially if this were ever to expand beyond schemas and into e.g. API Designs as well - I would want to have a group that included perhaps multiple OpenAPI documents as well as some JSON Schemas... and my OpenAPI would likely have `$ref`s to the JSON Schemas.

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+
+## 2. Schema Registry Elements
+
+This section further describes the elements enumerated in the introduction.
+
+### 2.1. Schema Group
+
```
EricWittmann, line 68: Just wanted to add that groups is a great concept that we don't (yet) have in Apicurio Registry, but may have been a mistake to not include.  It's important I think to organize these things into groupings.  I guess I'm just saying +1 to this concept from me. :)

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+Access control rules at the schema group level MUST NOT limit visibility of the
+schema groups themselves to users authorized to enumerate schema groups at the
+registry level or to create new schema groups in order to create unambiguous
+transparency about which `id` have already been taken. An implementation MAY
+withhold all details other than the `id` attribute from a user not authorized
+to access the schema group.
+
+This specification does not define management constructs for such access control
+rules.
```
EricWittmann, line 90: +1 Good

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+The data model for a schema group consists of three attributes:
+
+#### id
+
+- Type: `String`
+- Description: Identifies the schema group.
+- Constraints:
+  - REQUIRED
+  - MUST be a non-empty string
+  - MUST conform with RFC3986/3.3 `segment-nz-nc` syntax
```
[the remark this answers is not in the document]
EricWittmann: I think `segment-nz-nc` better supports some common ways you might want to name your groups to indicate a hierarchy where none exists.  Or an Organization + Project format - that sort of thing.  I'm thinking things like how a lot of NPM packages are now being named...

schemaregistry/schemaregistry.md
EricWittmann: I don't have a strong opinion here - but agree that media types can be problematic for the reason given by @evankanderson  - lots of formats don't have registered mime types. :(  Mime type does benefit from being nicely extensible (vs. something like an enum).  I presume it's not the goal of this spec to mandate the supported schema formats an implementation must support.

schemaregistry/schemaregistry.md
EricWittmann: > That being said, I'm curious to hear if there's a real world example where there is value in offering multiple schema formats for the same logical schema.

I am also very interested in a real-world example of this requirement.  It's interesting, but the various schema technologies can be so different and varied ... and I haven't seen any use cases where users need the same data format described in different schema technologies.  I can image lots of schemas where doing that would either be impractical or impossible.  But on the other hand, in a polyglot environment I can see the utility...

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+For simple scenarios, the API allows for version management to be automatic and
+transparent. Whenever a schema is updated, a new version number is assigned and
+prior schema versions are retained. The latest available schema is always the
+default version that is retrieved when the API is given just the schema `id`.
+
+A newer schema version might introduce breaking changes or it might only
+introduce careful changes that preserve compatibility. These strategies are not
+subject of this specification, but the API provides a conflict handling
+mechanism that allows an implementation to reject updates that do not comply
+with a compatibility policy, if one has been implemented.
```
EricWittmann, line 148: Schema evolution is a pretty significant concept in a schema registry.  I haven't seen how the spec facilitates configuring the compatibility (or validity) rules.  This paragraph mentions a compatibility policy, but if that's mentioned anywhere else I missed it. :(

This might be an important enough feature to include in the spec.  For Apicurio Registry we have the concept of "rules" that can be configured globally or per-artifact (in this spec I imagine rules could also be configured at the schemagroup level).  Right now we have only two rules:  Validity and Compatibility.  But perhaps "rule" is a concept to consider?

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+concurrently. That implies that the schema managed by this registry is not
+a single document, but MAY be a set of documents that all describe the exact
+same data structure, each in their own particular syntax and using their own 
+type model.
+
+The schema version therefore relates to this abstract notion of schema and the
+described data structure. All documents coexisting within the same version
+SHOULD describe the exact same data structure.
+
+### 2.2.2. Schema attributes
```
EricWittmann, line 163: It would be useful to keep some other meta-data for each artifact, particularly if anyone eve wants to create a Registry UI.  Name, description, labels, creationTime, modifiedTime, etc.

schemaregistry/schemaregistry.md
EricWittmann, line 186: Can validity be disabled or configured (e.g. syntax vs. semantic validity)?  Treating Validity and Compatibility in similar ways has proven useful in Apicurio Registry.

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,257 @@
...
+- Examples:
+  - 1
+  - 2
+
+#### format
+
+- Type: `String`
+- Description: 
+- Constraints:
+  - OPTIONAL. Can be used if and only if not format has been set for the schema
```
EricWittmann, line 211: Should be:  `if and only if format has not been set for the schema`

schemaregistry/schemaregistry.md
EricWittmann: Also update the OpenAPI to remove `/schemagroups` from paths.  Would make it easier to generate servers/clients from the OpenAPI at some future point.

schemaregistry/schemaregistry.md
duglin: They could optionally support the Discovery and Subscription API specs!  :-)

schemaregistry/schemaregistry.yaml
EricWittmann: Yes, in general deletes are problematic and should IMO be either discouraged or prevented.  Immutability has been a useful property in our registry for schema versions - we can deprecate and disable versions.  And I wish we had done the same for the entire schema (deprecate/disable) rather than allow deletes.

schemaregistry/schemaregistry.md
chrish42: Just discovered this project. Evaluating whether we could use this instead of writing our own schema registry. We want to be able to use semver to represent versions for a schema, where foo v1.y can be used as-is instead of foo v1.x (for y >= x), but foo v1.x and foo v2 are incompatible representations of the same data (and then maybe there'll be a converter registered in our system, etc.).

It's not clear to me which direction the conversation is leaning, but it'd be really lovely if it was possible to control the version id assigned to a new version of a schema. Otherwise, I mean, we can always add a proxy in front that remaps foo/versions/x.y to foo.x/versions/y, I guess. But it would be simpler if we could just control the version id to be created as we want. This is a very localized configuration point, too.

EricWittmann: One interesting thing I was thinking about with respect to semver support for versioning, is that compatibility checking could perhaps be customized with a different configuration based on a real understanding of semver.  So if a client tries to update the content of a schema that is currently at version `2.7.3`, uploading content to new version `2.7.4` then that is a patch version release and the compatibility checker can be very strict.  But if the client does the same thing to new version `3.0.0` then the compatibility checker can e.g. allow breaking changes.

Just a thought. :)

schemaregistry/schemaregistry.md
clemensv: @gunnarmorling You ultimately need a place where you can just go and look things up. A serialization framework that turns an object graph into a serialized Avro container needs to be able to fetch the correct schema in the current state of the art. This interface is literally for interoperable access to the place where the copies of the schemas are kept such that it can be accessed from multiple language runtimes and across the network. Schema distribution is a different concern, IMO, and because of the significant size of many schemas, that interaction is best done using a claim-check pattern where a schema version update gets announced and then the schema is fetched by interested parties. I do like the idea of standardizing those notifications, though; I will add change events that will let you do what you are looking for.

schemaregistry/schemaregistry.md
clemensv: @gunnarmorling I'm also generally leaning towards the simpler option, but I don't think schema languages needs to be perfectly aligned into all corners of their features for this to make sense. The majority of data structurs being passed around are fairly straightforward and do ok with a handful of common types and without getting much into sophisticated constraining expressions. Having one transfer-optimized format for a particular data structure (Avro, Parquet, Proto, etc) and then also a JSON representation for broader interop reach isn't terrible.

schemaregistry/schemaregistry.md
clemensv: Copying @gunnarmorling

After hearing all the arguments here, Im going to drop the option to keep multiple formats together under one name/version for the time being to keep things simple. If we would want to do this later, the path I chose here can indeed be added later without breaking backwards compatibility because you would just omit the now-required format declarations, which would then become optional.

schemaregistry/schemaregistry.md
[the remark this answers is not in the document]
clemensv: @EricWittmann I will add that as an XOR option, i.e. you can either define the formats at the group or schema level. If you define it at the group level, that is binding, meaning you can't override.

schemaregistry/schemaregistry.md
clemensv: New: MUST conform with RFC3986/3.3 `segment-nz-nc` syntax. This allows for "dot
    notation", e.g. `org.example.myapp.module` for logical organization of
    schema groups, even though the namespace is flat.

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,318 @@
...
+transparency about which `id` have already been taken. An implementation MAY
+withhold all details other than the `id` attribute from a user not authorized
+to access the schema group.
+
+This specification does not define management constructs for such access control
+rules.
+
+### 2.1.1. Schema group attributes
+
+The data model for a schema group consists of three attributes:
```
duglin, line 76: s/three/these/
since keeping this number up to date is hard.... in fact, it's already wrong :-)

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,318 @@
...
+  - OPTIONAL
+- Examples:
+  - "This group holds schemas for the fabulous example app."
+
+#### createdtimeutc
+
+- Type: `Timestamp`
+- Description: Instant when the schema group was added to the registry.
+- Constraints:
+  - OPTIONAL
```
duglin, line 126: Should probably add a MUST in here about the format/syntax

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,318 @@
...
+structures evolve over time, the schema describing them will also evolve over
+time. Therefore, a schema often has multiple versions.
+
+For simple scenarios, the API allows for version management to be automatic and
+transparent. Whenever a schema is updated, a new version number is assigned and
+prior schema versions are retained. This specification does not mandate a
+retention policy, but implementations MAY retire and remove outdated schema
+versions.
+
+The latest available schema is always the default version that is retrieved when
```
duglin, line 149: Might want to add a MUST in here someplace....
```
When the URL to a schema is used without a version string, the implementation MUST return the latest version of that schema.
```
perhaps?

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,318 @@
...
+- Examples:
+  - "This schema describes the toppings for a Pizza."
+
+  #### createdtimeutc
+
+- Type: `Timestamp`
+- Description: Instant when the schema was added to the registry.
+- Constraints:
+  - OPTIONAL
+  - Assigned by the server.
```
duglin, line 213: My OCD is kicking in... sometimes you have periods, sometime you don't on bulleted lists :-)  can we choose one? I'd prefer no periods.

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,318 @@
...
+
+#### id
+
+- Type: `String`
+- Description: Identifies the schema.
+- Constraints:
+  - REQUIRED
+  - MUST be a non-empty string
+  - MUST conform with RFC3986/3.3 `segment-nz-nc` syntax
+  - MUST be unique within the scope of the schema registry
```
duglin, line 177: all of the schema registry or just within this schema group?  I would think the schema group would be good enough

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,318 @@
...
+#### description
+
+- Type: `String`
+- Description: Explains the purpose of the schema.
+- Constraints:
+  - OPTIONAL
+- Examples:
+  - "This schema describes the toppings for a Pizza."
+
+  #### createdtimeutc
```
duglin, line 207: The leading spaces makes this header indented

schemaregistry/schemaregistry.md
duglin, line 213: as above, I think we need a MUST here specifying the format/syntax - or define "Timestamp" in some kind of "data types" section

schemaregistry/schemaregistry.md
duglin: Any reason you have "created" here but not on the Schema group?

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,318 @@
...
+  - Assigned by the server.
+
+### 2.2.3 Schema version attributes
+
+A schema version is a document. The "body" of a schema version MAY be a text
+document or binary stream. An implementation SHOULD validate whether a
+schema version is valid according to the rules of its format, for instance
+whether it is a valid Avro schema document when the format is Apache Avro.
+
+Within the scope of the schema set, the version is identified by the combination
```
duglin, line 230: use of the word "set" here might confuse people since it's new. Did you mean "group" ?

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,318 @@
...
+of a version number and an optional format identifier. The schema version MAY
+also have an additional, optional unique identifier within the scope of the
```
duglin, line 232: Can you elaborate on this last sentence? Do you mean they can add extensions that are unique identifier values? If so, why did you call this out? Just curious.

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,318 @@
...
+  the version in the scope of this schema within the schema group. The schema
+  document content MAY indicate a different versioning scheme.
+- Constraints:
+  - REQUIRED
+  - Assigned by server.
+- Examples:
+  - 1
+  - 2
+
+#### id
```
duglin, line 248: How do you see this being used by a consumer of the registry?

schemaregistry/schemaregistry.md
```diff
@@ -0,0 +1,318 @@
...
+part of this specification.
+
+### 3.1. Path hierarchy
+
+Schema groups contain schemas and those contain schema versions, which are the
+documents required for serialization or validation.
+
+These dependencies are reflected in the path structure:
+
+`[/schemagroups]/{group-id}/schemas/{schema-id}/versions/{version}`
```
duglin, line 289: Not sure if it matters, but the use of the word `id` got me wondering if it should be `name` instead. While the semantics would be the same either way, all examples of these IDs appear to be more like human friendly names rather than IDs (e.g GUIDs). So while it's not totally human friendly (eg no spaces, etc...), if we expect people to use meaningful "words" and not "random chars", then perhaps `name` would help guide them in that direction.

Just a thought

duglin: Do we need to add something here to indicate that the `/versions/{version}` part is optional?
`[/schemagroups]/{group-id}/schemas/{schema-id}[/versions/{version}]`
or text?

duglin:
@clemensv since you haven't gotten to the REST API yet, let me ask.... do you see people being able to do a PUT to replace an existing version of the schema or will it always be required to be a POST and get a new version?

I'm hoping they can do a PUT because I can image cases where people need to do non breaking changes (e.g. fixing typos in comments in the schema) and don't want to necessarily expose the equivalent of all "git commits" to their users.

chrish42:
@duglin Speaking for myself, for deployments of this, I would be more concerned with reproducibility than the ability of fix typos in comments, etc. So unless there's some kind of "pre-commit hook" to check that a PUT does not affect the schema, I'd much rather everything go through POST and create a new version number.

duglin:
@chrish42 sure - and as the owner of that schema you should have that choice... meaning you could always use a POST. But should we then ban someone from using a PUT if they want to hide their typos (or anything else they want to remove from the old versions)?

hmm I guess this might be less of a concern if the spec supports DELETE , but it's not in the list of ops for a version yet.  So, imagine a rogue employee uploading something bad on the way out the door and no way to fix it.

chrish42:
@duglin I'm building a data system where reproducibility is a top concern. That doesn't work it you leave that up to the owner of each schema in the system. I understand this may not be as strong a concern for everyone, so that might need to be made optional. But for me, I need a way to deploy this that is "append-only" at the schema registry level.

A "retract" feature could be useful in this mode (hide a schema version from LIST operations, but continue to allow access to it for operations that fetch that exact version). But that can probably be wait until after the first release.

Anyways, I'll let other chime in here.

duglin:
@chrish42 gotcha! Thanks for the insight!

EricWittmann:
Yeah we've found that immutability has worked better than allowing DELETE and PUT for versions - once a version is added it can't be removed or modified.  We have a state that can be 'deprecated' or 'disabled' to handle cases where a version shouldn't be used, or shouldn't be listed.  It would be awesome if the spec at least allowed for that concept/mode.

schemaregistry/schemaregistry.md
clemensv: Pointing to the core spec.

schemaregistry/schemaregistry.md
clemensv: I do

schemaregistry/schemaregistry.md
clemensv: I am referring to the CE spec for the data types now. Do you think we need to copy them?

schemaregistry/schemaregistry.md
clemensv: Any concrete schema document may also have a unique identifier for itself. There is either a path /group/xyz/schemas/abc/versions/1 or you could just address that exact doc with its ID. We want that to enable a URL shortener function.

schemaregistry/schemaregistry.md
clemensv: aka.ms/s/{id}   URL shortener option for greedy protocols.

schemaregistry/schemaregistry.md
duglin: I was just referring to the lack of consistency on the bulleted lists - some ending in a period and some not. I don't have an opinion on copying vs referencing the data types

schemaregistry/schemaregistry.md
duglin: Why isn't that just part of the URL shortener function's logic? E.g. tinyurl.com doesn't ask for an id - I just give it a full URL

duglin:
Approved on 6/18 call

chrish42:
@duglin @clemensv For folks like me who are interested in this, what is the next step here now that the schema registry proposal has been merged? What "thing" can I subscribe to? And where would implementation efforts be happening, so I could potentially help? Thanks!

duglin:
hi @chrish42, good question. I think this spec (and subscriptions) should follow the same path as discovery. Meaning, people should review it, open issues/PRs for changes, and then start to implement it to expose gaps. We've already started down this path for discovery, but not for subscriptions yet. I think people are just overloaded so it might take a while for the ball to really get rolling.

chrish42:
@duglin So, just to be sure I understand, the CloudEvents project itself is not working on an implementation of this spec, but member companies (maybe Microsoft given that the initiator @clemensv works there) potentially are writing their own implementation of this spec. Correct? I'm trying to figure out how best to learn about implementations of this as they appear, and maybe collaborate on one if it pops up early enough and our needs are aligned. Hence the questions. Thanks!

robinmeiss:
Hello all,

I really like the approach to standardize on this big missing topic (on an enterprise level).

May I propose to include an explicit metadata resource to additionally allow schema labels. So you can use a selector on the list operations (like K8s is doing it in a very flexible way: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors).

/schemagroups/mygroup/schemas/schemas/{schema-id}/metadata

```
"metadata": {
  <...>
  "creator": "Author1",
  "created_at": "2012-04-23T18:25:43.511Z",
  <...>
  "labels": {
    "key1" : "value1",
    "key2" : "value2",
    "schema-type": "public-api",
    "business-domain": "marketing",
    "PII": "false"
  }
}
```

This way anyone could just query `GET /schemagroups/mygroup/schemas/schemas/{schema-id}?labelSelector=environment%3Dproduction` getting all schemas with relevant labels, allowing for very flexible grouping and tagging variants.

So my interest is really on the "label selection" part (to group in common taxonomies) but there might be more common things to additionally put in a metadata resource (nevertheless you probably could "misuse" to put any additional information also as key/values).

zaheershaik:
Hi. I don't see this spec in the master branch anymore. Did it move to a different repo or project?

duglin:
It's there - see: https://github.com/cloudevents/spec/tree/master/schemaregistry
the default branch for the repo is now v1.0.1

Issue #576, opened 2020-02-29
Proposal for a new Protobuf spec


drtriumph:
In response to #504 I'd like to start discussion on a new spec for Protobuf (I'm not sure this is the right place but it seems as good as any)

Here's my initial pass for comment

```protobuf
syntax = "proto3";

package io.cloudevents.v1;

import "google/protobuf/any.proto";
import "google/protobuf/struct.proto";
import "google/protobuf/timestamp.proto";

message CloudEvent{
  // Required Fields
  string id = 1;
  string source = 2;
  string spec_version = 3 [json_name="specversion"];
  string type = 4;

  // Optional Fields
  string data_content_type = 5 [json_name="datacontenttype"];
  string data_schema = 6 [json_name="dataschema"];
  string subject = 7;
  google.protobuf.Timestamp time = 8;
  google.protobuf.Value data = 9;
  bytes data_binary = 10 [json_name="data_base64"];
  google.protobuf.Any data_proto = 11 [json_name="iocloudeventsprotodata"];
  google.protobuf.Struct extention_context = 12
    [json_name="iocloudeventsprotoextentioncontext"];
}
```

### Pros

- Serializes strictly to the JSON spec using default protobuf tooling
- Uses google.protobuf well-known types to store anything you could in a JSON event
- Strictly defines types for all required and optional attributes not data / extensions
- Handles time and base64 format conversion

### Cons

- Silently drops extension context attributes of a JSON event when serialized into a protobuf using standard tooling
- Is somewhat clumsy / unidiomatic in most languages to add and extract data attributes as they don't have specific or consistent types

If this is generally a good direction I'm happy to turn this into a PR for further comment.

## Example

Marshaling to and from a JSON event in Go yields:

Input:
```json
{
    "id" : "C234-1234-1234",
    "source" : "/mycontext",
    "specversion" : "1.0",
    "type" : "com.example.someevent",
    "datacontenttype" : "application/json",
    "time" : "2018-04-05T17:31:00Z",
    "comexampleextension1" : "value",
    "comexampleothervalue" : 5,
    "data" : {
        "appinfoA" : "abc",
        "appinfoB" : 123,
        "appinfoC" : true
    }
}
```

Output:
```json
{
    "id": "C234-1234-1234",
    "source": "/mycontext",
    "specversion": "1.0",
    "type": "com.example.someevent",
    "datacontenttype": "application/json",
    "time": "2018-04-05T17:31:00Z",
    "data": {
[...]

PR #626, opened 2020-05-20
Re-Introducing Protocol Buffer Representation

JemDay:
re-introducing protocol buffer support, this takes a slightly different approach to the one originally proposed.

I've used `one-of` constructs which I'm aware does have the potential to introduce some processing overhead but it does enable a level of structural correctness.

It's been a while since I've used Proto on a day-2-day basis so I'd appreciate comments.


duglin:
ping @evankanderson

clemensv:
Is the ownership story regarding Protobuf resolved, i.e. did Google put it into CNCF, yet?

markfisher:
ideally this could replace the `LiiklusEvent` as defined here: https://github.com/bsideup/liiklus/blob/master/protocol/src/main/proto/LiiklusService.proto

and possibly, we could then discuss Liiklus' proto `service` moving into the CloudEvents scope as well (along with unary service perhaps)

cc @bsideup

JemDay:
@markfisher - I toyed with add _grpc_ but that might be a more of a transport specification thing.

Do we consider gRPC a transport ?

I've updated the schema to represent the required and optional attributes as first-class fields.

evankanderson:
I believe gRPC is part of CNCF, so a proto format for CloudEvents would align with gRPC transport definitions.

I'll take a look later today; I think there was also an earlier protobuf definition PR, so we might want to reconcile the two.

protobuf-format.md
```diff
@@ -0,0 +1,181 @@
...
+      string ce_string = 3;
+      bytes ce_binary = 4;
+      string ce_uri = 5;
+      string ce_uri_reference = 6;
+      google.protobuf.Timestamp ce_timestamp = 7;
+    }
+}
+```
+
+NOTE: Field assignment numbers are non-normative.
```
evankanderson, line 88: Field assignment numbers absolutely *have* to be defined by the spec. Proto wire format encodes tag + type as a prefix, so if the tag doesn't match the expected type, the entire message will be discarded during decoding.

protobuf-format.md
```diff
@@ -0,0 +1,181 @@
...
+This approach allows for extensions to be represented and transported
+with no loss of *type* information.
+
+## 3. Data
+
+The specification allows for data payloads of the following types to be explicitly represented:
+
+* string
+* bytes
+* protobuf object/message
```
evankanderson, line 102: I thought this was a json structure (possibly equivalent to protobuf `Struct`).

protobuf-format.md
```diff
@@ -0,0 +1,181 @@
...
+oneof data_oneof {
+
+    // Binary data
+    bytes raw_data = 2;
+
+    // String data
+    string text_data = 3;
+
+    // Protobuf Message data
+    google.protobuf.Any proto_data = 4;
```
evankanderson, line 114: I don't think you want to use `Any` here; `Any` is basically a wrapper for a string type and an enveloped message. For CloudEvents, I think it would be more natural to put the message bytes into `raw_data`, and the message type into a CE extension attribute.

protobuf-format.md
```diff
@@ -0,0 +1,181 @@
...
+}
+```
+
+Before encoding, a protobuf serializer MUST first determine the runtime data type
+of the content. This can be determined by examining the data for invalid UTF-8
+sequences or by consulting the `datacontenttype` attribute.
+
+* Where the data is a protobuf message it MUST be stored in the `proto_data`
```
evankanderson, line 122: Why complicate this vs. just using `bytes` (and possibly needing to decode the payload a second time), vs needing to scan the payload to determine which field to put the content into?

Also, is it an error to receive a utf-8 string or a proto message as `raw_data`? If not, this seems like a case where the recipient will know what to do best.

protobuf-format.md
```diff
@@ -0,0 +1,181 @@
...
+   application/cloudevents+x-protobuf
+```
+
+## 5. Examples
+
+The following code-snippets show how proto representations might be constucted asuming the availability of some convenience methods ...
+
+```java
+public static Spec.CloudEvent plainTextExample() {
+    Spec.CloudEvent.Builder ceBuilder = Spec.CloudEvent.newBuilder();
```
evankanderson, line 143: You never define the enclosing CloudEvent type, only `extensions` and `data_oneof`.

protobuf-format.md
```diff
@@ -0,0 +1,181 @@
...
+message CloudEventAttribute {
+
+    oneof attr_oneof {
+      bool ce_boolean = 1;
+      sfixed32 ce_integer = 2;
+      string ce_string = 3;
+      bytes ce_binary = 4;
+      string ce_uri = 5;
+      string ce_uri_reference = 6;
+      google.protobuf.Timestamp ce_timestamp = 7;
```
evankanderson, line 83: How are unknown attribute keys handled? If I get a string for "datakey" with value "/path", should it be encoded as `ce_string`, `ce_binary`, or `ce_uri_reference`? (The same is also true for timestamp)

protobuf-format.md
```diff
@@ -0,0 +1,181 @@
...
+This section defines how CloudEvents attributes are represented in the protobuf
+[schema][proto-schema].
+
+Required and Optional attributes are represented explicitly as protobuf fields.
+
+CloudEvent extension attributes are represented using a map construct enabling
+direct support of the CloudEvent [type system][ce-types].
+
+```proto
+map<string, CloudEventAttribute> extensions = 1;
```
evankanderson, line 72: This map is labeled `extensions`. Are the core attributes handled the same way, or by unique tags?

The example at the end suggests that known attributes are assigned unique tags.

spec.proto
```diff
@@ -0,0 +1,58 @@
...
+  string type = 4;
+
+  // Optional
+  string data_content_type = 5; // RFC 2046
+  string data_schema = 6;       // URI
+  string subject = 7;
+  google.protobuf.Timestamp time = 8;
+
+  // -- CloudEvent Extension Attributes
+  map<string, CloudEventAttribute> extensions = 9;
```
evankanderson, line 31: See also my comments in https://github.com/cloudevents/spec/issues/576#issuecomment-612107502

You may want to define all optional attributes (core and extension) as part of the same string-keyed map, rather than putting a few as tags but not later-defined attributes.

evankanderson:
If expect a follow-on to proto definition would be a set of gRPC transports. At a minimum, this would be unary request-response and streaming request-response, but there might be other useful profiles.

evankanderson:
(oh, and see also #576)

bsideup:
@JemDay Nice! I have a couple of questions since I think I miss some important details:

1. I wonder why we have attributes as `map<string, value>` while they are strictly defined in the spec?
2. With `map`, how we plan to deprecate certain attributes? Or define required/optional?
3. Does it really makes sense to have `ce_uri` and `ce_uri_reference`, given that we're not gaining a type safety here, since protobuf is missing `URI` type? Perhaps we should add `message URI`?

Thanks in advance!

protobuf-format.md
bsideup, line 83: If we use `google.protobuf.Timestamp` here, we will always be using UTC for the timezone, while the spec says that it should be a timezoned datetime (which cares an extra information and can be used to align the notifications schedule, for example).

protobuf-format.md
JemDay: I think my comment was mis-leading ... what I was trying to imply was that the content in this document were 'representative' but the actual schema is defined in the spec.proto file .. I'll clean that up.

protobuf-format.md
JemDay: Not sure i follow your comment, My intention was to allow the transportation of Proto event data in a way that is more idiomatic for Proto. ie there's no need to marshal the data into a separate field, the pack/unpack feature of `Any` supports this directly.

I'll add an example..

protobuf-format.md
JemDay: See my previous comment ... if I'm dealing natively with protobuf it's not very idiomatic to marshal it separately.

I don't believe the spec calls for all event data to be encoded as a string.

protobuf-format.md
JemDay: I get your point but I'm not sure it complicates things. My thought process was that it was a bit odd to put textual content into a byte array when Proto allow us to be explicit.

protobuf-format.md
JemDay: Spec.CloudEvent *is* the enclosing type, the example method returns an instance of a CloudEvent.

protobuf-format.md
JemDay: The originator of the event presumably know the 'type', so they wold populate the extension attribute accordingly.

I'll add an example...

protobuf-format.md
JemDay: originally I had all the defined attributes in a _map_ structure, I flipped to a model when the defined attributes are explicit (as in the JSON schema).

So in that model only extension attributes need to be carried in the _map_.

spec.proto
JemDay: that's an interesting point ... we don't do that in that in the other format definitions but is something to consider.

It doesn't have the potential to introduce complexity where consumers/produces might be using different version of the spec as new extensions are added - Proto can cope with that but it would increase complexity for consumers.

I'd argue for sticking with the defined 'required' and 'optional' context attributes.

protobuf-format.md
JemDay: with respect to your prior comment (which I can't comment on directly)..

1 - I think I updated the spec after you commented, the context attributes defined in the spec are now represented explicitly.
2 - Again, with the re-structure I don't think that's an issue. In Proto typically everything is optional;i believe there is a way to flag 'required' by Proto purists may object :-)
3 - For me it's about retiring the type information, a receiver can determine that the context is URI because the value is in that field.

This might be a question for the larger group, my interpretation was that the intention was to ensure timezone information was included so representing the value as UTC met that requirement form my point of view.

On closer inspection there's an implication that timezones MUST be communicated as strings - which seems a bit inefficient especially when native representations are available.

bsideup: Re optional/required:

https://github.com/protocolbuffers/protobuf/blob/master/docs/field_presence.md

evankanderson: Proto2 has a notion of optional/required semantics. These were dropped in proto3 due to operational experience -- as protocol buffers evolved over time, fields which were previously "required" might no longer make sense, but converting a required field to optional was an extensive multi-stage rollout process across all consumers of the protocol buffer, even those which did not use the fields explicitly (like intermediate proxies).

evankanderson: What about a component which is adapting structured JSON input from e.g. AWS Kinesis to gRPC transport? This sort of intermediate component may not be aware of the types of all extensions (by design), but without type information from the upstream transport, a `datakey` attribute might be transcribed to a different encoding than the recipient is expecting.

protobuf-format.md
evankanderson: FWIW, the binary encoding of protobuf lets you "Pun" two different protos between different length-delimited field types, like so:

```proto
// Use this type if you are an intermediate router and don't need to care about interpreting or validating payload.
message Envelope {
  string name =1;
  bytes payload = 2;
}

message Data {
  map<string, int32> values = 1;
  double magnitude = 2;
}

// If you are an endpoint node, decode Envelope as Decoded and you can interpret payload as the Data message.
message Decoded {
  string name = 1;
  Data data = 2;
}
```

The main benefit of using `string` vs `bytes` is that Proto will generate the correct types from the auto-generated code, possibly verifying that the string is valid UTF-8 along the way (I don't recall if all implementations verify the UTF-8-ness of the bytestream). The cost is that any proto interpretation which understands the message will unpack and verify all the inner payload fields, whether or not they are used by the application.

protobuf-format.md
evankanderson: If you look at the various binary encodings, they encode the payload as a simple bytestream (not a string): [HTTP](https://github.com/cloudevents/spec/blob/v1.0/http-protocol-binding.md#31-binary-content-mode), [Kafka](https://github.com/cloudevents/spec/blob/v1.0/kafka-protocol-binding.md#32-binary-content-mode), [MQTT](https://github.com/cloudevents/spec/blob/v1.0/mqtt-protocol-binding.md#22-data), [AMQP](https://github.com/cloudevents/spec/blob/v1.0/amqp-protocol-binding.md#31-binary-content-mode). [AVRO](https://github.com/cloudevents/spec/blob/v1.0/avro-format.md#3-data) has some funniness, but I wouldn't try to emulate it.

protobuf-format.md
evankanderson: My concern is about what happens when core attributes are added, e.g. in a 1.1 spec.

Clients which were built before the 1.1 spec will map these attributes into the `extensions` map, while post-1.1 clients will map them into a unique tag-assigned field. Recipients of messages will then need to write complicated logic to map pre-1.1 messages to post-1.1 messages, losing most of the benefit of code auto-generation for protocol buffers.

Yes, this is a "perfect is the enemy of the good" case, but my experience has been that not planning for expansion tends to lead to "everything was good until we tried to touch something, and now it's all terrible."

Another acceptable scenario is to accept that we can't add further attributes to the 1.0 proto message definition, and all additional attributes (whether currently documented extensions or not-yet-defined extensions) must be stored in `extensions`.

The last option is to keep a global tag number registry mapping extension names to tags, but that only helps humans and doesn't really help conversion gateways, which would need to be able to load the list dynamically when transforming a message between proto and other encodings.

protobuf-format.md
evankanderson: For an example of what this looks like in another typed transport see the [AMQP spec](https://github.com/cloudevents/spec/blob/v1.0/amqp-protocol-binding.md#3132-amqp-application-property-values), where clients need to be prepared to accept either `string` or typed values for each attribute.

duglin:
@JemDay is this one ready for review?

JemDay:
@duglin - Probably ready for a broader conversation, I do need to go through the comments from @evankanderson a bit more.

Items to thing about :
- representing attributes explicitly or as a simple map (I believe I saw an issue/question relating to how this is done in Avro)
- Do we consider gRPC to be a transport, or is it a special case like http webhooks.

protobuf-format.md
JemDay: Maybe there's a half-way house solution where all the 'required' attributes are explicitly define but the optional ones aren't, on the basis that we can't add any required attributes without changing the major version..

protobuf-format.md
JemDay: I think we're conflating 'transport' and 'format' .. this spec is only applicable in structured mode and in those situations the payload is opaque (apologies if I mis-stated that somewhere, now you've got me worried).

In AMQP this would be transported as a application/cloudevents+x-protobuf content type.

It'd be the gRPC transport specification (if we consider it a transport) that would define how binary and structured payloads would be exchanged.

Switching between different structured formats (JSON->Proto, Avro->JSON) is a higher-order function and not a concern of the transport binding I believe.

protobuf-format.md
```diff
@@ -0,0 +1,251 @@
...
+### 1.1. Conformance
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in [RFC2119][rfc2119].
+
+### 1.2 Content-Type
+
+There is no official IANA *media-type* designation for protobuf, as such this
+specification uses 'application/x-protobuf' to identify such content.
```
tweing, line 45: There was a draft for protobuf media type here: https://tools.ietf.org/html/draft-rfernando-protocol-buffers-00.
We recently decided in our company to use the media type `application/protobuf` proposed in the draft above. 
Would you mind switching to that definition instead of the proposed `application/x-protobuf`?

```suggestion
specification uses 'application/protobuf' to identify such content based on the draft [Encoding rules and MIME type for Protocol Buffers](https://tools.ietf.org/html/draft-rfernando-protocol-buffers-00).
```

JemDay: Hi,

I certainly think this warrants further discussion, i've seen both application/protobuf and application/x-prototbuf used.

Looking at [Spring for example](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/converter/protobuf/ProtobufHttpMessageConverter.html) they appear to use the x-prototbuf.

I'm not sure there's one correct answer to this :-(

duglin:
@JemDay is this one ready for consideration on tomorrow's call?

protobuf-format.md
tweing: Sorry, my mistake - we mainly thought about the `datacontenttype` attribute.

I think the proposal made in the call today to use `application/cloudevents+protobuf` for the `Content-Type` http header field for structured mode is reasonable.

@JemDay : You can ignore my suggested change.

JemDay:
@evankanderson - I think all of the concerns have been covered now and I believe this is pretty closely aligned with the commentary in Issue #576.

Any comments ?

protobuf-format.md
evankanderson: Sorry, I thought this was handled, but I'm looking back, and it looks like we talked past each other.

Assume the following _structured_ JSON CloudEvent with no payload:

```json
{
    "specversion" : "1.0",
    "type" : "com.example.someevent",
    "source" : "/mycontext",
    "id" : "A234-1234-1234",
    "comexampleextension1" : "2018-04-05T17:31:00Z",
}
```

What does the equivalent _structured_ Proto CloudEvent look like? (using text format for readability)

```
id: "A234-1234-1234"
source: "/mycontext"
spec_version: "1.0"
type: "com.example.someevent"
attributes: [{
  key: "time"
  value: {
    ce_timestamp: {...}
}, {
  key: "comexampleextension1"
  value: {
    ce_string: "...."
  }
}]
```

Alternately, `comexamplextension1` could be encoded as:
```
{
  key: "comexampleextension1"
  value: {
    ce_timestamp: { ... }
  }
}
```

If there is a generic JSON -> Proto CloudEvents bridge:
* How does it know how to encode `comexampleextension1`?
* Do all downstream clients need to be able to handle both `ce_timestamp` and `ce_string` format?
* Does the same issue apply to the other types?

This is somewhat of a generic problem with the CloudEvents extension model, but mentioning it again here.

jskeet: Two notes here on details, while I'm still looking at the broader conversation. (Sorry for parachuting in; a colleague pointed me at this issue, and this particular subthread touches on things I know about.)

While proto3 still doesn't have "required" semantics, it has very recently regained the idea of "optional" fields. The word "optional" is pretty much historical now, but it's really about "does this field support the notion of being implicitly set vs explicitly set". An optional int32 proto3 field that has been explicitly set to 0 by user code is serialized even though 0 is the default value. A regular int32 proto3 field will never have its default value serialized, whether it was set by the user or not.

Separately, on time zones, I'd like us to be really clear on terminology. RFC 3339 does *not* allow the time zone to be conveyed. It allows a UTC offset to be specified. They're not the same thing, and that affects scheduling. If I send an event with timestamp that is in UTC+1 (as the UK is right now), and I promise to send another event one day later in local time, you can't tell whether that will be another timestamp with an offset of +1 or not. It's unfortunate that RFC 3339 (and ISO-8601 before it) conflates UTC offsets and time zones, but let's try not to in any CloudEvents documentation :)

JemDay:
@evankanderson .. A decision was taken prior to 1.0 **not** to encode type information for attributes, so in your example if an intermediary were to translate between JSON and Proto formats the `comexampleextension1` attribute can only be interpreted as a `String` and as such your first encoding example would be correct. This a general issue with custom extensions when attempting to move between different formats, or indeed from binary->structured transport representations.

I'm not clear on the thrust of your question regarding _downstream clients_ - I would assume there's some sort of negotiation prior to receiving events, if i can't handle events in proto format (or avro) then i wouldn't subscribe to them. if i understand proto then i would understand the type system it exposes. If i were an SDK writer and i exposed a 'setAttribute(type, value)' type method a protobuf encoder can retain the type information.

JSON <-> Proto Format Bridge - I think this would be _doable_

I don't believe the _type issue_ is specific to proto, it also exists in a pure JSON model where custom extensions are present, the receiver only sees a `string` it can't opine on it being a URI, a timestamp, etc.

I'll comment on the payload concern seperately...

JemDay:
@alexvanboxel .. Re your comments

1. - What you describe was my original submission, conversations lead to the current model where only required attributes are formally represented. Primarily this was to allow new optional attributes to be added in future minor spec revisions and supported in a consistent fashion.

2. - I think it's more idiomatic to support variants representations in this fashion to simplify SDK development. I don't see this is any different from the JSON Format supporting data payloads of binary (in base64), string, or Object. We could have required JSON data payloads to be encoded as a string but chose to be idiomatic in that format.

3. - I may be missing the point but are you suggesting that instead of an explicit `one_of` you'd prefer data to simply be an `Any` which would then natively support bytes, String, or indeed a proto message ? I can buy into that if that's the thrust of your comment ;-)

alexvanboxel:
> @alexvanboxel .. Re your comments
> 
> 1. * What you describe was my original submission, conversations lead to the current model where only required attributes are formally represented. Primarily this was to allow new optional attributes to be added in future minor spec revisions and supported in a consistent fashion.

Yee, but a lot of useful fields are optional and accessing them differently would be cumbersome if you're making a Proto specific CloudEvent system. I would vote to make all "defined" fields are native 

> 2. * I think it's more idiomatic to support variants representations in this fashion to simplify SDK development. I don't see this is any different from the JSON Format supporting data payloads of binary (in base64), string, or Object. We could have required JSON data payloads to be encoded as a string but chose to be idiomatic in that format.

well string fits into the proto binary (see answer in (3)) and message as well. So I would stick to binary for the data field. You can do anything with that and you don't have funky base64 encoding in strings if you do it the other way around. For the e

> 3. * I may be missing the point but are you suggesting that instead of an explicit `one_of` you'd prefer data to simply be an `Any` which would then natively support bytes, String, or indeed a proto message ? I can buy into that if that's the thrust of your comment ;-)

- messages can be packed into a binary
- string and bytes are compatible as long as the bytes are valid UTF-8
ref: https://developers.google.com/protocol-buffers/docs/proto3#updating

And having numbers 1 and 2 shuffled into the message would make it compatible with Any (https://developers.google.com/protocol-buffers/docs/proto3#any). That would make it:

```
message CloudEvent {

  // -- CloudEvent Context Attributes

  // Required
  string id = 3;
  string source = 4; // URI-reference
  string spec_version = 5;
  string type = 6;

  // *** yes, they are optional... but well know ***
  datacontenttype = 7;
  string dataschema = 1; // <-- notice # 1, so it aligns with Any
  string subject = 8;
  google.protobuf.Timestamp time = 9; // <-- or string if that better fits the spec
  bytes data = 2; // <-- notice # 2, so it aligns with Any

  // Optional & Extension
  repeated CloudEventAttribute attributes = 5;

  /**
   * The CloudEvent specification defines
   * seven atrribute types...
   */
  message CloudEventAttribute {
    string attribute_name = 1; 
    oneof attr_oneof {
      bool ce_boolean = 2;
      sfixed32 ce_integer = 3;
      string ce_string = 4;
      bytes ce_binary = 5;
      string ce_uri = 6;
      string ce_uri_reference = 7;
      google.protobuf.Timestamp ce_timestamp = 8;
    }
  }
} 
```

drtriumph:
I think the realignment of attribute order so a CloudEvent is a valid Any is clever and a really useful feature to be able to use existing tooling for unpacking typed payloads without adding nesting in the definition - I think it will make handling these messages much cleaner and allow binary, proto and string events to all be represented concisely within one definition. My concern is that `type.googleapis.com/_packagename_._messagename_` ([spec](https://developers.google.com/protocol-buffers/docs/proto3#any)) doesn't meet the requirements for `dataschema` - it needs to be a `URI` (i.e. include a scheme):

```
URI - Absolute uniform resource identifier.
String encoding: Absolute URI as defined in RFC 3986 Section 4.3.
```

Which I think means we need a separate field to be field number `1` (mapping to an extension context attribute in other representations). It's an interesting argument for well known types being top level attributes, but the concern would be (as @evankanderson described earlier) if a popular optional extension context attribute got promoted to optional attribute (or a new attribute got used before a new proto spec was created / deployed) - that would mean application code would need to check both the top level field and the attributes map. A hybrid approach where `data` is the only standard optional field and we have a proto-specific `type_url` field might be the way to go here.

Finally, I think we should adhere to [proto style](https://developers.google.com/protocol-buffers/docs/style#message_and_field_names) and use underscores in names. My style preference would be to write out the whole `cloud_events` for specific attributes for clarity, and to stick with the built-in map so that the tooling protects us from repeats of the same key and converts the message into a map or dict language-specific implementation for you without requiring everyone to range through the attributes and create the maps themselves:

```proto
syntax = "proto3";

package io.cloudevents.v1;

import "google/protobuf/timestamp.proto";

// next tag: 8
message CloudEvent {
  // Required Atrributes
  string id = 3;
  string source = 4;
  string spec_version = 5;
  string type = 6;

  // Optional Proto-Specific Attributes
  string proto_type_url = 1;
  bytes data = 2;

  // next tag: 8
  message CloudEventAttribute {
    oneof attribute {
      bool cloud_event_boolean = 1;
      sfixed32 cloud_event_integer = 2;
      string cloud_event_string = 3;
      bytes cloud_event_binary = 4;
      string cloud_event_uri = 5;
      string cloud_event_uri_reference = 6;
      google.protobuf.Timestamp cloud_event_timestamp = 7;
    }
  }

  // Optional and Extension Attributes
  map<string, CloudEventAttribute> attributes = 7;
}
```

JemDay:
Thanks for all the comments ... apologies for the slow response.

It sounds like we're pretty much aligned on the attribute representation, I will make the change to use the fully-qualified `cloud_event_ `prefix as suggested.

That leaves the discussion around data representation, would it be simpler to just say..

`google.protobuf.Any data = 8;`

ie .. removing my `one_of` construct with a more native one which, in a weird way, is more inline with the JSON format that allows for any valid JSON type to be carried.

I believe this would remove the need for additional 'type_url' properties as those wouldn't then be needed.

Can I get some +1's on this before I update all the artifacts accordingly.

cbraynor:
Perhaps the alignment of fields 1 and 2 to match an Any is too clever / unexpected / magical. Your simpler suggestion means you need your own separate proto definition to use this at all (rather than just throwing bytes or a string in there) but that's probably the 80% use-case and not too cumbersome if you're already working with protos / likely gRPC

JemDay:
@cbraynor - I believe there should be a schema to ensure interop, i may well have over-simplified now i think of it as an `Any` can't carry scalar types if my understanding is correct (i might be wrong)

So .. maybe the (2nd) compromise here is to move to ...

 ```
 // -- CloudEvent Data
  google.protobuf.Any proto_data = 6;
  bytes binary_data = 7;
```

Then by convention if the event data is json/xml/image then it gets put in `binary_data` and if it's a proto object it goes in `proto_data` (and maintains the proto type information).

This is similar to the original `one_of` suggestion but usage is enforced by convention not through the schema itself.

**Comment by cbraynor** [comment]
I'm interested in hearing what others think here - a single field and the requirement to define your own message for the payload (even if it's trivial) with marshalling to / from `google.protobuf.Any` should meet all technical requirements and not add too much overhead.

It's a little cumbersome but so too is handling a `oneof`, or handling the undefined behaviour of if both `binary_data` and `proto_data` have content, and the Any field alignment suggested is perhaps too undiscoverable.

**Comment by JemDay** [comment]
In the interest of transparency i really prefer the `one_of` model as it forces correctness, without it message exchanges become brittle IMHO. As you note the handling becomes potentially non-deterministic with the most recent proposal :-(

I'm talking myself into moving back to the `one_of` representation ;-)

**Comment by JemDay** [comment]
@cbraynor, @evankanderson - Do you guys want to jump on a call to discuss this, probably more efficient than going back-and-forth here. I could do 08:00 (Pacific) on Thursday (July 2nd) if that would work.

**Comment by cbraynor** [comment]
Sorry, I just realized I missed this. I have no authority in this repo and I'm generally happy with the latest few options that have been proposed.

I think we just need to make a decision and work out how to get the maintainers to drive this forward and get agreement. Let me know if there's anything I can do to help

**Comment by evankanderson** [comment]
Sorry, I've been on vacation. I think it's fine to either use `one_of` with Any and bytes, or simply use bytes and have the recipient of the message use the dataschema (possibly with a "proto://" scheme or the like) to determine the type _if_ the payload is a protocol buffer.

Efficiency should be about the same, because Any should be encoded as a string + bytes, so that's two length-delimited messages unless the client actually asks for a decoding of the payload.

My slight preference would be for dataschema + bytes, but neither format seems particularly onerous.

**Comment by hrasadi** [comment]
I am not sure if a new decision have been made about the oneof data field or no, but we tried to build a prototype using the proposed proto recently and realized that using protobuf.Any has a lot of drawbacks. Note that protobuf.Any can only be used as a wildcard if the concrete proto file is available (and compiled in the same buildable unit). Otherwise, the proto.Any has no advantage to a byte[] blob since the SDK cannot parse it.

In order to support structured messages, we cas as well consider using protobuf.Struct, as opposed to protobuf.Any. With Structs, we basically forgo some of the features of the protobuf but guarantee that CloudEvent intermediaries that rely on the proto representation can handle the structured format.

**Comment by JemDay** [comment]
@cbraynor / @evankanderson - Thanks for the comments, I've moved back to the one_of construct as I believe it's more explicit and shouldn't add much processing overhead.

@hrasadi - My assumption is (and I'm open to being swayed) that if you need to decode/encode a proto data payload then you probably would need to understand the schema - correct me if I'm wrong, I've never used protobuf in a scenario where the messages I was dealing with were unknown.

The schema of the data proto only needs to be be available at time of "unpack()" so I believe an intermediary can accept the CloudEvent message and introspect the attributes etc without any knowledge of the schema. 

Similarly, and SDK writer can handle the marshalling/unmarshalling of the structured payload without knowledge of the underlying data schema. An SDK that support dprotobuf would probably also expose the ability for application code to register the various data schemas it understood allowing the the _unpack()_ to occur as well.

**Comment by JemDay** [comment]
@alexvanboxel - Ack, will take a stab at that..

**Comment by hrasadi** [comment]
I see that there is no activity on this PR for a while. Is it blocked by anything? I was hoping that we can start writing some code soon that uses this spec.

**Comment by duglin** [comment]
@hrasadi I think people are just swamped with other stuff (work and vacations) right now. But it is on their todo list to wrap it up

**Comment by hrasadi** [comment]
Thanks Doug for the update. Sounds good.

**Comment by evankanderson** [comment]
> @hrasadi - My assumption is (and I'm open to being swayed) that if you need to decode/encode a proto data payload then you probably would need to understand the schema - correct me if I'm wrong, I've never used protobuf in a scenario where the messages I was dealing with were unknown.

I've used proto in many cases where intermediaries didn't know the full proto definition ("unknown messages"). It actually works quite well for data pass through (as this was part of the design). It does not work well for message enrichment. It's possible to do this by passing along a descriptor as well as the message, but I'd look elsewhere if you actually need to enrich messages whose schema you don't know in advance. All the use cases I've seen simply pass along the unknown fields, so the final recipient who has the schema can unpack them.

> The schema of the data proto only needs to be be available at time of "unpack()" so I believe an intermediary can accept the CloudEvent message and introspect the attributes etc without any knowledge of the schema.

The intermediary can accept the message and read the CloudEvents attributes, because it knows the outer schema. I wouldn't recommend this pattern, but it's possible to do define two different proto messages, one for the outer envelope and one for envelope + payload:

```proto
message EnvelopeOnly {
  string key = 1;
  int counter = 2;

  // NOTE: tags 10-100 are reserved!
}

message EnvelopeAndBody {
  string key = 1;
  int counter = 2;

  OtherMessage payload = 10;
  AnotherMessage metadata = 11;
}
```

An intermediary could build in the `EnvelopeOnly` definition, parse the message and act on `key` and `counter` and not be aware of or try to validate `OtherMessage` or `AnotherMessage` at tags 10 and 11. If the same proto message was sent along further, these unknown fields would be serialized and passed on to the next recipient, which could unpack them if they had the `EnvelopeAndBody` message definition.

> Similarly, and SDK writer can handle the marshalling/unmarshalling of the structured payload without knowledge of the underlying data schema. An SDK that supported protobuf would probably also expose the ability for application code to register the various data schemas it understood allowing the the _unpack()_ to occur as well.

I'm assuming this would be via some sort of registration argument, i.e.

```python
@protobody(protodef.OtherMessage)
def handle(event):
  // event.payload is of type protodef.OtherMessage
```

```java
public void handle(CloudEvent<protodef.OtherMessage> event) {
  // Not sure if this works because type erasure is awesome...
```

```golang
func Handle(ctx context.Context, payload protodef.OtherMessage) error {
  // uses reflection to DTRT
```

spec.proto
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * The CloudEvent specification defines
+   * seven atrribute types...
+   */
+
+  message CloudEventAttribute {
+
+    oneof attr_oneof {
+      bool cloud_event_boolean = 1;
+      sfixed32 cloud_event_integer = 2;
```
evankanderson, line 47: Not sure if you need/want `sfixed32` here -- this is a fixed-length encoding rather than a varint, which is typical. This means it will always be 4 bytes long, rather than being 1-5 bytes long depending on content. If integers follow a power law distribution, `int32` will almost certainly be more compact; if you have negative integers more than about 20% of the time, then `sint32` will probably be more compact.

spec.proto
```diff
@@ -0,0 +1,55 @@
...
+      string cloud_event_string = 3;
+      bytes cloud_event_binary = 4;
+      string cloud_event_uri = 5;
+      string cloud_event_uri_reference = 6;
+      google.protobuf.Timestamp cloud_event_timestamp = 7;
```
evankanderson, line 52: I'd remove the `cloud_event_` prefix from these fields -- since it's inside a CloudEvent, I worry that you'll get ReallyLongNames.With.ReallyLongStutters.

protobuf-format.md
```diff
@@ -0,0 +1,222 @@
...
+
+    // String data
+    string text_data = 3;
+
+    // Protobuf Message data
+    google.protobuf.Any proto_data = 4;
+}
+```
+
+* Where the data is a protobuf message it MUST be stored in the `proto_data` property.
```
hrasadi, line 121: @evankanderson I think part of my concern is caused by this clause. If a passthrough intermediary (that doesn't know about the target type) receives the message, what's the point in "trying" to populate the 'proto_data' property? Why not just populate the 'binary_data' with the bytes of the serialized proto and pass it on?

I understand the previous comment @JemDay made about natively working with protos and I like that idea. However, I think the spec should allow intermediaries to opt-out when the concrete type is not known.

spec.proto
```diff
@@ -0,0 +1,55 @@
...
+
+  // -- CloudEvent Context Attributes
+
+  // Required
+  string id = 1;
+  string source = 2; // URI-reference
+  string spec_version = 3;
+  string type = 4;
+
+  // Optional & Extension
```
hrasadi, line 28: Why treating optional attributes similar to the extensions? How about we define optionals using proto optional fields (e.g. 'optional string subject = 5')?

nirmal-mehta: Agree with hrasadi@ in principle. 
I understand that with the latest proto spec there's no distinction between required vs optional fields and strictly speaking all fields are optional to set. However, I am not sure if that warrants abandoning the proto spec for fields that are 'optional to set' and instead put them inside one big map.

I think the way to model CloudEvents spec in proto would be that the fields that are 'optional to set' can have empty (default) values when a consumer reads that field. A writer can continue to not set an optional field if it doesn't need to (or when the value is empty).

// Required
string id = 1;  // must not be empty
...

// Optional
string subject = 5;  // empty, when not set
...

In such an implementation, there's no difference between when `subject` wasn't set and when `subject` was explicitly set an empty value (though writers shouldn't do that). In other words, empty value for `subject` means only one thing - an invalid value and not an actual subject. I believe, this will be in line with CloudEvents spec; at-least in principle.

jskeet:
> I understand that with the latest proto spec there's no distinction between required vs optional fields

That's not quite true any more - proto3 recently added the ability to explicitly state that a primitive field is optional. An optional field that is set to the default value will still be serialized, and consumers can detect whether a field has been explicitly set.

So it's not really "required vs optional", but "with or without the ability to detect field presence". But the "optional" keyword was reused for simplicity.

spec.proto
nirmal-mehta: Per jskeet@ response on the main thread: 
Apparently, `optional` keyword is being reintroduced in proto3 for the sake of field presence check. So, I think it is even more compelling to have optional fields (like `subject`) directly in the proto as opposed to putting them in the `attributes` map.
https://github.com/protocolbuffers/protobuf/blob/master/docs/implementing_proto3_presence.md

JemDay:
HI All, yet again i've not been able to give this thread the attention is deserves, i will start walking through all the commentary.

Given the amount of interest in this topic i'd be happy to setup a conf. call to discuss (which is probably more time-efficient than this method) or we could try and get some blocked time at a future WG call - yet again i find myself triple-book this week.. :-(

spec.proto
JemDay: Ack .. I think i was going for _predictability_ at the time, _efficiency_ is probably a better driver..

spec.proto
JemDay: Ack

spec.proto
JemDay: The concern here has really been around how the schema should evolve, it's possible that a future minor version of the spec might introduce a new 'optional' attribute. Formally declaring the 'required' CE attributes and then carrying the optional ones in the map structure seemed the least brittle way to allow evolution.

protobuf-format.md
JemDay: This seems to be my most contentious proposal ;-) It becomes relevant when moving between formats (eg JSON <-> Proto).

Assuming a JSON formatted CloudEvent (containing a protobuf data payload) is received by an intermediary that wants to output a CE in proto format ...

The required attributes map directly (and hopefully simply).
- The received CE has a `datacontenttype` of `application/protobuf`
- The received CE has a `data_base64` containing the encoded serialized representation.
- The received CE has a 'dataschema` of something like `proto:io.cloudevent.examples.SomeData` (it needs to be a URI)

The intermediary decodes the data_base64 and that becomes the `Any.Builder.setValue()`, the path portion of `dataschema` is passed into `Any.Builder.setTypeUrl().

I believe that works ...

spec.proto
hrasadi: Thanks for the reply. I guess my question is why not introduce it now? 

Totally unaware of other risks that you might have in mind, having the optional attributes as first class members of the proto improves the usability and reduces future toil of migration etc.

I think you mentioned there will be a conf. call. I am happy to provide more context there.

nirmal-mehta: Hi @JemDay, Thank you for moving this change forward!

Why is storing optional attributes in a map least brittle way? What's wrong with using the proto `optional` keyword? As mentioned in the last comment, it sounds like there's a clean and future-proof way of specifying optional attributes in the proto schema. So why not use the proto features as opposed to bypassing them and storing everything (optional) in a map.

My concern is that we are circumventing the proto spec and making it hard for every proto user that may want to consume CloudEvents. Rather than using proto generated methods, the user will have to access a loosely-typed map. Also, users will have to remember the right string values for the attribute names - which are not listed anywhere in this spec.

evankanderson: The set of CloudEvents attributes is unbounded -- while the spec defines certain attributes, it also leaves open both public and private extensions which might be standardized later.

With respect to proto, there's no consistent requirement for registering tag values for these extensions (and since this would be a proto-unique requirement, I think @duglin has been reluctant to require this -- it also makes generic translation bridges between other formats and proto more difficult), so moving an extension from "allowed but not specifically known to proto" to "known to proto and assigned a tag number" is a **breaking** wire-format change.

Consider the case of promoting a "signature" attribute from extension to core. If we had a definition like so:

```
message CloudEvent {
  string datacontenttype = 5;
  // etc
  map<string, CloudEventAttribute> extensions = 12;
  // new standard attributes!
  bytes signature = 13;
}
```

Consider what recipients who need to handle messages from old and new senders will need to do:

```
// Oh look, we wrote our own accessor wrapper on top of Proto!
byte[] GetSignature(CloudEvent ce) {
  if ce.getSignature().Length() > 0 {
    return ce.getSignature();
  }
  return ce.getAttributesOrDefault("signature", new CloudEventAttribute()).getCloudEventBinary();
}
```

And the sender needs to encode the data in **both** locations if it doesn't know if all recipients have been upgraded yet:

```
// We also need to provide our own setters, which we might choose to do anyway to get type-safe attribute setters.
void setSignature(CloudEvent ce, byte[] signature) {
  ce.setSignature(signature)
  // Also set "signature" in extensions for clients who aren't aware of the signature field
  // Note that this doubles the storage size of these attributes until everything in the
  // ecosystem has upgraded. It also introduces a source-of-truth problem if the two
  // fields somehow ended up being set to different values.
  ce.putAttributes("signature", CloudEventAttribute.newBuilder().setCloudEventBinary(signature).Build());
}
```


@JemDay -- reading the above code, I think you may want to name `map<string, CloudEventAttribute>` as `attribute` rather than `attributes`, so that the code reads more pleasantly: `ce.putAttribute("signature", sig)` seems better than `ce.putAttributes("signature", sig)` for setting a single attribute.

JemDay: @evankanderson

Thanks, i was about to embark on crafting a similar explanation and you saved me some work :-)

I'll tweak the naming, the resultant code does read better - the proto map name might not!)

JemDay:
@hrasadi, @jskeet, @nirmal-mehta - Does this work for you now?  @evankanderson's [earlier response](https://github.com/cloudevents/spec/pull/626#discussion_r478616801) provides some good context around the reason for _maps_ vs _named properties._

Is everybody able to make the WG session on Thursday (Sept 2nd) to discuss or would people prefer a separate chat.

jskeet:
@JemDay: I won't be able to make the WG session. I'll try to have another look at this during the week.

hrasadi:
> @hrasadi, @jskeet, @nirmal-mehta - Does this work for you now? @evankanderson's [earlier response](https://github.com/cloudevents/spec/pull/626#discussion_r478616801) provides some good context around the reason for _maps_ vs _named properties._
> 
> Is everybody able to make the WG session on Thursday (Sept 2nd) to discuss or would people prefer a separate chat.

Thanks for the context. I understand the reasoning behind this decision but the UX we will get as the result of this doesn't spark joy in me! I will attend the WG this week to discuss this a bit more.

spec.proto
nirmal-mehta: @evankanderson 
Thanks for the details, this is very helpful! However, I disagree with the conclusion.

First, note that I was only asking to take the standard optional attributes out of the map not extension attributes that can be different for different providers. Leaving extensions attributes in the map LGTM.

My argument is twofold:
1. I challenge the premise that replacing an extension attribute with a standard attribute must be backwards compatible. I think that is impossible in real world because there's no semantic-consistency guarantee for an extension attribute named "foo" across providers.  
What is the rationale for this goal? And how will CNCF spec guarantee that an extension attribute "foo" has the same type and semantics across different event providers before graduating it to a standard optional attribute?
2. Even if the aforementioned is deemed as a worthy goal, merging the namespace of standard optional attributes and extension attributes by putting everything in a single map is not a good solution IMHO. This is because they have different scopes (extensions can be experimental or highly provider-specific). 

If an existing extension attribute "foo" is graduated to a standard attribute with the same name, how is the CNCF spec going to guarantee that the type and semantics of this new standard attribute is exactly the same as that of *all extension attributes named "foo"*? Otherwise, it'll force the providers to update their extension attribute (meaning a backwards incompatible change) - or opening the door for changes that can silently break existing user (meaning a silent change in semantics, format, etc). 

Overall, I still think that by bypassing proto spec we will end up accidentally punishing the proto users.

nirmal-mehta:
> @hrasadi, @jskeet, @nirmal-mehta - Does this work for you now? @evankanderson's [earlier response](https://github.com/cloudevents/spec/pull/626#discussion_r478616801) provides some good context around the reason for _maps_ vs _named properties._
> 
> Is everybody able to make the WG session on Thursday (Sept 2nd) to discuss or would people prefer a separate chat.

I added my response to @evankanderson. I am not sure if we are making the right trade-off here by putting everything in a map, but I may be missing something still. I'll attend the WG as well; I assume you meant Sep 3rd.

Thanks again for driving this change!

evankanderson:
Unfortunately, I have conflicts 830-915 and 915-1000 Pacific this Thursday.

Open-ended flexibility like http headers isn't a great fit for the proto tagging model. I see three flavors solutions:

1. Assign tags to all well-known attributes. This puts the onus on early adopters to migrate their existing working systems from string-typed keys to tag-based keys.

1. Define a fixed set of tags that will be converted, and close the door to future tag numbers for attributes. This means that access to attributes will depend on when the attribute was defined vis-a-vis finalization.

1. Don't use tag numbers at all for attributes; have only string typed keys. This is least efficient wire-protocol wise and least helpful to IDE tooling.

jskeet:
One thing to consider: library support for CloudEvents could help with any ambiguity between "is it in the map or is it a known attribute"... at least for the attributes which are known (and tagged) at library-generation time. (I assume any further attributes would be added at a spec revision boundary anyway; I think it's *probably* reasonable for a 1.0 library to not understand some attributes specified using tags with a 1.1 event - although that does make it an interoperability boundary, and we should consider that carefully.)

So, do we expect "helper" libraries to be part of the developer experience when it comes to proto-based CloudEvents, or would we expect developers to want to take the .proto file, generate code with protoc and work directly with that, and still have a great experience?

(My experience is that manually-written helper libraries can make a huge difference to the usability where additional semantics are desirable beyond the raw data model.)

I still need to do a more thorough PR review though.

protobuf-format.md
```diff
@@ -0,0 +1,220 @@
...
+* If the implementation determines that the type of the data is binary, the value
+MUST be stored in the `binary_data` property.
+
+## 4. Transport
+
+Transports that support content identification MUST use the following designation:
+
+```text
+   application/cloudevents+protobuf
+```
```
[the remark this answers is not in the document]
jskeet: I was assuming that this was the binary encoding, given that JSON is already covered.

I'd strongly discourage using textprotos - they're not supported in all languages, and not well-defined externally as far as I'm aware.

hrasadi:
> > @hrasadi, @jskeet, @nirmal-mehta - Does this work for you now? @evankanderson's [earlier response](https://github.com/cloudevents/spec/pull/626#discussion_r478616801) provides some good context around the reason for _maps_ vs _named properties._
> > Is everybody able to make the WG session on Thursday (Sept 2nd) to discuss or would people prefer a separate chat.
> 
> Thanks for the context. I understand the reasoning behind this decision but the UX we will get as the result of this doesn't spark joy in me! I will attend the WG this week to discuss this a bit more.

I also gave this a little bit more thought. I think making two changes to the spec might help:

1- Move the attributes to the bottom of the message, also support optional attributes to be tagged:

```
  // -- CloudEvent Data (Bytes, Text, or Proto)
  oneof data_oneof {
    bytes binary_data = 1;
    string text_data = 2;
    google.protobuf.Any proto_data = 3;
  }

  // Optional & Extension Attributes
  map<string, CloudEventAttribute> attribute = 4;

  // Required Attributes
  string id = 5;
  string source = 6; // URI-reference
  string spec_version = 7;
  string type = 8;

  // Optional attributes
  optional string subject = 9;
  // ...
```

The change ensures that continuity of tags and the makes it easier to maintain the proto overtime. But it doesn't solve the problem of promoting extensions to optionals. Read on.

2- Expand the spec as follows:

"An optional attribute SHALL be added in the `attribute` map, as a tagged field (if exists) or both. If a producer decided to populate an optional attribute both in the map and as a tagged field, the two values MUST be identical".

We should also explicitly callout that using this capability is discouraged and reserved only for the cases when new optionals are introduced (that have naming collision with some provider's extensions).

In doing so, we optimize our spec for the main use case and not the corner case, but still provides a path forward for (rare) cases where an extension is being promoted to a top-level field.

amrynsky:
Just started following the project because we are trying to adapt cloudevents in our system. We are using kafka with protobuf serde and come up with a similar proto schema. 
I agree with @jskeet that working directly with generated code is not the best dev experience because it exposes some protobuf internals (i.e. `pack`/`unpack` data for `Any`). In our case created several helpers to map POJO to protobuf objects and back but still not really happy with the solution.

jskeet:
@amrynsky: To be clear, I would still advocate for working with the "raw" proto generated code - but with augmentation. That's somewhat easier in C# due to partial classes than it is for other languages, admittedly. Wrapping everything can be a good solution in some cases, but you often end up with an impedance mismatch in my experience.

amrynsky:
@jskeet 
> you often end up with an impedance mismatch in my experience

agree, and I'm not really happy with our current solution because of this. At the same time in some cases with "wrapping" generated code, we could provide more clear API. For example, it allows us to make application-specific attributes first-class citizens and then put them into cloudevents extensions, or provide some constraints for data. Still playing with API to find the right balance.

JemDay:
Thanks for the healthy debate, we can discuss more at the call tomorrow but i'll go ahead and see if we can get a side-bar call setup that we can get better attendance at.

w.r.t the commentary around 'helper' libraries, i'm sure you're aware of the SDK work that's going on where the idea is to provide a mechanism to insulate application code from the quirks of the wire representation and transport. I'm no C# expert but their SDK model seems more flexible in supporting data payload representations beyond 'bytes'.

jskeet:
> w.r.t the commentary around 'helper' libraries, i'm sure you're aware of the SDK work that's going on where the idea is to provide a mechanism to insulate application code from the quirks of the wire representation and transport. I'm no C# expert but their SDK model seems more flexible in supporting data payload representations beyond 'bytes'.

Yes, and for C# I'd expect that we'd have a protobuf-specific package with a `ProtobufEventFormatter` for example. I'd expect that package to contain the generated CloudEvent proto representation too. I just didn't want to make too many assumptions about that being the case for *all* languages.

evankanderson:
> @amrynsky: To be clear, I would still advocate for working with the "raw" proto generated code - but with augmentation. 

My hope was that end-users would be able to use the proto-generated code directly too many gotchas, particularly since the current set of CloudEvents SDKs doesn't have support for nearly the [range of languages supported by protoc](https://github.com/protocolbuffers/protobuf/blob/master/docs/third_party.md).

JemDay:
HI .. If you'd like to discuss this proposal, please select a time [here that works for you](https://doodle.com/poll/v76rvbvdgpp265mp)...

JemDay:
Meeting scheduled for 11:00 PT, Weds Sept 9th .. Normal Zoom Id : https://zoom.us/my/cncfserverlesswg

JemDay:
@jskeet - Apologies for the timing, hopefully you can make it despite the timing. I'll endeavor to present your comments if you can't make it.

jskeet:
I may be able to come - I'll see. I'll try to take another pass and leave comments anyway.

protobuf-format.md
```diff
@@ -0,0 +1,227 @@
...
+[schema][proto-schema].
+
+## 2.1 Type System
+
+The CloudEvents type system is mapped to protobuf as follows :
+
+| CloudEvents   | protobuf |
+| ------------- | ---------------------------------------------------------------------- |
+| Boolean       | [boolean][proto-scalars] |
+| Integer       | [sint32][proto-scalars] |
```
jskeet, line 60: `sint32` is very rarely used within Google; we normally stick to `int32`. While `sint32` is more compact for negative numbers, it's (potentially) larger for non-negative numbers.

protobuf-format.md
```diff
@@ -0,0 +1,227 @@
...
+direct support of the CloudEvent [type system][ce-types].
+
+```proto
+map<string, CloudEventAttribute> attribute = 1;
+
+message CloudEventAttribute {
+
+    oneof attr_oneof {
+      bool ce_boolean = 1;
+      sfixed32 ce_integer = 2;
```
jskeet, line 83: This is in contrast to the `sint32` above. I'd suggest just `int32` though.

protobuf-format.md
jskeet: This part definitely needs more discussion - if the format needs to be able to store a timestamp *with offset from UTC* as per RFC 3339, `Timestamp` will not do that on its own. We could create our own message for that concept, but it wouldn't be as easy to handle - `Timestamp` has built-in goodness in terms of a readable JSON representation, and utility methods in at least some languages (e.g. C#).

I would suggest that in retrospect, the CE spec deciding that a timestamp doesn't *just* represent an instant in time may have been a mistake - but it's unclear what we need to do now. Being able to convert between formats without loss of information would certainly be a useful property...

protobuf-format.md
```diff
@@ -0,0 +1,227 @@
...
+      google.protobuf.Timestamp ce_timestamp = 7;
+    }
+}
+```
+
+In this model an attribute's name is used as the map *key* and is
+associated with its *value* stored in the appropriately typed property.
+
+This approach allows for attributes to be represented and transported
+with no loss of *type* information.
```
jskeet, line 97: Unlike the JSON format, presumably - where for an extension attribute (where we don't necessarily know the type inherently in the way that we do for the built-in attributes) we don't know whether a given attribute value is a string that also happens to be a URI but shouldn't be treated as one, or is *really* intended to be a URI. (Hope that makes sense...)

I think we're missing some detail in terms of requirements of a CloudEvent format - when designing one, it would be nice to have guidance about what's required, desirable etc.

protobuf-format.md
```diff
@@ -0,0 +1,227 @@
...
+
+    // Binary data
+    bytes binary_data = 2;
+
+    // String data
+    string text_data = 3;
+
+    // Protobuf Message data
+    google.protobuf.Any proto_data = 4;
+}
```
jskeet, line 118: I think it's at least worth *considering* a field of

```text
Value json_data = 5;
```

`Value` is intended to be "any JSON" value; compare that with `Struct` which is "a JSON object".

If we're trying to convert from a JSON formatted event with a JSON payload to a Protobuf-formatted event with the same payload, `Value` would be a more "native" representation than `string` in my view. But we really need to work out the use cases here.

jskeet: (Note that this would avoid any confusion between a text payload that happens to be valid JSON, and one which was designed to be JSON.)

protobuf-format.md
```diff
@@ -0,0 +1,227 @@
...
+  ceBuilder
+    .setId(UUID.randomUUID().toString())
+    .setSpecVersion("1.0")
+    .setType("io.cloudevent.example")
+    .setSource("producer-2")
+
+    // Add the proto data into the CloudEvent envelope.
+    .setProtoData(Any.pack(dataBuilder.build()));
+
+    // Add the protto type URL
```
jskeet, line 198:
```suggestion
    // Add the proto type URL
```

protobuf-format.md
```diff
@@ -0,0 +1,227 @@
...
+    withAttribute(ceBuilder, "dataschema", ceBuilder.getProtoData().getTypeUrl());
```
jskeet, line 199: I'm not keen on `with` as a name - I'd prefer `set`. I generally expect a `with` method to return a new instance with the specified change applied, leaving the original object as it was before.

protobuf-format.md
```diff
@@ -0,0 +1,227 @@
...
+
+Transports that support content identification MUST use the following designation:
+
+```text
+   application/cloudevents+protobuf
+```
+
+## 5. Examples
+
+The following code-snippets shows how proto representations might be constucted asuming the availability of some convenience methods ...
```
jskeet, line 143:
```suggestion
The following code-snippets shows how proto representations might be constructed assuming the availability of some convenience methods ...
```

spec.proto
```diff
@@ -0,0 +1,55 @@
...
+  string id = 1;
+  string source = 2; // URI-reference
+  string spec_version = 3;
+  string type = 4;
+
+  // Optional & Extension Attributes
+  map<string, CloudEventAttribute> attribute = 5;
+
+  // -- CloudEvent Data (Bytes, Text, or Proto)
+  oneof data_oneof {
```
jskeet, line 32:
```suggestion
  oneof data {
```

I believe that will generate nicer code.

spec.proto
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  /**
+   * The CloudEvent specification defines
+   * seven atrribute types...
+   */
+
+  message CloudEventAttribute {
+
+    oneof attr_oneof {
```
jskeet, line 45:
```suggestion
    oneof value {
```

(Same comment as before about `_oneof` not being necessary... and there's no need for the "attr" side given that this is already within a `CloudEventAttribute` message.)

spec.proto
```diff
@@ -0,0 +1,55 @@
...
+    string text_data = 7;
+    google.protobuf.Any proto_data = 8;
+  }
+
+  /**
+   * The CloudEvent specification defines
+   * seven atrribute types...
+   */
+
+  message CloudEventAttribute {
```
jskeet, line 43:
```suggestion
  // The value of a CloudEvent attribute, with a separate oneof case
  // for each type defined in the CloudEvent specification.
  message CloudEventAttributeValue {
```

Note that I've specified that it's the value, because this isn't the whole attribute - an attribute has both a name and a value, and this is *only* the value part.

An alternative design would be to include the name here and make the `attributes` field a repeated field instead.

spec.proto
```diff
@@ -0,0 +1,55 @@
...
+  // -- CloudEvent Context Attributes
+
+  // Required Attributes
+  string id = 1;
+  string source = 2; // URI-reference
+  string spec_version = 3;
+  string type = 4;
+
+  // Optional & Extension Attributes
+  map<string, CloudEventAttribute> attribute = 5;
```
jskeet, line 29:
```suggestion
  map<string, CloudEventAttribute> attributes = 5;
```

(Old Google naming conventions used singular names even for repeated fields and maps... we've moved to plural names now.)

spec.proto
```diff
@@ -0,0 +1,55 @@
...
+package io.cloudevents.v1;
+
+import "google/protobuf/any.proto";
+import "google/protobuf/timestamp.proto";
+
+option java_package = "io.cloudevents.v1.proto";
+
+message CloudEvent {
+
+  // -- CloudEvent Context Attributes
```
jskeet, line 20: We should revisit all the comments. I'd expect a comment for every field and message.

protobuf-format.md
jskeet: (I note that it *is* `int32` in the spec.proto file.)

spec.proto
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+syntax = "proto3";
+
+package io.cloudevents.v1;
+
+import "google/protobuf/any.proto";
+import "google/protobuf/timestamp.proto";
+
+option java_package = "io.cloudevents.v1.proto";
```
jskeet, line 16: We should work out what we want in terms of options for other languages... which may partly depend on what level of support we expect in SDKs. For example, for C# I'd expect a NuGet package called `CloudNative.CloudEvents.Protobuf`, so it would be nice to have the `csharp_namespace` option for that here.

protobuf-format.md
```diff
@@ -0,0 +1,227 @@
...
+    // Protobuf Message data
+    google.protobuf.Any proto_data = 4;
+}
+```
+
+* Where the data is a protobuf message it MUST be stored in the `proto_data` property.
+  * `datacontenttype` MAY be populated with `application/protobuf`
+  * `dataschema` SHOULD be populated with the type URL of the protobuf data message.
+
+* When the type of the data is text, the value MUST be stored in the `text_data` property.
```
jskeet, line 125: It's not really clear what's meant by "the type of the data" here - does JSON data count as text by default, for example?

jskeet (commented):
Lots of comments ahead of the meeting.

More broadly, I think there's some higher level work to do in defining what we want a CloudEvent format to do:

- What use cases do we envisage?
- What are the requirements for each of those?
- What are the goals for particular event formats?
- How much can/should we lean on SDK support?

This also speaks to the concern I have with the CE spec in general in terms of data representation - the degree to which there's a separation of concerns between CE format and data type seems to vary, and I believe that's reflected in the various discussions that we've seen around SDK representations and operations on data.

spec.proto
```diff
@@ -0,0 +1,55 @@
...
+
+message CloudEvent {
```
jskeet, line 18: Should we have a major version somewhere? I can imagine CE v2 having a different set of required fields to CE v1, for example. Do we have a document around how the spec is envisaged to evolve?

jskeet: (Resolving - it's in the package name.)

hrasadi:
Thanks for setting up the meeting @JemDay. Based on what I have learned, I am convinced that the current proposal for extension and optional attributes placement is the best path forward. Thanks for driving this.

nirmal-mehta:
Based on offline sync with @JemDay, @clemensv and @hrasadi, keeping all non-required attributes in a map is fine by me. 

----- rest of the message is an optional read -----
However, I think it is less than ideal for the following reasons:

Thanks to @clemensv for clarifying this: 
There are three kinds of non-required attributes involved here - standard optional attributes, common extension attributes (meant to common across providers/vendors) and provider-specific extension attributes. Currently, they are all in the same map.

1. I don't think we can realistically graduate provider-specific extension attributes to standard attributes. See my previous comment in spec.proto for more details.
2. By putting standard optional attributes and common extension attributes in the same map, we are already removing any distinction between those two. So, if we were to put both standard optional and common extension attributes as proto fields then we would achieve the same effect as putting them in the same map.
3. I think there's a tangible value in using proto fields as that would provide a better, streamlined use experience (named fields, auto-generated methods, etc). And the provider-specific attributes can continue to be in the map.

evankanderson:
I'm terribly sorry for missing the meeting today -- I don't have the CloudEvents meetings on my calendar, and didn't get a direct invite, so completely forgot that the meeting was happening. I'm satisfied with the outcome in any case; I was asked for an opinion with respect to proto, but I suspect many of you will have to live with these decisions at least as much as I will.

protobuf-format.md
evankanderson: (I suggested the switch to `int32`, but I think it only got corrected in the `.proto` file. It seems like we should make sure that changes in the `.proto` get reflected back into the text document.)

protobuf-format.md
evankanderson: This is `int32` in the proto as well.

spec.proto
JemDay: Hey Jon; I only have working knowledge via Java, if there are other language specific annotations that should be added we should do that earlier rather than later.

JemDay:
@evankanderson - Sorry you didn't see the invite, I sent it out over as many channels as I could think of :-(

protobuf-format.md
JemDay: Ack

protobuf-format.md
JemDay: Ack

spec.proto
JemDay: Ack

spec.proto
JemDay: We moved from plural to a singular name as the generated code read more naturally ... I'm happy either way, I think we just need to pick one.

If plurals are now part of a style-guide then we should switch back.

@evankanderson

spec.proto
JemDay: Ack ... Will address...

spec.proto
evankanderson: I'm fine either way.

evankanderson:
> @evankanderson - Sorry you didn't see the invite, I sent it out over as many channels as I could think of :-(

This was totally my fault. In any case, it seems like you all got to a good conclusion, so I'm happy to let you all get on with the work. Having this will be helpful for some demo ideas I've got floating around. 😁

spec.proto
jskeet: SGTM. Here's an example with Java, Go, Ruby, PHP and C#. (I chose this API as it has a name that should be capitalized in the middle, for DataCatalog):

https://github.com/googleapis/googleapis/blob/master/google/cloud/datacatalog/v1/common.proto

I'm not sure what would be needed for other languages.

jskeet:
Btw, we didn't talk about timestamps on Wednesday (at least, not before I had to leave) and whether we need to be able to store UTC offsets. We should keep discussing that.

JemDay:
I've added some language specific package options as-per the example from @jskeet - i'm not conversant in some of these languages so if anybody can sanity check that would be good.

@jskeet - On the subject of _Timestamp_, my understanding is that the specification simply demands a time instant. I don't believe there's any requirement that any original offsets be retained across processing boundaries so long as the semantic is retained.

jskeet:
@JemDay: The specification isn't as clear on this as it might be. The description is "Date and time expression using the Gregorian Calendar" - which is interesting given that the calendar system is actually irrelevant for an instant in time until you choose a text representation. The text representation is specified to be RFC 3339, which does allow for UTC offsets.

I'm happy to create a PR to clarify the specification here, in terms of:

- It's just intended to be an instant in time
- Any UTC offset present in a local representation need not be preserved across processing boundaries
- We should specify a required level of precision (milliseconds?) but again allow for more information to be present-but-not-necessarily-preserved

Once all of those are in place (or at least agreed to in principle), I'm fine using Timestamp from protos.

Does that sound like the kind of PR that would be welcomed?

JemDay:
@jskeet - I think this a specification clarity issue, believe it is meant to be an "instant in time", and that we follow RFC 3330 when encoding into a string. Might be simpler to discuss it tomorrow in the WG call before doing a PR (I could event make it part of this one).

With respect to your suggestion regarding using the 'Value' type as a more first-class way to support JSON data; I understand the model but I'm unclear on how this would get realized in code. Looking at the Java language spec I don't see an option to fo something like `mergeFrom(javax.json.JsonValue jsonValue);` on the Value builder. Without these sort of constructs I'm not sure what we gain.

jskeet:
@JemDay: Timestamps - I'll try to find a WG that I can come to, but it definitely won't be tomorrow. (The WG timing tends to be clash with all kinds of things for me - which is to be expected for an international meeting, of course.)

For the `Value` part supporting JSON data - based on last week's call, I don't think it's worth doing.

JemDay:
@jskeet : Thanks, i believe we're aligned.

I chatted with @clemensv and there is no requirement for the offset information to be propagated, i'll try tweaking the spec to clarify that - so we're good-to-go with the `Timestamp` usage as defined.

jskeet:
@JemDay: Excellent. Please cc me on the PR; as a date/time nerd I'm happy to be picky :)

JemDay:
All - I believe we've reached consensus so could I asked for a few LGTM to allow this PR to be merged and closed. I will open a separate PR to clarify the expectations related to _Timestamps_ in the specification, that does not impact this proposal.

spec.proto
jskeet: Ideally it would be good to get this done before it's merged, but I'm okay for us to do it later.

spec.proto
jskeet: Ping on this one.

spec.proto
jskeet: And again.

jskeet (commented):
There are a few things I'd still like to see addressed - but if others want to get the first draft in and then iterate, that's fine by me. It's still only going to be a draft, right?

JemDay:
@jskeet - I believe your concerns are covered now. I re-pluralized the 'attributes' and removed the c# directive, we can add that in as a seperate PR. I'm no C# expert but i believe the major version must be encoded into the package to enable V1 and V2 generated code to live side-by-side. That packaging should be independant (but not clash with) the C# SDK packages.

spec.proto
jskeet: Still no sign of this or the later oneof being fixed.

jskeet:
Definitely happy to sort out the C# namespace later, and I agree that we should include the version in the namespace. Not entirely convinced about it being independent of the SDK, but we can decide that a little later.

duglin:
@jskeet are ok with this PR as-is? It sounds like there might be some follow-on discussions, but I wanted to make sure we didn't merge it if there are still things that need to be resolved first.

jskeet:
@duglin: I would *prefer* to get the oneof names nicer to start with - is there any controversy about those? I don't have any fundamental objections though.

duglin:
I'll need @JemDay to speak to that :-)

JemDay:
@jskeet - Can you suggest alternatives ... we need to differentiate between a "binary", "text", and "Proto Message" data payload.

jskeet:
@JemDay: All I'm suggesting is removing the `_oneof` suffix from `data_oneof` and `attr_oneof`. Keep the oneofs themselves - it's just the name that will be nicer without the suffix.

evankanderson:
Note that if you're only talking about binary Protobuf representation, then changing the name of `data_oneof` and `attr_oneof` is a wire-compatible change, since the name of the oneof is not transmitted and is only an affordance for library code.

This also means _those_ naming fixups could be made in a subsequent PR if desired.

jskeet:
I'm assuming *anything* can still be changed in the near term - merging this isn't defining a 1.0 for the protobuf representation is it? Because if it is, that should include source compatibility as well, which *would* require the names to be finalized.

evankanderson:
There are two different things -- this is defining the wire-format protocol; there are separately per-language SDKs. The Golang SDK, for example is at v2.2.0 supporting the v1 and v0.3 wire-format specs.

jskeet:
> this is defining the wire-format protocol

Well it's defining a .proto file, which I'd *expect* to be the canonical source for SDKs. It would be pretty silly for someone to create a wire-compatible-but-separate proto file. So why not aim to get both right to start with?

jskeet:
(I'd also suggest that we should determine that the JSON serialization format of the protobuf data is also standardized - that doesn't affect oneof names as far as I remember, but it *does* affect tag names, basically making tag renaming a breaking change.)

JemDay:
@jskeet - Tidied the names as suggested.

The JSON format has its own specification, we could tweak that if needed - I don't think we did that for Avro.

jskeet:
I'm not suggesting a change to the standard JSON CloudEvent format. I'm suggesting that we make sure that the proto representation doesn't change in any way that breaks source or proto3-JSON format, once we've declared it as 1.0. That's important because it should be reasonable for folks to use the proto3-JSON format in test data, and not be broken by changes.

I don't think it's a major imposition - it just means that we'd be standardizing *the .proto file* rather than *just* the binary wire format. I'm definitely up for discussing that, but it would be my recommendation.

Will double-check the oneofs and LGTM this PR though.

jskeet (approved):
I'm happy to have a pass at reformatting comments and adding more of them in a future PR if that would be helpful.

duglin:
We're in kind of a weird situation. The PR was approved on today's call with the assumption that all parties involved in the PR were ok with the current state. After a few more commits it looks like that is true. However, we technically didn't approve this one, which mean we should really wait. BUT, based on the lack of comments on today's call I suspect the people who really care are part of this PR's thread and not the weekly call. So, rather than waiting for next week's call I'm going to let this sit until EOD tomorrow to see if there are any lingering concerns - then I'll merge. 

Hope that's ok with everyone.

duglin:
ok ... sat long enough. Merging!!!  
Thanks to all, especially @JemDay , for all of the effort on pushing this thru!

Approved on the 9/24 call

Issue #622, opened 2020-05-07
[Discovery] Event Subscriber term is unused


duglin:
As of today this term is unused - if in a few weeks/months this is still the case then we should consider removing it. If not... we can then close this issue.

PR #631, opened 2020-05-26
Remove unused 'subscriber'

duglin:
Closes: #622 

Clean-up - Just removing a term that is unused - we can add it back later if needed.


duglin:
Approved on the 5/28 call

PR #632, opened 2020-05-26
Add pointers to CE terms in CE spec

### LINKED ISSUE #621: [Discovery] Terminology section - dups with CE
I think the reader might be helped if we could make it clear which concepts are new to Discovery and which are copied from CE just for easy reference. Two options come to mind:
1 - on the same line that shows each term add something like `[From CloudEvents]` with a link to the term in the CE spec
2 - have two sections - one for new terms and one for duplicated CE terms

I'm learning more towards 1 and I think it'll flow nicer.

We then should include text to say that each CE term that is copied into this spec is done so for convenience, and ease of reading. Any difference between the two docs is an error and the CE spec is the authoritative definition.

duglin:
Closes #621

And moved a few terms to order them in the order in which the event flows,
starting with the Service.  If people want alphabetical, I'd be ok with that
too.

### DIFF
--- a/discovery.md
+++ b/discovery.md
@@ -68,34 +68,41 @@ interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
 
 ### Terminology
 
+Note: some of the terms defined below are taken from the
+[CloudEvents](spec.md) specification, and are marked with a reference to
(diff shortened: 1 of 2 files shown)


duglin:
Approved on the 5/28 call

PR #636, opened 2020-05-28
Add property "sourcetemplate" to eventprovider/service

deissnerk:
This PR adds the property _sourcetemplate_ to the eventprovider that provides a means to describe how the `source` attribute of a service/eventprovider will look like. It is referring to [RFC 6570](https://tools.ietf.org/html/rfc6570) which is also referenced by the definition of parameters in OpenAPI.

### DIFF
--- a/discovery.md
+++ b/discovery.md
@@ -243,9 +243,7 @@ entity.
   - OPTIONAL
   - If present, MUST be a non-empty absolute URI
 - Examples:
-  - `http://cloud.example.com/docs/blobstorage`
-
-###### specversions
+  - "http://cloud.example.com/docs/blobstorage"
 
 - Type: Array of `Strings` per [RFC 2046](https://tools.ietf.org/html/rfc2046)
 - Description: CloudEvents [`specversions`](https://github.com/cloudevents/spec/blob/master/spec.md#specversion)
@@ -254,7 +252,7 @@ entity.
   - REQUIRED
   - MUST be a non-empty array or non-empty strings
 
-###### subscriptionurl
+##### subscriptionurl
 
 - Type: `URL`
 - Description: An absolute URL indicating where CloudSubscriptions `subscribe`
@@ -262,7 +260,7 @@ entity.
 - Constraints:
   - REQUIRED
 
-###### subscriptionconfig
+##### subscriptionconfig
 
 - Type: `Map` of `String` to `String`
 - Description: A map indicating supported options for the `config` parameter
@@ -275,7 +273,7 @@ entity.
 - Examples:
   - ??
 
-###### authscope
+##### authscope
 
 - Type: `String`
 - Description: Authorization scope needed for creating subscriptions.
@@ -285,7 +283,7 @@ entity.
 - Example:
   - `storage.read`
 
-###### protocols
+##### protocols
 
 - Type: `List` of `String`
 - Description: This field describes the different values that might be passed
@@ -315,7 +313,7 @@ entity.
   - `com.github.pull.create`
   - `com.example.object.delete.v2`
 
-##### description (Type)
+###### description (type)
 
 - Type: `String`
 - Description: Human readable description.
@@ -323,7 +321,7 @@ entity.
   - OPTIONAL
   - If present, MUST be a non-empty string
 
-##### datacontenttype
+###### datacontenttype
 
 - Type: `String`
 - Description: CloudEvents [`datacontenttype`](https://github.com/cloudevents/spec/blob/master/spec.md#datacontenttype)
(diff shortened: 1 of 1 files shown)


duglin:
If we add an attribute like this then this LGTM.

Do you think we'd need more info someplace to allow people to programmatically map the variable to some like a spec that defines what each variable means? Or should we just assume that the docs for the service will have that?

duglin:
would we need to place any restrictions on the characters used for the variable name?

deissnerk:
@duglin 
>Do you think we'd need more info someplace to allow people to programmatically map the variable to some like a spec that defines what each variable means? Or should we just assume that the docs for the service will have that?

There is one difference to OpenAPI and similar standards. In our case the template would only be used to document, how the producer fills the source attribute. In OpenAPI the template is also used to explain to an API consumer, how to fill in values for the variable. In our case it would be interesting to get some information, how to parse the source attribute to get again the values. This is not possible to derive from all URI templates. To my understanding it should be possible for level 1 templates. 
Specifying the possible variable values could still provide some benefits. Something like what is used in JSON Schema (description, examples etc.). 

> would we need to place any restrictions on the characters used for the variable name?

RFC 6570 describes which characters are supposed to be used in variable name [here](https://tools.ietf.org/html/rfc6570#section-2.3). I wonder, if we have a reason to be more restrictive. As the RFC describes 4 levels of templates, I wonder if we should restrict usage to level 1. On the other hand people should know what they do, if they decide to use really complex templates. I could add a sentence that it is RECOMMENDED to only use templates of level 1.

duglin:
Given the chat on today's call... should we consider moving `sourcetemplate` to be under `type` since it seems very possible that the source's value might change based on what the type of event is? e.g. bucket.create vs issue.create

deissnerk:
@duglin 
If you have multiple `services/eventproviders` that emit the same `types` but with different `sourcetemplates`, what would you expect to be returned by a REST call to `/types/{name}`

deissnerk:
One more thing:
I think subscription UIs and developer tooling could be a good use case for source templates. A UI could use the variables in the template and render input fields for them. It could then take the input and generate subscription filters. 
For this use case it would be good to specify the variables similar to how it is done in OpenAPI/AsyncAPI with type, constraints, description and examples.

duglin:
> @duglin
> If you have multiple `services/eventproviders` that emit the same `types` but with different `sourcetemplates`, what would you expect to be returned by a REST call to `/types/{name}`?

Hmm, not 100% sure what @mikehelmick had in mind for the `/types` query, but it says the return value of that if `/types/{name}/services`, which seems like it should be something like:
```
GET /types/github-pull

- type:
  - name: github-pull
    - services:
      - service: github.com
        description: github-pull
        specversion: 1.0
        subscriptionuri: https://github.com/subscribe
        protocols:
        - protocol: http
        types:
        - type: github.com/pull
          description: github-pull
          datacontenttype: application/json
          sourcetemplate: https://api.github.com/repos/{org}/{repo}/pulls/{id}
```
The real question is whether it should include just "github-pull" types or all types for the services that support "github-pull"? I'd say it should include all.

WDYT?

mikehelmick:
I think that source template would be specific to both the provider + the event type.

Two different blob storage providers might use the same event type, but have different formats for the `source` attribute.

Sadly, I don't think that this is a property that is tied exactly to the just the event type.

duglin:
thanks @mikehelmick 

Did my response to the `/types/{name}` query align with your thinking? I wasn't sure if having 'type' at two levels like that was correct.

mikehelmick:
`Service` and `Type` are intended to narrow down to the exact event type + service combo that you want and the `Producer` entity is meant to give you the information you need to actually get a subscription created.

If you run the queries specified here -> https://github.com/mikehelmick/discovery-prototype

I think it illustrates my thinking well. Applies equally well to a rest implementation.

duglin:
> I think that source template would be specific to both the provider + the event type.
> 
> Two different blob storage providers might use the same event type, but have different formats for the `source` attribute.
> 
> Sadly, I don't think that this is a property that is tied exactly to the just the event type.

I think my latest proposal #642 should address this by moving the 'type' under 'service' for all queries.

deissnerk:
@duglin I suppose the change introduced with #642 will solve the issue.

discovery.md
```diff
@@ -402,13 +415,13 @@ entity.
       "type": "com.example.storage.object.create",
       "specversions": [ "1.x-wip" ],
       "datacontenttype": "application/json",
-      "dataschema": "http://schemas.example.com/download/com.example.storage.object.create.json"
+      "dataschema": "http://schemas.example.com/download/com.example.storage.object.create.json",
+      "sourcetemplate":"https://storage.example.com/service/storage/{objectID}"
```
duglin, line 419: space after colon

duglin:
LGTM

duglin:
Approved on 6/25 call

PR #641, opened 2020-06-09
Draft of DS-Primer

duglin:
Direct copy from: https://gist.github.com/duglin/5920f706ee0ce608b0be1fa2e3669ce5

I took the liberty of includeing Klaus' sourcetemplate stuff since it
seems like people are ok with and I wanted to show what it would look like
when in use.

### DIFF
--- a/ds-primer.md
+++ b/ds-primer.md
@@ -1,0 +1,168 @@
+# Discovery and Subscription Primer - WIP
+
+## Abstract
+
+This non-normative document provides an overview of the Discovery and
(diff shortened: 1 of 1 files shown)


duglin:
ping @deissnerk 

All,just a first draft but feel free to suggest additional text/use-cases. We can always add more via additional PRs, but if you feel like there are some that need to be there from day-one, let me know.

duglin:
Approved on the 6/11 call

PR #642, opened 2020-06-11
Discovery Revamp

duglin:
- Restructued things so that it didn't feel like we were repeating the
      definition of the attributes. For example, we now just define Service
      and its attributes once and then simply point back to that, and its
      serialization, from the Query section
- got rid of "provider" in the model. Technically, that's an impl detail
      and consumers should not care what generated the CE - it's things like
      the source and service that they'll care about (or even know about)
- introduced the term "Discovery Endpoint" as the "thing" that implements
      the spec
- started an JSON encoding definition
- Modified the query stuff so that all data is returned in one shot.
      I believe the current spec implied that people would need to do one
      query to get the list of Types of interest and then another one to pull
      back the details of each Service. I think it's more natural to do it all
      in one. If this is too large then we should consider query flags to
      indicate if people want the deep result-set or an shallow result-set.
    - I made Service and Type have a non-human friendly main keys since names
      will change over time but these keys need to be static and therefore
      machine usable. `description` can be used for human readable names.
    
Probably other minor things that I can't remember. I basically just
went from top to bottom touching stuff that felt odd as I was reading
but I think it's closer now to being something that someone could
sit down and implement w/o too many gaps.
    
### DIFF
--- a/discovery.md
+++ b/discovery.md
@@ -144,6 +144,7 @@ Service:
 ```
 {
   "url": "[unique URL to this service]",
+  "name": "[unique name for this services]",
   "description": "[human string]", ?
   "docsurl": "[URL reference for human documentation]", ?
   "specversion": "[ce-specversion value]",
@@ -177,7 +178,7 @@ An example:
 ```json
 {
   "url": "https://example.com/services/widgetService",
-  "specversion": "v1.0",
+  "specversion": "1.0",
   "subscriptionurl": "https://events.example.com",
   "protocols": [ "HTTP" ],
   "type": [
@@ -369,14 +370,24 @@ entity.
 
 ###### extensions
 
-- Type: `Map` of `String` to `String`
-- Description: Associative map of CloudEvents
+- Type: `Array` of structures
+- Description: An array or CloudEvents
   [Extension Context Attributes](https://github.com/cloudevents/spec/blob/master/spec.md#extension-context-attributes)
-  that are used for this event `type`. Keys MUST be confirm to the extension
-  context attributes naming rules and value are the type of the extension
-  attribute, conforming to the CloudEvents [type system](./spec.md#type-system).
+  that are used for this event `type`. The structure contains the following
+  attributes:
+  - `name` - the CloudEvents context attribute name used by this extension.
+    It MUST adhere to the CloudEvents context attrbibute naming rules
+  - `type` - the data type of the extension attribute. It MUST adhere to the
+    CloudEvents [type system](./spec.md#type-system)
+  - `specurl` - an attribute pointing to the specification that
+    defines the extension
 - Constraints:
   - OPTIONAL
+  - if present, the `name` attribute in the structure is REQUIRED
+  - if present, the `type` attribute in the structure is REQUIRED
+  - if present, the `specurl` attribute in the structure is OPTIONAL
+- Examples:
+  - `{ "name": "dataref", "type": "URI-reference", "specurl": "https://github.com/cloudevents/spec/blob/master/extensions/dataref.md" }`
 
 #### Service Examples
 
@@ -465,7 +476,7 @@ whose `name` attribute contains the `search term` value (case insensitive).
 * Type: `string`
 * Description: Search term that provides case insensitive match against
   the Service's `name` attribute. The parameter can match any portion
-  of the service's `description` value.
+  of the service's `name` value.
 * Constraints:
   * OPTIONAL
   * If present, MUST be non-empty
--- a/discovery.md
+++ b/discovery.md
@@ -3,8 +3,8 @@
 ## Abstract
 
 CloudSubscriptions Discovery API is a vendor-neutral API specification for
-determining what events an Event Producer has available, as well as how to
-subscribe to those events.
+determining what events are available from a particular system, as well as
+how to subscribe to those events.
 
 ## Status of this document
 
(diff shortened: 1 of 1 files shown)


discovery.md
```diff
@@ -109,296 +123,217 @@ to execute some logic, which might lead to the occurrence of new events.
...
-    subscriptionuri: https://events.example.com
-    protocols:
-    - protocol: HTTP
-    types:
-    - type: com.example.widget.create
-    - type: com.example.widget.delete
+```json
+{
+  "url": "https://example.com/services/widgetService",
+  "specversion": "v1.0",
```
tweing, line 180: `specversion` should be `"1.0"` not `"v1.0"`

discovery.md
```diff
@@ -109,296 +123,217 @@ to execute some logic, which might lead to the occurrence of new events.
...
-This section details the fields that make up each of the entities referenced
-earlier in this document.
-
-#### `service` entity
-
-Used in discovery for enumerating the different services represented in this
-discovery system.
-
-##### `service` entity attributes
+##### name
```
tweing, line 201: I did not see the `name` in the pseudo json above. Was this intentionally?

discovery.md
duglin: fixed

discovery.md
duglin: fixed

discovery.md
```diff
@@ -432,78 +367,105 @@ requirements for the query API.
...
 
 - Type: `Map` of `String` to `String`
-- Description: Associative map of CloudEvents [Extension Context Attributes](https://github.com/cloudevents/spec/blob/master/spec.md#extension-context-attributes)
+- Description: Associative map of CloudEvents
+  [Extension Context Attributes](https://github.com/cloudevents/spec/blob/master/spec.md#extension-context-attributes)
   that are used for this event `type`. Keys MUST be confirm to the extension
   context attributes naming rules and value are the type of the extension
   attribute, conforming to the CloudEvents [type system](./spec.md#type-system).
 - Constraints:
   - OPTIONAL
```
tweing, line 380: Please add examples here as well. 
Did you imagine something like
- "dataref" (from [Dataref extension](https://github.com/cloudevents/spec/blob/master/extensions/dataref.md))
- "traceparent" (from [Distributed Tracing extension](https://github.com/cloudevents/spec/blob/master/extensions/distributed-tracing.md))
?

discovery.md
```diff
@@ -432,78 +367,105 @@ requirements for the query API.
...
+Same as `/services` but the array MUST be limited to just those Services
+whose `name` attribute contains the `search term` value (case insensitive).
+
+###### matching
 
 * Type: `string`
 * Description: Search term that provides case insensitive match against
-  `service` names. The parameter can match any portion of the service name.
+  the Service's `name` attribute. The parameter can match any portion
+  of the service's `description` value.
```
tweing, line 469: On the previous lines above you said 

> whose `name` attributes contains
<br>
but here you were saying

> The parameter can match any portion of the service's `description` value.

Was the intend that the result should return any match to `name` or `description`?

discovery.md
duglin: Done - had to convert it from a map to an array.

discovery.md
duglin: Fixed - good catch. I switch from `description` to `name` late yesterday.

discovery.md
```diff
@@ -109,296 +123,218 @@ to execute some logic, which might lead to the occurrence of new events.
...
-
-##### `type` entity attributes
-
-###### type
+###### specversion
 
-- Type: `String`
-- Description: CloudEvents [`type`](https://github.com/cloudevents/spec/blob/master/spec.md#type)
-  attribute.
+- Type: `String` per [RFC 2046](https://tools.ietf.org/html/rfc2046)
```
deissnerk, line 250: I wonder, if this could also be an array, if multiple specversions are supported, but that would also require the subscription API to support subscribing for an event in a specific specversion.

discovery.md
```diff
@@ -109,296 +123,218 @@ to execute some logic, which might lead to the occurrence of new events.
...
 - Constraints:
-  - REQUIRED
-  - MUST be a non-empty string
+  - OPTIONAL
+  - If present, MUST be a non-empty string
 
-###### datacontenttype
+##### datacontenttype
 
 - Type: `String`
```
deissnerk, line 328: Same here as above for `specversion`, but the subscription API does not support it, yet.

discovery.md
```diff
@@ -3,8 +3,8 @@
 ## Abstract
 
 CloudSubscriptions Discovery API is a vendor-neutral API specification for
-determining what events an Event Producer has available, as well as how to
-subscribe to those events.
+determining what events are available from a particular system, as well as
```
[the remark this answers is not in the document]
duglin: switched to "service"

discovery.md
```diff
@@ -21,33 +21,39 @@ version.
...
-advertise the event types that are available, provide the necessary information
-to consume that event (schema / delivery protocol options), and the necessary
-information to create a subscription. The output of the API ought to be such
-that tooling can be built where all possible event producers and types aren’t
-known in advance.
+In order for consumers to receive events from event producers, they need
+to first subscribe, or ask for, events from those producers. To do so, there
+is often a process necessary that involves steps such as discovering which
+event producer is of interest, what events it generates and how to create the
+subscription for those events.
```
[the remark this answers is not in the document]
duglin: I struggled with this because in CE we use, and define, the term "producer" to mean the entity that creates the CE - more as a piece of middleware that's acting as a serialization engine. But, then there are times when it makes sense to use the term "event producer" to mean "something that generates events" - the more generic meaning outside of CE. And I can try to use the term "service" but it might feel awkward at time.... but we'll see....

discovery.md
```diff
@@ -21,33 +21,39 @@ version.
...
-The second case becomes relevant if multiple sources support the same event
+The second case becomes relevant if multiple services support the same event
 types. Use case 1 is likely the dominant use case. Given the example of a public
-cloud provider where all producers produce events, there might be dozens of
+cloud provider where all services generate events, there might be dozens of
 sources (producer systems) and hundreds of event types. The discovery funnel of
-producer first, then event types helps users navigate without having to see large
-lists of event types. Both of these cases show the importance of using filters
-in the discovery API to narrow down the selection of available events.
+serviecs first, then event types helps users navigate without having to see
```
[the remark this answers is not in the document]
duglin: fixed

discovery.md
```diff
@@ -109,296 +123,218 @@ to execute some logic, which might lead to the occurrence of new events.
...
-      dataschematype: [string per RFC 2046] ?
-      dataschemacontent: [schema] ?
-      dataschemacontent:
-      extensions: *
-      - name: [CE context attribute name]
-        type: [CE type string]
-        spec: [URI to specification defining the extension] ?
+{
+  "url": "[unique URL to this service]",
+  "name": "[unique name for this services]",
```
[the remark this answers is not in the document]
duglin: I mention this must be unique within the scope of the Discovery Endpoint down in the more formal definition of this attribute below.

I think the aggregator will just have to solve that itself.... somehow :-)

discovery.md
duglin: For some reason this makes me nervous, but I can't explain why.  It does make sense though so I'll make this `specversions` and see what people think.

discovery.md
duglin: Would this apply to all of the `dataschema...` attributes?

duglin: If so, I'll do this in a follow-on PR so we can get the bulk of this merged. ok?

deissnerk: It's probably worth a separate discussion. 
+1 for a separate PR

discovery.md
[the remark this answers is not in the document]
deissnerk: The name could become a path with a segment per aggregation step. This way an aggregator would have to add a single path segment unique for the origins it is aggregating. But we can discuss this, when we are getting to this challenge. 😃

duglin:
Approved on 6/18 call

Issue #643, opened 2020-06-12
HTTP binding MUST content modes?


c-pius:
Hi folks,

in the documentation of the HTTP binding it states ([1.3. Content Modes](https://github.com/cloudevents/spec/blob/v1.0/http-protocol-binding.md#13-content-modes))
> Every compliant implementation SHOULD support the structured and binary modes.

In the section below ([1.4. Event Formats](https://github.com/cloudevents/spec/blob/v1.0/http-protocol-binding.md#14-event-formats)) it states
> All implementations of this specification MUST support the non-batching JSON event format,...

Does this second statement make structured content mode a logical MUST? I assume it is intended to say that: when supporting structured content mode, then JSON event format is a MUST?

Anyway, would it be sensible to have binary and structured mode as a MUST instead of a SHOULD?

Thanks!

duglin:
I think your reading of section 1.4 is correct... I think the intent is that the MUST is within the scope of doing "structured".  We should probably clear that up.  Interested in a PR? :-)  If not, someone else will do it.

If I'm remembering correctly, I think we made it a SHOULD because it's a really good idea, but there are going to be times where an implementation just can't support anything but one (e.g. binary due to resource constraints) and so making it a MUST might prevent someone from adopting the spec.

Anyone else for either question?  @clemensv @n3wscott @deissnerk @cneijenhuis

c-pius:
Sounds reasonable, thanks! About the PR... I would have to go through our corporate contributor enablement process so I guess that is a bit too much of overhead for this small contribution :)

deissnerk:
@c-pius  @duglin  I think this issue can be closed, as my proposal of a clarification got merged.

duglin:
great - thanks for circling back to this

PR #644, opened 2020-06-15
Clarified MUST requirement for JSON format

deissnerk:
Addresses #643

### DIFF
--- a/kafka-protocol-binding.md
+++ b/kafka-protocol-binding.md
@@ -76,16 +76,17 @@ the Kafka message's value section as-is, with the `content-type` header value
 declaring its media type; all other event attributes MUST be mapped to the
 Kafka message's [header section][kafka-message-header].
 
-Implementations that use Kafka 0.11.0.0 and above MAY use either _binary_
-or _structured_ modes. Implementations that use Kafka 0.10.x.x and below
-MUST use only use _structured_ mode and encode the event in JSON. This is
-because older versions of Kafka lacked support for message level headers.
+Implementations that use Kafka 0.11.0.0 and above MAY use either _binary_ or
+_structured_ modes. Implementations that use Kafka 0.10.x.x and below MUST only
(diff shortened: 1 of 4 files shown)


duglin:
I believe this was the original intent, so LGTM

deissnerk:
I realized that the wording is pretty much the same for AMQP, MQTT and Kafka. As MQTT and Kafka do only support structured mode for a certain protocol version, the situation is different there. Nevertheless I think the wording also fits there. 

@clemensv In the MQTT binding I removed a sentence stating that MQTT 5.0 MAY support additional formats. I think that was redundant, as the whole concept of structured mode already suggests that additional formats may be used.

kafka-protocol-binding.md
```diff
@@ -76,16 +76,17 @@ the Kafka message's value section as-is, with the `content-type` header value
...
 Kafka message's [header section][kafka-message-header].
 
-Implementations that use Kafka 0.11.0.0 and above MAY use either _binary_
-or _structured_ modes. Implementations that use Kafka 0.10.x.x and below
-MUST use only use _structured_ mode and encode the event in JSON. This is
-because older versions of Kafka lacked support for message level headers.
+Implementations that use Kafka 0.11.0.0 and above MAY use either _binary_ or
+_structured_ modes. Implementations that use Kafka 0.10.x.x and below MUST only
+use _structured_ mode and encode the event in JSON. This is because older
+versions of Kafka lacked support for message level headers.
```
duglin, line 82: just checking... this was just a formatting change, right?

duglin:
still LGTM

kafka-protocol-binding.md
deissnerk: There was a typo. the word "use" occurred twice. I thought I could just fix it, while doing my edits.

duglin: ah missed that - thanks!

duglin:
Approved on 6/18 call

PR #652, opened 2020-06-26
Schema Registry - Change version to comply with OpenAPI spec

tomkerkhove:
Signed-off-by: Tom Kerkhove <kerkhove.tom@gmail.com>

### DIFF
--- a/schemaregistry/schemaregistry.yaml
+++ b/schemaregistry/schemaregistry.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Cloud Native Data Schema Registry
-  version: 0.1
+  version: "0.1"
 paths:
   /schemagroups:
     get:


duglin:
Did a quick search and found this: https://swagger.io/docs/specification/basic-structure/ 
Notice no quotes when in yaml. Does it matter?

tomkerkhove:
Odd, I was using their [editor](https://editor.swagger.io/) which threw the error:
image

Probably because of this:
>  `version` is an arbitrary string that specifies the version of your API (do not confuse it with file revision or the openapi version).

Which matches [this doc](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#infoObject) but I'm happy to abandon if you want.

duglin:
Nope, i was just wondering. I think the quotes are nice :-)

LGTM

duglin:
Approved on the 7/2 call

PR #655, opened 2020-06-28
fix some types

### LINKED ISSUE #654: Terminology: Eventing vs. Remote Procedure Calls - a critique
The current primer of Cloud Events [v1.0](https://github.com/cloudevents/spec/blob/308ba596627266976945afe4a68c43868bd06384/primer.md) describes the purpose of Cloud Eventing with a graphic

source-event-action

The spec presents an [example of an event](https://github.com/cloudevents/spec/blob/308ba596627266976945afe4a68c43868bd06384/spec.md#example)

```
{
    "specversion" : "1.0",
    "subject" : "123",
    "id" : "A234-1234-1234",
    "comexampleothervalue" : 5,
    "datacontenttype" : "text/xml",
}
```
where the event type is named `com.github.pull.create`. This is rather unfortunate.

#### The Critique

The primer states

> CloudEvents should increase interoperability and decouple the producer and consumer of events

But then the primer writes:

> The source generates a message where the event is encapsulated in a protocol. The event arrives to a destination, triggering an action *which is provided with the event data.*

This is clearly the opposite of *decoupling*. A producer of an event *must not* prescribe what a consumer is supposed to do. It *must not* need to know about the code executed and the actions triggered in another system. Something being created in a System *A* might have the effect of something being deleted in a system *B*. If both systems are decoupled a producer *A* can only inform about what happend within its own context but can't tell its consumers what they should do.

From a consumer perspective, when receiving an "event" `com.github.pull.create` I would immediately ask the following questions:

1. does the producer want *me* to *create* a pull request? 
2. or does it intend to *notify* me that a pull request *has been created*?

If the answer is...

- ... "1." then gRPC or Apache Avro RPC were likely the better protocols for those systems rather than *Cloud Events*. Otherwise the result will be a system pretending to be event-driven but (ab)using an event protocol for realising (async.) *remote procedure calls*.
- ... "2." then naming the example event type `com.github.pull.created` would be much more precise about the fact that the actual *occurrence* *has happend* and any consumer is meant to *react* to that.

**Action vs. Effect**

When it comes to Domain Driven Design (Eric Evans) and event-driven architectural styles like CQRS (Greg Young et al.) or Event Sourcing, there is a [commonly accepted convention to name (Domain-) Events in **past tense**](https://docs.microsoft.com/en-us/dotnet/architecture/microservices/microservice-ddd-cqrs-patterns/domain-events-design-implementation). 

In Event Sourced systems an action is something that results in an event, rather than something that is caused by an event in the first place.. E.g. in system *A* an action *CreatePullRequest* results in the system publishing an event `com.github.pull.created` which when consumed by another system *B* may have the **effect** of deleting a flag "No pull request available" - for example by an action *DeleteStateNoPullRequest* (which may result in a new event *NoPullRequestStateDeleted* published by System B).

I suspect the event name `com.github.pull.create` has been directly influenced by the general use of the term *Action* in this spec. I propose consistently choosing a past-tense grammar for event names in examples and talking about *Effects* rather than *Actions*. This better aligns with other disciplines, system architectures and design techniques and makes it easier to draw the line between loosely coupled event driven systems and logically coupled RPC protocols.

#### The proposal

- **Use past tense in the specification to name events in examples**
   - because it helps to avoid designing RPC systems on top of Cloud Events (with logical coupling of producer and consumer)
   - because an event "create" could also have the effect of something being deleted
   - because it aligns well with event-driven data systems, DDD and patterns like Event Sourcing and CQRS
- **Replace the term *action* with *effect*** 
   - because *actions* can mislead to designing producers that *prescribe what a consumer should do* rather than describe *what has happend*
- **Consider extending the specification with a  *recommendation* to name events in past tense rather than naming actions** 
   - even though the actual event names are part of the application domain not part of the protocol, of course (therefore *recommendation*)

duglin:
Closes #654

- Uses past tense
- Update to match the github adapter doc

/cc @about-code

### DIFF
--- a/adapters/couchdb.md
+++ b/adapters/couchdb.md
@@ -19,7 +19,7 @@ based on the event type.
 | `id`                  | The event sequence identifier (`seq`) |
 | `source`              | The server URL / `db`                 |
 | `specversion`         | `1.x-wip`                             |
-| `type`                | `org.apache.couchdb.document.update`  |
+| `type`                | `org.apache.couchdb.document.updated` |
 | `datacontenttype`     | `application/json`                    |
 | `subject`             | The document identifier (`id`)        |
 | `time`                | Current time                          |
@@ -32,7 +32,7 @@ based on the event type.
 | `id`                  | The event sequence identifier (`seq`) |
 | `source`              | The server URL / `db`                 |
 | `specversion`         | `1.x-wip`                             |
-| `type`                | `org.apache.couchdb.document.delete`  |
+| `type`                | `org.apache.couchdb.document.deleted` |
 | `datacontenttype`     | `application/json`                    |
 | `subject`             | The document identifier (`id`)        |
 | `time`                | Current time                          |
@@ -47,7 +47,7 @@ based on the event type.
 | `id`                  | The event sequence identifier (`seq`) |
 | `source`              | The server URL                        |
 | `specversion`         | `1.x-wip`                             |
-| `type`                | `org.apache.couchdb.database.create`  |
+| `type`                | `org.apache.couchdb.database.created` |
 | `subject`             | The database name (`db_name`)         |
 | `time`                | Current time                          |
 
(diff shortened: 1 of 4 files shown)


duglin:
Approved on the 7/2 call

PR #658, opened 2020-07-01
Bring 'datadef' definition into line with specification

### LINKED ISSUE #647: JSON schema allows only object and string for `data` attribute.
JSON schema allows only object or string as data of CloudEvent:
```
   "datadef": {
      "type": ["object", "string"]
    },
```
however spec says that data may be any JSON value.

```
Unlike attributes, for which value types are restricted to strings per the type-system mapping, the resulting data member JSON value is unrestricted, and MAY contain any valid JSON.
```

JemDay:
Fixes #647 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .idea/
+gen/
 .vscode/
 .DS_Store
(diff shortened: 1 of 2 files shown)


duglin:
approved on the 7/9 call

PR #659, opened 2020-07-02
typo and return type mismatch on /types endpoint for discovery spec

loopingz:
While implementing the discovery spec: https://github.com/cloudevents/sdk-javascript/pull/236

I notice the result of /types/* endpoint seems to be map instead of array, not sure if the intent was a map but seems more logical to me so did the change this way.

Another question that raised was the permission scheme, as I was unsure if we needed to filter types within the service object based on permission, I did implement this type of filtering as I supposed it is a possibility that we want to restrain the types of event visible for some audience

### DIFF
--- a/discovery.md
+++ b/discovery.md
@@ -505,13 +505,13 @@ whose `name` attribute contains the `search term` value (case insensitive).
 
 ##### `/types`
 
-This MUST return an array of zero or more Types values, where each Type's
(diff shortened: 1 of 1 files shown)


duglin:
Not sure why it was an array.
LGTM

duglin:
Approved on the 7/9 call

PR #664, opened 2020-07-07
[Discovery] Add an ID field that can not change

### LINKED ISSUE #661: [Discovery] Make Service.Name be immutable
In order for a client to know when a new service appears in the list, vs it's an old one with new metadata, we need an immutable/unique field. I'm proposing we use `service.Name`. If people think that the name of the service might change, then we should introduce an `ID` field.

duglin:
Closes #661 

### DIFF
--- a/discovery.md
+++ b/discovery.md
@@ -143,6 +143,7 @@ its basic structure:
 Service:
 ```
 {
+  "id": "[a globally unique UUID]",
   "url": "[unique URL to this service]",
   "name": "[unique name for this services]",
   "description": "[human string]", ?
@@ -177,7 +178,9 @@ Service:
 An example:
 ```json
 {
+  "id": "cbdd62e8-c095-11ea-b3de-0242ac130004",
   "url": "https://example.com/services/widgetService",
+  "name": "widgets",
   "specversions": [ "1.0" ],
   "subscriptionurl": "https://events.example.com",
   "protocols": [ "HTTP" ],
@@ -198,6 +201,18 @@ with the wire format/API defined by this specification.
 
 The following sections define the attributes that appear in a Service
 entity.
+
+##### id
+- Type: `String`
+- Description: A unique identifier for this Service. This value MUST be
+  globally unique. While other metadata within this Service MAY change,
+  this value MUST NOT. Clients can use this value to ensure that the Service
+  entity returned from a query refers to a previously retrieved Service.
+- Constraints:
+  - REQUIRED
+  - MUST be a valid UUID per RFC4122.
+- Examples:
+  - `bf5ff5cc-d059-4c79-a89a-2513e45a1340`
 
 ##### name
 - Type: `String`
@@ -406,6 +421,7 @@ entity.
 
 ```json
 {
+  "id": "3db60532-e839-417e-8644-e255f338776a",
   "url": "https://storage.example.com/service/storage",
   "name": "storage",
   "description": "Blob storage in the cloud",
@@ -456,7 +472,8 @@ When encoded in JSON, the response format MUST adhere to the following:
 ```
 [
   {
-    "url": "SERVICE-URL",
+    "id": "{id}",
+    "url": "{url}",
     "name": "{name}",
     ... remainder of Service attributes ...
   }
@@ -473,7 +490,8 @@ When encoded in JSON, the response format MUST adhere to the following:
 
 ```
 {
-  "url": "SERVICE-URL",
+  "id": "{id}",
+  "url": "{url}",
   "name": "{name}",
   ... remainder of Service attributes ...
 }
@@ -514,7 +532,8 @@ When encoded in JSON, the response format MUST adhere to the following:
 {
   "TYPE-VALUE": [
     {
-      "url": "SERVICE-url",
+      "id: "{id}",
+      "url": "{url}",
       "name: "{name}",
       ... remainder of Service attributes ...
     }
@@ -534,9 +553,10 @@ When encoded in JSON, the response format MUST adhere to the following:
 
 ```
 {
-  "TYPE-VALUE": [
+  "{type}": [
     {
-      "url": "SERVICE-url",
+      "id: "{id}",
+      "url": "{url}",
       "name: "{name}",
       ... remainder of Service attributes ...
     }
--- a/discovery.md
+++ b/discovery.md
@@ -206,13 +206,37 @@ entity.
 - Type: `String`
 - Description: A unique identifier for this Service. This value MUST be
   globally unique. While other metadata within this Service MAY change,
-  this value MUST NOT. Clients can use this value to ensure that the Service
-  entity returned from a query refers to a previously retrieved Service.
+  this value MUST NOT. Because it is expected that Clients will use this
+  value to know whether a Service returned by a query is the same Service
+  returned by a previous query, Discovery Endpoints MUST ensure this value
+  is immutable for the lifetime of a Service entity.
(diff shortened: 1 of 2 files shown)


discovery.md
```diff
@@ -199,6 +202,18 @@ with the wire format/API defined by this specification.
 The following sections define the attributes that appear in a Service
 entity.
 
+##### id
+- Type: `String`
+- Description: A unique identifier for this Service. This value MUST be
+  globally unique. While other metadata within this Service MAY change,
+  this value MUST NOT. Clients can use this value to ensure that the Service
+  entity returned from a query refers to a previously retrieved Service.
```
deissnerk, line 210: I wonder about the lifecycle of this id. Does the UUID have to be the same, if discovery of the same service is possible via different discovery endpoints? Who is responsible for assigning this id?

duglin: I'm leaning towards saying a little as possible on this to allow for the most flexibility. Meaning, since it must be globally unique, then if the implementation chooses to use the same ID across different discovery endpoints, then they are basically telling all clients that it is the same service. If they then also choose to change the metadata about that service based on the discovery endpoint, then the spec is silent as to what that means - beyond they are the same service.  So, whether one endpoint's metadata should be used in place of another one's metadata, or whether the metadata is supposed to be different because each endpoint is assuming a different type of user is doing the query, is out of scope for the spec to decide.

wdyt?  Is this something we can resolve in a follow-on PR or do you think it needs to be nailed down now?

deissnerk: @duglin I wonder, if this metadata belongs to the discovery endpoint or the service. Does the ID identify the service or the service entry in a specific discovery endpoint? We might also defer this decision until later, when we understand better, how things will be used and interact. We might not be able to clarify this directly with a follow-on PR, but we could accept the PR and create an issue to remind us later about this question.

duglin:
I tweaked it a bit last night - see what you think @deissnerk 
Probably needs more wordsmithing, but I wanted to see if the general direction starts to address your concerns.

deissnerk:
@duglin I'm still not convinced that this is needed. Where possible, I prefer names for identifying resources. Did you run into trouble when trying to implement the API?

duglin:
Yes. If I needed to change the name, I had no way for the client to distinguish between a new service and a rename

deissnerk:
I wonder what makes the name of a service so different from an event type? We never considered using UUIDs for event types. I doubt that we can hide the UUIDs in the producer and consumer developer flows. I understand, if there is a need for a "display name", but dealing with UUIDs in development and debugging is error prone and painful. It is often necessary, when something is instantiated and there is really no meaningful symbolic identifier available, or as a key that is purely internal to a system. Perhaps we have just very different ideas about how a discovery service may be used.
What about a (namespaced) symbolic ID and a display name?

deissnerk:
The description of the `url` field states:
> This value MUST be usable in subsequent requests, by authorized clients, to retrieve this Service entity.

and under restrictions:
> MUST end with fsegments (per RFC1738) of: /services/{name} where name is the Service's name attribute.

Would that still work, if the name was changed?

duglin:
@deissnerk yup - I was thinking about that too. I just updated the PR so that the URL contains the ID instead of NAME. I then had to modify the query section a bit.

All - please take a look.

discovery.md
```diff
@@ -199,31 +202,68 @@ with the wire format/API defined by this specification.
...
   - REQUIRED
   - MUST be a non-empty URL
-  - MUST end with `fsegments` (per RFC1738) of: `/services/{name}` where
-    `name` is the Service's `name` attribute.
+  - MUST end with `fsegments` (per RFC1738) of: `/services/{id}` where
+    `id` is the Service's `id` attribute.
 - Examples:
-  - `http://example.com/services/storage`
-  - `http://discovery.github.com/services/github`
+  - `http://example.com/services/bf5ff5cc-d059-4c79-a89a-2513e45a1340`
```
tweing, line 265: While I like the idea of having a unique id per service, it doesn't make sense to put the id in the URL in my opinion. The service name needs to be unique per discovery endpoint, so the name can be used in the URL (which of course limits the name to a single word).
If the name changes, the new URL can be retrieved through the discovery API (..../services/{id}).

discovery.md
```diff
@@ -456,15 +497,16 @@ When encoded in JSON, the response format MUST adhere to the following:
...
+    "url": "{url}",
     "name": "{name}",
     ... remainder of Service attributes ...
   }
   ...
 ]
 ```
 
-##### `/services/{name}`
+##### `/services/{id}`
```
tweing, line 509: I would keep name here.

tweing:
> but dealing with UUIDs in development and debugging is error prone and painful

@deissnerk : I would say the opposite is true from my experience. We used names in one of our products as the identifier and then the name got changed - We had to find where it was used as identifier and where as pure text and then change it EVERYWHERE in the code, database, etc. This was really painful.

I agree though that UUID's are not the most convenient identifiers, but they are globally unique. 

We also use [OID's](http://www.oid-info.com/), but they are on a similar level like UUID's with a meaning behind in a registry. Here is an example:  2.16.840.1.113995.3.110.3.6281

discovery.md
duglin: My main motivation for putting the ID into the URL instead of the NAME is because if the NAME changes I wanted a static URL that a client could use w/o worrying about it changing over time. Often these URLs appear as refs in data and asking someone to find/change all possible refs could be quite costly (and racy).

I did think about allowing both ../services/ID and ../services/NAME, but then you run into the fun issue of people using UUID as NAMES and getting conflicts :-)    But we could say "be smart"... dunno

discovery.md
```diff
@@ -473,13 +515,14 @@ When encoded in JSON, the response format MUST adhere to the following:
...
-  "url": "SERVICE-URL",
+  "id": "{id}",
+  "url": "{url}",
   "name": "{name}",
   ... remainder of Service attributes ...
 }
 ```
 
-##### `/services?matching=[search term]`
+##### `/services?name=[search term]`
```
manuelstein, line 525: Two things:

1. I think ?name= and ?matching= is not the same
2. the "contains"-style query does not functionally replace /service/{name} (from above) because here you'd get all services whose name contains the search term

I'd suggest to keep ?name= here but change the description. REST query parameters like "?name=", to me personally, provide only exact matches anyhow, so I would no longer expect it to return all Services that "contain" the search term.
But using ?matching, IMHO is also not a good encoding of a search operation operation. Personally, I don't like encoding the operation type (match) in a URL query parameter ("?matching") and at the same time imply the searched field (name). Swagger suggests to encode operations as endpoints, e.g. /services/matching.

So, with ?name= it might be better to change the functionality here to offer an **exact** match in order to provide an equivalent to /services/{name}.
But then, a replacement for the ?matching= functionality is also required to return "contains" matches. Now that /services/{name} is gone, it could be /services/matchName, /services/match?name=, /services/find?name= or any /services/{operation} as long as it's not a UUID.

duglin: I agree with your logic on "name" - defining it to mean "matching' is a bit awkward. 
I'm not sure yet how I feel about `/service/{operation}` - I think that might require a bit more thinking by the group. In the interest of making forward progress though, I'd like to propose that as part of this PR we use "name" to mean an exact match - then we still allow for exact-name-based searching. Then I'll open an issue for the "matching" stuff so we can brainstorm on how people would like to add that.
What do people think?

duglin:
Approved on the 8/6 call

PR #668, opened 2020-07-14
First pass at pagination

duglin:
Just a first draft of a "pagination" spec. Rather than put it into Discovery, I thought that we might be able to define it once and reuse it for all of the specs.

### DIFF
--- a/pagination.md
+++ b/pagination.md
@@ -1,0 +1,193 @@
+# Pagination
+
+This document describes a mechanism by which a server can return a set of
+records to a client in an incremental fashion. Often this will be used when
+a client is doing a query for a set of records and the result set is too large
+to return in one response.
+
+## Notations and Terminology
+
+### Notational Conventions
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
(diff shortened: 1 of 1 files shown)


duglin:
on 7/16 call @clemensv recommended we look at RFC7232 - section 3.3

duglin:
> s/sever/server/

@drewpca fixed!

pagination.md
```diff
@@ -0,0 +1,178 @@
...
+
+The follow list of attributes MAY be included in a client's request for
+a set of records from a server. The client MUST only specify these on
+the initial request for set. If the results can not fit within one
+message then these attributes MUST NOT be added to subsequent requests
+by the client. Note, the server MAY include these attributes within
+the URI-reference returned in a response message, but the client MUST NOT
+modify those values.
+
+#### size
```
sdatspun2, line 33: suggestion: limit (size) and offset (page) are more commonly used in pagination queries

duglin: ok done

duglin:
I reworked the 'expires' stuff to leverage the HTTP "Expires" header. Most of the rest of the doc remains the same (just s/size/limit/ and s/page/offset/) but that's just a syntax change, no semantic change.

duglin:
All - sorry for not pushing the latest before the call.  I just updated it and made it clear that the use of things like "offset" in the URLs is an impl choice. The spec doesn't mandate how the server does its chunking.

duglin:
Approved on 8/13 call

PR #672, opened 2020-07-16
Clarify difference between message mode and HTTP content mode

cneijenhuis:
Another attempt to fix https://github.com/cloudevents/spec/issues/645

See also https://github.com/cloudevents/spec/pull/660

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -136,6 +136,9 @@ a stand-alone event format and stored in the message body.
 A "binary-mode message" is one where the event data is stored in the message
 body, and event attributes are stored as part of message meta-data.
 
+While protocol bindings typically use "binary-mode messages" directly on the
+wire, "structure-mode messages" are often embedded in an envelope.
+
 #### Protocol
 
 Messages can be delivered through various industry standard protocol (e.g. HTTP,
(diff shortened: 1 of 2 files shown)


spec.md
```diff
@@ -136,6 +136,9 @@ a stand-alone event format and stored in the message body.
 A "binary-mode message" is one where the event data is stored in the message
 body, and event attributes are stored as part of message meta-data.
 
+While protocol bindings typically use "binary-mode messages" directly on the
```
grant, line 139: Not sure if this note belongs in the spec. It seems more like a usage/user guide detail.

cneijenhuis:
FYI I won't make todays meeting due to travel, and will be on vacation for the two weeks after.

I'm ok if someone else takes over, or putting it on hold while I'm gone. Sorry for the inconvenience!

duglin:
Approved on the 7/30 call w/o these lines: https://github.com/cloudevents/spec/pull/672/files#diff-958e7270f96f5407d7d980f500805b1bR139-R140

@cneijenhuis there was some concern about those lines... both from a perspective that it wasn't clear what it was trying to address, and some folks didn't think it was necessary.

Would you be ok with this PR w/o those lines? If so, we can merge. But if you'd like to keep them then we should it more in the PR and on next week's call.

duglin:
@cneijenhuis ping! :-)

cneijenhuis:
I'm back from vacation and have removed those lines. Thanks for your patience! 😉

duglin:
@cneijenhuis thanks and welcome back!

Merging.

PR #675, opened 2020-07-22
Adding use cases for producers and intermediaries

deissnerk:
So far the subscription and discovery effort was mostly discussed from the consumer perspective. I divided the use cases into use cases for consumers, intermediaries, and producers and added cases for intermediaries and producers.


duglin:
Approved on 7/30 call

Issue #685, opened 2020-08-12
[Discovery] service.types[0].type should be .name


n3wscott:
The field inside the types array is currently named `type` to represent the name of the type. This field should be renamed to `name`, as it is the name of the type, not the type of the type.

duglin:
I can certainly see that.... I think we might have picked `type` to align it with the CloudEvents field called `type` - to mentally establish the connection.

Issue #684, opened 2020-08-12
[Discovery] Drop /types endpoint.


n3wscott:
The `/types` endpoint is nothing more than data processing, forcing the implementor to do this data processing should be reconsidered. I would recommend dropping the `/types` endpoint in favor of some other solution that can do the pivot filtering that is out of scope for CloudEvents. CloudEvents should have the responsibility stop at providing the raw data, not an opinionated projection of the data.

nachocano:
I agree with @n3wscott. I was building a prototype of this in my [eventing fork](https://github.com/nachocano/eventing/tree/discovery/pkg/serverless/discovery) to better understand how we can leverage this spec for custom events with our Broker/Trigger/EventType objects, and all the grouping by types was extra processing and also extra burden to implement, and I'm not sure it added much value. 

Also, it seems to me that we have some kind of cycle... type --> service --> types. 

According to this output of `/types`:

```
{
  "TYPE-VALUE": [
    {
      "id: "{id}",
      "url": "{url}",
      "name: "{name}",
      ... remainder of Service attributes ...
    }
    ...
  ]
  ...
}
```

Say for example I have service `my-svc` that emits types `t1` and `t2`, then the output would look something like this:

```
{
  "t1": [
    {  ...
      "name: "my-svc",
       ...
      "types": [
         { "type": "t1",
            ...
         },   
         { "type": "t2",
            ...
         }, 
       ],
    }
  ],
  "t2": [
    { ...
      "name: "my-svc",
       ...
      "types": [
         { "type": "t1",
            ...
         },   
         { "type": "t2",
            ...
         }, 
       ],
    }
  ],
}
```

Is my interpretation correct?

duglin:
Yes it is a bit of a cycle, but it also allows you to go from "one type" -> "service(s)" -> "what other types does that service support".  So it's not 100% redundant, more like additive.

But I do prefer the idea of getting rid of /types and just using /services?type=...  type of thing instead.

Issue #679, opened 2020-07-29
[Discovery] Sources of truth in RESTful APIs


n3wscott:
I just reread the spec with the aim to implement it and one thing really sticks out as weird to me. One of the RESTful principles is there is only 1 way to get an individual resource (or a collection). The issue I am seeing is in the `/types` request. You can get the full definition of a service in `/services` request, but also in a `/types` request. This violates the RESTful principles of single source of truth.

I understand the intent is for mapping types to services. I think we need to take a look at this and understand why `/types`  needs to return the full service definition, or if it can return a map of type to service id and let the `/services` endpoint be to source of truth for services.

duglin:
I'm not sure REST requires only one way to get at metadata, but returning a list of URLs instead is possible. However, there are problems with that w.r.t. performance. Let's say we get back a list of 1000 services. Unless people can formulate some (potentially complex) query to `/services` that allows them to ask for that exact list of 1000, they'll need to call `/services/ID` 1000 times. That's not only going to be really slow compared to the `/types` stuff (which will, at worst, have several requests due to pagination), but people might run into rate-limiting issues.

We could allow for people to choose whether they want the deep list vs shallow list via a query parameter tho.

n3wscott:
related to #684

deissnerk:
I agree to @duglin in that I don't think REST requires that there is always only a single URL to retrieve a certain piece of information. In order to enable caching, clients usually should retrieve the same information using the same URL, though. Afaik it's a principle of REST that requests can be cached. 
I also agree to @n3wscott that we may need to take a look at why `/types` a list of full service definitions. There might still be some ambiguities in the data model.

@duglin In OData there is the `$expand` query parameter: 
[https://docs.microsoft.com/en-us/aspnet/web-api/overview/odata-support-in-aspnet-web-api/using-select-expand-and-value](https://docs.microsoft.com/en-us/aspnet/web-api/overview/odata-support-in-aspnet-web-api/using-select-expand-and-value)
[https://www.odata.org/getting-started/basic-tutorial/#expand](https://www.odata.org/getting-started/basic-tutorial/#expand)

PR #689, opened 2020-08-20
Rename services types[].type to types[].name; Drop /types endpoint

n3wscott:
Fixes: https://github.com/cloudevents/spec/issues/685
Fixes: https://github.com/cloudevents/spec/issues/684
Fixes: https://github.com/cloudevents/spec/issues/679

- Rename services types[].type to types[].name
- Drop /types endpoint

Signed-off-by: Scott Nichols <snichols@vmware.com>

### DIFF
--- a/discovery.md
+++ b/discovery.md
@@ -143,9 +143,9 @@ Service:
   },
   "authscope": "[string]", ?
   "protocols": [ "[string]" + ],
-  "types": [ ?
+  "events": [ ?
     { *
-      "name": "[ce-type value]",
+      "type": "[ce-type value]",
       "description": "[human string]", ?
       "datacontenttype": "[ce-datacontenttype value]", ?
       "dataschema": "[ce-dataschema URI]", ?
@@ -173,12 +173,12 @@ An example:
   "specversions": ["1.0"],
   "subscriptionurl": "https://events.example.com",
   "protocols": ["HTTP"],
-  "types": [
+  "events": [
     {
-      "name": "com.example.widget.create"
+      "type": "com.example.widget.create"
     },
     {
-      "name": "com.example.widget.delete"
+      "type": "com.example.widget.delete"
     }
   ]
 }
@@ -334,7 +334,7 @@ The following sections define the attributes that appear in a Service entity.
   - `[ "HTTP" ]`
   - `[ "HTTP", "AMQP", "KAFKA" ]`
 
-##### name
+##### type
 
 - Type: `String`
 - Description: CloudEvents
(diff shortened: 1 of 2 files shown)


discovery.md
```diff
@@ -41,19 +41,8 @@ a Discovery Endpoint could choose to be implemented as part of a Service, or
...
-
-The second case becomes relevant if multiple services support the same event
-types. Use case 1 is likely the dominant use case. Given the example of a public
-cloud provider where all services generate events, there might be dozens of
-sources and hundreds of event types. The discovery funnel of services first,
-then event types helps users navigate without having to see large lists of event
-types. Both of these cases show the importance of using filters in the discovery
-API to narrow down the selection of available events.
+The main usecase to condider from the viewpoint of the event consumer is what
+services are available, and what event types do they generate?
```
deissnerk, line 45: Why do we drop the second use case entirely? I think it is still relevant.

discovery.md
```diff
@@ -156,7 +145,7 @@ Service:
   "protocols": [ "[string]" + ],
   "types": [ ?
     { *
-      "type": "[ce-type value]",
+      "name": "[ce-type value]",
```
deissnerk, line 148: I wonder, if it is the attribute `type` that needs to be renamed to `name`, or if the name `types` of the array is misleading. Is it really a list of types? We are specifying how context attributes (including `type`) are used. So isn't it a list of event pattern or definitions?

deissnerk: One more thing:
Would `name` be unique within the elements of `types`, or could there be multiple entries for the same event type?

discovery.md
```diff
@@ -41,19 +41,8 @@ a Discovery Endpoint could choose to be implemented as part of a Service, or
...
-2. What event types are available, and from which services?
-
-The second case becomes relevant if multiple services support the same event
-types. Use case 1 is likely the dominant use case. Given the example of a public
-cloud provider where all services generate events, there might be dozens of
-sources and hundreds of event types. The discovery funnel of services first,
-then event types helps users navigate without having to see large lists of event
-types. Both of these cases show the importance of using filters in the discovery
-API to narrow down the selection of available events.
+The main usecase to condider from the viewpoint of the event consumer is what
```
duglin, line 44: s/condider/consider/

discovery.md
duglin: hmmm s/types/events/ might sounds interesting.  If we did that, would we keep "type"?  Seems like it might be ok then, no?

n3wscott: Yeah events and type seems fine to me.

discovery.md
n3wscott: I dropped it because the answer to the question is suppose to be calls to the /types endpoint, which I dropped.

duglin: It does beg the question... how does someone get a list of all of the 'types'?  Or is that even a valid question? Should they be able to ask for the list of all "sourcetemplates" too?  With the rename of "types" to "events", is "type" no longer special? Should it be?

I like the idea of treating all attributes equally. So, perhaps we should ask if getting the list of types is really a query people want to do?  Or is it more likely they people have a type in mind (e.g. com.example.db.created) and are looking for services that emit that?  Is a list of random strings (types) something that's interesting and useful?

duglin: I guess if someone can explain why the list of types is useful then that would help us define the `/types` API that this PR is removing.

duglin:
Approved on the 8/27 call

PR #690, opened 2020-09-06
Add more clarity around our REST APIs

duglin:
Just adding more detail about our REST API


discovery.md
```diff
@@ -521,14 +545,184 @@ When encoded in JSON, the response format MUST adhere to the following:
 }
 ```
 
-##### `/services?name={name}`
+Note: in the case of `406 Not Acceptable` or `409 Conflict`, it is an
+implementation choice as to how long these responses will be returned. But it
+is expected that a `404 Not Found` will eventually be returned.
```
lance, line 550: I'm not sure I understand what is meant here. Does this mean that, as a side effect of a `GET` request, a `Service` may be created? And if so, that process may fail, so subsequent requests for the same service will eventually return `404 Not Found`? I suspect that reading is incorrect because it seems wrong to me that anything would get created as a side effect here. But if that's the case, I don't understand the `406` and `409` return codes.  

Alternatively, it could be that the `Service` is being created as a result of a `POST` request as shown below. If so, this would mean that the Discovery API must maintain some sort of state around the creation of services that can be shared across requests. In all, this is confusing.

Update: it's becoming clearer as I continue reading. But I think it might be worth calling out that these responses can only exist in the case where a `POST` or `PUT` has recently been attempted with a previous request.

discovery.md
```diff
@@ -521,14 +545,184 @@ When encoded in JSON, the response format MUST adhere to the following:
...
+
+The follow responses are defined by this specification:
+- `201 Created` if the new Service was added. The response MUST include
+  an HTTP `Location` Header with a value that points to the new Service.
+  The HTTP Response Body MUST include the JSON represenation of the
+  newly created Service.
+- `202 Accepted` to indicate that the request has been accepted but not
+  processed yet. The response MUST include an HTTP `Location` Header with
+  a value that can be used to determine the status of the request. The
+  HTTP Response Body MAY be empty.
```
lance, line 569: This starts to clarify my questions noted in `GET /services/{id}` and does indicate that the discovery service must be stateful to some degree.

discovery.md
```diff
@@ -521,14 +545,184 @@ When encoded in JSON, the response format MUST adhere to the following:
...
+
+In the case of `202 Accepted` the Discovery Endpoint MUST include an
+HTTP `Location` Header in the response pointing to an endpoint that the
+client can use to determine the status of the delete.
+
+A successful HTTP `GET` to the `Location` endpoint MUST either return:
+- `200 OK` meaning the the original Service was successfully deleted
+- `204 No Content` to indicate that the `DELETE` request is not completed yet
+- `409 Conflict` to indicate that the original `DELETE` request message failed.
+  The HTTP Response Body SHOULD contain information as to why the create
```
lance, line 702:
```suggestion
  The HTTP Response Body SHOULD contain information as to why the delete
```

discovery.md
```diff
@@ -521,14 +545,184 @@ When encoded in JSON, the response format MUST adhere to the following:
...
+If an error is not generated, the Discovery Endpoint MUST return one of:
+- `200 OK` if the new Service was deleted
+- `202 Accepted` to indicate that the request has been accepted but not
+  processed yet
+
+Other responses are allowed, but not defined by this specification.
+
+In the case of `202 Accepted` the Discovery Endpoint MUST include an
+HTTP `Location` Header in the response pointing to an endpoint that the
+client can use to determine the status of the delete.
```
lance, line 696: What does this URL look like - what is the endpoint?

discovery.md
```diff
@@ -521,14 +545,184 @@ When encoded in JSON, the response format MUST adhere to the following:
...
+
+##### `DELETE /services/{id}`
+
+This MUST delete the Service at the referenced URL. If the Service can not
+be deleted then an error MUST be generated.
+
+If an error is not generated, the Discovery Endpoint MUST return one of:
+- `200 OK` if the new Service was deleted
+- `202 Accepted` to indicate that the request has been accepted but not
+  processed yet
```
lance, line 690: What should the response be if there is not a `Service` corresponding to `id`?

discovery.md
```diff
@@ -113,10 +113,9 @@ to execute some logic, which might lead to the occurrence of new events.
 
 The request for events from a Service.
 
-## API Specification
+## Resource Model
 
-This API is specified as a REST API with well defined entities and relationships
-between those entities.
+This sections defines the resource model of a Discovery Endpoint.
```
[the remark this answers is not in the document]
duglin: Done

discovery.md
```diff
@@ -459,11 +458,10 @@ The following sections define the attributes that appear in a Service entity.
 }
 ```
 
-### REST Paths
+## API Specification
 
-Each path in the REST API represents either a list (or search) over the
-Discovery Endpoint, or the retrieval of an individual entity. All of these
-operations MUST be supported by compliant discovery implementations.
+All of the APIs specified below MUST be supported by compliant Discovery
```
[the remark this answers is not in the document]
duglin: Done

discovery.md
duglin: I added a '404' line

discovery.md
duglin: that's an impl detail - I'm not sure we need to standardize it. Do we?

discovery.md
duglin: Done

discovery.md
lance: Maybe not. My brain kept wanting to try and map something in this document to what that URL should be though. Would it be worth noting that the URL in this case is an implementation detail outside of the scope of the spec?

duglin: added

discovery.md
duglin: I tried to add more text to help clarify why we're talking about this stuff... see if that helps.

discovery.md
```diff
@@ -474,26 +472,30 @@ https://example.com/services
...
-the entity did not exist (e.g. for HTTP the response would be `404 Not Found`).
-In cases where the response is an array, then the response SHOULD return a
-successful status with an array, even if that array is empty. As Discovery
-service can be decoupled from Services permission checks, the above requirement
-is OPTIONAL. If the information is available to the Discovery service, it is
-highly RECOMMENDED.
+If a Discovery Endpoint can perform authorization checks to determine
+which client can see which Service, and the requesting client is not allowed
+access to a particular Service, then the Discovery Endpoint MUST respond
+as if that Service did does exist. For example, it would be excluded from
```
lance, line 478:
```suggestion
as if that Service does not exist. For example, it would be excluded from
```

duglin: fixed

discovery.md
lance: It does - thanks.

discovery.md
```diff
@@ -459,11 +458,10 @@ The following sections define the attributes that appear in a Service entity.
 }
 ```
 
-### REST Paths
+## API Specification
 
-Each path in the REST API represents either a list (or search) over the
-Discovery Endpoint, or the retrieval of an individual entity. All of these
-operations MUST be supported by compliant discovery implementations.
+All of the API endpoints specified below MUST be supported by compliant
```
duglin, line 463: remove this

discovery.md
```diff
@@ -521,14 +545,190 @@ When encoded in JSON, the response format MUST adhere to the following:
...
+  failed.
+- `409 Conflict` to indicate that the request message is syntactically valid,
+  but the value of one or more of the attributes are invalid - for example, the
+  `name` is not unique.
+  The HTTP Response Body SHOULD contain information as to why the create
+  failed.
+
+Other responses are allowed, but not defined by this specification.
+
+If a `202 Accepted` was returned, an HTTP `GET` to the `Location` endpoint
```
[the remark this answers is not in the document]
deissnerk: And this is how OData is doing it:
[http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_AsynchronousRequests](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_AsynchronousRequests)

discovery.md
```diff
@@ -521,14 +545,190 @@ When encoded in JSON, the response format MUST adhere to the following:
...
+Other responses are allowed, but not defined by this specification, however
+they are related to the `GET` itself and not the original add request.
+
+How long the Discovery Endpoint will support the original `Location` URL is
+an implementation choice, however, it MUST be available immediately upon
+return of the `202 Accepted`. Meaning, a `404 Not Found` MUST NOT be returned
+in response to a `GET` to the `Location` URL during the time the creation
+request is being processed.
+
+##### `PUT /services/{id}`
```
n3wscott, line 615: I would break this into another interface, creating services is not about discovering them.

duglin: yes I'm working on putting the management stuff into a separate section - just couldn't get to it last night...

duglin:
Updated

discovery.md
```diff
@@ -474,26 +473,33 @@ https://example.com/services
...
 
-#### `/services`
+#### `GET /services`
 
-This MUST return an array of zero or more Services. The list MUST contain all
+This MUST return an array of zero or more Services. The array MUST contain all
 Services available via this Discovery Enpoint.
 
-When encoded in JSON, the response format MUST adhere to the following:
+In the case of `200 OK`, the response format MUST adhere to the following:
```
duglin, line 497: Should we mention that a missing entry should be considered deleted?

duglin: Open a new issue to discuss

discovery.md
```diff
@@ -537,17 +563,118 @@ When encoded in JSON, the response format MUST adhere to the following:
...
+MAY choose to include the `epoch` attribute. If present then the Discovery
+Endpoint MUST ensure that the value on the incoming request matches the
+current value on the Service. If they do not match then an error MUST be
+generated and the Service MUST NOT be updated. However, if the incoming
+request does not include this attribute then this verification MUST NOT
+be done.
+
+Additionally, when a Service with the specified `id` aleady exists, the
+`epoch` value MUST be updated with a larger value, indicating that the
+Service has been updated.
```
grantr, line 646: I'm a bit confused by this and the preceding paragraph. It seems to be saying that when updating an existing service, the singular `epoch` attribute must be both equal to AND larger than the current value.

Perhaps the intent is that the epoch value is incremented by the Discovery Endpoint, not the requester, and the new value is returned in the response? In this case, is there no way for the requester to choose an epoch value?

discovery.md
```diff
@@ -521,14 +545,190 @@ When encoded in JSON, the response format MUST adhere to the following:
...
 
-##### `/services?name={name}`
+Note: the `406 Not Acceptable` and `409 Conflict` cases are for situations
+where a create request returned a `202 Accepted` and the client is doing a
+`GET` to this endpoint to determine the status of that create request.
+It is an implementation choice as to how long these 2 responses will be
+returned. But it is expected that a `404 Not Found` will eventually be
+returned instead.
+
+#### `POST /services`
```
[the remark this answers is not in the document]
nachocano: +1

discovery.md
duglin: Opened: https://github.com/cloudevents/spec/issues/700

discovery.md
duglin: yes and yes. I'll make it clearer that the D.E. is the one doing the update. The only way a client can set an epoch value is during an import.  But how a D.E. knows it's an import vs an update is an open question as of now.

discovery.md
```diff
@@ -537,17 +563,118 @@ When encoded in JSON, the response format MUST adhere to the following:
...
+The follow responses are defined by this specification:
+- `201 Created` if the new Service was added. The response MUST include
+  an HTTP `Location` Header with a value that points to the new Service.
+  The HTTP Response Body MUST include the JSON represenation of the
+  newly created Service.
+- `404 Not Found` if there is no Service with the specified `id`.
+
+Other responses are allowed, but not defined by this specification.
+
+#### `PUT /services/{id}`
```
duglin, line 621: I was going to update this PR to try to deal with "import" vs "update" but I decided to open an issue instead so we can deal with it on its own.  I'm leaning towards having the client add a query parameter to indicate which they're doing. But if someone has any other ideas please speak up.

https://github.com/cloudevents/spec/issues/701

discovery.md
grantr: Based on this and #699, it feels like this "update" operation would be more clear semantically if it is actually a creation of an epoch resource owned by the service resource. Update is then a POST to the service's resource path that returns as its Location the resource path of the epoch that was created:

`POST /services/{id}`

This also solves the problem of import vs update, because you can allow PUT to a specific epoch:

`PUT /services/{id}/{epoch}`

grantr: I realized there's another issue just for this #701 so I cross-posted the above there.

duglin: I'd like to deal with your suggestion in a separate issue/PR if that's ok. Mainly because I think the proposal will require a lot of discussion and I'd like to get _something_ in that we can morph and test-out rather than not have anything at all for potentially several more weeks.

duglin:
Approved on 9/24 call

PR #691, opened 2020-09-06
Add a version/timestamp attribute

duglin:
Per @n3wscott's suggestion, this adds a new attribute that can be used to
know when an entry has been updated. It can be used as an e-tag type of
thing or to know when, during an import, the entity being imported is
older than what the discovery endpoint already has.

I wasn't sure if a counter or a timestamp would be better, so I added
both so people can see how both look.

I'm leaning more towards a timestamp since it give a bit more information.

### DIFF
--- a/discovery.md
+++ b/discovery.md
@@ -202,14 +202,6 @@ The following sections define the attributes that appear in a Service entity.
   NOT so that clients can use this attribute to know whether a Service returned
   by a query is the same Service returned by a previous query.
 
-  Typically, this value is defined by the Discovery Endpoint, or one of the
-  components behind it. However, there might be cases where the value is
-  provided to the Discovery Endpoint, for example, during an "import" type of
-  operation. In these "import" cases the attribute is REQUIRED to be
-  in the client's request. However, in cases where the Service has no
-  associated `id` that needs to be retained, this attribute MUST NOT be
-  present in the client's request.
-
   Whether a change to a Service would result in changing of the Service's
   metadata (except `id`) and thus be just an update of an existing Service, or
   whether the change would result in a brand new Service (with a new `id`) is
@@ -229,7 +221,7 @@ The following sections define the attributes that appear in a Service entity.
   See the Primer for more information.
 
 - Constraints:
-  - conditionally REQUIRED. See above.
+  - REQUIRED in responses from the Discovery Endpoint.
   - MUST be a valid UUID per RFC4122.
 - Examples:
   - `bf5ff5cc-d059-4c79-a89a-2513e45a1340`
@@ -246,26 +238,8 @@ The following sections define the attributes that appear in a Service entity.
   integer comparision to determine which version of this Service Entry is the
   latest - meaning, the one with the larger integer value.
 
-  While this attibute is REQUIRED to be part of the the Service's description
-  when returned from a Discovery Endpoint, depending on the action from the
-  client it might not be mandatory to appear in a client's request. For a
-  "create" operation, this attribute MUST NOT appear. For an "import" type
-  of operation, where it is important for the entry to retain a previously
-  defined value, this attribute MUST appear.
-
-  A client MAY choose to include this attribute on an "update" operation.
-  If present then the Discovery Endpoint MUST ensure that the value
-  on the incoming request matches the current value within the Discovery
-  Endpoint, and if it does not then an error MUST be generated and the Service
-  Entry MUST NOT be updated. However, if the incoming request does not include
-  this attribute then the Discovery Endpoint MUST skip the version matching
-  check.
-  Note: regardless of whether the incoming request has this attribute or not,
-  a subsequent retrieval of this Service Entry MUST have an updated value
-  for this attribute.
-
 - Constraints:
-  - Conditionally REQUIRED. See above.
+  - REQUIRED in responses from the Discovery Endpoint.
 - Examples:
   - `42`
   - `915148800`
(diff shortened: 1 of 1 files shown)


discovery.md
```diff
@@ -203,7 +207,10 @@ The following sections define the attributes that appear in a Service entity.
   Typically, this value is defined by the Discovery Endpoint, or one of the
   components behind it. However, there might be cases where the value is
   provided to the Discovery Endpoint, for example, during an "import" type of
-  operation.
+  operation. In these "import" type of cases the attribute is REQUIRED to be
```
lance, line 210: Nit
```suggestion
  operation. In these "import" cases the attribute is REQUIRED to be
```

discovery.md
```diff
@@ -224,11 +231,85 @@ The following sections define the attributes that appear in a Service entity.
   See the Primer for more information.
 
 - Constraints:
-  - REQUIRED
+  - conditionally REQUIRED. See above.
   - MUST be a valid UUID per RFC4122.
 - Examples:
   - `bf5ff5cc-d059-4c79-a89a-2513e45a1340`
 
+##### discversion
```
lance, line 239: What is the rationale for not going with simply `version`?

discovery.md
```diff
@@ -224,11 +231,85 @@ The following sections define the attributes that appear in a Service entity.
...
+  client it might not be mandatory to appear in a client's request. For a
+  "create" operation, this attribute MUST NOT appear. For an "import" type
+  of operation, where it is important for the entry to retain a previously
+  defined value, this attribute MUST appear
+
+  A client MAY choose to include this attribute on an "update" operation.
+  If present then the Discovery Endpoint MUST ensure that the value
+  on the incoming request matches the current value within the Discovery
+  Endpoint, and if it does not then an error MUST be generated and the Service
+  MUST NOT be updated. However, if the incoming request does not include this
```
lance, line 261: Doesn't an "update" operation alter the metadata for the Service? If so, is the expectation that a subsequent `GET /services/{service}` would have a discversion that has increased from the provided value after this update?

discovery.md
```diff
@@ -224,11 +231,85 @@ The following sections define the attributes that appear in a Service entity.
...
+  client it might not be mandatory to appear in a client's request. For a
+  "create" operation, this attribute MUST NOT appear. For an "import" type
+  of operation, where it is important for the entry to retain a previously
+  defined value, this attribute MUST appear.
+
+  A client MAY choose to include this attribute on an "update" operation.
+  If present then the Discovery Endpoint MUST ensure that the value on the
+  incoming request matches the current value within the Discovery
+  Endpoint, and if it does not then an error MUST be generated and the Service
+  MUST NOT be updated. However, if the incoming request does not include this
```
lance, line 291: The same question as above. The update for an entry alters the metadata, so presumably subsequent `GET` requests for the updated service will have a new `updated` value generated by the endpoint. Should this be explicitly called out?

discovery.md
duglin: I originally did use `version` but I wanted it to be clear that this is NOT the version of the service itself - this is the "Discovery Version". Note that we may have other versions later on... for example, we already have 'specversion" - so it might be better to not let anyone grab the generic "version" and avoid the possible confusion.

discovery.md
duglin: done

discovery.md
duglin: yes, a subsequent GET would have a new discversion value

discovery.md
duglin: I added a sentence to make this clear

discovery.md
```diff
@@ -203,7 +207,10 @@ The following sections define the attributes that appear in a Service entity.
   Typically, this value is defined by the Discovery Endpoint, or one of the
   components behind it. However, there might be cases where the value is
   provided to the Discovery Endpoint, for example, during an "import" type of
-  operation.
+  operation. In these "import" cases the attribute is REQUIRED to be
+  in the client's request. However, in cases where the Service has no
+  associated `id` that needs to be retained, this attribuet MUST NOT be
```
duglin, line 212: typo

discovery.md
```diff
@@ -168,6 +170,8 @@ An example:
 ```json
 {
   "id": "cbdd62e8-c095-11ea-b3de-0242ac130004",
+  "discversion": 1,
+  "updated": "2018-04-05T17:31:00Z",
```
duglin, line 174: consider "epoc"

duglin: Proposal: "epoch" - int64,must go up each time
s/client/....

duglin:
ok I updated it with `epoch`

discovery.md
```diff
@@ -224,11 +229,47 @@ The following sections define the attributes that appear in a Service entity.
...
+  represents a timestamp or could choose to simply use a monotomically
+  increasing number. The only requirement is that the value MUST always
+  increase each time the Service Entry is updated. This allows for a quick
+  integer comparision to determine which version of this Service Entry is the
+  latest - meaning, the one with the larger integer value.
+
+  While this attibute is REQUIRED to be part of the the Service's description
+  when returned from a Discovery Endpoint, depending on the action from the
+  client it might not be mandatory to appear in a client's request. For a
+  "create" operation, this attribute MUST NOT appear. For an "import" type
```
n3wscott, line 252: There is no create in the discovery spec.

duglin: Yep I didn't think I'd find a good way to talk about it in the management API stuff but I did, so I removed this text now.

duglin:
Updated

duglin:
Approved on the 9/24 call

PR #707, opened 2020-09-25
[Discovery] import and bulk POST support

duglin:
Changes include:
- tweaked async text to say that the response to the async's status URL should look the same as the sync response on the original request
- added support for POST to take a list of services instead of just one
- added explicit support for "import" via a "?import" query param - split each into their own section so we don't confuse people between add/update and "import". Did this for both POST and PUT


discovery.md
```diff
@@ -600,83 +604,233 @@ A `GET` MAY be sent to the status endpoint to determine the status of the
...
+the presence of the `id` and `epoch` attributes are OPTIONAL. If either is
+present then those value MUST be preserved by the Discovery Endpoint and
+not modified as a result of this operation. If they are not present then the
+Discovery Endpoint MUST assign appropriate values.
+
+If the Discovery Endpoint is unable to successully add all of the Services
+in the incoming request then an error MUST be generated and none of the
+specified Services are to be added, or updated, in the Discovery Endpoint.
+
+The Services in the request message MUST be processed in the order in which
```
duglin, line 690: please read this carefully

duglin: while I do think this is useful, and maybe even necessary in a real-world scenarios, bulk updates are always problematic and potentially complex.

duglin:
Approved on the 10/8 call

PR #709, opened 2020-09-30
docs: do not use code ticks for data that may actually be data_base64

grant:
The spec currently uses ticks around `data` as if this is a literal field.

This isn't quite accurate, we don't mean the literal key `data`. We state the data can be stored in `data` or `data_base64 ` right after, so it's a little confusing saying `data` before. I suggest removing the ticks to make this clearer.

See the diff to understand the change the best.

Related: https://github.com/cloudevents/spec/issues/520

### DIFF
--- a/json-format.md
+++ b/json-format.md
@@ -123,12 +123,12 @@ the [type system mapping](#22-type-system-mapping).
 Before taking action, a JSON serializer MUST first determine the runtime data
 type of the `data` content.
 
-If the implementation determines that the type of `data` is `Binary`, the value
+If the implementation determines that the type of data is `Binary`, the value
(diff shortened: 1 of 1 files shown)


clemensv:
That looks like a good clarification. LGTM.

grant:
Cool. I believe we discussed this today and are fine with this micro PR? @duglin

duglin:
While this came in after the Tuesday deadline, it's a fair small (typo-ish) change so I'll wait until EOD Friday (tomorrow) before merging. If anyone has any concerns please don't hesitate to speak up.

duglin:
Approved on the 10/1 call

Issue #703, opened 2020-09-23
HTTP webhook signature-based proof-of-origin


alexec:
Currently the web hook spec mandates an `Authorization` header. However, many common web hooks use signature-based header. This is typically a header that contains a hashed digest of a secret and the request payload. This acts as proof of origin. Unfortunately, this is in now way standardised. Github, Githlab, BitBucket, Stripe all use different headers and algos meaning you must implement X and X as many bugs and security issues. Nor does it mandate any of these systems abide by best security practices, use the best crypto, future proof themselves (i.e. how does one migrate Github from RSA to ECC - impossible?).

Because signature-based proof-of-origin is common, CloudEvents can and should standardise this.

Some ideas:

* The body is just a JWT where the payload is the data?
* The signature is a JWT?
* Standard header names `Signature`,  `Signature-Alg` for the algo, `Signature-Type` for the type .
* Standard error responses for unsupported algos.

References::

* https://developer.github.com/webhooks/securing/
* https://github.com/cloudevents/spec/blob/master/http-webhook.md

RichiCoder1:
There's a pretty long running spec draft for request signatures in the IETF, but it's focused on specifically HTTP Request Signing: https://www.ietf.org/id/draft-ietf-httpbis-message-signatures-00.html

That said, it could also provide some inspiration here.

clemensv:
The `Authorization` header is a generic mechanism defined in https://tools.ietf.org/html/rfc7235#section-4.2 

There are no restrictions to what you can put into that field, it only needs to work for both client and server. The upside of using "Authorization" is that RFC7235 prescribes that the header must not be cached, which custom schemes do not.

The point of the WebHooks spec is to constrain the HTTP binding further in order to provide broader interop. Accommodating arbitrary, proprietary AuthNZ schemes would be contra-productive. 

At the same time, I don't see us being in the security business. OAuth2 bearer tokens with TLS seems to be where the security folks have largely landed for securing these kinds of interactions and that's why we lean on that.

alexec:
I'm not sure about this "authorization" and "signature" are semantically different. A message signature would not, in itself, be an authorization. And there is clearly a use case for signatures instead of authorization, otherwise it would not be more common that authorization. 

That said, I would expect the RFC to cover all signature use cases and CloudEvents could just point to that.

manuelstein:
[...]

Issue #379, opened 2019-01-29
Event Confidentiality


erikerikson:
As part of #276, protections of the data in an event can be lost once the connection across which the event is sent gets terminated.  Does partial or complete encryption of the event solve this?  Are there means to specify the protections required for data that touches disk?  Are solutions for addressing event confidentiality in scope for CloudEvents?

clemensv:
End-to-end encryption is a battlefield I would urge the group to stay out of. I realize that the requirement is real, but binding an encryption and signing model to this standard has a good chance of blowing a hole into the ship and sinking it. 

Exhibit A: SOAP and WS-Security.

There are several options for end-to-end security. A comprehensive recent framework is Javascript Object Signing and Encryption (JOSE: JWS/JWE/JWT), which includes JWE https://tools.ietf.org/html/rfc7516 and JWS https://tools.ietf.org/html/rfc7515. 

S/MIME is another option.

In messaging and eventing we largely see proprietary solutions today, also due the fact that key management is not trivial for one-way messaging without the ability to negotiate session keys, because of the degradation of key strength through frequent use, and because the receiver needs to be able to get at the required key material and through several layers of routing that might be a nontrivial challenge.

As I would love to see a standards-based model for messaging/eventing all up that builds on something like JOSE or S/MIME, I absolutely would want to avoid diving down into that rabbit hole in the core CloudEvents effort, certainly not before version 1.0.

JemDay:
From a  cloud-events perspective i would hope we can make statement along the lines of:

_"... sensitive information SHOULD NOT be placed in event context attributes .."_
_"... data SHOULD be encoded/encrypted when sensitive information is present in the payload"_

I don't think we need to into the weeds of the end-to-end mechanisms at present.

duglin:
@erikerikson two questions for you:
1 - if we address https://github.com/cloudevents/spec/issues/373, will that cover this issue?
2 - what do you think of this proposal: https://github.com/cloudevents/spec/issues/373#issuecomment-470351631  ?

erikerikson:
1 - yes, sorry for duplication
2 - broadly good proposal, response will come there

JemDay:
See PR #399

duglin:
Per the above discussion, and the closing of #373 (and merging of #399), I believe we can close this. If I'm mistaken please let me know and I'll reopen it.

PR #712, opened 2020-10-19
[Primer] Adding a Non-Goal w.r.t Security

anishj0shi:
in #703 it was asked to mandate certain fields in the spec to enhance the security model, However based on the discussion we all came to a conclusion that the spec doesn't have to get into the security business. Hence, the Group decided to add this information into the Primer.

### DIFF
--- a/primer.md
+++ b/primer.md
@@ -140,6 +140,7 @@ The following are considered beyond the scope of the specification:
 - Selecting a single identity/access control system
 - Inclusion of protocol-level routing information
 - Event persistence processes
+- Mechanism for Authorization, Data Integrity and Confidentiality
 
 The CloudEvents spec will not include protocol-level routing information (e.g.
 a destination URL to which the event is being sent). This is a common suggestion
@@ -173,6 +174,15 @@ available to attackers and thereby reduces security. It is expected that
 attributes may be defined that facilitate meeting persistence requirements but
 it is expected that these will continuously evolve along with industry best
 practices and advancements.
+
+The CloudEvents spec currently does not mandate or advocates any specific
+mechanism or principles in the matter of Authorization, Data Integrity and
+Confidentiality, as the core intention behind this spec is not to define
+Security principles with regards to CloudEvents. Every implementor has a
+different principle for enhancing their security model. We leave it up to the
+implementor of the spec to provide additional details for hardening their
+security model as an extension field, which can be furthermore interpreted by
+the components implemented by the implementor of the specification themselves.
 
 ## Architecture
 
--- a/primer.md
+++ b/primer.md
@@ -175,14 +175,18 @@ attributes may be defined that facilitate meeting persistence requirements but
 it is expected that these will continuously evolve along with industry best
 practices and advancements.
 
-The CloudEvents spec currently does not mandate or advocates any specific
+The CloudEvents spec currently does not mandate or advocate any specific
 mechanism or principles in the matter of Authorization, Data Integrity and
-Confidentiality, as the core intention behind this spec is not to define
+Confidentiality, as the current intention behind this spec is not to define
 Security principles with regards to CloudEvents. Every implementor has a
 different principle for enhancing their security model. We leave it up to the
 implementor of the spec to provide additional details for hardening their
 security model as an extension field, which can be furthermore interpreted by
 the components implemented by the implementor of the specification themselves.
+However, if the community observes a pattern in usage of certain extension
+fields, as a standard way to deal with the topic of data integrity. In that
+case, such extension fields can be declared as official extension to the
+CloudEvent specification.
 
 ## Architecture
 


primer.md
```diff
@@ -140,6 +140,7 @@ The following are considered beyond the scope of the specification:
 - Selecting a single identity/access control system
 - Inclusion of protocol-level routing information
 - Event persistence processes
+- Vendor specific Authorisation principles
```
duglin, line 143: I think we should rat-hole on "Authorisation" vs "Authorization"  :-)

primer.md
```diff
@@ -174,6 +175,13 @@ attributes may be defined that facilitate meeting persistence requirements but
 it is expected that these will continuously evolve along with industry best
 practices and advancements.
 
+The Cloudevents spec will not mandate any vendor specific Authorisation model, As the
```
duglin, line 178: - Can you wrap these at 80 columns?
- s/, As/, as/

primer.md
```diff
@@ -174,6 +175,13 @@ attributes may be defined that facilitate meeting persistence requirements but
...
+core intention behind this spec is not to define Security principles with regards to Cloudevents. 
+Every vendor has a different principle for enhancing their security model. We
+leave it up to the implementor of the spec on providing additional details for hardening
```
duglin, line 181: s/on/to/

primer.md
anishj0shi: *Authorization* it is!

primer.md
anishj0shi: done

primer.md
anishj0shi: done

primer.md
```diff
@@ -174,6 +175,14 @@ attributes may be defined that facilitate meeting persistence requirements but
 it is expected that these will continuously evolve along with industry best
 practices and advancements.
 
+The Cloudevents spec will not mandate any vendor specific Authorisation model,
```
bryeung, line 178: CloudEvents (capital E) here and below?

eclipselu: s/Authorisation/Authorization

primer.md
```diff
@@ -140,6 +140,7 @@ The following are considered beyond the scope of the specification:
 - Selecting a single identity/access control system
 - Inclusion of protocol-level routing information
 - Event persistence processes
+- Vendor specific Authorization principles
```
erikerikson, line 143: I would suggest that vendor specificity is less direct than the concerns of [authorization](#378), [integrity](#143), and [confidentiality](#379).  I would recommend directly identifying those concerns.  Consider that open source standards can be unattached to a vendor but that we are not now declaring a correct mechanism for satisfying those requirements.

As a side note, recall that we support satisfying those concerns in more standardized forms through informal extensions or in individualized forms through the data payload.

Side note: the potential mention I declared during the call was tied up in the persistence comment and did not directly address this concern.  Apologies for that noise.

primer.md
anishj0shi: done

primer.md
anishj0shi: I've rephrased the entire thing. please have a look.

primer.md
```diff
@@ -174,6 +175,15 @@ attributes may be defined that facilitate meeting persistence requirements but
 it is expected that these will continuously evolve along with industry best
 practices and advancements.
 
+The CloudEvents spec currently does not mandate or advocates any specific
+mechanism or principles in the matter of Authorization, Data Integrity and
+Confidentiality, as the core intention behind this spec is not to define
```
erikerikson, line 180: I totally agree with the sentiment here.  I appreciate you putting in the effort to craft its statement - thank you.  Regarding this specific wording, I'm uncomfortable with the statement that the core intention of the spec excludes authorization/integrity/confidentiality.

I would suggest that those are core concerns but also that the diversity and necessary ongoing evolution of valid user preferences in that decision space produces the circumstance that while we allow for and perhaps help guide users to solve those requirements we are also not yet specifying the mechanisms they should or must use for solving them.  I would guess that current consensus (as you capture) expects that more standardized solutions for these concerns will enter the spec as unofficial and then official extensions that may never reach the core spec unless the whole industry can permanently agree on a single set mechanisms for these needs.  I rate the probability of that as approaching zero though so that leaves us with where we are.

primer.md
```diff
@@ -174,6 +175,15 @@ attributes may be defined that facilitate meeting persistence requirements but
...
+Security principles with regards to CloudEvents. Every implementor has a
+different principle for enhancing their security model. We leave it up to the
```
erikerikson, line 182: I appreciate that this speaks to the individual reading the primer.  It might be improved by also speaking to the community around that user.  The best choice for the reader is probably in the context of the community's offerings of concrete extensions and more abstract techniques so while the use of extensions seems like good guidance, a statement of the expectation that extensions will emerge might be a good nudge toward discovering what is on offer by the particular integrator or the community at large.

erikerikson (commented):
Sorry this doesn't offer more help.  Writing these can be a bear so thank you for your effort.

duglin:
Any update on this?

anishj0shi:
> Any update on this?

I wanted to talk about this in the next WG meeting.. I'm not sure what else I can add to this. so a brush up on the discussion might be helpful.

primer.md
anishj0shi: How about if we state that the **Currently, the core intention.....and is subject to change in the future**

primer.md
anishj0shi: Any suggestions on how can we improve the wording to address it to the community instead of the user?

primer.md
erikerikson: The core intention of the spec, to me, includes supporting users in satisfying their authorization, integrity, and confidentiality requirements.  This stands in contrast with its current state which neither offers nor mandates any mechanisms for satisfying those requirements.  FWIW, I am a little concerned I am being overly pedantic and I would be happy enough to back off the point.

primer.md
erikerikson: I like that it was addressing the individual.  I intended to suggest it was forgoing an opportunity to suggest a community opportunity and to direct users in looking to what the community will offer over time.  I have in mind a tiered offering of 1) attributes defined in the spec, 2) official extensions 3) unofficial but standardized extensions 4) individually defined/non-standard extensions.  You are saying we are at 4 but it might be worth suggesting 3 exists to be found via a search and that 2 should exist at some point (without making any promises).  I would expect this to encourage people to look which I would expect to be increasingly helpful over time but also provides incentive for offering such things and indicates there is an opportunity to impactfully do so.

Totally informally:
"hey, there should be extensions to find"
"maybe you'd like to make some for implementers to use"
"a business opportunity for hosting a turnkey and portable solution for these problem exists"

I suspect at the root of my comments is the durability of the text over time but mostly I'm trying to be helpful.  I made off comments and Doug skillfully used that to get me to comment.  Feel entirely welcome to disregard, I only mean to suggest possible improvements and if it doesn't feel like these changes would make it better, go with it as is.  :D

primer.md
```diff
@@ -174,6 +175,15 @@ attributes may be defined that facilitate meeting persistence requirements but
 it is expected that these will continuously evolve along with industry best
 practices and advancements.
 
+The CloudEvents spec currently does not mandate or advocates any specific
```
sdatspun2, line 178: or advocate

duglin:
Approved on 11/19 call

Issue #702, opened 2020-09-23
Are null values for attributes permitted?


jskeet:
This is in the context of https://github.com/cloudevents/sdk-csharp/pull/68

I believe the spec should be clear about whether the concept of a "null" attribute value is supported - and then SDKs should be consistent in supporting or prohibiting it.

(My personal feeling is that they should be prohibited everywhere. Supporting them requires encoding that information in all transports etc, which is likely to be really tricky.)

duglin:
Are there specific attributes you have in mind?  We tried to add text like "MUST be a non-empty string" and "If present, MUST be a non-empty URI" to deal with that.

jskeet:
No specific attributes, no - and it's not about empty vs non-empty, either... it's about whether an attribute can be expressed with the absence of a value... a null reference in Java or C#, basically.

I'd argue that if the concept of expressing "this attribute is definitely not present" isn't supported by the overall spec, then SDKs for languages with the concept of null should either prohibit its use, or use it as an alternative way of removing an attribute. (I'd personally far prefer the former. If you want to remove an attribute from an in-language CloudEvent, call a suitable "remove" function. That leaves less room for ambiguity.)

duglin:
This sounds a bit like an SDK issue, no?  I say this because CE itself doesn't have the concept of "removing an attribute"... there's nothing to remove an attribute from. It's either in the CE on the wire, or it's not. There's no state involved here. Or am I not following?

jskeet:
It's an SDK issue, but one which should IMO be informed in the spec - or at least the SDK spec - so that it's handled consistently.

I agree that the CE itself doesn't have the concept of removing an attribute - but if we're explicitly about saying that it *also* doesn't have the concept of "an attribute with no value" then that makes it clearer that you shouldn't have a null value (which is effectively the absence of a meaningful value) in the SDK representation.

deissnerk:
@jskeet I agree with you, and I remember having that discussion with @duglin and @clemensv on Slack quite a while ago, when preparing the demo for KubeCon EU 2019. I think we agreed that this should be clarified but forgot to follow-up. 😄 
I suppose it could be clarified in the section describing the [type system](https://github.com/cloudevents/spec/blob/master/spec.md#type-system).

duglin:
[...]

PR #713, opened 2020-10-19
Allow JSON values to be null, meaning unset.

n3wscott:
Fixes: https://github.com/cloudevents/spec/issues/702

### DIFF
--- a/json-format.md
+++ b/json-format.md
@@ -70,6 +70,10 @@ with exceptions noted below.
 | URI-reference | [string][json-string] following [RFC 3986][rfc3986]            |
 | Timestamp     | [string][json-string] following [RFC 3339][rfc3339] (ISO 8601) |
 
+Unset attributes MAY be encoded to the JSON value of `null`. When decoding
+attributes and a `null` value is encountered, it MUST be treated as the
+equivilent of unset or omitted.
+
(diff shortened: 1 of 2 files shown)


json-format.md
```diff
@@ -70,6 +70,10 @@ with exceptions noted below.
 | URI-reference | [string][json-string] following [RFC 3986][rfc3986]            |
 | Timestamp     | [string][json-string] following [RFC 3339][rfc3339] (ISO 8601) |
 
+Unset attributes MAY be encoded to the JSON value of `null`. When decoding
+attributes and a `null` value is encountered, it MUST be treated as the
+equivilent of unset or omitted.
+
```
deissnerk, line 76: LGTM
@clemensv  Do we need a similar rule for [AMQP Binary Mode](https://github.com/cloudevents/spec/blob/master/amqp-protocol-binding.md#3132-amqp-application-property-values)? Strictly speaking, `null` values are not allowed there, as `null` in AMQP is a primitive type, and the type mapping does not mention it.

duglin:
LGTM

json-format.md
n3wscott: IDK, @clemensv might know better, if it is a problem.

PR #717, opened 2020-10-25
tweak url description

duglin:
Minor wording change to ensure that people know that this URL points to the Service entry in the Discovery Endpoint and NOT to the Service itself


duglin:
Approved on the 10/29 call

PR #718, opened 2020-10-25
Minor tweaks in prep for interop

duglin:
- fixed psuedo schema in discovery - config has 'types' not 'values
    - typos
    - clarify that URL points to the DE entry and not the actual service
    - show sample JSON serializaton of a Subscription
    - outline for HTTP REST API
    - move "dialect" under "filter"
    - s/filter/filters/
    - make "dialect" an URI-reference
    - add "subscriptiondialects" to the DE spec

### DIFF
--- a/subscriptions-api.md
+++ b/subscriptions-api.md
@@ -264,7 +264,8 @@ how many subscriptions are supported is implementation specific.
 Each subscription is represented by an object that has the following properties:
 
 - **id** (string) – REQUIRED. The unique identifier of the subscription in the
-  scope of the subscription manager.
+  scope of the subscription manager. This value MUST be unique within the scope
+  of the subscription manager and MUST be immutable.
 - **protocol** (string) - REQUIRED. Identifier of a delivery protocol. Because
   of WebSocket tunneling options for AMQP, MQTT and other protocols, the URI
   scheme is not sufficient to identify the protocol. The protocols with existing
@@ -280,7 +281,7 @@ Each subscription is represented by an object that has the following properties:
   using the selected protocol. The format of the address MUST be valid for the
   selected protocol or one of the protocol's own transport bindings (e.g. AMQP
   over WebSockets).
-- **filter** - OPTIONAL. A filter is an expression of a particular filter
+- **filters** - OPTIONAL. One or more filter expressions of a particular filter
   dialect that evaluates to true or false and that determines whether an
   instance of a CloudEvent will be delivered to the subscription's sink. If a
   filter expression evaluates to false, the event MUST NOT be sent to the sink.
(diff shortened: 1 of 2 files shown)


duglin:
ping @clemensv

subscriptions-api.md
```diff
@@ -280,7 +281,7 @@ Each subscription is represented by an object that has the following properties:
   using the selected protocol. The format of the address MUST be valid for the
   selected protocol or one of the protocol's own transport bindings (e.g. AMQP
   over WebSockets).
-- **filter** - OPTIONAL. A filter is an expression of a particular filter
+- **filters** - OPTIONAL. One or more filter expressions of a particular filter
```
duglin, line 284: Keep it as one

subscriptions-api.md
```diff
@@ -444,14 +471,12 @@ This filter will select events only with the event type equal to
 
 ```JSON
 {
+  "filters": [{
     "dialect": "basic",
-    "filters": [
```
deissnerk, line 448: I think this should be `conditions`. So there is just a single filter, but a filter could have multiple `conditions`. So it is not possible to mix dialects but to have multiple `conditions` combined with an implicit logical AND.

subscriptions-api.md
deissnerk: It's one filter, but the filter may contain multiple conditions, as described [here](https://github.com/cloudevents/spec/blob/master/subscriptions-api-openapi.yaml). I think the confusion results from an inconsistency in [3.2.3.1](https://github.com/cloudevents/spec/blob/master/subscriptions-api.md#3231-basic-filter-dialect), where `filters` instead of `conditions` is used.

duglin:
Added `susbcriptiondialect` to the Discovery metadata - so the subscriber knows which dialects are supported.

subscriptions-api.md
```diff
@@ -289,7 +290,35 @@ Each subscription is represented by an object that has the following properties:
...
+
+```JSON
+{
+  "id": "sub-193-18365",
+  "protocol": "HTTP",
+  "protocolsettings": {
+    "method": "POST"
+  },
+  "sink": "http://example.com/event-processor",
+  "filters": [{
```
deissnerk, line 310: Originally it was supposed to be
```
"filter":{
  "dialect":"basic",
  "conditions":[
    {
      "type":"prefix",
      "property":"type",
      "value":"com.example"
    }
  ]
}

```

This limits a subscription to the use of a single filter dialect. The fact that there may be multiple `conditions` is specific to the filter dialect `basic`.

duglin: I'd prefer if we didn't limit it to just one dialect per Subscription. It feels really limiting for a Subscription Manager to support 5 dialects but then the user can only use one at a time.I don't see the benefit to either side of that.

subscriptions-api.md
duglin: see above

subscriptions-api.md
deissnerk: Well, it might make it easy to users to create expensive and/or inefficient subscriptions. As an infrastructure developer I usually lean towards being restrictive first and keep the option to broaden the scope later. 😄 
@clemensv What do you think?

subscriptions-api.md
```diff
@@ -496,9 +539,10 @@ supported.
 
 The **Create** operation SHOULD be supported by compliant Event Producers. It
 creates a new Subscription. The client proposes a subscription object which MUST
-contain all REQUIRED properties. The subscription manager then realizes the
-subscription and returns a subscription object that also contains all OPTIONAL
-properties for which default values have been applied.
+contain all REQUIRED properties with the exception of the `ID` property, which
+will be defined by th subscription manager. The subscription manager then
```
duglin, line 543: e

subscriptions-api.md
```diff
@@ -605,6 +649,52 @@ Errors:
...
+
+{
+  "protocol": "...",
+  "protocolsettings": { ... }
+  "sink": "...",
+  "filters": [{
+    "dialect": "...",
+    "type": "...",
+    "property": "...",
+    "value": "i..."
```
duglin, line 666: i

duglin:
Approved on the 11/05 call

PR #720, opened 2020-10-28
refactor: change subscription API to resource based

loopingz:

Current spec is using GET/DELETE/PUT/POST on /subscriptions with id url parameter or id in body.
This PR move to:
GET /subscriptions: List subscriptions
POST /subscriptions: Create subscription
GET /subscriptions/{id}: Retrieve a subscription
PUT /subscriptions/{id}: Update a subscription
DELETE /subscriptions/{id}: Delete a subscription

Some questions popped while working on it: 
 - in my opinion, the proxy settings should be the application choice, not the subscription choice.
 - the creation request should not contain an id, as the id should be generated by the application (so I added a SubscriptionRequest schema)

Screen Shot 2020-10-27 at 5 39 26 PM

### DIFF
--- a/subscriptions-api-openapi.yaml
+++ b/subscriptions-api-openapi.yaml
@@ -7,55 +7,25 @@ servers:
 paths:
   /subscriptions:
     get:
-      description: Retrieve one or multiple subscriptions
-      parameters:
-        - in: "query"
-          name: "id"
-          description: identifier of a specific subscription
-          required: false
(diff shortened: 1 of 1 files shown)


duglin:
FYI I think this goes with the other PR #718 which moves the subscription ID into the PATH (away from the query param)

duglin:
Approved on the 11/05 call

PR #721, opened 2020-10-29
Add protobuf format as a sub-protocol

JemDay:
Tweak the WS protocol binding to support protobuf fomat.


duglin:
@JemDay while this PR is ok - I just noticed that the proto spec is listed as v1.0 and I'm wondering if it should be something less than v1.0 until it's had time to be tested? Once we give it the official v1.0 label we can't make any breaking changes.

duglin:
Approved on the 11/05 call

PR #722, opened 2020-10-30
schema registry replication

clemensv:
This update adds change notification and replication support to the schema registry.

### DIFF
--- a/schemaregistry/schemaregistry.md
+++ b/schemaregistry/schemaregistry.md
@@ -163,8 +163,8 @@ prior schema versions are retained. This specification does not mandate a
 retention policy, but implementations MAY retire and remove outdated schema
 versions.
 
-The latest available schema is always the default version that is retrieved when
-the URI to the schema is used without the version having been specified.
+The latest available schema MUST always the default version that is retrieved
+when the URI to the schema is used without the version having been specified.
 
 ### 2.2.1 Schema authority
 
@@ -173,8 +173,8 @@ is a person or group of people or system that controls the schema. For the most
 common cases, the authority is whoever manages the schemas in a registry.
 
 For the [replication model](#4-replication-model-and-state-change-events), the
-authority information is used to disambiguate schemas from different origins
-that need to be consolidated in central schema registries or in caches.
+authority information MUST be used to disambiguate schemas from different
+origins that need to be consolidated in central schema registries or in caches.
 
 In complex business systems, the authority might be centralized and schemas are
 explicitly designed, reviewed and approved for use. In simpler scenarios, the
@@ -216,7 +216,7 @@ address, as permitted for the authority URI in the previous section.
 
 In those scenarios, an event consumer SHOULD be configured with a fixed schema
 registry endpoint for it to use, and obtain schemas identified, for example, in
-the [dataschema](../spec.md#dataschema) attruibute, using the
+the [dataschema](../spec.md#dataschema) attribute, using the
 ["getSchemaVersionByURI"](#342-get-a-specific-schema-version) API operation on
 that endpoint, rather than trying to resolve the URI directly.
 
@@ -500,10 +500,16 @@ The HEAD method SHOULD also be implemented.
 
 #### 3.4.3. Get a specific schema version by schema version URI
 
-Each schema version has a URI, as defined in [2.2.2](#222-schema-version-uri).
+Each schema version has a URI as a unique identifier, as defined in
+[2.2.2](#222-schema-version-uri). This URI can be used as a lookup key on any
+schema registry that holds a copy of that schema version.
 
-This operation is a GET on `/schema?uri={uri}`, with the required parameter
-being the schema version URI.
+As discussed in 2.2.2, the URI might not be network resolvable or the network
+location may not be reachable from everywhere. That is a key motivation for
+replication.
+
+The operation to obtain a schema version is a GET on `/schema?uri={uri}`, with
+the required parameter being the schema version URI.
 
 The returned payload is the schema document. Further attributes such as the
 `description` and the `format` indicator are returned as HTTP headers.
@@ -521,8 +527,8 @@ quite common that event producers and event consumers do not share the same
 network scope.
 
 In complex systems made up of subsystems owned by different organizations, the
-subsystem owners might also want to decouple the systems such that one or not
-dependent on the other system's availability whereever possible, and therefore
+subsystem owners might also want to decouple the systems such that they are not
+dependent on the other system's availability wherever possible, and therefore
 introduce redundant footprint for metadata stores such as a schema registry.
 
 We therefore cannot assume that an event consumer has access to the same
@@ -593,11 +599,20 @@ This replication model uses the registry API as defined above on the target
 registry and the configuration of what aspects of the registry are pushed are
 implementation specific.
 
+The names of the schema groups and of the schemas MAY mirror those from the
+source registry, but they MAY also differ. Each schema version is always
+accessible under both [its local path a given
+registry](#342-get-a-specific-schema-version) and under its globally unique
+[schema version URI](#343-get-a-specific-schema-version-by-schema-version-uri).
+
 ### 4.4 Pull replication
 
 In an imperative "pull" replication model, a target registry will once or periodically
 traverse the source registry contents and apply differences between the source
 state and its own state.
+
+The names of the schema groups and of the schemas MAY mirror those from the
+source registry, but they MAY also differ.
 
 The pulling registry might be configured to copy changes of the full registry, of
 a specific schema group, or just of a specific schema.
--- a/schemaregistry/schemaregistry.md
+++ b/schemaregistry/schemaregistry.md
@@ -1,10 +1,13 @@
-# CNCF Schema Registry API Version 0.1-wip
+# CNCF Schema Registry API Version 0.2-wip
 
 ## Abstract
 
 This specification defines a simple API for storing, organizing, and accessing
 schema documents for use with serialization and data validation frameworks such
 as Apache Avro and JSON Schema.
+
+It also defines a set of CloudEvent event type definitions and topology models
+for synchronizing schemas across different schema registries.
 
 ## Status of this document
 
@@ -62,7 +65,7 @@ schema groups. For instance, a user or a group of users might be given write
 access only to a particular schema group that their organization owns. If trade
 secret protection is required for an application or parts of it and the schema
 structure would give some of those away, read access to a group of schemas might
-likewise be restricted. 
+likewise be restricted.
 
 Access control rules at the schema group level MUST NOT limit visibility of the
 schema groups themselves to users authorized to enumerate schema groups at the
@@ -78,7 +81,7 @@ rules.
 
 The data model for a schema group consists of these attributes:
 
-#### id
+#### `id` (schema group id)
 
 - Type: `String`
 - Description: Identifies the schema group.
@@ -95,7 +98,7 @@ The data model for a schema group consists of these attributes:
   - org.example.myapp.module
   - events@example.com
 
-#### format
+#### `format` (schema group format)
 
 - Type: `String`
 - Description: Defines the schema format managed by this schema group. The
@@ -112,7 +115,7 @@ The data model for a schema group consists of these attributes:
   - "json-schema"
   - "xsd11"
 
-#### description
+#### `description` (schema group description)
 
 - Type: `String`
 - Description: Explains the purpose of the schema group.
@@ -121,7 +124,7 @@ The data model for a schema group consists of these attributes:
 - Examples:
   - "This group holds schemas for the fabulous example app."
 
-#### createdtimeutc
+#### `createdtimeutc` (schema group created time)
 
 - Type: `Timestamp`
 - Description: Instant when the schema group was added to the registry.
@@ -129,9 +132,9 @@ The data model for a schema group consists of these attributes:
   - OPTIONAL
   - Assigned by the server.
 
-#### updatedtimeutc
+#### `updatedtimeutc` (schema group updated time)
 
-- Type: `Timestamp` 
+- Type: `Timestamp`
 - Description: Instant when the schema group was last updated
 - Constraints:
   - OPTIONAL
@@ -143,14 +146,10 @@ Conceptually, a schema is a description of a data structure. Since data
 structures evolve over time, the schema describing them will also evolve over
 time. Therefore, a schema often has multiple versions.
 
-For simple scenarios, the API allows for version management to be automatic and
-transparent. Whenever a schema is updated, a new version number is assigned and
-prior schema versions are retained. This specification does not mandate a
-retention policy, but implementations MAY retire and remove outdated schema
-versions.
-
-The latest available schema is always the default version that is retrieved when
-the URL to the schema is given without the version specified.
+In this specification, the ["schema version"](#225-schema-version-attributes)
+refers to an individual schema document. The ["schema"](#224-schema-attributes)
+is a management bracket for those documents that helps enforcing consistency
+across versions, including compatibility policies.
 
 A newer schema version might introduce breaking changes or it might only
 introduce careful changes that preserve compatibility. These strategies are not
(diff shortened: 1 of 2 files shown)


schemaregistry/schemaregistry.md
```diff
@@ -404,4 +496,195 @@ The returned payload is the schema document. Further attributes such as the
...
+        "value": "io.cloudevents.schemaregistry.update"
+    },{
+        "type": "prefix",
+        "property": "subject",
+        "value": "/schemagroups/mygroup"
+    }
+}
+```
+
+#### io.cloudevents.schemaregistry.update
```
duglin, line 640: "update" or "updated" ?
I seem to remember that we changed: https://github.com/cloudevents/spec/blob/master/adapters/couchdb.md to put things into the past-tense. Should we here too?

schemaregistry/schemaregistry.md
```diff
@@ -404,4 +496,195 @@ The returned payload is the schema document. Further attributes such as the
...
+
+This event notifies the subscriber of an element of the registry having been
+created or updated. Because schema documents might be very large, and because
```
duglin, line 643: Any reason to not have two different events, one for "created" and one for "updated" ? Worried about events arriving out of order?

schemaregistry/schemaregistry.md
```diff
@@ -404,4 +496,195 @@ The returned payload is the schema document. Further attributes such as the
...
+disclose trade secrets, the event carries no data, but requires the consumer to
+fetch the indicated change from the registry.
+
+* **source**: Base URI of the schema registry
+* **eventType**: `io.cloudevents.schemaregistry.update`
+* **subject**: [API path](#31-path-hierarchy) of the object that has been created or updated
+* **time**: Time of the change, exactly as recorded in `updatedtimeutc` (or
+  `createdtimeutc`for schema versions)
+
+`id` MUST be set to a unique value relative to the scope formed by these three
```
duglin, line 654: which 3 fields?I assume you mean source, type and subject.  Should probably be explicit

famartinrh:
how does the schema authority affect the schema listing?
I mean, in a replication scenario it could be possible to have schemas from different authorities but the same group id and schema id, will the schema groups collide? should maybe the authority concept be reflected at a higher level concept in the schema registry API hierarchy?.
I understand the whole point of the new authority concept, and I think it's pretty useful as it allows for the creation of schema hubs for safely sharing schemas, and of course it should help in gathering schemas from different sources in replication scenarios... 
I would suggest to continue exploring this because I see some design flaws, for instance, the inclusion of the new `getSchemaVersionByURI` API operation plus the dual logic proposed to obtain schemas identified by the [dataschema](https://github.com/cloudevents/spec/blob/master/spec.md#dataschema) attruibute look unnecessarily complex.

schemaregistry/schemaregistry.md
n3wscott: +1 to updated.

n3wscott, line 640: We should add `.v1` to this and all other event types.

schemaregistry/schemaregistry.md
```diff
@@ -404,4 +496,195 @@ The returned payload is the schema document. Further attributes such as the
...
+to subscribe to change events.
+
+The subscription's scope MUST be root of the schema registry. A [prefix
+filter](../subscriptions-api.md#3231-basic-filter-dialect) can be used to to
+scope the subscription to a particular schema group or schema.
+
+The subscribing target registers a notification sink of its choosing.
+
+```HTTP
+POST https://source.example.com/subscriptions HTTP/1.1
```
n3wscott, line 618: As a result of this post, we should also add an event type `io.cloudevents.schemaregistry.subscribed.v1` to give direct feedback when the subscription is registered and the subscription does not filter out that kind of event type.

schemaregistry/schemaregistry.md
```diff
@@ -143,33 +146,102 @@ Conceptually, a schema is a description of a data structure. Since data
...
+For simple scenarios, the API allows for version management to be automatic and
+transparent. Whenever a schema is updated, a new version number is assigned and
+prior schema versions are retained. This specification does not mandate a
+retention policy, but implementations MAY retire and remove outdated schema
+versions.
+
+The latest available schema is always the default version that is retrieved when
+the URI to the schema is used without the version having been specified.
+
+### 2.2.1 Schema authority
```
famartinrh, line 169: WDYT about removing the Schema Version URI concept and make the schema authority a parameter in `getSchemaVersion` api operation, or just create an api operation where authority is a explicit parameter...

Here two examples showing the CloudEvent `dataschema` attribute and how I suggest the Schema Authority could be implemented.

##### Example 1 (explicit authority)
`dataschema: corp.example.com/mygroup/schemas/myschema/versions/2`

`authority: corp.example.com`
`schema-group: mygroup`
`schema-id: myschema`
`schema-version: 2`

##### Example 2 (implicit authority)
`dataschema: /mygroup/schemas/myschema/versions/2`

`authority: `(not specified - schema registry should return it's default implicit authority)
`schema-group: mygroup`
`schema-id: myschema`
`schema-version: 2`

##### Option 1
both cases will use `getSchemaVersion` api operation 
`getSchemaVersion` will accept a new header in which authority can be specified

##### Option 2
add a new api operation with request path like:
`authorities/{authority}/schemagroups/{group-id}/schemas/{schema-id}/versions/{version-number}`
for cases when no authority is specified in the dataschema a `default` authority could be used

schemaregistry/schemaregistry.md
```diff
@@ -404,4 +496,195 @@ The returned payload is the schema document. Further attributes such as the
...
+POST https://source.example.com/subscriptions HTTP/1.1
+Content-Type: application/cloudevents-subscription+json
+
+{
+  "id" : "mysubscription",
+  "scope": "https://source.example.com",
+  "protocol" : "HTTPS",
+  "sink" : "https://target.example.com/notifications",
+  "filter" : {
+    "dialect": "basic",
```
duglin, line 627: This needs to be changed based on the recent PR to the sub spec.

schemaregistry/schemaregistry.md
clemensv: Makes sense

clemensv: I'll change that.

schemaregistry/schemaregistry.md
clemensv: The semantics are "the registry has been updated, look at the subject to find out what"

Schema versions are append-only and immutable, so that is always "create" and that is the majority of the events that would be raised. Schema groups may be updated. I think "upsert" semantics are sufficient here. "create" vs. "update" really only complicates matters here, IMO.

schemaregistry/schemaregistry.md
clemensv: will fix

schemaregistry/schemaregistry.md
clemensv: I don't understand what you want, but what you want most likely belongs into the subscription spec, @n3wscott  :)

schemaregistry/schemaregistry.md
clemensv: The Schema Version URI exists such that we can have a unique identifier for a schema that works for lookup of the schema independent of which registry contains it: It is a unique identifier. 

"dataschema" must have an absolute URI to reference, because events get forwarded. If you publish into an event broker and that forwards into Kafka and then you have another event broker that also forwards into Kafka, the Kafka consumer has no idea what a relative URI means for "dataschema"

The way you would set up that topology is for there to be a registry that belongs with the Kafka broker and that has replication relationships with those two upstream brokers' registries. Since the upstream event brokers otherwise have no relationship except feeding into the same Kafka brokers, we need to disambuigate the authorities and that needs to be reflected in "dataschema"

schemaregistry/schemaregistry.md
```diff
@@ -246,18 +333,28 @@ within the scope of the registry.
   - 1
   - 2
 
-#### id
+#### `id` (schema version id)
```
famartinrh, line 336: should this be schema version uri? I think it's being mentioned like that in other parts of the document

schemaregistry/schemaregistry.md
```diff
@@ -404,4 +496,195 @@ The returned payload is the schema document. Further attributes such as the
...
+
+Event-driven replication uses CloudEvents to notify targets of changes in the
+source such that the targets can [pull](#44-pull-replication) changes
+immediately and as they become available.
+
+Each registry SHOULD offer a [Subscription API](../subscriptions-api.md)
+endpoint, either directly or using some middleware, to allow replication targets
+to subscribe to change events.
+
+The subscription's scope MUST be root of the schema registry. A [prefix
```
erikerikson, line 611: `A prefix filter` => `Filter(s)`?

schemaregistry/schemaregistry.md
```diff
@@ -404,4 +496,195 @@ The returned payload is the schema document. Further attributes such as the
...
+```HTTP
+POST https://source.example.com/subscriptions HTTP/1.1
+Content-Type: application/cloudevents-subscription+json
+
+{
+  "id" : "mysubscription",
+  "scope": "https://source.example.com",
+  "protocol" : "HTTPS",
+  "sink" : "https://target.example.com/notifications",
+  "filter" : {
```
erikerikson, line 626: Is filter meant to be an array here?  Every other example show  it as one.  Should it be explicitly typed as `array` or `array | object` in the filter definitions section?  Even if the latter, this seems to declare two filters rather than one.  I couldn't find a statement or a schema.

schemaregistry/schemaregistry.md
```diff
@@ -404,4 +496,195 @@ The returned payload is the schema document. Further attributes such as the
...
+}
+```
+
+#### io.cloudevents.schemaregistry.delete
+
+This event notifies the subscriber of an element of the registry having been
+deleted.
+
+* **source**: Base URI of the schema registry
+* **eventType**: `io.cloudevents.schemaregistry.update`
```
erikerikson, line 674: s/update/delete/ here and next line?

erikerikson (commented):
Almost entirely me asking for accomodations to my thickness.

It's not entirely clear how a target registry initializes and correctly transitions to materialization maintenance via a subscription.  I worry about timing and consistency.  Probably subscribing but not processing events, initializing, and then processing events would suffice but this seems like a significant in scope scenario.  I may have just been too thick to catch it.   Also, this is a great start so maybe that sort of thing should come in subsequent PRs.

duglin:
@clemensv any update on this?

clemensv:
@duglin I need help from a Git Jedi to fix the DCO issue.

duglin:
it wants you to sign all commits - best is to just squash and sign the resulting one

duglin:
@clemensv no idea if it's right but I had to add my signature to get past the DCO.  Check to make sure I handled the merge conflicts properly

schemaregistry/schemaregistry.md
```diff
@@ -121,17 +124,17 @@ The data model for a schema group consists of these attributes:
 - Examples:
   - "This group holds schemas for the fabulous example app."
 
-#### createdtimeutc
+#### `createdtimeutc` (schema group created time)
```
duglin, line 127: any reason we have the type (utc) here? We don't have `descriptionstring`  :-)

schemaregistry/schemaregistry.md
```diff
@@ -143,33 +146,102 @@ Conceptually, a schema is a description of a data structure. Since data
...
 
 A newer schema version might introduce breaking changes or it might only
 introduce careful changes that preserve compatibility. These strategies are not
 subject of this specification, but the API provides a conflict handling
 mechanism that allows an implementation to reject updates that do not comply
 with a compatibility policy, if one has been implemented.
 
-### 2.2.1. Multi-format schema groups
+For simple scenarios, the API allows for version management to be automatic and
+transparent. Whenever a schema is updated, a new version number is assigned and
```
duglin, line 161: Do you mean for this to be normative? I wondering if it should be "a new version number MUST be assigned"  ?

schemaregistry/schemaregistry.md
```diff
@@ -143,33 +146,102 @@ Conceptually, a schema is a description of a data structure. Since data
...
+prior schema versions are retained. This specification does not mandate a
+retention policy, but implementations MAY retire and remove outdated schema
+versions.
+
+The latest available schema is always the default version that is retrieved when
```
duglin, line 166: MUST always be the default...  ?

schemaregistry/schemaregistry.md
```diff
@@ -405,3 +497,196 @@ The returned ´Content-Type´ is the same that was passed when the schema versio
...
+
+## 4. Replication Model and State Change Events
+
+Many eventing scenarios require that events cross from private networks into
+public networks (and possibly back into a different private network) and it is
+quite common that event producers and event consumers do not share the same
+network scope.
+
+In complex systems made up of subsystems owned by different organizations, the
+subsystem owners might also want to decouple the systems such that one or not
```
duglin, line 524: typo here?  maybe ... such that they are not dependent...   ?

schemaregistry/schemaregistry.md
```diff
@@ -405,3 +497,196 @@ The returned ´Content-Type´ is the same that was passed when the schema versio
...
+
+1. The Schema Registry API explained in [section 3](#3-http-rest-api) and
+   formally defined in the [OpenAPI spec](schemaregistry.yaml) allows for
+   schema information to be read from a source and written imperatively to a
+   target: A target can pull changes or a source can push changes.
+2. A set of change events, defined in this section, can be emitted by the source
+   whenever any aspect of its state changes, and the target can update the
+   replica of said state by subscribing to these events.
+
+### 4.1 Producer authority or central authority
```
duglin, line 548: If we wanted to support replication for Discovery Endpoints, do you think we'd end up using a similar mechanism to what you've defined in this PR? If so, could these concepts be abstracted into a common spec/doc?

schemaregistry/schemaregistry.md
```diff
@@ -143,33 +146,102 @@ Conceptually, a schema is a description of a data structure. Since data
...
+["getSchemaVersion"](#342-get-a-specific-schema-version) API operation, which is
+convenient for simple usage scenarios.
+
+In [replication scenarios](#4-replication-model-and-state-change-events) where
+those attributes are set, the URI might not correspond to a resolvable network
+address, as permitted for the authority URI in the previous section.
+
+In those scenarios, an event consumer SHOULD be configured with a fixed schema
+registry endpoint for it to use, and obtain schemas identified, for example, in
+the [dataschema](../spec.md#dataschema) attruibute, using the
```
loopingz, line 219: `attruibute` to `attribute`

schemaregistry/schemaregistry.md
```diff
@@ -405,3 +497,196 @@ The returned ´Content-Type´ is the same that was passed when the schema versio
...
+
+#### io.cloudevents.schemaregistry.updated.v1
+
+This event notifies the subscriber of an element of the registry having been
+created or updated. Because schema documents might be very large, and because
+schema information might be subject to special authorization since it might
+disclose trade secrets, the event carries no data, but requires the consumer to
+fetch the indicated change from the registry.
+
+* **source**: Base URI of the schema registry
```
loopingz, line 650: The Base URI is defined as the authority by default.
In the case of replication is it the original authority or the hub authority in `source`?

schemaregistry/schemaregistry.md
```diff
@@ -405,3 +497,196 @@ The returned ´Content-Type´ is the same that was passed when the schema versio
...
+endpoint, either directly or using some middleware, to allow interested parties
+to subscribe to these change events.
+
+The subscription's `source` MUST be the root of the schema registry. A [prefix
+filter](../subscriptions-api.md#prefix-filter-dialect) on the `subject`
+attribute MAY be used to scope the subscription to a particular schema group or
+schema.
+
+The subscribing party registers a notification sink of its choosing.
+
```
erikerikson, line 625: The models all seem to imply a snapshot or ETL type of interaction. A history is not offered nor is a request for "all changes since offset/epoch/timestamp X". Is that intentional?

schemaregistry/schemaregistry.md
```diff
@@ -405,3 +497,196 @@ The returned ´Content-Type´ is the same that was passed when the schema versio
 was registered.
 
 The HEAD method SHOULD also be implemented.
+
+#### 3.4.3. Get a specific schema version by schema version URI
+
+Each schema version has a URI, as defined in [2.2.2](#222-schema-version-uri).
+
+This operation is a GET on `/schema?uri={uri}`, with the required parameter
```
duglin, line 505: I'm probably missing something, but why can't they just use the URI instead of doing it indirectly via the "uri" query string?

duglin:
Conditionally approved on the 5/27 call.  @clemensv will address all open comments/questions by early tomorrow and then we just need approval from the questioners to merge.  @loopingz please look for an update tomorrow and give your LGTM or not.

schemaregistry/schemaregistry.md
clemensv: As discussed, I think we should copy the mechanism, but a standalone abstract spec seems odd.

schemaregistry/schemaregistry.md
clemensv: I hopefully managed to clarify that in 4.3 and 4.4.  Here, it is the BaseURI indeed. It is the path to the "local copy" in the registry raising the event.

clemensv:
@duglin I don't understand the check-in validation complaints.

duglin:
@clemensv can you check to make sure my forced push didn't lose anything?
@loopingz are you ok with the latest round of edits?

schemaregistry/schemaregistry.md
duglin: did you want to update this?

clemensv:
@duglin Thank you for untangling. Let's get this in. I assume the commit lint error is okay?

duglin:
Merging!!

PR #729, opened 2020-11-19
talk about rc releases

duglin:
Talk about how we'll use "rc" release version strings and how we'll never update the "minor" version number.


duglin:
@JemDay are you ok with pushing the protobuf spec back to v1.0-rc1 until it can be fully tested?

duglin:
Approved on the 12/3 call

PR #735, opened 2020-11-28
Fix reference link name.

coryan:


duglin:
@coryan thanks. Can you make the PR against the 'master' branch. By default github picks the v1.0 branch now

PR #739, opened 2020-11-30
docs: add a note about data to the primer

lance:
This commit adds some text around a CloudEvent's data attribute. Specifically,
it highlights that the data may be encoded as base64 for transmission across
the wire, but is otherwise left unchanged for the event consumer.

Signed-off-by: Lance Ball <lball@redhat.com>

### DIFF
--- a/primer.md
+++ b/primer.md
@@ -289,6 +289,28 @@ context, for the purposes of this CloudEvent attribute those meanings are not
 relevant and therefore using `id` for some other purpose beyond uniqueness
 checking is out of scope of the specification and not recommended.
 
+### data
+
+The `data` attribute of a CloudEvent may carry the event data itself. Event
+consumers can use the `datacontenttype` attribute of a CloudEvent to determine
+what the type is, for example `application/json` or `image/png`. There are no
+restrictions on the type of data that may be included with a CloudEvent. Event
+consumers can expect that the event data is transported unaltered due to any
+encoding used by an SDK to facilitate transport over a given protocol. When a
+CloudEvent is sent by an event producer over a supported protocol, the event
+`data` may be encoded in Base64 form for transmission. However, on the receiving
+end of the transmission this data will be decoded into its original unaltered
+form. CloudEvent SDKs are responsible for encoding and decoding this data.
+
+Consider an example. An IoT device is creating CloudEvents with JSON expressed
+as a sequence of bytes, i.e. in binary form. It may set the `datacontenttype`
+attribute on the event to `application/json` even though the data itself is in
+binary form. A receiver of this event should expect to find the data attribute
+still in its original binary form even after having been transmitted over the
+wire. It is the responsibility of the event consumer to convert this data from
+its binary form into JSON or other data type as specified by the
+`datacontenttype` attribute.
+
 ## CloudEvent Attribute Extensions
 
 In order to achieve the stated goals, the specification authors will attempt to
--- a/primer.md
+++ b/primer.md
@@ -291,16 +291,16 @@ checking is out of scope of the specification and not recommended.
 
 ### data
 
-The `data` attribute of a CloudEvent may carry the event data itself. Event
-consumers can use the `datacontenttype` attribute of a CloudEvent to determine
-what the type is, for example `application/json` or `image/png`. There are no
-restrictions on the type of data that may be included with a CloudEvent. Event
-consumers can expect that the event data is transported unaltered due to any
-encoding used by an SDK to facilitate transport over a given protocol. When a
-CloudEvent is sent by an event producer over a supported protocol, the event
-`data` may be encoded in Base64 form for transmission. However, on the receiving
-end of the transmission this data will be decoded into its original unaltered
-form. CloudEvent SDKs are responsible for encoding and decoding this data.
+Event consumers can use the `datacontenttype` attribute of a CloudEvent to
+determine what the type is, for example `application/json` or `image/png`. There
+are no restrictions on the encoding format of data that may be included with a
+CloudEvent. Event consumers can expect that the event data is transported
+unaltered due to any encoding used by an SDK to facilitate transport over a
+given protocol. When a CloudEvent is sent by an event producer over a supported
+protocol, the event `data` may be encoded in Base64 form for transmission.
+However, on the receiving end of the transmission this data will be decoded into
+its original unaltered form. CloudEvent SDKs are responsible for encoding and
+decoding this data.
 
 Consider an example. An IoT device is creating CloudEvents with JSON expressed
 as a sequence of bytes, i.e. in binary form. It may set the `datacontenttype`
--- a/primer.md
+++ b/primer.md
@@ -292,24 +292,15 @@ checking is out of scope of the specification and not recommended.
 ### data
 
 Event consumers can use the `datacontenttype` attribute of a CloudEvent to
-determine what the type is, for example `application/json` or `image/png`. There
-are no restrictions on the encoding format of data that may be included with a
-CloudEvent. Event consumers can expect that the event data is transported
-unaltered due to any encoding used by an SDK to facilitate transport over a
-given protocol. When a CloudEvent is sent by an event producer over a supported
-protocol, the event `data` may be encoded in Base64 form for transmission.
-However, on the receiving end of the transmission this data will be decoded into
-its original unaltered form. CloudEvent SDKs are responsible for encoding and
-decoding this data.
-
-Consider an example. An IoT device is creating CloudEvents with JSON expressed
-as a sequence of bytes, i.e. in binary form. It may set the `datacontenttype`
-attribute on the event to `application/json` even though the data itself is in
-binary form. A receiver of this event should expect to find the data attribute
-still in its original binary form even after having been transmitted over the
-wire. It is the responsibility of the event consumer to convert this data from
-its binary form into JSON or other data type as specified by the
-`datacontenttype` attribute.
+determine the type of `data`, for example `application/json` or `image/png`.
+There are no restrictions on the encoding format of data that may be included
+with a CloudEvent. Event consumers can expect that the event data is
+transported unaltered due to any encoding used by an SDK to facilitate
+transport over a given protocol.  CloudEvent SDKs should avoid modifying the
+event data in any way.  This includes actions such as syntax reformatting.
+Even though it might appear to not influence the semantic meaning of the
+information, the SDK can not be sure if the original format is meaningful to
+the consumer's processing.
 
 ## CloudEvent Attribute Extensions
 


primer.md
```diff
@@ -289,6 +289,28 @@ context, for the purposes of this CloudEvent attribute those meanings are not
 relevant and therefore using `id` for some other purpose beyond uniqueness
 checking is out of scope of the specification and not recommended.
 
+### data
+
+The `data` attribute of a CloudEvent may carry the event data itself. Event
```
deissnerk, line 294: Actually, `data` is not regarded as attribute in v1.0.  If I recall it correctly, that was changed between 0.3 and 1.0, due to a lot of discussions around the CloudEvents type system and the attribute `datacontentencoding` that existed in 0.3.

duglin: https://github.com/cloudevents/spec/blob/master/spec.md#event-data
I wonder if we should just drop the first sentence entirely - doesn't seem to add much.

primer.md
```diff
@@ -289,6 +289,28 @@ context, for the purposes of this CloudEvent attribute those meanings are not
...
+consumers can use the `datacontenttype` attribute of a CloudEvent to determine
+what the type is, for example `application/json` or `image/png`. There are no
+restrictions on the type of data that may be included with a CloudEvent. Event
```
duglin, line 297: might want to talk about encoding format instead of "type" - so it's not confused with the "type" attribute

lance: Done

primer.md
lance: >  I wonder if we should just drop the first sentence entirely - doesn't seem to add much.

Done

primer.md
```diff
@@ -289,6 +289,28 @@ context, for the purposes of this CloudEvent attribute those meanings are not
 relevant and therefore using `id` for some other purpose beyond uniqueness
 checking is out of scope of the specification and not recommended.
 
+### data
+
+Event consumers can use the `datacontenttype` attribute of a CloudEvent to
+determine what the type is, for example `application/json` or `image/png`. There
```
tweing, line 295: Proposal
... to determine the type of `data`, for example ...

primer.md
```diff
@@ -289,6 +289,28 @@ context, for the purposes of this CloudEvent attribute those meanings are not
...
+decoding this data.
+
+Consider an example. An IoT device is creating CloudEvents with JSON expressed
+as a sequence of bytes, i.e. in binary form. It may set the `datacontenttype`
+attribute on the event to `application/json` even though the data itself is in
+binary form. A receiver of this event should expect to find the data attribute
+still in its original binary form even after having been transmitted over the
+wire. It is the responsibility of the event consumer to convert this data from
+its binary form into JSON or other data type as specified by the
+`datacontenttype` attribute.
```
tweing, line 312: > A receiver of this event should expect to find the data attribute still in its original binary form even after having been transmitted over the wire. It is the responsibility of the event consumer to convert this data from its binary form into JSON or other data type as specified by the `datacontenttype` attribute.

Did I miss something here? This doesn't make sense for me. As the consumer, I would expect that the `datacontenttype` describes the encoding e.g. `JSON`. How would the consumer be able to detect whether the `data` is in binary form or in character form with a certain encoding?

primer.md
```diff
@@ -289,6 +289,28 @@ context, for the purposes of this CloudEvent attribute those meanings are not
...
 checking is out of scope of the specification and not recommended.
 
+### data
+
+Event consumers can use the `datacontenttype` attribute of a CloudEvent to
+determine what the type is, for example `application/json` or `image/png`. There
+are no restrictions on the encoding format of data that may be included with a
+CloudEvent. Event consumers can expect that the event data is transported
+unaltered due to any encoding used by an SDK to facilitate transport over a
+given protocol. When a CloudEvent is sent by an event producer over a supported
```
duglin, line 299: I wonder if after the sentence on lines 297-299 we should avoid being too specific since the base64 processing can end up confusing people and distracting them from the main point. And instead, what if we added something like:
```
In other words, SDKs should avoid modifying the event data and try to keep the original format sent.
This would include actions such as syntax reformatting. Even though it might appear to not influence
the semantic meaning of the information, the SDK can not be sure if the original format would be
meaningful to the consumer's processing.
```
Not sure we would need the rest of the text.

duglin:
@lance any update on this one?

primer.md
```diff
@@ -289,6 +289,19 @@ context, for the purposes of this CloudEvent attribute those meanings are not
...
+Event consumers can use the `datacontenttype` attribute of a CloudEvent to
+determine the type of `data`, for example `application/json` or `image/png`.
+There are no restrictions on the encoding format of data that may be included
+with a CloudEvent. Event consumers can expect that the event data is
+transported unaltered due to any encoding used by an SDK to facilitate
+transport over a given protocol.  CloudEvent SDKs should avoid modifying the
+event data in any way.  This includes actions such as syntax reformatting.
+Even though it might appear to not influence the semantic meaning of the
+information, the SDK can not be sure if the original format is meaningful to
+the consumer's processing.
```
jskeet, line 303: That recommendation is fine when the data is provided as a byte array to start with - it becomes a lot less clear if the SDK allows the data to be specified as "an arbitrary object" which an event format can then serialize in the most natural format (for example, as a JSON object within a structured mode message). Grappling with this has been one of the trickiest aspects of working on the C# SDK. We *could* just say "the data is always a byte array" but that makes all kinds of things less pleasant for end-users.

That's primarily about how code that produces events in memory works, of course. Code that propagates a CloudEvent from one transport to another (possibly using a different event format) may have different challenges.

I don't think I'm particularly arguing against this paragraph as an ideal - just noting that life is a lot more complicated in practice.

lance: @jskeet in the Javascript SDK we face the same issue, and are definitely not compliant with this statement.

duglin: Are you guys talking about things changing due to the language or due to the SDK going out of its way to tweak it? I interpret this text to be talking about an SDK trying to be smart and going out of it's way to reformat stuff - which I think we should discourage. But if something about the language itself ends up tweaking stuff, and that's normal expectations for devs using that language, then that would seem ok to me. Or am I missing the point?

jskeet: I would be surprised if an SDK tried explicitly to reformat things... but it *could* very easily lose formatting due to lossy parse/reformat cycle.

For example, support you have a structured JSON event of this form:

```json
{
  "spec_version": "1.0",
  [ other attributes ]
  "data": {
    "x": 5
  }
}
```

I suspect that "Parse that using a JSON event format, then use the resulting CloudEvent object later to create a new HTTP message, still using structured mode" could end up with:

```json
{
  "spec_version": "1.0",
  [ other attributes ]
  "data": { "x": 5 }
}
```

In the C# SDK at least, the data here would be parsed into a `JObject` or `JsonElement` (depending on which package you use), without explicitly trying to preserve formatting.

That feels okay to me, even though it *is* reformatting.

I suspect the SDK is doing the right thing here, but it does violate the text of the primer. (It *is* only a primer rather than a spec, of course...)

tweing: The two examples are for me syntactically identical. They are both valid `json`. The only difference is `whitespace` which should be ignored by deserializers. Another thing are the different line endings which are `CR/LF` under Windows and `LF` under Linux. Usually not visible but a noticable difference in the byte stream.

To throw in some more oil into the fire, we could go one step further and output the message as
```
{
  [ other attributes ],
  "spec_version": "1.0",
  "data": { "x": 5 }
}
```

Is this message above still equivalent to the original one?

My question would be what is meant in the text with the term `syntax reformatting`?

deissnerk: My first thought was:
[RFC 8259](https://tools.ietf.org/html/rfc8259) desribes a JSON text as *a serialized JSON value*. So I would say that just this value MUST be unchanged. Whitespaces that do not change the value should be fine. 

But:
What happens, when we start discussing digital signatures?

duglin: hmm I feel like I'm missing something though. If the data is parsed into some language specific object/type, and that type is passed on to the consumer, then I agree that formatting will change because you're not giving the user "raw json".... meaning a string. You're giving them a runtime object. And in some languages I think the whitespace/noise will be preserved and in others it wont - but that's ok for each language.But, if you deserialize the json/string into an object and then back into a json/string and give that to the consumer, I have to ask... why? Unless you're planning on modifying the attributes, why aren't you passing the original string on to the consumer

I guess it's the "parse/reformat cycle" that I need to understand more.

jskeet: I wouldn't expect parse/reformat to happen just as part of "parse a CloudEvent". What I'm imagining is something like this (pseudo-code - not what the actual C# would look like):

```csharp
void ProxyRequest(HttpRequest input)
{
    CloudEvent event = HttpRequestBinding.ParseCloudEvent(input, jsonEventFormatter);
    HttpRequest proxiedRequest = new HttpRequest(targetUrl);
    // Populate some other headers etc here, perhaps - we might
    // choose the targetUrl conditional on parts of the event
    HttpRequestBinding.PopulateRequest(proxiedRequest, event, jsonEventFormatter);
    proxiedRequest.Send();
}
```

In other words: we have an incoming HTTP request, which we parse to the in-language CloudEvent representation, and then use that to populate a *new* HTTP request (or perhaps a new message queue message, or whatever). I suspect there are many ways where the resulting HTTP request body may not be byte-for-byte identical to the original request.

I suspect the only way of fixing that use case (e.g. preserving the exact bytes) would be to make other use cases much harder.

Binary mode messages make this easier, admittedly - and an SDK may well provide multiple parsing approaches, some of which are designed to preserve opaque data and some of which are designed to make application data processing easier.

Does that clarify the concern a bit more?

duglin: yes that helps and is sort of consistent with what I was thinking. Meaning, in your above example I don't see the value in parsing it at all - in fact it is probably a performance issue to do so. It seems like it would be better to just pass along the bytes.

If the SDK needs to deal with the JSON, for some reason, it seems like it could parse it, examine it, but then still pass along the original bytes.

The only reason to parse it and pass along the parsed object is if you want to edit it, but then you're no longer in the simple "it's a proxy" type of situation where the "do not touch the formatting" would apply. Although, if you don't touch the "data" I would hope that's passed along "untouched" in that case too.

Of course, this is for cases where the output (next hop) is HTTP. In cases where the event is passed along as something like a language Object, then I think we're back to what I said above, each language will determine whether whitespace is preserved. 

But Lance, wasn't the original situation here one where the data was a string that just happened to look like json?  And in that case I think it should not be parsed or touched.

lance: > But Lance, wasn't the original situation here one where the data was a string that just happened to look like json? And in that case I think it should not be parsed or touched.

We are pretty far afield from the original use case at this point, and to be quite honest, I think the original use case was flawed. In some of the early tests written for the Javascript SDK, the `data` attribute was JSON sent as binary in base64 encoding. The test case was contrived and as we talked through this issue, I realized that it was actually incorrect, as the `content-type` header was `application/json`. Those tests have since been removed. Here is an example. I think it's probably fair to say that this violates the spec.

https://github.com/cloudevents/sdk-javascript/blob/d70274d4045a351e4e2081f119212e6cfa8ce3f0/test/bindings/http/receiver_binary_1_tests.js#L377-L402

duglin: do we still need this PR then?

lance: I'm happy to close it, to be honest

lance:
Closing per agreement in weekly call

duglin:
Agreed to close on 4/1 call

PR #741, opened 2020-12-03
Proposal: Error Adapter

yordis:
## Purpose

Errors are everywhere. However, error's producers tend to describe errors
differently.

The lack of a common way of describing errors means developers are continually
re-learning how to consume errors. That also limits the potential for libraries,
tooling, and infrastructure to aid of error data across delivery environments,
like SDK or logger systems. The portability and productivity that can be
achieved from error data are hindered overall.

This ADR is a specification for describing error data in standard formats to
provide interoperability across services, platforms, and systems.

### DIFF
--- a/adapters/error/README.md
+++ b/adapters/error/README.md
@@ -36,7 +36,7 @@ provide interoperability across services, platforms, and systems.
 
 - Constraints:
   - REQUIRED
-  - MUST be https://github.com/cloudevents/spec/tree/v1.0/adapters/error/schema.json"
+  - MUST be https://github.com/cloudevents/spec/tree/v1.0/adapters/error/dataschema.json"
 
 ### data
 
--- a/adapters/error/README.md
+++ b/adapters/error/README.md
@@ -1,0 +1,121 @@
+# Error CloudEvents Adapter
+
+This document describes how to Errors events into a CloudEvents.
+
+## Table of Contents
+
+- [Error CloudEvents Adapter](#error-cloudevents-adapter)
+  - [Table of Contents](#table-of-contents)
+  - [Purpose](#purpose)
+  - [Context Attributes Extension](#context-attributes-extension)
+    - [dataschema](#dataschema)
+    - [data](#data)
+  - [Event Data](#event-data)
+    - [errormessage](#errormessage)
+    - [errorlink](#errorlink)
+    - [metadata](#metadata)
+  - [Example](#example)
+
+## Purpose
+
+Errors are everywhere. However, error's producers tend to describe errors
+differently.
+
+The lack of a common way of describing errors means developers are continually
+re-learning how to consume errors. That also limits the potential for libraries,
+tooling, and infrastructure to aid of error data across delivery environments,
+like SDK or logger systems. The portability and productivity that can be
+achieved from error data are hindered overall.
+
+This ADR is a specification for describing error data in standard formats to
+provide interoperability across services, platforms, and systems.
+
+## Context Attributes Extension
+
+### dataschema
+
+- Constraints:
+  - REQUIRED
+  - MUST be https://github.com/cloudevents/spec/tree/v1.0/adapters/error/schema.json"
+
+### data
+
+- Constraints:
+  - REQUIRED
+
+## Event Data
+
+### errormessage
+
+- Type: String
+- Description: contains a generic description of the error condition in English.
+  It is intended for a human audience. Simple programs display the message
+  directly to the end user if they encounter an error condition they don't know
+  how or don't care to handle. Sophisticated programs with more exhaustive error
+  handling and proper internationalization are more likely to ignore the error
+  message.
+- Constraints:
+  - OPTIONAL
+  - If present, MUST be a non-empty string
+
+### errorlink
+
+- Type: URI-reference
+- Description: URL to documentation related to the error.
+- Constraints:
+  - OPTIONAL
+  - If present, MUST be a non-empty URI
+
+### metadata
+
+- Type: Object
+- Description: The metadata associated to the error.
+- Constraints:
+  - OPTIONAL
+  - If present, MUST be an object
+
+## Example
+
+The following example shows a CloudEvent Error serialized as JSON:
+
+```json
+{
+    "specversion" : "1.x-wip",
+    "id" : "YH9UUdlCjQztui7untrQE",
+    "type" : "com.straw-hat-team.sso.not_found",
+    "source" : "https://github.com/straw-hat-team/sso",
+    "time" : "2018-04-05T17:31:00Z",
+    "subject" : "/account",
+    "datacontenttype" : "application/json",
+    "data" : "{
+      \"errormessage\":\"Resource not found\",
+      \"errorlink\":\"https://github.com/straw-hat-team/sso/docs/errors/E0000008.md\",
+      \"metadata\":{
+        \"account_id\":\"something\"
+      }
+    }"
+}
+```
+
+You may use a `subject` that identify particular field from the input data.
+The following example shows a CloudEvent Error serialized as JSON:
+
+```jsonc
+{
+    "specversion" : "1.x-wip",
+    "id" : "e8QIbN7KDaQDSagrWM8d8",
+    "type" : "com.straw-hat-team.payment.insufficient_fund",
+    "source" : "https://github.com/straw-hat-team/payment",
+    "time" : "2018-04-05T17:31:00Z",
+    "subject" : "/payment/amount",
+    "datacontenttype" : "application/json",
+    "data" : "{
+      \"errormessage\":\"You can not transfer an amount greater than the current balance\",
+      \"errorlink\":\"https://github.com/straw-hat-team/payment/docs/errors/E0000001.md\",
+      \"metadata\":{
+        \"amount_requested\":\"500000\",
+        \"current_balance\":\"200000\"
+      }
+    }"
+}
+```
--- a/adapters/error/README.md
+++ b/adapters/error/README.md
@@ -60,7 +60,7 @@ provide interoperability across services, platforms, and systems.
 
 ### errorlink
 
-- Type: URI-reference
+- Type: URI
 - Description: URL to documentation related to the error.
 - Constraints:
   - OPTIONAL
--- a/adapters/error/dataschema.json
+++ b/adapters/error/dataschema.json
@@ -10,6 +10,5 @@
     "metadata": {
       "type": "object"
     }
-  },
-  "required": ["errormessage"]
+  }
 }
--- a/adapters/error/dataschema.json
+++ b/adapters/error/dataschema.json
@@ -1,0 +1,15 @@
+{
+  "$schema": "http://json-schema.org/schema",
+  "type": "object",
+  "properties": {
+    "errormessage": { "type": "string" },
+    "errorlink": {
+      "type": "string",
+      "pattern": "uri"
+    },
+    "metadata": {
+      "type": "object"
+    }
+  },
+  "required": ["errormessage"]
+}
--- a/adapters/error/dataschema.json
+++ b/adapters/error/dataschema.json
@@ -1,13 +1,20 @@
 {
   "$schema": "http://json-schema.org/schema",
+  "title": "Error",
+  "description": "A single error",
   "type": "object",
   "properties": {
-    "errormessage": { "type": "string" },
+    "errormessage": {
+      "description": "Human readable error message, intended to be used by developers",
+      "type": "string"
+    },
     "errorlink": {
+      "description": "A URL to the more detailed information about the error",
       "type": "string",
       "pattern": "uri"
     },
     "metadata": {
+      "description": "A object containing metadata associated with the error",
       "type": "object"
     }
   }
--- a/adapters/error/dataschema.json
+++ b/adapters/error/dataschema.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "http://json-schema.org/schema",
+  "$schema": "http://json-schema.org/schema#",
   "title": "Error",
   "description": "A single error",
   "type": "object",


hiebj:
In practice, I am not sure if this will really work out. My team uses CloudEvents for error reporting as well, both as domain events in our pub/sub systems and also when reporting errors to API consumers.

We almost made a proposal very similar to this, but ultimately decided against it. I'll try to outline the reasons.

Ultimately, it really boils down to one problem: **domain errors just don't have enough structural overlap** - particularly, structural overlap that can't be handled by the existing CloudEvents spec and proper use of existing schemas.

**prior art**

This has actually been tried before, for HTTP requests specifically: see [Problem JSON](https://tools.ietf.org/html/rfc7807) and [JSON:API errors](https://jsonapi.org/format/#error-objects). However, those solutions are not suitable for all environments; they are too reliant on JSON and HTTP to be effective for an event-driven error scheme.

Even when compared to the above - probably the closest anyone has ever gotten to a universal error schema - there are essentially only two fields not covered by CloudEvents: a **human-readable message** and a **detail link**.

Everything else must be provided as either **a metadata object** (as you have here) or **additional attributes** (the way Problem JSON handles it).

**how to provide schema for domain-specific error data?**

Lots of errors have highly domain-specific data that needs to be passed along, and that data is often crucial to the handling of the error - not simply "metadata" that can be dumped into a log for human debugging purposes. It is not sufficient to leave that data untyped - middleware and other abstract event processors need to be able to figure out what kind of data is there if you want to have any kind of generic handling, such as a common way of dealing with data rollback.

The problem is: how do you provide a schema definition for those properties? There really is no way to provide a schema definition for expansion attributes if the `dataschema` is fixed. With metadata, you could certainly provide an additional `metadataschema` attribute that defines an additional schema for the metadata - but to what avail? Why not simply provide a `dataschema` that captures your domain-specific error concerns?

**suggestions**

Given the above, there are basically two things I think that can be done to improve the CloudEvents spec in this regard:

1. create _specific_ adapters for particular error structures, such as existing schema constraint violation errors like `ajv`
2. create `message` and `link` Expansion Context Attributes for the CloudEvents spec. Human-readable messages and a detail link are useful in far more use cases than only the error case. They are not critical for routing, which is what context attributes are for, but they are close enough to being universal that I could see a compelling argument for adding them to the spec as expansions. (I may submit a PR requesting feedback on exactly this proposal in the next couple of days).

Beyond that, I think what the industry really needs is a collaborative effort on a broader mission of creating standardized error schemas for the myriad domain errors that occur. Then we can simply attach those schemas to CloudEvens and be done with this once and for all.

yordis:
> particularly, a structural overlap that can't be handled by the existing CloudEvents spec and proper use of existing schemas.

Which schemas?

I agree with you that CloudEvents has most of the data that we would need. That is why I re-wrote my ADR (which is how the conversation started) to take advantage of CloudEvents's structure and add more alignment around the event's data.

> they are too reliant on JSON and HTTP to be effective for an event-driven, cross-platform error scheme.

Would you mind expanding on that?

> In many cases, it is sufficient to put a human-readable message directly in data with datacontenttype set to text/plain.

Totally,

Just that at that point, you lost the ability to add some data that would allow you to do i18n (if you want to send the event out to some clients that may care about that) or do i18n all together, which many people opt-in to do so, you can still use the `errormessage` if you would like to ignore i18n.

If that is your case, I think it is totally valid, but personally, I much rather bring more alignment. In case you don't care, you can ignore the field. It is not mandatory.

> The problem is: how do you provide a schema definition for those properties?

That is why you have other fields that allow you to identify the error itself, such as the `type` field. `type` field will allow you to understand better the context of the "unstructured" metadata value. If you want a centralized way to deal with this, you will end up with a centralized mapping that understands all the contexts 🤷🏻 I am not sure how to avoid.

> I think it will probably turn out to be best to leave the data attribute alone

We can't if we want to have some alignment and consider multiple concerns like i18n or helping programmers.

Which we could remove `errormessage` and `errorlink` for sure, leaving `metadata` (it could be just data) behind it.

Still, from my experience, there is always a balance between "what is technically correct" (whatever that means) and dealing with the fact that they are always a trade-off.
I much rather have alignment around giving programmers some useful information. In the past, we tried to be technically correct and brought resentment (for valid reasons) towards the decisions because it didn't take into consideration their needs as programmers.

> use the other attributes creatively

That is where I strongly disagree. That is where misguided decisions are made since you rely on people's experiences and ability to "creative" (take my proposal as an example)

Would you mind helping me to identify how I accomplish the intention without following the proposal?  What would be an alternative?

hiebj:
You said you rewrote your ADR - out of curiosity, what was your original proposal?

> Which schemas?

> Would you mind helping me to identify how I accomplish the intention without following the proposal? What would be an alternative?

I think I'd love to see an open-source project called `ErrorSchemas` or something that does work similar to what CloudEvents has done here - gather together a huge collection of existing error shapes from various libraries and applications, group them into categories, and try to align on a limited number of error shapes - or perhaps a single error shape - that eliminates the need for an arbitrary untyped "metadata" object, either by giving us a single solution that works in every case, or else a set of schemas to choose from for different types of error cases.

So far every attempt to generalize on errors has basically gone the route of "it's a message and a link, and then a bucket of who-knows-what" and one of my favorite things about CloudEvents is the capability to use schema to define the shape of your payload.

I realize that's not a workable suggestion in any sort of near term. Perhaps instead I'll share what we were thinking of proposing.

My first instinct was actually to open a PR suggesting that we actually create a `humanreadable` extension attribute, or perhaps more simply, `message`. The `type` or `subject` field could contain a detail link (this is what I meant by "use the existing attributes creatively", though it's not really a stretch from the intended use) and `data` would contain your metadata, typed by the given `dataschema`. Originally we had thought to name the attribute `error` which would allow you to use it for duck-typing, but in all of the use cases we could think of, it was sufficient to just create a different `type` to represent errors. And `humanreadable` or `message` is more flexible - the idea was, there are many types of events that might benefit from having a human-readable message attached, not just errors; and as an extension attribute, it wouldn't be too offensive.

We elected not to suggest that after reading carefully the documentation, [particularly this](https://github.com/cloudevents/spec/blob/v1.0/primer.md#cloudevent-attribute-extensions):

> Extension attributes to the CloudEvent specification are meant to be additional metadata that needs to be included to help **ensure proper routing and processing** of the CloudEvent. Additional metadata **for other purposes, that is related to the event itself and not needed in the transportation or processing of the CloudEvent**, should instead be placed within the proper extensibility points of the event (data) itself.

It's hard to make the argument that a human-readable message falls under routing and processing concerns. It seems like it belongs in `data`, where you have placed `errormessage` and `errorlink`. However, given your comment here:

> I much rather have alignment around giving programmers some useful information. In the past, we tried to be technically correct and brought resentment (for valid reasons) towards the decisions because it didn't take into consideration their needs as programmers.

It seems pretty clear to me that adding an extension attribute `humanreadable` or `message` to the event body, then allowing your metadata to sit in the schema-backed `data` attribute would be a pretty ideal developer experience. It's not "technically correct" pursuant to the comments in the spec about extension attributes being meant for routing... but it seems so useful that I'm inclined to think it might be forgiven.

What do you think?

hiebj:
Nevermind to my question above about your ADR, I found the PR against your ADR repository since you tagged this PR in a comment there.

Seems you are interested in essentially the same problem I am. I would be interested in pursuing some comprehensive research into existing error patterns as I mentioned above. I don't think the resulting work necessarily belongs in the CloudEvents spec, but I do think the projects would be complimentary - essentially providing a set of schemas that are used in CloudEvent data - particularly errors, since they can often be generalized to a significant extent across domains.

cneijenhuis:
Can you define error?

An error could be of technical nature - something that a developer needs to have a look at and fix. Given that a human has to have a look, interoperability and portability don't strike me as the most important goals.

An error could also be a state in the domain. E.g. if you're trying to charge a credit card, your payment provider may return either a `charge.succeeded` or `charge.failed` event. Surely, the latter case is an error - we prefer the charge to be successful. However, the rejection is to be expected as well. Our code has to handle both cases.

If we encounter an event/case our code doesn't handle yet, we need to figure out how to handle the event - read the documentation, look at the schema etc. However, that is independent of whether the event represents a success or an error in our domain. I am not sure why we should treat these differently.
E.g. i18n - you need to solve it for both cases. If you want to inform the user the charge was successful and you'll e.g. send a package, the mechanism is the same as when informing the user the charge failed they'll e.g. need to provide a different payment method or cancel their order. Adding a textual description into the event may or may not be a good idea, but that doesn't depend on whether the event is an error state in your domain.

yordis:
> hat eliminates the need for an arbitrary untyped "metadata" object, either by giving us a single solution that works in every case, or else a set of schemas to choose from for different types of error cases.

Personally, I need to update my ADR to add even more constraints where the metadata is a Map<string, string>, so you can't have nested values, and everything is a string.

> So far every attempt to generalize on errors has basically gone the route of "it's a message and a link, and then a bucket of who-knows-what" and one of my favorite things about CloudEvents is the capability to use schema to define the shape of your payload.

Right, agree, that is why I am leveraging the spec to handle errors.

> suggesting that we actually create a `humanreadable` extension attribute, 

The original spec was using extensions, I proposed the last meeting and some people suggest that we don't need to use extensions when we could use an adapter instead, and add the schema of the data.

yordis:
>  interoperability and portability don't strike me as the most important goals.

We do 🤷🏻 

I would like to hear more precise feedback since your comment is amazing, and I would like to remove the subjectivity as much as possible. Proposing some alternative would be amazing.

hiebj:
> An error could also be a state in the domain. E.g. if you're trying to charge a credit card, your payment provider may return either a `charge.succeeded` or `charge.failed` event. Surely, the latter case is an error - we prefer the charge to be successful. However, the rejection is to be expected as well. Our code has to handle both cases.

Exactly. There are basically two different types of errors: **Errors that are actually expected domain results**, and **unexpected errors that the user can't do anything about.**

In the case of the latter, **there is really nothing useful to show to the user** other than "whoops, something went wrong, try again later or call support with event ID `xyz-1234`". And IMO, that message does not actually need to be provided by the backend at all. If something unexpected, and therefore unrecoverable has happened, there is little value in even sending metadata about the error to the client.

In the case of the former, we're not really talking about an "error". **An error in this case is, in fact, a domain result** - indicating that the user's request couldn't be executed. When I said that we use CloudEvents to represent errors in pub/sub, this is what I meant - domain-specific, expected events (which nevertheless may be consumed by parties who would benefit from having a schema attached).

This is what I meant by "there is not sufficient structural overlap in error types". **Errors that are actually domain results are going to have domain specific concerns**.

If we're talking about messages sent to a client app, and your domain chooses to provide the user with a link to an error page, that's one approach; if your domain instead encodes a list of steps the user needs to take, such as a list of missing fields they need to fill out before their user profile is "complete", that's another approach. The UI has to know what that shape is and it is likely to be context-specific. If your API is consumed by third parties, it is useful to have a well-defined schema, and send it along with your message.

If we're talking about domain events being published to pub/sub, you're going to need to provide some information that allows consuming services to handle the event somehow... and that information should have a schema.

There are relatively few cases, notably in the case of developer tools, where you actually want to take "a bucket of who-knows-what" in the form of untyped metadata and just show it to the user. And if you are building a developer tool, "a bucket of who-knows-what" _is domain specific_!

So in the end, what we are left with is:

- an unexpected, unrecoverable error in which case providing additional information has questionable value; or
- an expected domain result, in which case the schema is domain-specific and local to your project or team

It's possible that sending `errormessage` and `errorlink` and `metadata` is an effective solution for all of the domain concerns encountered by your team. It's not really generalizable to "all errors, everywhere", though.

Really the only thing about errors that can be generalized is that there is a human-readable string involved - but **even that is domain-specific**, because some domains want to translate it, and some domains don't want to provide a message at all because the front-end decides what to show the user based on the encoded `type`. I'm not sure about you, but most "human readable" error messages I get from third party APIs are not something I actually want to show the user.

This is why we didn't go through with suggesting the `humanreadable` extension. Our domain failure shapes are specific to our domain and often don't even include a human-readable string. Many of our events do; in those cases, it's encoded as part of the domain-specific `dataschema`.

> Proposing some alternative would be amazing.

I think the alternative is that we don't solve this problem with changes to the CloudEvents spec. Trying to create a schema to solve "errors" is **out of scope for this spec**. The error schema in your PR is specific to your domain and belongs in your application or a team repository. A Map of strings is not sufficient for all domain use cases, and the "general" solution is what already exists in CloudEvents - a `data` attribute that has an associated `dataschema`.

If you want to tackle the problem of consistency and interoperability of error shapes across domains, I think a lot more work needs to be done - specifically, analyzing existing errors across lots of systems and coming up with categories and schemas for domain errors. The resulting schemas could be shared by the open source community and used as payloads in CloudEvents. But they are complimentary solutions. The goal of CloudEvents isn't to come up with a unified set of domain failure schemas. It's to define a consistent envelope shape for inter-service communication so that services utilizing pub/sub can be deployed to multiple platforms, and so that systems that span platforms don't have to constantly be mapping between Azure/Google/AWS/IBM event shapes.

yordis:
> I think the alternative is that we don't solve this problem with changes to the CloudEvents spec. Trying to create a schema to solve "errors" is out of scope for this spec.

Which CloudEvents spec? This is not part of CloudEvents spec directly but an adapter.

Does that mean CloudEvents shouldn't tell you how to define GitHub events? https://github.com/cloudevents/spec/blob/v1.0/adapters/github.md#github-cloudevents-adapter

I think it is important to understand the differences between CloudEvents Spec and Adapters.

> analyzing existing errors across lots of systems and coming up with categories and schemas for various well-known and oft-used domain errors.

Hint where we are right now. It is hard for me to disagree with you, but for the wrong reasons.

Most of your concerns have been thought out before, I am not sure if you understand that, none of this comes from a misunderstanding of the situation.

Although what you are saying is completely valid, when it comes to trying to tackle the situation, most of your concerns fall into such schema decision (hint, as an adapter, that is **not** part of the spec), or at least my understanding from what you are saying suggest that.

I would like to hear more clear examples (potentially sharing some concrete examples based on real cases) where the Error Adapter would become a problem, and wouldn't work for you.

hiebj:
You're right, Yordis. As an adapter, this doesn't have an effect on me unless I choose to use it. However, I think you know what I'm getting at here. You are proposing `Error Adapter`, not `Yordis Error Adapter` because you are trying to create a generalization for all errors.

The intent of Adapters, while perhaps not explicitly stated, is to provide a mapping from various platform-specific event types to CloudEvents. Is there a Yordis platform I am not aware of?

Ultimately what this amounts to is a single suggestion of one possible way to represent errors using CloudEvents. My issue with it is that it is neither specific enough to adapt a specific event type (`Yordis Error Adapter`) nor general enough to solve the general case of adapting cross-domain failure events to CloudEvents. I am just not sure why this needs to be here. Should I be suggesting adapters for my own way of representing errors? What about domain success events? Should I propose an adapter for a `UserUpdatedEvent` in my AWS IAM system? Does my representation of `User` merit official documentation as the solution for all `User` events?

The problem that this creates for me is that accepting a proposal that quite clearly intends to be a generalized solution for all errors precludes, or makes difficult, the adoption of a more comprehensive solution later; such as a set of defined adapters to work with the generalized error schemas that would be created by a project like the one I outlined above.

It's not actually my job to decide what does or does not belong in this spec or its documented extensions and adapters, since I'm not a voting contributor. At this point I'll bow out and leave it up to the maintainers to decide what is best.

hiebj:
> I would like to hear more clear examples (potentially sharing some concrete examples based on real cases) where the Error Adapter would become a problem, and wouldn't work for you.

So, the onus is actually not on me to explain why this doesn't work for me. The onus is on you to explain why your opinionated error schema that does not represent an adaptation of an existing common platform event should be officially documented as an adapter for the CloudEvents spec.

- But since you asked, let's talk about the specific case I already mentioned. I have a REST API that allows users to toggle their professional profile and become "searchable" on my organization's job search platform. The back end performs a fairly common type of business rule validation and then comes back to the user with a response that is either successful, or contains an array of objects indicating, in detail, the set of minimum requirements they have not yet met with their profile. Perhaps they need to upload a picture. Maybe they haven't confirmed their email address. They may not have provided the minimum of 5 professional skills.

How exactly do I represent this with your schema? Particularly after your comment about constraining the metadata object to a Map of strings?

- Our API is used by third party integrations for job placement organizations that create user profiles automatically for their clients. It is important that these consumers be able to understand the data we are sending back to them, because it's complex and not necessarily obvious how it should be handled. On the client side, we provide a library that generates the reconciliation UI using the schema. This makes it possible for the user to fix errors and resubmit, and we don't have to keep updating our Web Component every time we add new requirements.
- Or another example: we allow companies to define custom business rules around requirements to apply for a job. These business rules are defined as a special type of schema. When a job applicant fails the company's business rules, we use the schema and the data payload to explain to the user what went wrong, and what requirements they have to meet.
- Or another example: when an unexpected failure occurs and events go to a DLQ, we have a common middleware that inspects the schema of the failed payload and resolves/hydrates identifiers for logging purposes so that the poor soul debugging at 2am doesn't have to go look up every sparse identifier by hand to see the entity's full object graph.

All of these applications are failure events that require both complex data and a schema, and that's not to mention the basic point that it's just damn useful to deliver a meaningful schema with your payloads, for data integrity validation and debugging purposes.


In the end, it's not actually my job to decide what does or does not belong in this spec or its documented extensions and adapters, since I'm not a voting contributor. At this point I'll bow out and leave it up to the maintainers to decide what is best.

sdatspun2:
all, I am running late in raising PR against @yordis's fork for JSON schema for error. Here are the schemas associated with event propagating an error. Imo, these should cover both validation errors (error in an event sent earlier that failed validation for example) as well as errors in a system that consumers might be interested in receiving. The errorinstance.json is derived from https://json-schema.org/draft/2019-09/output/schema.  pl. review. thanks.

**error.json**
```
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "title": "Error",
  "description": "Error conveys one or more error conditions in CloudEvent",
  "properties": {
    "name": {
      "type": "string",
      "description": "Human readable, unique name of the error."
    },
    "id": {
      "type": "string",
      "description": "A unique identifier for this error for correlation purposes."
    },
    "message": {
      "type": "string",
      "description": "Message describing the error."
    },
    "errors": {
      "type": "array",
      "description": "Additional details for error instances if more than ones.",
      "items": {
        "$ref": "errorinstance.json"
      }
    },
    "links": {
      "type": "array",
      "description": "Links providing more information related to error(s). These links could point to FAQ or support pages providing help in order to resolve the error(s).",
      "items": {
        "$ref": "linkdescription.json"
      }
    }
  },
  "required": [
    "name",
    "id",
    "message"
  ]
}
```

**errorinstance.json**
```
{
	"$schema": "http://json-schema.org/draft-07/schema#",
	"type": "object",
	"title": "Error Instance",
	"description": "Error instance.",
	"properties": {
		"keywordlocation": {
			"type": "string",
			"description": "Location of keyword entity involved in error. If the keyword is in `data`, set this value to the JSON pointer to the property."
		},
		"instancelocation": {
			"type": "string",
			"description": "The location of the error instance if qualification is needed beyond the keyword location. Could be used to provide index of entry in array."
		},
		"instancevalue": {
			"type": "string",
			"description": "The value of the keyword entity in error, if available."
		},
		"in": {
			"type": "string",
			"description": "The location of the entity that is in error. A valid value is `context`, `data` or `extension`. Default is `data`."
		},
		"message": {
			"type": "string",
			"description": "Error message for the entity in error."
		},
		"link": {
			"description": "Link providing more information related to this error instance.",
			"$ref": "linkdescription.json"
		}
	},
	"required": [
		"in",
		"message"
	]
}
```

**linkdescription.json**
```
{
	"$schema": "http://json-schema.org/draft-07/schema#",
	"title": "Link Description Object (JSON Schema)",
	"type": "object",
	"required": [
		"href",
		"rel"
	],
	"properties": {
		"href": {
			"description": "a URI template, as defined by RFC 6570, with the addition of the $, ( and ) characters for pre-processing",
			"type": "string"
		},
		"rel": {
			"description": "relation to the target resource of the link.",
			"type": "string"
		},
		"title": {
			"description": "title for the link",
			"type": "string"
		},
		"mediatype": {
			"description": "media type (as defined by RFC 2046) describing the link target",
			"type": "string"
		}
	}
}
```

yordis:
@sdatspun2 hey, happy to see you here. I have some concerns based on your response.

1. From the perspective of the ADR as of today, the event itself is an error event, rather than the payload of the event.
    - Why would you do that you may ask?

    Because most of the fields that you may propose could fall into the CloudEvent itself, and either the `data` or the event itself 
    containing all the information is unimportant from my perspective.

    I don't mind adding all the metadata into `data`, but then we will be replicating most of the fields of the event into the `data` such as `id`, `source`, `subject`, `time`; and I can't see a strong reason to do so.


2. (help me here) Your example has a list of errors, where this adapter is for a single error event instance, which means that a list of error events is out of the scope. How CloudEvents handles a list of events sent at once? Could you do that?

PR #747, opened 2020-12-10
Fix version URI typo for the 3.4.2 section

cgfalcon:
To get a specific schema version, the URI is donated as `/schemagroups/mygroup/schemas/myschema/versions/myversion` may more reasonable.


duglin:
Approved on the 12/10 call

PR #752, opened 2021-01-08
[subscriptions] clarify our various subscription properties

### LINKED ISSUE #724: [Subscription-API] Outlining Difference b/w subscriptionconfig and protocolsettings
it seems that the property [subscriptionconfig](https://github.com/cloudevents/spec/blob/master/discovery.md#subscriptionconfig) in discovery api is somewhat similar to the [protocolsettings](https://github.com/cloudevents/spec/blob/master/subscriptions-api.md#321-subscription-object) property in the subscription api. 
Are there any fundamental differences b/w these two if yes, then we probably should outline it and if there's no difference b/w the two then, it'd be a nice idea to remove one of them and stay consistent.

duglin:
Closes #724 


duglin:
Approved on the 1/21 call

PR #754, opened 2021-01-11
fix href checking

duglin:


duglin:
minor tweak to fix our ci

PR #757, opened 2021-01-20
Update http-webhook.md

bsnyder:
Updated the URL for the original and now missing blog post on webhooks from progrium.com. The new URL now points to a snapshot of the old blog post at archive.org.


duglin:
Approved on the 1/21 call

PR #760, opened 2021-01-21
Discovery fixes

### LINKED ISSUE #759: Align discovery OAS and discovery spec
To use client code generators etc, the discovery OAS should be aligned with the spec implemented in the interop effort.

- specversions is specified a list of strings, but OAS uses an object.
- subscriptiondialects is described in the spec as "`Array` or `URI-reference` valuies", where in fact it should be strings (note that OAS uses the term List and Array interchangeably, protocols parameter is described as a "List"), suggesting "List of `String` values"
- OAS is missing id, epoch and subscriptiondialect altogether
- the spec currently doesn't distinguish the subscription object from the embedded event types object (it lists type and other event type parameters alongside subscription parameters), suggesting to separate

manuelstein:
fixes #759 

(align ce_discovery.yaml with discovery.md and separate Subscription schema parameters from EventType parameters)

### DIFF
--- a/ce_discovery.yaml
+++ b/ce_discovery.yaml
@@ -108,7 +108,7 @@ components:
           example: '[ "HTTP", "AMQP", "KAFKA" ]'
         events:
           type: array
-          description: List of event types available in this service.
+          description: List of EventType objects that specify the CloudEvent types available from this service.
           items:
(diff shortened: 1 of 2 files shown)


duglin:
Approved on the 1/28 call

PR #761, opened 2021-01-21
Subscription OAS use of required and anyOf

### LINKED ISSUE #758: Subscription OAS use of required and anyOf
I've noticed the subscription-api-openapi.yaml doesn't validate and found a couple of bugs:

The operation /subscriptions/{id}, which has id as a path field, has two occasions trying to make {id} optional (GET and OPTIONS), which is not allowed per OAS (see [4.7.12.2 parameter fixed field 'required'](https://spec.openapis.org/oas/v3.0.3#fixed-fields-9)).

The content body when creating a subscription is rightfully required, but the spec is missing the required attribute - allowing to send POST /subscriptions operations without a body.
Same goes for updating (PUT /subscriptions/{id}) operations whose content (proposed subscription object) is required.

The protocolsettings currently use anyOf in the context of specifying a single object. It's correct to have only one protocolsettings object held by the parameter as the subscription can only choose a single protocol. Semantically, oneOf is a better fit here.

manuelstein:
fixes #758 

- fix "required: false" on path parameters which can't be omitted
- missing required field for body of creating and updating operation
- use oneOf instead of anyOf

### DIFF
--- a/subscriptions-api-openapi.yaml
+++ b/subscriptions-api-openapi.yaml
@@ -20,6 +20,7 @@ paths:
                   $ref: "#/components/schemas/Subscription"
     post:
       requestBody:
+        required: true
         content:
           application/json:
             schema:
@@ -78,6 +79,7 @@ paths:
           description: The id of an existing subscription
 
       requestBody:
+        required: true
         content:
           application/json:
             schema:
--- a/subscriptions-api-openapi.yaml
+++ b/subscriptions-api-openapi.yaml
@@ -55,7 +55,7 @@ paths:
         - in: "path"
           name: "id"
           description: identifier of a specific subscription
-          required: false
+          required: true
           schema:
             type: string
       responses:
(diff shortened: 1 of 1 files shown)


subscriptions-api-openapi.yaml
```diff
@@ -135,7 +137,7 @@ components:
           example: "HTTP"
         protocolsettings:
           type: object
-          anyOf:
+          oneOf:
```
duglin, line 140: Are you sure we want this change? As I understand it, this will fail validation if it input matches more than one of the $ref's. But what if that's true? What if the settings for two protocols look the exact same - that seems legal to me. I think "anyOf" might be better since we just want to make sure it matches "at least one", not "exactly one".

manuelstein: Yes, good catch, I think you're right. Short answer: the OAS spec isn't very clear about oneOf rendering ambiguous objects invalid, but the implementation of the openapi-generator, for example, matches your description. The issue I am experiencing with the openapi-generator for golang seems to be a separate problem.

duglin:
Approved on the 2/28 call

PR #762, opened 2021-01-23
Syntax fixes for v1.0.1

duglin:
- removed WIP specs (e.g. discovery, subscriptions...)
- removed the "working draft" text from the v1.0.1 spec

I decided to leave the protobuf spec in here since it's basically done and
we're just waiting for final testing - but I did leave the "rc1" suffix.

Thanks to @JemDay and @clemensv for noticing.


duglin:
Approved on the 2/28 call

PR #768, opened 2021-02-08
changes from deep dive

duglin:
Changes based on our deep-dive sessions.

I decided to just jump to the more compact form of filters.


duglin:
Approved on the 2/11 call

Issue #565, opened 2020-02-03
Message Producer Signature


jkozlick:
I was reading through some of the issues that have been closed, namely #378 and #379, so sorry if this is a duplicate issue.

Using Cloudevents to publish events across systems where the publisher does not control the consumer endpoint seems to be problematic, as there is no way to verify within context attributes that the producer is actually the one that is sending the message. Is there a proposal being discussed on adding something along the lines of a "eventsignature" field within the the context attributes so that an event can be verified as coming from a specific producer without having to unpack the event data to do so. Would this be better served by extension context attributes?

duglin:
This topic has come up a few times and each time the group decided to defer it. Sometimes out of fear of the work becoming what WS-Security turned into, or because it was thought that it would be better handled by a layer on top of CloudEvents - e.g. perhaps via a secondary spec/extension as you suggested.

IMO I think the group might be open to discussing this as long as it doesn't block any of the existing work efforts. However, I think if someone really wanted to drive this then it would be best to start with a concrete proposal (even if everyone hates it 😄 ) because then it'll focus the discussion and not turn into endless rat-holes.

Anyone else have any more thoughts on this?

jkozlick:
Been a while since I made this comment, but the group that I was working with solved this by creating a canonical version of the event, signing it and using a JWT to pass the signed claims.

This has worked well, and allowed us to stay complaint with the Cloudevents specification without having to introduce any secondary specifications or extensions. 

I can outline this in more detail if anyone is interested, but I am OK with this issue being closed if no one else has any comment.

duglin:
@jkozlick thanks - I think I'll leave it open for a while and tag it for post v1.0 work to remind us to look it over.

duglin:
Been a while - I'd like to suggest that we close this one.  Any objections?

erikerikson:
Hi @jkozlick, I am curious if you've seen #770 or might have thoughts to add there?  A comparison might be valuable.  Perhaps better would be: how might your approach have been supported or hindered  by that extension?

cicorias:
> Been a while since I made this comment, but the group that I was working with solved this by creating a canonical version of the event, signing it and using a JWT to pass the signed claims.
> 
> This has worked well, and allowed us to stay complaint with the Cloudevents specification without having to introduce any secondary specifications or extensions.
> 
> I can outline this in more detail if anyone is interested, but I am OK with this issue being closed if no one else has any comment.

Could you share your approach here? Or DM me direct?

[...]

PR #770, opened 2021-02-12
Proposal : Signature Extension

JemDay:
Proposal for an extension to allow signatures to be transported in a consistent fashion.

For review & comment (be gentle, I'm not a security expert!)

### DIFF
--- a/extensions/signature.md
+++ b/extensions/signature.md
@@ -15,8 +15,7 @@ of selection.
 
 This extension supports a few general usage models:
 
-- Where the algorithm used to produce an embeded signature needs to be shared (
-  eg the `data` is a signed JWT).
+- Where the algorithm used to produce an attached signature needs to be shared.
 - Where the `data` is signed with a detached signature.
 - Where the context attributes are signed.
 
@@ -24,9 +23,8 @@ This extension supports a few general usage models:
 
 ### sigdata
 
-- Type: `String`
-- Description: The Base64 encoded value (per [RFC4648](rfc4648)) of the binary
-signature generated for the event `data`.
+- Type: `Binary`
+- Description: The binary signature generated for the event `data`.
 - Constraints:
 
   - OPTIONAL
@@ -42,9 +40,9 @@ signature generated for the event `data`.
 
 ### sigattr
 
-- Type: `String`
-- Description : The Base64 encoded value [RFC4648](rfc4648) of the binary
-signature generated for the context attributes (as-per method described below).
+- Type: `Binary`
+- Description : The binary signature generated for the context attributes 
+(as-per method described below).
 - Constraints:
 
   - OPTIONAL
--- a/extensions/signature.md
+++ b/extensions/signature.md
@@ -1,0 +1,110 @@
+# Signature Extension
+
+This extension defines a standard mechanism to carry signatures allowing
+event consumers to ensure the identity of the event producer and verify
+the event-data has not been tampered with during transport.
+
+This specification does not:
+
+- Declare a mechanism for public key exchange.
+- Declare a mechanism for public key selection during signature verification.
+Implementors may consider leveraging the CloudEvent `source` as a method
+of selection.
+
+## Usage
+
+This extension supports a few general usage models:
+
+- Where the algorithm used to produce an embeded signature needs to be shared (
+  eg the `data` is a signed JWT).
+- Where the `data` is signed with a detached signature.
+- Where the context attributes are signed.
+
+## Attributes
+
+### sigdata
+
+- Type: `String`
+- Description: The Base64 encoded value (per [RFC4648](rfc4648)) of the binary
+signature generated for the event `data`.
+- Constraints:
+
+  - OPTIONAL
+
+### sigdataalgo
+
+- Type : `String`
+- Description : The algorthim used to generate the `data` signature.
+- Constraints:
+
+  - OPTIONAL
+  - *Should* be present when `sigdata` is present.
+
+### sigattr
+
+- Type: `String`
+- Description : The Base64 encoded value [RFC4648](rfc4648) of the binary
+signature generated for the context attributes (as-per method described below).
+- Constraints:
+
+  - OPTIONAL
+
+### sigattralgo
+
+- Type: `String`
+- Description : The alorithm used to generate the attribute signature
+- Constraints:
+
+  - OPTIONAL
+  - *Should* be present when `sigattr` ifs present.
+
+### sigattrtype
+
+- Type: `String`
+- Description: Defines the familly of attributes that participated in
+the signing function. See [SigAttrType Values](sigattrtype-values)
+- Constraints:
+
+  - OPTIONAL
+  - *MUST* be present when `sigattr` is present.
+
+## SigAttrType Values
+
+This specification defines the following values for `sigattrtype`. Additional
+values MAY be defined by other specifications.
+
+### REQUIRED
+
+- When `sigattrtype` is set to `REQUIRED` it indicates that only *required*
+context attributes particpated in the signing process.
+
+### OPTIONAL
+
+- When `sigattrtype` is set to `OPTIONAL` it indicates that the *required
+and optional* context attributes participated in the signing process.
+
+### ALL
+
+- When `sigattrtype` is set to `ALL` it indicates that the *required, optional,
+and extension* context attributes participated in the signing process.
+
+NOTE: atributes defined by this extension *ARE NOT* considered during the
+signing process.
+
+## Attribute Signing & verification process
+
+- The set of attributes to be considered is constructed based on the scheme
+defined by `sigattrtype`.
+
+  - Any attribute with a null or empty value is ignored.
+  - Atrributes defined in this extension are ignored.
+
+- The attributes are sorted, by name, into acending alphabetical order.
+
+- The string value of each attribute are concatenated according to sorted
+order.
+
+- The signature is generated from the UTF-8 byte representation of the
+concatenated string.
+
+[rfc4648]: https://tools.ietf.org/html/rfc4648
--- a/extensions/signature.md
+++ b/extensions/signature.md
@@ -17,7 +17,7 @@ This extension supports a few general usage models:
 
 - Where the algorithm used to produce an attached signature needs to be shared.
 - Where the `data` is signed with a detached signature.
-- Where the context attributes are signed.
+- Where the CloudEvent context attributes are signed.
 
 ## Attributes
 
@@ -41,8 +41,8 @@ This extension supports a few general usage models:
 ### sigattr
 
 - Type: `Binary`
-- Description : The binary signature generated for the context attributes 
-(as-per method described below).
+- Description : The binary signature generated for the CloudEvent context
+attributes (as-per method described below).
 - Constraints:
 
   - OPTIONAL
@@ -89,7 +89,7 @@ and extension* context attributes participated in the signing process.
 NOTE: atributes defined by this extension *ARE NOT* considered during the
 signing process.
 
-## Attribute Signing & verification process
+## Attribute signing & verification process
 
 - The set of attributes to be considered is constructed based on the scheme
 defined by `sigattrtype`.
@@ -97,7 +97,7 @@ defined by `sigattrtype`.
   - Any attribute with a null or empty value is ignored.
   - Atrributes defined in this extension are ignored.
 
-- The attributes are sorted, by name, into acending alphabetical order.
+- The selected attributes are sorted by name into acending alphabetical order.
 
 - The string value of each attribute are concatenated according to sorted
 order.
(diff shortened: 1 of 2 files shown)


extensions/signature.md
```diff
@@ -0,0 +1,108 @@
...
+
+This extension defines a standard mechanism to carry signatures allowing
+event consumers to ensure the identity of the event producer and verify
+the event-data has not been tampered with during transport.
+
+This specification does not:
+
+- Declare a mechanism for public key exchange.
+- Declare a mechanism for public key selection during signature verification.
+Implementors may consider leveraging the CloudEvent `source` as a method
```
duglin, line 11: s/may/MAY/  or use "can"

extensions/signature.md
```diff
@@ -0,0 +1,108 @@
...
+  - OPTIONAL
+
+### sigdataalgo
+
+- Type : `String`
+- Description : The algorthim used to generate the `data` signature.
+- Constraints:
+
+  - OPTIONAL
+  - *Should* be present when `sigdata` is present.
```
duglin, line 39: s/Should/SHOULD/
No need for the italics

extensions/signature.md
```diff
@@ -0,0 +1,108 @@
...
+  - OPTIONAL
+
+### sigattralgo
+
+- Type: `String`
+- Description : The alorithm used to generate the attribute signature
+- Constraints:
+
+  - OPTIONAL
+  - *Should* be present when `sigattr` ifs present.
```
duglin, line 57: s/Should/SHOULD/
No need for the italics

extensions/signature.md
```diff
@@ -0,0 +1,108 @@
...
+
+### sigattrtype
+
+- Type: `String`
+- Description: Defines the familly of attributes that participated in
+the signing function. See [SigAttrType Values](#sigattrtype-values)
+- Constraints:
+
+  - OPTIONAL
+  - *MUST* be present when `sigattr` is present.
```
duglin, line 67: no need for the italics

extensions/signature.md
```diff
@@ -0,0 +1,108 @@
...
+- When `sigattrtype` is set to `OPTIONAL` it indicates that the *required
+and optional* context attributes participated in the signing process.
+
+### ALL
+
+- When `sigattrtype` is set to `ALL` it indicates that the *required, optional,
+and extension* context attributes participated in the signing process.
+
+NOTE: atributes defined by this extension *ARE NOT* considered during the
+signing process.
```
duglin, line 90: How does a receiver know which extension http headers are part of the signature since some may have been added after the signature was created? I wonder if you need a list of them and include that new attribute in the signature?

extensions/signature.md
```diff
@@ -0,0 +1,108 @@
...
+### sigdataalgo
+
+- Type : `String`
+- Description : The algorthim used to generate the `data` signature.
+- Constraints:
+
+  - OPTIONAL
+  - *Should* be present when `sigdata` is present.
+
+### sigattr
```
erikerikson, line 41: Are you accommodating for the potential for intermediaries to add or alter attributes?

extensions/signature.md
```diff
@@ -0,0 +1,108 @@
...
+
+- Type: `Binary`
+- Description : The binary signature generated for the CloudEvent context
+attributes (as-per method described below).
+- Constraints:
+
+  - OPTIONAL
+
+### sigattralgo
```
erikerikson, line 50: Do you imagine a use case where users will select different algorithms for signing the data attributes as opposed to the others?

Perhaps the previous comment explains, expecting multiple signers perhaps using separate algorithms?

extensions/signature.md
```diff
@@ -0,0 +1,108 @@
...
+- Description : The alorithm used to generate the attribute signature
+- Constraints:
+
+  - OPTIONAL
+  - *Should* be present when `sigattr` ifs present.
+
+### sigattrtype
+
+- Type: `String`
+- Description: Defines the familly of attributes that participated in
```
erikerikson, line 62: s/familly/family/

extensions/signature.md
```diff
@@ -0,0 +1,108 @@
...
+
+### sigdata
+
+- Type: `Binary`
+- Description: The binary signature generated for the event `data`.
+- Constraints:
+
+  - OPTIONAL
+
+### sigdataalgo
```
erikerikson, line 32: Does this provide an ID of the materials allowing verification (e.g. certificate common name, other type of public key ID, or even a shared key ID)?

The signature will need to be decrypted with the right material(s) to verify possession of identity by the sender.  How are those selected?

erikerikson (commented):
Thank you for offering this, I believe it is a really important one.  I am also not an expert.  However, I would like suggest that an array of signatures, not differentiating between data and other attributes, where each signature would identify the set of attributes signed, any overwritten attributes (including original values), the signature algorithm, the cryptographic materials used with the algorithm, and the signature itself.  I expect this would allow a potential chain of edits (including deletions) to be verified by verifying the last signature, modifying altered values to those overwritten, and moving on to the next signature until either a failure is detected (including on the initial signature) or no signatures remain.

erikerikson:
I nearly forgot to ask, what "content" is hashed and signed?  Is it the JSON stringification of the attributes identified attributes, taken in one at a time and in the given order and then appended into a single string without delimiters, the utf-8 bytes of which are then signed?  I wouldn't expect the algorithm to specify this as they take an array of bytes in my experience.  Like a transport, the "on the wire" format used for signing might need to be clearly specified so that consistent data is fed into the algorithm.

extensions/signature.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Attribute signing & verification process
+
+- The set of attributes to be considered is constructed based on the scheme
+defined by `sigattrtype`.
+
+  - Any attribute with a null or empty value is ignored.
+  - Atrributes defined in this extension are ignored.
+
+- The selected attributes are sorted by name into acending alphabetical order.
```
pri-naik, line 100: s/acending/ascending

extensions/signature.md
```diff
@@ -0,0 +1,108 @@
...
+- Constraints:
+
+  - OPTIONAL
+  - *Should* be present when `sigdata` is present.
+
+### sigattr
+
+- Type: `Binary`
+- Description : The binary signature generated for the CloudEvent context
+attributes (as-per method described below).
```
pri-naik, line 45: Can a reference be added to the exact heading where the method is defined?

extensions/signature.md
```diff
@@ -0,0 +1,108 @@
...
+
+- When `sigattrtype` is set to `OPTIONAL` it indicates that the *required
+and optional* context attributes participated in the signing process.
+
+### ALL
+
+- When `sigattrtype` is set to `ALL` it indicates that the *required, optional,
+and extension* context attributes participated in the signing process.
+
+NOTE: atributes defined by this extension *ARE NOT* considered during the
```
pri-naik, line 89: s/atributes/attributes

extensions/signature.md
```diff
@@ -0,0 +1,108 @@
...
+signing process.
+
+## Attribute signing & verification process
+
+- The set of attributes to be considered is constructed based on the scheme
+defined by `sigattrtype`.
+
+  - Any attribute with a null or empty value is ignored.
+  - Atrributes defined in this extension are ignored.
```
pri-naik, line 98: s/Atrributes/Attributes

extensions/signature.md
```diff
@@ -0,0 +1,108 @@
...
+- Type: `Binary`
+- Description: The binary signature generated for the event `data`.
+- Constraints:
+
+  - OPTIONAL
+
+### sigdataalgo
+
+- Type : `String`
+- Description : The algorthim used to generate the `data` signature.
```
pri-naik, line 35: Given the size constraints on each event, can there be a set of algorithms that are strongly recommended by this specification?

extensions/signature.md
```diff
@@ -0,0 +1,108 @@
+# Signature Extension
+
+This extension defines a standard mechanism to carry signatures allowing
```
pri-naik, line 3: In a typical scenario, where the events generated could be relayed by middleware to the eventual consumers, would it be appropriate to ask about multiple signers?

pri-naik (changes requested):

extensions/signature.md
JemDay: Ack

extensions/signature.md
JemDay: Ack

extensions/signature.md
JemDay: Ack

extensions/signature.md
JemDay: Ack

clemensv:
Someone should go down into the RFC mines and see whether we can pick up S/MIME in a reasonable way.

I would like to use something that is specifically not novel and not unique to CloudEvents and that has library coverage and tie-in into security infra like PKI.

extensions/signature.md
JemDay: Hi @erikerikson - I think that's related to a comment made by @duglin.

I'll be tweaking to support that use-case.

extensions/signature.md
JemDay: Ack

extensions/signature.md
JemDay: Good point .... there's a bit of a hand-wavy disclaimer at the begging around the selection of keys.

I'm open to suggestions ;-)

extensions/signature.md
```diff
@@ -0,0 +1,108 @@
...
+
+## SigAttrType Values
+
+This specification defines the following values for `sigattrtype`. Additional
+values MAY be defined by other specifications.
+
+### REQUIRED
+
+- When `sigattrtype` is set to `REQUIRED` it indicates that only *required*
+context attributes particpated in the signing process.
```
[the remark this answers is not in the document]
JemDay: ack

extensions/signature.md
JemDay: Ack (my spellings getting worse) !

extensions/signature.md
JemDay: Ack

extensions/signature.md
JemDay: ack

extensions/signature.md
JemDay: I'm loathe to get down to that level of detail, I see CE's as containers to enable data interchange, not to specification the "how" - does that make sense?

extensions/signature.md
JemDay: this seems similar to comments made by @erikerikson.

I'm open to suggestions as to the best way to represent this with a flat attribute structure...

extensions/signature.md
JemDay: I was wondering that as well ....

JemDay:
@clemensv 

> Someone should go down into the RFC mines and see whether we can pick up S/MIME in a reasonable way.
> 
> I would like to use something that is specifically not novel and not unique to CloudEvents and that has library coverage and tie-in into security infra like PKI.

Doesn't S/MIME work by using multipart messages - that way they're able to append the signature without monkeying with the original payload ?

Attributes are the only way I can see to transport this sort of information.

extensions/signature.md
erikerikson: Yes, I appreciate that you did.  The challenge is that keys roll over time.  While one could try and use a hopefully synchronized wall clock as a proxy for key roll over alignments over time but providing an attribute to write an identifier into is way easier if also duplicative across events.  So while I think it's right for the specification to stay out of the business of declaring the nature of the materials, facilitating the declaration of their identities (optionally if it makes sense) seems reasonable.

On a wider scope, I've made a suggestion on the top line comment of this review: that the signatures be an ordered array identifying the signed content, the resulting signature, the algorithm used, the id of the cryptographic materials used, and any altered values that were previously signed long with their prior values.  The algorithm for verification would take the event as given, verify it against the last signature of the event, modify the event to revert any alterations identified in the last signature, and pass the altered/reverted event along with the signature array (less the just verified signature) back into the same algorithm.  This process would exited with failure if any signature verification fails (or potentially only a subset of signatures that the receiver "cares about") but otherwise, once no more signatures exist to verify then verification is successful.  I apologize for offering what amounts to a bit of a rewrite of what you've offered but it seems like it might support a wider array of our scenarios.  Happy to own that work if you would prefer.

erikerikson:
It might be insufficient but perhaps this closes #565?  It's been a bit @JemDay.  Any chance you'll have time to bring it home?

duglin:
@JemDay what's the status of this one? If nothing else it needs a rebase

JemDay:
Yes .. I was hoping to get the _"easy"_ one done first ;-=)

duglin:
@JemDay just checking on the status of this one. Still interested?

JemDay:
PR Closed .. will re-work proposal into new document structure.

alexec:
Relates to #1093

PR #771, opened 2021-02-16
CloudEvents SQL Expression Language

### LINKED ISSUE #730: CloudEvents Expression Language
I want to propose to start a new effort to develop an expression language to filter CloudEvents.

## Motivation

One of the main drivers is the ability to have more powerful filtering capabilities in the subscription API, where the existing filter dialect `basic` is not powerful enough for most of production use cases. I foresee an eventual application of such language in the Discovery API and the Workflow spec. 

## Expected outcome

Our goal is to develop a spec for an expression language tailored for CloudEvents. Some goals of this language are, but not limited to, the following features:

* [Total functional programming language](https://en.wikipedia.org/wiki/Total_functional_programming), hence the termination of the evaluation of the expression is always guaranteed
* Type system conform to the [CloudEvents type system](https://github.com/cloudevents/spec/blob/master/spec.md#type-system)
* Boolean operators (and/or/nor/xor)
* Comparison operators on the types provided by the type system
* Regex
* Function invocation semantics + basic built-ins (e.g. type assertions/casting, 

It's out of the scope of the effort to develop a generic expression language, hence the input of the expression will always be a `CloudEvent`.

## Notes

* We could help the implementers providing the antlr grammar for it.
* We should implement this language in our sdks (I'm willing to help for java, golang, rust).
* I would prefer to exclude payload filtering from the initial efforts, since I consider it a complex argument (It comes to my mind problems like payload too big, supporting different content types, ...)
* We don't want anything particularly fancy regarding the syntax, so my feeling is that we should a C-like syntax or similar, something familiar for cloud developers out there

slinkydeveloper:
Fix #730

### DIFF
--- a/expression-language.md
+++ b/expression-language.md
@@ -55,7 +55,7 @@ producer, in an intermediary, and it can be implemented using any tech stack.
 The CloudEvents Expression Language assumes the input always includes, but it's not limited to, a single valid and type
 checked CloudEvent instance. An expression MUST NOT mutate the value of the input CloudEvent instance, nor any of the
 other input values. The evaluation of an expression observes the concept of [referential
-transparency][referential-transparency-wiki].
+transparency][referential-transparency-wiki]. The output of a CESQL expression evaluation is always a boolean _true_ or _false_ and it might include an error.
 
 The CloudEvents Expression Language doesn't support the handling of the data field of the CloudEvent instances, due to
 its polymorphic nature and complexity. We strongly encourage users that needs this functionality to use other more
--- a/expression-language.md
+++ b/expression-language.md
@@ -178,9 +178,9 @@ function-invocation ::= function-identifier "(" parameter-list? ")"
 
 The type system contains 3 types:
 
-- _String_: Sequence of allowable Unicode characters
-- _Integer_: A signed 32-bit integer
-- _Boolean_: A boolean value of "true" or "false"
+- _String_: Sequence of Unicode characters.
+- _Integer_: A whole number in the range -2,147,483,648 to +2,147,483,647 inclusive. This is the range of a signed, 32-bit, twos-complement encoding.
+- _Boolean_: A boolean value of "true" or "false".
 
 The [types defined in the CloudEvents specification][ce-type-system] URI, URI Reference and String are represented as
 _String_.
@@ -240,13 +240,18 @@ Corresponds to the syntactic rule `binary-operation`:
 | `x > y: Integer x Integer -> Boolean`   | Returns `true` if `x` is strictly greater than `y`                                              |
 | `x >= y: Integer x Integer -> Boolean`  | Returns `true` if `x` is greater or equal to `y`                                                |
 | `x * y: Integer x Integer -> Integer`   | Returns the product of `x` and `y`                                                              |
-| `x / y: Integer x Integer -> Integer`   | Returns the rounded division of `x` and `y`. Returns `0` if `y = 0` and raise an error          |
-| `x % y: Integer x Integer -> Integer`   | Returns the remainder of the division of `x` and `y`. Returns `0` if `y = 0` and raise an error |
+| `x / y: Integer x Integer -> Integer`   | Returns the truncated division of `x` and `y`. Returns `0` if `y = 0` and raise an error          |
+| `x % y: Integer x Integer -> Integer`   | Returns the remainder of the truncated division of `x` and `y`. Returns `0` if `y = 0` and raise an error |
 | `x + y: Integer x Integer -> Integer`   | Returns the sum of `x` and `y`                                                                  |
 | `x - y: Integer x Integer -> Integer`   | Returns the difference of `x` and `y`                                                           |
 | `x = y: String x String -> Boolean`     | Returns `true` if the values of `x` and `y` are equal                                           |
 | `x != y: String x String -> Boolean`    | Same as `NOT (x = y)`                                                                           |
 | `x <> y: String x String -> Boolean`    | Same as `NOT (x = y)`                                                                           |
+
+The modulo and divisions MUST follow the [truncated divisions definition][modulo-operation-wiki], that is:
+
+* The remainder of the modulo MUST have the same sign as the dividend
+* The quotient MUST be rounded towards zero _truncating_ the decimal part.
 
 #### 3.4.3. Like operator
 
@@ -373,3 +378,4 @@ INT(hop) < INT(ttl)
 [ce-id-attribute]: ./spec.md#id
 [subscriptions-filter-dialect]: ./subscriptions-api.md#3231-filter-dialects
 [ebnf-xml-spec]: https://www.w3.org/TR/REC-xml/#sec-notation
+[modulo-operation-wiki]: https://en.wikipedia.org/wiki/Modulo_operation
--- a/expression-language.md
+++ b/expression-language.md
@@ -1,0 +1,369 @@
+# CloudEvents SQL Expression Language
+
+## Abstract
+
+The goal of this specification is to define an expression language SQL-like which can be used to express predicates on
+CloudEvents instances.
+
+## Status of this document
+
+This document is a working draft.
+
+## Table of Contents
+
+1. [Introduction](#1-introduction)
+
+- 1.1. [Conformance](#11-conformance)
+- 1.2. [Relation to Subscriptions API](#12-relation-to-the-subscriptions-api)
+
+2. [Language syntax](#2-language-syntax)
+
+- 2.1. [Expression](#21-expression)
+- 2.2. [Value identifiers and literals](#22-value-identifiers-and-literals)
+- 2.3. [Operators](#23-operators)
+- 2.4. [Functions invocation](#24-functions-invocation)
+
+3. [Language semantics](#3-language-semantics)
+
+- 3.1. [Type System](#31-type-system)
+- 3.2. [CloudEvent context identifiers and types](#32-cloudevent-context-identifiers-and-types)
+- 3.3. [Errors](#33-errors)
+- 3.4. [Operators](#34-operators)
+- 3.5. [Functions](#35-functions)
+- 3.6. [Evaluation of the expression](#36-evaluation-of-the-expression)
+
+4. [Examples](#4-examples)
+5. [References](#5-references)
+
+## 1. Introduction
+
+CloudEvents SQL expressions (also known as CESQL) allow matching of CloudEvent attributes against complex expressions
+that lean on the syntax of Structured Query Language (SQL) `WHERE` clauses. Using SQL-derived expressions for message
+filtering is in widespread implementation use because the Java Message Service (JMS) message selector syntax also leans
+on SQL. Note that neither the SQL standard (ISO 9075) nor the JMS standard nor any other SQL dialect are used as a
+normative foundation or constrain the expression syntax defined in this specification, but the syntax is informed by
+them.
+
+CESQL is a _[Total pure functional programming language][total-programming-language-wiki]_ in order to guarantee the
+termination of the evaluation of the expression. It features a type system correlated to the [CloudEvents type
+system][ce-type-system] and it features boolean and arithmetics operations, as well as built-in functions for string
+manipulation.
+
+The language is not constrained to a particular execution environment, which means it might run in a source, in a
+producer, in an intermediary, and it can be implemented using any tech stack.
+
+The CloudEvents Expression Language assumes the input always includes, but it's not limited to, a single valid and type
+checked CloudEvent instance. An expression cannot mutate the value of the input CloudEvent instance, nor any of the
+other input values. The evaluation of an expression observes the concept of [referential
+transparency][referential-transparency-wiki].
+
+The CloudEvents Expression Language doesn't support the handling of the data field of the CloudEvent instances, due to
+its polymorphic nature and complexity. We strongly encourage users that needs this functionality to use other more
+appropriate tools.
+
+### 1.1. Conformance
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and
+"OPTIONAL" in this document are to be interpreted as described in [RFC2119][rfc2119].
+
+### 1.2. Relation to the Subscriptions API
+
+The CESQL can be used as a [filter dialect][subscriptions-filter-dialect] to filter out the input values
+
+<!-- TODO -->
+
+## 2. Language syntax
+
+The grammar of the language is defined using the EBNF Notation from [W3C XML specification][ebnf-xml-spec].
+
+Although in the EBNFs keywords are defined using uppercase characters, they are case-insensitive. For example:
+
+```
+int(hop) < int(ttl) and int(hop) < 1000
+```
+
+Has the same syntactical meaning of:
+
+```
+INT(hop) < INT(ttl) AND INT(hop) < 1000
+```
+
+### 2.1 Expression
+
+The root of the expression is the `expression` rule:
+
+```ebnf
+expression ::= value-identifier | literal | unary-operation | binary-operation | function-invocation | ( "(" expression ")" )
+```
+
+Nested expressions must be correctly parenthesized.
+
+### 2.2. Value identifiers and literals
+
+Value identifiers in CESQL have the same restrictions of the [Attribute Naming
+Convention][ce-attribute-naming-convention] from the CloudEvents spec. A value identifier cannot be greater than 20
+characters in length.
+
+```ebnf
+lowercase-char ::= [a-z]
+value-identifier ::= lowercase-char ( lowercase-char | digit )*
+```
+
+CESQL defines 3 different literal kinds: integer numbers, `true` or `false` booleans and `''` or `""` delimited strings.
+
+```ebnf
+digit ::= [0-9]
+number-literal ::= digit?
+
+boolean-literal ::= "true" | "false"
+
+string-literal ::= ( "'" ( [^'] | "\'" )* "'" ) | ( '"' ( [^"] | '\"' )* '"')
+
+literal ::= number-literal | boolean-literal | string-literal
+```
+
+Because string literals can be either `''` or `""` delimited, in the former case, the `"` has to be escaped, while in
+the latter the `"` has to be escaped.
+
+### 2.3. Operators
+
+CESQL defines boolean unary and binary operators, arithmetic unary and binary operators and the `LIKE`, `IN`, `EXISTS`
+operators.
+
+```ebnf
+not-operator = "NOT"
+unary-logical-operator = not-operator
+binary-logic-operator = "AND" | "OR" | "XOR"
+
+unary-numeric-operator = "-"
+binary-numeric-comparison-operator = "=" | ">=" | "<=" | "<" | ">"
+binary-numeric-arithmetic-operator = "+" | "-" | "*" | "/" | "%"
+
+like-operator = "LIKE"
+exists-operator = "EXISTS"
+in-operator = "IN"
+
+unary-operation ::= (unary-numeric-operator | unary-logical-operator) expression
+
+binary-operation ::= expression (binary-numeric-comparison-operator | binary-numeric-arithmetic-operator | binary-logic-operator) expression
+
+like-operation ::= expression not-operator? like-operator string-literal
+
+exists-operation ::= exists-operator value-identifier
+
+set-expression ::= "(" expression ("," expression)* ")"
+in-operation ::= expression not-operator? in-operator set-expression
+```
+
+Note: the `set-expression` supports only compile-time set definition, hence the language doesn't include a _set_ type
+that can be manipulated at runtime.
+
+### 2.4. Functions invocation
+
+CESQL supports N arity function invocation:
+
+```ebnf
+char ::= [A-Z] | [a-z]
+parameter ::= expression
+
+function-identifier ::= char ( "_" | char )*
+
+parameter-list ::= parameter ("," parameter)*
+function-invocation ::= function-identifier "(" parameter-list? ")"
+```
+
+## 3. Language semantics
+
+### 3.1. Type System
+
+The type system contains 3 types:
+
+- _String_: Sequence of allowable Unicode characters
+- _Integer_: A signed 32-bit integer
+- _Boolean_: A boolean value of "true" or "false"
+
+The [types defined in the CloudEvents specification][ce-type-system] URI, URI Reference and String are represented as
+_String_.
+
+### 3.2. CloudEvent context identifiers and types
+
+Each CloudEvent context attribute and extension MUST be addressable from an expression using its identifier, as defined
+by the spec. For example, using `id` in an expression will address to the CloudEvent [id attribute][ce-id-attribute].
+
+Unless otherwise specified, every attribute and extension MUST be represented by the _String_ type. Through explicit
+casting, the user can convert the addressed value instances to _Integer_ and _Boolean_.
+
+When addressed an attribute not included in the input event, an empty _String_ MUST be assumed as value
+
+### 3.3. Errors
+
+Because every operator and function is total, an expression evalution flow is defined statically and cannot be modified
+by expected or unexpected errors. Nevertheless CESQL includes the concept of errors: when an expression is evaluated, in
+case an error arise, the evaluator collects a list of errors, referred in this spec as _error list_, which is then
+returned together with the evaluated value of the CESQL expression.
+
+Whenever possible, some errors checks SHOULD be done at compile time by the expression evaluator, in order to prevent
+runtime errors.
+
+### 3.4. Operators
+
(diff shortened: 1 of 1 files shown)


expression-language.md
```diff
@@ -0,0 +1,369 @@
...
+### 1.1. Conformance
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and
+"OPTIONAL" in this document are to be interpreted as described in [RFC2119][rfc2119].
+
+### 1.2. Relation to the Subscriptions API
+
+The CESQL can be used as a [filter dialect][subscriptions-filter-dialect] to filter out the input values
+
+<!-- TODO -->
```
slinkydeveloper, line 73: I would love to revisit this paragraph after we merge this PR, in order to align with Subscription api (which I guess will need to contain a little text about this filter dialect)

expression-language.md
```diff
@@ -0,0 +1,369 @@
...
+The language is not constrained to a particular execution environment, which means it might run in a source, in a
+producer, in an intermediary, and it can be implemented using any tech stack.
+
+The CloudEvents Expression Language assumes the input always includes, but it's not limited to, a single valid and type
+checked CloudEvent instance. An expression cannot mutate the value of the input CloudEvent instance, nor any of the
+other input values. The evaluation of an expression observes the concept of [referential
+transparency][referential-transparency-wiki].
+
+The CloudEvents Expression Language doesn't support the handling of the data field of the CloudEvent instances, due to
+its polymorphic nature and complexity. We strongly encourage users that needs this functionality to use other more
```
skonto, line 61: From a user perspective it could be useful for some cases like json.

expression-language.md
```diff
@@ -0,0 +1,369 @@
...
+| ------------------------------------------------------ | ------------------------------------------------------------------------ |
+| `x IN (y1, y2, ...): String x String^n -> Boolean`     | Returns `true` if `x` is an element included in the set of `yN` elements |
+| `x NOT IN (y1, y2, ...): String x String^n -> Boolean` | Same as `NOT (x IN set)`                                                 |
+
+The set of values `yN` can be of arbitrary length, and the matching is done using the same semantics of the equal `=`
+operator.
+
+### 3.5. Functions
+
+The following tables show the built-in functions that MUST be supported by a CESQL evaluator.
```
skonto, line 283: UDFs will be too heavy?

expression-language.md
```diff
@@ -0,0 +1,369 @@
...
+- 3.4. [Operators](#34-operators)
+- 3.5. [Functions](#35-functions)
+- 3.6. [Evaluation of the expression](#36-evaluation-of-the-expression)
+
+4. [Examples](#4-examples)
+5. [References](#5-references)
+
+## 1. Introduction
+
+CloudEvents SQL expressions (also known as CESQL) allow matching of CloudEvent attributes against complex expressions
```
skonto, line 40: Expression are going to be visible on the yaml definitions eg. trigger?

expression-language.md
```diff
@@ -0,0 +1,369 @@
...
+When addressed an attribute not included in the input event, an empty _String_ MUST be assumed as value
+
+### 3.3. Errors
+
+Because every operator and function is total, an expression evalution flow is defined statically and cannot be modified
+by expected or unexpected errors. Nevertheless CESQL includes the concept of errors: when an expression is evaluated, in
+case an error arise, the evaluator collects a list of errors, referred in this spec as _error list_, which is then
+returned together with the evaluated value of the CESQL expression.
+
+Whenever possible, some errors checks SHOULD be done at compile time by the expression evaluator, in order to prevent
```
skonto, line 205: What are some examples of the compile time expression evaluation? If the SQL parser is going to be part of the sdk so people can parse the events they receive for example, then we could have some fluent api with some type safety eg. catch an invalid cast at compile time.

slinkydeveloper: > catch an invalid cast at compile time.

That's an example: you're trying to convert `time` to _Boolean_ with `BOOL(time)`. This can fail at compile time, since it never makes sense to convert `time` to _Boolean_

expression-language.md
slinkydeveloper: This spec is about the language per-se, which then vendors will figure out how to integrate in their offerings

expression-language.md
slinkydeveloper: Definitely too heavy and out of the scope. Note: the spec allows vendors to extend the set of built-ins, so I assume an engine can have a way to add custom functions.

But the language per-se doesn't allow function definition, it's too heavy and it might lead us into the turing completeness trap.

expression-language.md
slinkydeveloper: I see it, but as already discussed in previous Serverless WG meetings, it's just too hard to filter on the body. That's also something that middlewares might not want to do, because it's generally heavy.

Also note that for body filtering, depending on the "body nature", there might be tools better suited for that purpose: e.g. to filter a json payload a tool like json path/jsoniq definitely fits better than this language

expression-language.md
```diff
@@ -0,0 +1,369 @@
...
+Corresponds to the syntactic rule `unary-operation`:
+
+| Definition                  | Semantics                       |
+| --------------------------- | ------------------------------- |
+| `NOT x: Boolean -> Boolean` | Returns the negate value of `x` |
+| `-x: Integer -> Integer`    | Returns the minus value of `x`  |
+
+#### 3.4.2. Binary operators
+
+Corresponds to the syntactic rule `binary-operation`:
```
matejvasek, line 225: I don't see `<>` or `!=` operator, is that deliberate?

expression-language.md
```diff
@@ -0,0 +1,369 @@
...
+| `x = y: Integer x Integer -> Boolean`   | Returns `true` if the values of `x` and `y` are equal                                           |
+| `x < y: Integer x Integer -> Boolean`   | Returns `true` if `x` is strictly lower than `y`                                                |
+| `x <= y: Integer x Integer -> Boolean`  | Returns `true` if `x` is lower or equal to `y`                                                  |
+| `x > y: Integer x Integer -> Boolean`   | Returns `true` if `x` is strictly greater than `y`                                              |
+| `x >= y: Integer x Integer -> Boolean`  | Returns `true` if `x` is greater or equal to `y`                                                |
+| `x * y: Integer x Integer -> Integer`   | Returns the product of `x` and `y`                                                              |
+| `x / y: Integer x Integer -> Integer`   | Returns the rounded division of `x` and `y`. Returns `0` if `y = 0` and raise an error          |
+| `x % y: Integer x Integer -> Integer`   | Returns the remainder of the division of `x` and `y`. Returns `0` if `y = 0` and raise an error |
```
matejvasek, line 240: For clarity it might be useful to also add example where lhs is negative. Sometimes reminder/modulo get confused. For instance in `C` and `Python` the `%` operator have different semantic.

expression-language.md
```diff
@@ -0,0 +1,369 @@
...
+```ebnf
+lowercase-char ::= [a-z]
+value-identifier ::= lowercase-char ( lowercase-char | digit )*
+```
+
+CESQL defines 3 different literal kinds: integer numbers, `true` or `false` booleans and `''` or `""` delimited strings.
+
+```ebnf
+digit ::= [0-9]
+number-literal ::= digit?
```
matejvasek, line 116: shouldn't it be `digit+`?

expression-language.md
matejvasek: > rounded division

does that mean:
* truncated division -- rounded to zero
* floored division -- rounded downward
* rounded to nearest integer

https://en.wikipedia.org/wiki/Modulo_operation

expression-language.md
```diff
@@ -0,0 +1,369 @@
+# CloudEvents SQL Expression Language
+
+## Abstract
+
+The goal of this specification is to define an expression language SQL-like which can be used to express predicates on
```
duglin, line 5: at some point run it thru that md prettier tool so it wraps stuff at 80 columns

duglin: This is based on an existing SQL RFC, right? If so, perhaps we should mention it and say whether it's compliant or where the differences are.

expression-language.md
```diff
@@ -0,0 +1,369 @@
...
+4. [Examples](#4-examples)
+5. [References](#5-references)
+
+## 1. Introduction
+
+CloudEvents SQL expressions (also known as CESQL) allow matching of CloudEvent attributes against complex expressions
+that lean on the syntax of Structured Query Language (SQL) `WHERE` clauses. Using SQL-derived expressions for message
+filtering is in widespread implementation use because the Java Message Service (JMS) message selector syntax also leans
+on SQL. Note that neither the SQL standard (ISO 9075) nor the JMS standard nor any other SQL dialect are used as a
+normative foundation or constrain the expression syntax defined in this specification, but the syntax is informed by
```
duglin, line 44: How far away from them are we? Is there a way to soften this statement?

expression-language.md
```diff
@@ -0,0 +1,369 @@
...
+appropriate tools.
+
+### 1.1. Conformance
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and
+"OPTIONAL" in this document are to be interpreted as described in [RFC2119][rfc2119].
+
+### 1.2. Relation to the Subscriptions API
+
+The CESQL can be used as a [filter dialect][subscriptions-filter-dialect] to filter out the input values
```
duglin, line 71: s/out/on/   maybe?

expression-language.md
```diff
@@ -0,0 +1,369 @@
...
+
+### 2.1 Expression
+
+The root of the expression is the `expression` rule:
+
+```ebnf
+expression ::= value-identifier | literal | unary-operation | binary-operation | function-invocation | ( "(" expression ")" )
+```
+
+Nested expressions must be correctly parenthesized.
```
duglin, line 99: upper case the RFC2119 keywords

expression-language.md
```diff
@@ -0,0 +1,369 @@
...
+
+### 2.2. Value identifiers and literals
+
+Value identifiers in CESQL have the same restrictions of the [Attribute Naming
+Convention][ce-attribute-naming-convention] from the CloudEvents spec. A value identifier cannot be greater than 20
```
duglin, line 104: s/cannot/MUST NOT/

expression-language.md
erikerikson: In support of Dug, I had thought that in discussion the plan for this path was to use an existing standard so as to be able to reuse existing implementations and knowledge.  Were there particular reasons for departing from a pre-existing canon?

expression-language.md
slinkydeveloper: Ops, lemme add them

expression-language.md
slinkydeveloper: I wrapped it at 120, is that ok? I don't want to have tables too much shrinked

expression-language.md
slinkydeveloper: I think that's the major point of the discussion we had during the last meeting, where I pointed out the difference between the other JS like language and this language. We just can't claim we have a SQL subset, because otherwise:

* We have to choose **a specific SQL dialect** and **stick with it**: there are many out there and, in particular when it comes to built-ins, they all have different definitions.
* It gives us more freedom of movement when evolving this language: we might decide to add new features, maybe from another specific SQL dialect, that fits our needs. In particular when it comes down to _able to reuse existing implementations_, it literally means that we just can't move away from that particular standard and just use it as is (and evolve together with it).

Hence I believe this statement should not be soften, it should remain exactly as is to show our intention of giving you a SQL-like language, but not SQL-X or SQL-Y.

expression-language.md
```diff
@@ -0,0 +1,375 @@
...
+
+like-operation ::= expression not-operator? like-operator string-literal
+
+exists-operation ::= exists-operator value-identifier
+
+set-expression ::= "(" expression ("," expression)* ")"
+in-operation ::= expression not-operator? in-operator set-expression
+```
+
+Note: the `set-expression` supports only compile-time set definition, hence the language doesn't include a _set_ type
```
clemensv, line 158: I don't think that statement is necessary, because the language only allows for expressions. You can't manipulate anything at runtime.

expression-language.md
```diff
@@ -0,0 +1,375 @@
...
+```
+
+## 3. Language semantics
+
+### 3.1. Type System
+
+The type system contains 3 types:
+
+- _String_: Sequence of allowable Unicode characters
+- _Integer_: A signed 32-bit integer
```
clemensv, line 182: That should be bigger. Since we allow for arithmetic here, it's pretty easy to get into overflow situations even if CloudEvents metadata is constrained to 32-bit value space.

expression-language.md
```diff
@@ -0,0 +1,375 @@
...
+- _Boolean_: A boolean value of "true" or "false"
+
+The [types defined in the CloudEvents specification][ce-type-system] URI, URI Reference and String are represented as
+_String_.
+
+### 3.2. CloudEvent context identifiers and types
+
+Each CloudEvent context attribute and extension MUST be addressable from an expression using its identifier, as defined
```
clemensv, line 190: That section is a bit more complicated than it needs to be, IMO.

expression-language.md
```diff
@@ -0,0 +1,375 @@
...
+Unless otherwise specified, every attribute and extension MUST be represented by the _String_ type. Through explicit
+casting, the user can convert the addressed value instances to _Integer_ and _Boolean_.
+
+When addressed an attribute not included in the input event, an empty _String_ MUST be assumed as value
+
+### 3.3. Errors
+
+Because every operator and function is total, an expression evalution flow is defined statically and cannot be modified
+by expected or unexpected errors. Nevertheless CESQL includes the concept of errors: when an expression is evaluated, in
+case an error arise, the evaluator collects a list of errors, referred in this spec as _error list_, which is then
```
clemensv, line 202: I think you throw or you don't. If you throw you get an error. If you don't, you get a result.

expression-language.md
```diff
@@ -0,0 +1,375 @@
...
+| `x != y: Integer x Integer -> Boolean`  | Same as `NOT (x = y)`                                                                           |
+| `x <> y: Integer x Integer -> Boolean`  | Same as `NOT (x = y)`                                                                           |
+| `x < y: Integer x Integer -> Boolean`   | Returns `true` if `x` is strictly lower than `y`                                                |
+| `x <= y: Integer x Integer -> Boolean`  | Returns `true` if `x` is lower or equal to `y`                                                  |
+| `x > y: Integer x Integer -> Boolean`   | Returns `true` if `x` is strictly greater than `y`                                              |
+| `x >= y: Integer x Integer -> Boolean`  | Returns `true` if `x` is greater or equal to `y`                                                |
+| `x * y: Integer x Integer -> Integer`   | Returns the product of `x` and `y`                                                              |
+| `x / y: Integer x Integer -> Integer`   | Returns the rounded division of `x` and `y`. Returns `0` if `y = 0` and raise an error          |
```
clemensv, line 243: Rounding is trouble.

expression-language.md
```diff
@@ -0,0 +1,375 @@
...
+The set of values `yN` can be of arbitrary length, and the matching is done using the same semantics of the equal `=`
+operator.
+
+### 3.5. Functions
+
+The following tables show the built-in functions that MUST be supported by a CESQL evaluator.
+
+#### 3.5.1. Casting and type checking
+
+| Definition                      | Semantics                                                                                                                        |
```
clemensv, line 293: You are punting on type inference here. I realize that's a grueling topic, but type casts should be a last resort and we should avoid having them completely, IMO.

expression-language.md
```diff
@@ -0,0 +1,375 @@
...
+or the subject with value 'Francesco Guardiani'_
+
+```
+(firstname = 'Francesco' AND lastname = 'Guardiani') OR subject = 'Francesco Guardiani'
+```
+
+_CloudEvent including the extension 'sequence' with numeric value 10_
+
+```
+INT(sequence) = 10
```
clemensv, line 355: 10 is an integer and that implies the type cast by inference.

expression-language.md
```diff
@@ -0,0 +1,375 @@
...
+```
+
+_CloudEvent including the extension 'hop' and 'ttl', where 'hop' is smaller than 'ttl'_
+
+```
+INT(hop) < INT(ttl)
```
clemensv, line 361: < is only allowed for numerical types and therefore the type casts can also be inferred

clemensv (commented):
I left a few comments. Right direction overall.

expression-language.md
```diff
@@ -0,0 +1,375 @@
...
+
+CESQL is a _[Total pure functional programming language][total-programming-language-wiki]_ in order to guarantee the
+termination of the evaluation of the expression. It features a type system correlated to the [CloudEvents type
+system][ce-type-system] and it features boolean and arithmetics operations, as well as built-in functions for string
+manipulation.
+
+The language is not constrained to a particular execution environment, which means it might run in a source, in a
+producer, in an intermediary, and it can be implemented using any tech stack.
+
+The CloudEvents Expression Language assumes the input always includes, but it's not limited to, a single valid and type
```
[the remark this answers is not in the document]
slinkydeveloper: I wanted to leave room for vendors to add other "context dependent" input variables. output is true or false.

expression-language.md
slinkydeveloper: Well, if we decide to have a set type, then an expression could return a set. I'm just underlining that, given this type doesn't exist, sets are only a compile time thing.

expression-language.md
slinkydeveloper: Copy paste from the ce spec integer definition?

expression-language.md
slinkydeveloper: I didn't wanted to place this constraint: the evaluation always ends, nevertheless an error happened or not. In other words, the evaluation can return:

* true, no error
* false, no error
* true, error
* false, error

This gives to the user the flexibility of deciding between a "strict" and a "soft" mode, and whether some errors may be more "critical" than others and treating such failures as `false`

slinkydeveloper:
@clemensv I saw different comments about type inference. In my view of this language, types are always explicit, but I'm open to add implicit type casting if you think it's necessary. Do you want to draft a paragraph about it and PR it on my branch?

expression-language.md
```diff
@@ -0,0 +1,375 @@
+# CloudEvents SQL Expression Language
```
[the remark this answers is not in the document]
slinkydeveloper: How about _CloudEvents SQL Predicate Language_? I don't like having _Filter_ in it, nothing stops you from using predicates defined with CESQL for other purposes

expression-language.md
slinkydeveloper: Any ideas on how to simplify it?

expression-language.md
lionelvillard: can you reorder the EBNF rules so that the child nodes appear after the parent nodes? For instance `unary-operation` and `binary-operation` should appear first.

lionelvillard: since the language allows the creation of set then there must be a type set. You can't have one a without the other. Otherwise is the type of the value constructed by `set-expression` ?

expression-language.md
matejvasek: I would picked `truncated division` since it seems that what `SQLite`, `Go` and `Clojure` does. Also `%` should behave accordingly.

@clemensv this is probably related to your rounding comment.

expression-language.md
lionelvillard: I agree, but I think this is unavoidable because CloudEvent extensions don't have type annotations. They can only be inferred from the lexical value, which sometimes leads to ambiguous case, eg. `true as string` vs `true as boolean`.

expression-language.md
slinkydeveloper: > since the language allows the creation of set then there must be a type set. You can't have one a without the other. Otherwise is the type of the value constructed by set-expression ?

The language doesn't allow for the definition of a set, that's exactly what i was talking about. Syntactically and semantically speaking, the set-expression is just part of the _in_ operator definition. I can just take the set-expression ebnf and collapse it in side the in-operation and things won't change.

lionelvillard: > The language doesn't allow for the definition of a set,

isn't what `set-expression` is doing, creating a set?  

>  I can just take the set-expression ebnf and collapse it in side the in-operation and things won't change

Inlining does not change the fact the comma operator in `(" expression ("," expression)* ")"` constructs a set (as you call it).

expression-language.md
```diff
@@ -0,0 +1,375 @@
...
+normative foundation or constrain the expression syntax defined in this specification, but the syntax is informed by
+them.
+
+CESQL is a _[Total pure functional programming language][total-programming-language-wiki]_ in order to guarantee the
+termination of the evaluation of the expression. It features a type system correlated to the [CloudEvents type
+system][ce-type-system] and it features boolean and arithmetics operations, as well as built-in functions for string
+manipulation.
+
+The language is not constrained to a particular execution environment, which means it might run in a source, in a
+producer, in an intermediary, and it can be implemented using any tech stack.
```
duglin, line 53: s/tech/technology/

expression-language.md
```diff
@@ -0,0 +1,375 @@
...
+```ebnf
+expression ::= value-identifier | literal | unary-operation | binary-operation | function-invocation | ( "(" expression ")" )
+```
+
+Nested expressions MUST be correctly parenthesized.
+
+### 2.2. Value identifiers and literals
+
+Value identifiers in CESQL have the same restrictions of the [Attribute Naming
+Convention][ce-attribute-naming-convention] from the CloudEvents spec. A value identifier MUST NOT be greater than 20
```
duglin, line 104: since the ce spec only has this as a SHOULD, perhaps we should as well?

expression-language.md
```diff
@@ -0,0 +1,375 @@
...
+characters in length.
+
+```ebnf
+lowercase-char ::= [a-z]
+value-identifier ::= lowercase-char ( lowercase-char | digit )*
```
duglin, line 109: oddly enough, CE allows for names to start with a digit. I'm not sure if that is correct or not, but we should align with it.

expression-language.md
slinkydeveloper: Ah I wasn't aware that the size limit is a SHOULD in the cloudevents spec...

I see your point on compatibility and it makes sense to me, aka _what happens if somebody doesn't respect that SHOULD of the length?_, but on the other hand, the syntactic rules of the language must define some bound for the identifier length IMO. 

So or we prefix some greater arbitrary number, say:

> A value identifier SHOULD NOT be greater than 20 characters in length and MUST NOT be greater than 40 characters in length.

Or we just assume events not compliant with the 20 characters attribute name bound won't work with this language. WDYT?

expression-language.md
slinkydeveloper: Ah ok I wasn't aware of that, let me fix it

expression-language.md
slinkydeveloper: check it out now

expression-language.md
slinkydeveloper: Look now

expression-language.md
slinkydeveloper: For me explicit type casting should definitely be there, for the reasons mentioned by @lionelvillard. Given that, as discussed during the meeting, I'll try to draft some inference rules here

expression-language.md
slinkydeveloper: > isn't what set-expression is doing, creating a set?

Yes, but you can't use it anywhere else than in the _IN_ operator. For me this is a detail of the definition of _IN_, not an actual type to be included in the type system. It's just a list of parameters for the _IN_ operator to compute.

slinkydeveloper:
@clemensv I drafted the type inference, can you check it out?

expression-language.md
matejvasek: I think it's good now.

expression-language.md
slinkydeveloper: done, check it out now

expression-language.md
lionelvillard: > Yes, but you can't use it anywhere else than in the IN operator

It does not matter how often it is used. This is perfectly normal for a programming language to define more types than the input it can process (eg. SQL, XPath, XQuery, etc..). 

> It's just a list of parameters for the IN operator to compute.

Variadic parameter lists are typically syntactic sugar for constructing a list/set via the function call syntax.

slinkydeveloper: So how about I just remove this sentence and that's it? Or do you think we should have a proper definition in the _Type system_ paragraph when we talk about the supported types?

lionelvillard: +1 to remove this sentence and to have a proper definition in the type system paragraph.

slinkydeveloper: @lionelvillard look at this commit https://github.com/cloudevents/spec/pull/771/commits/b1d75246ceff146fc83bc21497a18affaca1d751

I've defined set as a collection of strings, in order to restrict it to exactly what `IN` is capable of

duglin:
Approved on the 3/18 call

PR #780, opened 2021-03-02
Fix typo in schema registry spec

pierDipi:
As per title, fx typo in schema registry spec, schwema -> schema.


duglin:
executive (hard) decision... merging

PR #783, opened 2021-03-10
fix build

duglin:
Remove missing files from checkers


duglin:
executive decision - merging just to fix the build

Issue #755, opened 2021-01-15
Changelog / Version Differences


grant:
Hello CE folks,

Does the CE spec have a changelog? I wish to know the differences between 1.0.0 and 1.0.1 for our libraries. I assume there are no breaking changes, but I cannot find a changelog summarizing the changes.

A `CHANGELOG.md` or summarized version cut in https://github.com/cloudevents/spec/releases/tag/v1.0.1 would be nice.

## Example

https://github.com/facebook/react/blob/master/CHANGELOG.md

(There are tools to help like https://github.com/apps/semantic-pull-requests)

It seems like we could categorize PRs with conventional commit (like `docs:`, `fix:`) and then group changes appropriately.

## Current note:

```
Add protobuf format as a sub-protocol - #721
Allow JSON values to be null, meaning unset - #713
[Primer] Adding a Non-Goal w.r.t Security - #712
WebSockets protocol binding - #697
Clarify difference between message mode and HTTP content mode - #672
add missing sdks to readme - #666
New sdk maintainers rules - #665
move sdk governance and cleanup - #663
Bring 'datadef' definition into line with specification - #658
Add CoC and move some governance docs into 'community' - #656
Add blog post around understanding Cloud Events interactions - #651
SDK governance draft - #649
docs: add common processes for SDK maintainers and contributors - #648
Adding Demo for Cloud Events Orchestration - #646
Clarified MUST requirement for JSON format - #644
Re-Introducing Protocol Buffer Representation - #626
Closes #615 - #616
Reworked Distributed Tracing Extension - #607
Minor updates to Cloud Events Primer - #600
Kafka clarifications - #599
Proprietary binding spec inclusion guide - #595
Adding link to Pub/Sub binding - #588
Add some clarity around SDK milestones - #584
How to determine binary CE vs random non-CE message - #577
Adding Visual Studio Code extension to community open-source doc - #573
Specify encoding of kafka header keys and values and message key - #572
Fix distributed tracing example - #569
Paragraph about nested events to the primer - #567
add rules for changing Admins- #564
Updating JSON Schema - #563
Say it's ok to ignore non-MUST recommendations - at your own risk - #562
Update Distributed Tracing extension spec links - #550
Add Ruby SDK to SDK lists - #548
```

An SDK wants support 1.0.1, and I'd like to see changes between 1.0.1 and 1.0.0.

duglin:
I'm confused - isn't the list you pasted above the changelog you're looking for?

[...]

Issue #687, opened 2020-08-19
Is there a spec version history available?


jskeet:
I'm looking at the spec version support in the C# SDK - I'd like to *at least* remove support for 0.1 and 0.2, and I want to check the differences between 0.3 and 1.0 to see how awkward it is to support both. (The spec *recommends* but doesn't *require* supporting 0.3, so I think it's reasonable to explore how tricky that is.)

Currently I can see https://github.com/cloudevents/spec/releases/tag/v1.0 which gives the differences between 1.0 and 1.0-rc1, but not in a particularly helpful semantic way... and there's nothing similar earlier than that.

Is there a document I've missed somewhere that gives the information in a form useful to SDK developers, or am I best off just doing a diff of the 0.3 and 1.0 spec? If that *is* the best option available, would a version history doc be welcomed? If I'm going to have to write up notes myself, I might as well share it...

duglin:
@jskeet good questions.  Unfortunately, the list of github PRs per release that we put into the release notes is really all we have now. If you have the time/desire to write-up a more human friendly description of the major diffs between v0.3 and v1.0 I think that would be great!

jskeet:
Righto - will do. I don't expect it'll be *soon* (as this is only one of the many pies I have my fingers in), but hopefully I'll have something in the next couple of weeks.

duglin:
@jskeet any update on this?  No worries if you're busy... just doing a quick issue triage.

jskeet:
I'm afraid not - I'm hoping to get to it next week. Still on it.

jskeet:
Right, some genuine progress now. I've *only* compared `spec.md` in v1.0 vs v0.3, but here are the changes - I believe this is an exhaustive list aside from reformatting and moving paragraphs around a bit:

- Use "producer" and "consumer" instead of "sender" and "receiver"
- Clarification that intermediaries should forward optional attributes
- Remove constraint that attribute names must start with a letter
- Remove suggestion that attributes names should be descriptive and terse
- Clarify that a single occurrence may result in more than one event
- Add an Event Data section (replacing `data`), making event data a top level concept rather than an attribute
- Introduce an Event Format section
- Define structured-mode and binary-mode messages
- Define protocol binding
- Add extension attributes into "context attributes" description
- Move mention of attribute serialization mechanism from "context attributes" description into
  "type system"
- Change "transport" to "protocol" universally
- Introduce the Boolean, URI and URI-reference types for attributes
[...]

PR #784, opened 2021-03-12
ci: conventional commits

grant:
Per issues:

- https://github.com/cloudevents/spec/issues/755
- https://github.com/cloudevents/spec/issues/687

It is difficult to determine the differences between spec versions and progression to the spec as time goes on.

---

One way we can help developers understand the spec changes is by using and enforcing [conventional commits](www.conventionalcommits.org).

With these types of commits, we can generate more detailed changelogs in the future. Unfortunately, there's not a whole lot we can do with unstructured commit messages from the past.

### Proposed Improvement in this PR

This PR adds a GitHub Action to this repo to add a check to pushes and pull requests to see if the last commit is a conventional commit. Note: PRs when squashed and merged can use a single commit message.

- Adds `.github/workflows/commitlint.yaml`
- Updates `.github/CONTRIBUTING.md`, mentioning how contributors should use conventional commits.
  - Note: Updated TOC was done automatically in VS Code

PR mergers should merge PRs with a conventional commit message to keep the CI green.

CC @duglin this is following up the CNCF chat about this yesterday.

### DIFF
--- a/.github/workflows/commitlint.yaml
+++ b/.github/workflows/commitlint.yaml
@@ -1,0 +1,11 @@
+name: Lint Commit Messages
+on: [pull_request]
+
+jobs:
+  commitlint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: wagoid/commitlint-github-action@v2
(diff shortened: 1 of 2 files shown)


duglin:
approved on 3/25 call

PR #787, opened 2021-03-16
Add a deprecated timestamp

### LINKED ISSUE #705: [DISCOVERY] Service Lifecycle Clarity
The service definition does not support the state of a service (ACTIVE, DEPRECATED, RETIRED) etc.

Clients should be able to search/filter by such states enabling them to proactively plan to move off services/endpoints as they are decommissioned. Ideally a state of DEPRECATED would be associated with a date for final retirement.

Additionally (and maybe this is a separate issue).. Is a _subscriptionurl_ immutable? If not what how would a client know that it has changed, and what notice would they get that it's about to change?.

duglin:
Closes #705

### DIFF
--- a/discovery/discovery.yaml
+++ b/discovery/discovery.yaml
@@ -234,6 +234,21 @@ components:
           type: string
           description: Absolute URL that provides a link to additional documentation about the service. This is intended for a developer to use in order to learn more about this service's events produced
           example: http://cloud.example.com/docs/blobstorage
+        deprecated:
+          type: object
+          properties:
+            effectivetime:
+              type: string
+              description: The time at which the service will enter a deprecated state
+            removaltime:
+              type: string
+              description: The time at which the service will be removed
+            alternative:
+              type: string
+              description: A URL to an possible alternative service
+            docsurl:
+              type: string
+              description: A URL to additional information concerning the deprecation of this service. Possible information might include rationale behind the action, or discussion of alternatives
         specversions:
           type: array
           description: CloudEvents specversions that can be used for events published for this service
--- a/discovery/spec.md
+++ b/discovery/spec.md
@@ -132,6 +132,12 @@ Service:
   "url": "[unique URL to this service]",
   "description": "[human string]", ?
   "docsurl": "[URL reference for human documentation]", ?
+  "deprecated": { ?
+	"effectivetime": "[RFC3339 timestamp of when the service will enter a deprecated state]", ?
+    "removaltime": "[RFC3339 timestamp of when service will be removed]", ?
+    "alternative": "[URL to possible alternative service]", ?
+    "docsurl": "[URL reference to additional information]" ?
+  },
   "specversions": [ "[ce-specversion value]" + ],
   "subscriptionurl": "[URL to which the Subscribe request will be sent]",
   "subscriptionconfig": { ?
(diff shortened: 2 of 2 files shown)


discovery.md
```diff
@@ -136,6 +136,7 @@ Service:
   "url": "[unique URL to this service]",
   "description": "[human string]", ?
   "docsurl": "[URL reference for human documentation]", ?
+  "removaltime": "[RFC339 timestamp of when service will be removed]", ?
```
tomkerkhove, line 139: Shouldn't we give guidance on what to migrate to?
```suggestion
  "removal": {
    "time": "[RFC339 timestamp of when service will be removed]", ?
    "alternative": "", ?
  }
```

duglin: Questions from the group:

John:
- what kind of guarantees are there around "alt"?  does it have it to exist before the time?
- do subscriptions carry over to the alt?

- Erik: any kind of transition?

Jennifer:
- what's the value? id or url? machine or human readable?
- or how does someone interpret it?

duglin: My recommendation, if we choose to include "alt":
- the "alt" must be available immediately 
- subscriptions do not carry over and there's no guarantee that the 'alt' is a semantic replacement for the service. It is simply a recommendation for consideration and the user is expected to investigate if it's a good match.
- no transitioning or transferring of subscriptions to "alt"
- "alt" must be the URL of the alt service. MAY be in another DE, but no guarantee that the client can access it.

duglin: s/removal/deprecated/

duglin:
ok ready for another round of reviews
@cloudevents/notify

discovery/spec.md
```diff
@@ -132,6 +132,10 @@ Service:
   "url": "[unique URL to this service]",
   "description": "[human string]", ?
   "docsurl": "[URL reference for human documentation]", ?
+  "deprecated": { ?
+    "removaltime": "[RFC339 timestamp of when service will be removed]", ?
```
JemDay, line 136: Following the spirit of the main comment thread to follow the HTTP semantics shouldn't we call this `sunsettime` ??

JemDay (commented):
Suggest following the other convention mention to use the term `sunset` to represent the point at which the item is no longer available.

discovery/spec.md
duglin: done

duglin:
This one is ready for review on this week's call.
@cloudevents/notify

discovery/discovery.yaml
```diff
@@ -238,6 +238,15 @@ components:
           type: string
           description: Absolute URL that provides a link to additional documentation about the service. This is intended for a developer to use in order to learn more about this service's events produced
           example: http://cloud.example.com/docs/blobstorage
+        deprecated:
+          type: object
+          properties:
+            sunsettime:
+              type: string
+              description: The time at which the service will be removed
+            alterntaive:
```
jskeet, line 247: Typo

discovery/spec.md
```diff
@@ -304,6 +308,44 @@ The following sections define the attributes that appear in a Service entity.
 - Examples:
   - `http://cloud.example.com/docs/blobstorage`
 
+##### deprecated
+
+- Type: Object containing 2 properties:
+  - sunsettime<br>
+    An OPTIONAL property indicating the time when the Service MAY be removed.
```
jskeet, line 315: Given that we don't use the word "sunset" anywhere in the description, maybe "removaltime"? That also avoids the easily-typo'd double t.

discovery/spec.md
```diff
@@ -304,6 +308,44 @@ The following sections define the attributes that appear in a Service entity.
...
+    The Service MUST NOT be deleted before this time. If this property is not
+    present then client can not make any assumption as to when the Service
+    might be removed. Note: as with most Service properties, this property
+    is mutable. If present, this MUST be an [RFC339][rfc3339] timestamp.
```
duglin, line 319: Add a 'started' time

discovery/spec.md
jskeet: Possibly earliestremovaltime to indicate that it's a lower bound...

discovery/discovery.yaml
duglin, line 247: typo

discovery/spec.md
duglin: per the call today we switched to "removaltime"

duglin:
Made all of the updates based on today's call.

duglin:
approved on the 5/5 call

PR #790, opened 2021-03-26
[CESQL] Better default value for division by 0

slinkydeveloper:
Changed the default value of division by `0` to a more reasonable default.


slinkydeveloper:
Since this PR will have conflicts with #788, I'll rebase it and run prettier after the #788 is merged

duglin:
On the 4/8 call there wasn't consensus on whether returning max/min int or zero would be more intuitive. So we decided to keep it at zero for now until we more data. Let's reopen after we get more experience if we want to change it.

PR #791, opened 2021-03-26
[CESQL] Return any value

slinkydeveloper:
Allow the expression to return any value, not just boolean.

This makes the language usable for a broader range of use cases, but at the same time I clarify that for filtering, the return value should always be bool (either is bool, or it must be casted to bool)

### DIFF
--- a/expression-language.md
+++ b/expression-language.md
@@ -38,7 +38,7 @@ This document is a working draft.
 
 ## 1. Introduction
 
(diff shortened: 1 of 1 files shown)


slinkydeveloper:
Same considerations as https://github.com/cloudevents/spec/pull/790#issuecomment-808079454

duglin:
Approved on the 4/1 call

PR #792, opened 2021-03-30
[CESQL] Clarification of the IN semantics

slinkydeveloper:
The `IN` operator semantics were unclear, in particular its relation with the implicit type casting rule.

This clarification specified that no implicit casting rules are applied for `IN`, that is:

```
'123' IN (123)
```

Evaluates to `FALSE`. One can rewrite this expression using explicit cast, like:

```
'123' IN (STRING(123))
```

### DIFF
--- a/expression-language.md
+++ b/expression-language.md
@@ -290,8 +290,9 @@ assumed valid, e.g. `EXISTS id` MUST always return `true`.
 | `x IN (y1, y2, ...): String x String^n -> Boolean`     | Returns `true` if `x` is equal to an element in the _Set_ of `yN` elements |
 | `x NOT IN (y1, y2, ...): String x String^n -> Boolean` | Same as `NOT (x IN set)`                                                   |
 
-The matching is done using the same semantics of the equal `=` operator, except for the implicit casting rules described in the [Type casting paragraph](#37-type-casting).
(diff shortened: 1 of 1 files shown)


slinkydeveloper:
I've did a little investigation on how the implicit type casting behaves in conjunction with the `IN` operator...

Oracle SQL coerces the types in the condition, hence this predicate `'8' IN (2, 3, 5, 8)` returns true and this predicate `'8' IN (2, 3, 5)` returns false. Because oracle sql doesn't have the boolean keywords, I wasn't able to test with booleans `true` and `false`.

In MySQL I see a similar behavior with numbers, and this is what happens with booleans:

* `'true' IN (TRUE, 'false')` -> false
* `TRUE IN (TRUE, 'false')` -> true
* `TRUE IN ('true', 'false')` -> false

So it seems to me it coerces only integers and strings in the in operator...

My suggestion is that, when comparing, we just coerce to the left argument type. I think it's simpler to understand for users, it's predictable and it resembles the way other SQLs work.

slinkydeveloper:
I've changed this PR accordingly

slinkydeveloper:
These are the tests I've developed in the implementation, they should give you the idea: https://github.com/cloudevents/sdk-java/pull/363/files#diff-200757bd9f261f88b374ca648b8d98ab420184561e5f7e6efea2161baf996aeaR77-R108

duglin:
Approved on the 4/8 call

PR #799, opened 2021-03-31
docs: Expand versioning suggestions

jskeet:
Addresses #778

### DIFF
--- a/primer.md
+++ b/primer.md
@@ -18,7 +18,7 @@ This document is a working draft.
 - [CloudEvents Concepts](#cloudevents-concepts)
 - [Design Goals](#design-goals)
 - [Architecture](#architecture)
-- [Versioning of Attributes](#versioning-of-attributes)
+- [Versioning of CloudEvents](#versioning-of-cloudevents)
 - [CloudEvent Attributes](#cloudevent-attributes)
 - [CloudEvent Attribute Extensions](#cloudevent-attribute-extensions)
 - [Creating CloudEvents](#creating-cloudevents)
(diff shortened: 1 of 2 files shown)


jskeet:
Note: I'm definitely *not* expecting this to just be approved and merged. I'm very happy to take either specific suggestions, or a more general change in direction ("give concrete examples" or "don't ramble on for so long") to work on. I just wanted to make *some* kind of progress :)

primer.md
```diff
@@ -238,23 +238,54 @@ there are errors during the processing of a CloudEvent, the software
...
-event producer. However, when a version-specific string is included, care should
-be taken whenever its value changes as event consumers might be reliant on the
-existing value and thus a change could be interpreted as a "breaking change".
-Some form of communication between producers and consumers should be established
-to ensure the event consumers know what possible values might be used. In
-general, this is true for all CloudEvents attributes as well.
+## Versioning of CloudEvents
+
+For many CloudEvents, the schema of the data within the event might
+change over time. The CloudEvents specification does mandate any
```
malevy, line 244: did you mean to say "...does not mandate..."?

jskeet: Absolutely - well spotted. Not sure how I fouled up that C&P. Will fix and repush.

primer.md
```diff
@@ -238,23 +238,54 @@ there are errors during the processing of a CloudEvent, the software
...
+"backwardly-compatible" will vary by data content type.
+
+The `dataschema` attribute is expected to be informational, largely
+to be used during development and by tooling that is able to provide
+diagnostic information over arbitrary CloudEvents with a data
+content type understood by that tooling.
+
+When a CloudEvent's data changes in a backwardly-compatible way, the
+value of the `type` attribute should generally stay the same. The
+value of the `dataschema` attribute may change, or it may stay the
```
erikerikson, line 269: I would suggest `s/may change, or/should change, though/`.  The implication otherwise is that assistive tooling (e.g. IDE plugins) or event consumers might have to poll for schema content changes rather than be able to depend on the unique resource identifier that a URI can provide.  I appreciated we are not forcing implementer's hands but the use of semver minor and patch version designators in URIs seems like a good best practice to encourage.  As stated, the mays seem equivalent and the implication I take away (when placing a probably undue level of attention here) is that we don't have an opinion.  At least **I** have a strong preference for the former, I can only speak for myself.

primer.md
```diff
@@ -238,23 +238,54 @@ there are errors during the processing of a CloudEvent, the software
...
+same but the content served from the URI can change to reflect the
+updated schema.
+
+When a CloudEvent's data changes in a backwardly-incompatible way,
+the value of the `type` attribute should generally change. The event
+producer is encouraged to produce both the old event and the new
+event for some time (potentially forever) in order to avoid
+disrupting consumers.
```
erikerikson, line 277: nit: this paragraph of content seems to belong in the paragraph starting at 253 where the first half discussing changes both breaking and non-breaking was given.  Putting them together may increase readability.

primer.md
```diff
@@ -238,23 +238,54 @@ there are errors during the processing of a CloudEvent, the software
...
+
+When considering the value of the `type` attribute, event producers
+may choose any versioning scheme they wish, such as a version number
+(`v1`, `v2`) or a date (`2018-01-01`) as part of the value. They may
+also use the `type` attribute to convey that a particular version is
+not yet stable, and may take breaking change.
```
erikerikson, line 283: nit: this too seems better grouped with the section addressing type.  I suspect you're trying to get into the meat rather than hide it behind too much detail.  I might suggest that the top level description paragraph ending on line 251 could benefit from the content starting at 267 and having that stated up front would allow for a slightly deeper indulgence in the full details of best practice for both type and dataschema separately.

primer.md
```diff
@@ -238,23 +238,54 @@ there are errors during the processing of a CloudEvent, the software
...
+
+Event producers are encouraged to consider versioning from the
+outset, before first declaring a particular CloudEvent to be
+"stable". Documentation of the chosen versioning scheme, including
+the rationale behind it, is likely to be appreciated by consumers.
```
erikerikson, line 288: ❤️

erikerikson (approved):
Thanks for these improvements.  This guidance is better than what it replaces.

I offer mostly nits and munging (sorry but hope they can help comprhension) but also one point on 269 suggesting a stronger declaration of best practice and personal preference.

primer.md
jskeet: I've moved things around quite a bit now... as you suggested, it was a matter of "top-level, then more detail" but it's now reorganized to treat each attribute separately. That's also involved moving the general encouragement to earlier on, which I'm generally pleased about. See what you think.

primer.md
jskeet: I'm more on the fence - partly aware that I'm pretty sure our `dataschema` attributes *aren't* versioned that way.

I've made it "should" but acknowledged that the alternative is simpler... along with the reasons why it's more consumer-friendly to version.

I deliberately *haven't* called out semver minor changes here, partly as I can imagine one simple schema for `dataschema` attributes being a link to a file on GitHub, including the commit. That still has the "cache-busting" property, but doesn't require every schema change to be considered as "minor or patch" based on whether it's just a refactoring/comment change, and doesn't require tagging etc.

As ever, feel free to push back - and thanks for the input.

erikerikson: Should feels like a win and as much as I can reasonably ask.  Thank you.  I think there's a lot of room for a "latest"/"evergreen" link that one would find and use through the discover service/portal but have thought that a strictly versioned dataschema would be best for each event, if one is even sending the attribute.  I have to admit not having thought of using commits in the URI but I now in hindsight it seems indefensible.  :D

erikerikson:
Thank you for the updates, it feels a lot easier to absorb now.  LGTM

primer.md
```diff
@@ -238,23 +238,69 @@ there are errors during the processing of a CloudEvent, the software
...
+value of the `type` attribute should generally stay the same.
+
+When a CloudEvent's data changes in a backwardly-incompatible way,
+the value of the `type` attribute should generally change. The event
+producer is encouraged to produce both the old event and the new
+event for some time (potentially forever) in order to avoid
+disrupting consumers.
+
+When considering the value of the `type` attribute, event producers
+may choose any versioning scheme they wish, such as a version number
```
duglin, line 281: overall this seems ok, but it does drop the idea of no versioning scheme at all for "type". Was that intentional? The use of the phrase "may choose any" kind of implies they need to choose one. The old text implied it was optional.

jskeet: No, that wasn't intentional - although I do want to encourage it.

How would this be:

> When considering the value of the `type` attribute, event producers
> may choose any versioning scheme they wish, such as a version number
> (`v1`, `v2`) or a date (`2018-01-01`) as part of the value. Alternatively, they
> may choose not to include a version in the `type` value at all.

duglin: sure that works - thanks!

jskeet: Done in an extra commit. Thanks for the careful review :)

jskeet:
Now squashed into a single commit. I *suspect* (contrary to my expectations yesterday) that this can be merged as-is rather than waiting for #810 (the master branch equivalent) to be merged.

duglin:
Approved on the 4/29 call

PR #800, opened 2021-03-31
Introducing XML Format.

### LINKED ISSUE #662: XML Format
Hello.

I would enjoy having an XML Format for Cloud Events. I have some existing XML processing systems, and trying to integrate Cloud Events messages from other systems with these XML systems relies on some informal JSON->XML glue. It would be lovely to have a real specification for what Cloud Events look like in XML.

JemDay:
feat: Introduce XML Format

Adds support for XML as a format for CloudEvents (including batch format), XML schema provided along with textual specification.

Closes #662 

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ The following documents are available ([Release Notes](misc/RELEASE_NOTES.md)):
 | NATS Protocol Binding         | [v1.0.2](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/bindings/nats-protocol-binding.md)  |    [WIP](https://github.com/cloudevents/spec/blob/main/cloudevents/bindings/nats-protocol-binding.md)    |
 | WebSockets Protocol Binding   |                                        -                                        | [WIP](https://github.com/cloudevents/spec/blob/main/cloudevents/bindings/websockets-protocol-binding.md) |
 | Protobuf Event Format         |                                                                                 | [v1.0-rc1](https://github.com/cloudevents/spec/blob/main/cloudevents/formats/protobuf-format.md)                                  |
-| XML Event Format              | - | [WIP](https://github.com/cloudevents/spec/blob/main/cloudevents/working-drafts/xml-foramt.md) |
+| XML Event Format              | - | [WIP](https://github.com/cloudevents/spec/blob/main/cloudevents/working-drafts/xml-format.md) |
 | Web hook                      |      [v1.0.2](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/http-webhook.md)      |        [WIP](https://github.com/cloudevents/spec/blob/main/cloudevents/http-webhook.md)         |
 |                               |
 | **Additional Documentation:** |
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ The following documents are available:
 | NATS Protocol Binding         | [v1.0.1](https://github.com/cloudevents/spec/blob/v1.0.1/nats-protocol-binding.md)  |    [WIP](https://github.com/cloudevents/spec/blob/main/cloudevents/bindings/nats-protocol-binding.md)    |
 | WebSockets Protocol Binding   |                                        -                                        | [WIP](https://github.com/cloudevents/spec/blob/main/cloudevents/bindings/websockets-protocol-binding.md) |
 | Protobuf Event Format         |                                                                                 | [v1.0-rc1](https://github.com/cloudevents/spec/blob/main/cloudevents/formats/protobuf-format.md) |
-| XML Event Format              | - | [WIP][xml-format-working |]
+| XML Event Format              | - | [WIP][xml-format-working] |
 | Web hook                      |      [v1.0.1](https://github.com/cloudevents/spec/blob/v1.0.1/http-webhook.md)      |        [WIP](https://github.com/cloudevents/spec/blob/main/cloudevents/http-webhook.md)         |
 |                               |
 | **Additional Documentation:** |
--- a/README.md
+++ b/README.md
@@ -40,7 +40,8 @@ The following documents are available:
 | MQTT Protocol Binding         | [v1.0.1](https://github.com/cloudevents/spec/blob/v1.0.1/mqtt-protocol-binding.md)  |    [WIP](https://github.com/cloudevents/spec/blob/main/cloudevents/bindings/mqtt-protocol-binding.md)    |
 | NATS Protocol Binding         | [v1.0.1](https://github.com/cloudevents/spec/blob/v1.0.1/nats-protocol-binding.md)  |    [WIP](https://github.com/cloudevents/spec/blob/main/cloudevents/bindings/nats-protocol-binding.md)    |
 | WebSockets Protocol Binding   |                                        -                                        | [WIP](https://github.com/cloudevents/spec/blob/main/cloudevents/bindings/websockets-protocol-binding.md) |
-| Protobuf Event Format         |                                                                                 | [v1.0-rc1](https://github.com/cloudevents/spec/blob/main/cloudevents/formats/protobuf-format.md)                                  |
+| Protobuf Event Format         |                                                                                 | [v1.0-rc1](https://github.com/cloudevents/spec/blob/main/cloudevents/formats/protobuf-format.md) |
+| XML Event Format              | - | [WIP][xml-format-working |]
 | Web hook                      |      [v1.0.1](https://github.com/cloudevents/spec/blob/v1.0.1/http-webhook.md)      |        [WIP](https://github.com/cloudevents/spec/blob/main/cloudevents/http-webhook.md)         |
 |                               |
 | **Additional Documentation:** |
@@ -141,3 +142,6 @@ Periodically, the group may have in-person meetings that coincide with a major
 conference. Please see the
 [meeting minutes doc](https://docs.google.com/document/d/1OVF68rpuPK5shIHILK9JOqlZBbfe91RNzQ7u_P7YCDE/edit#)
 for any future plans.
+
+[xml-format-working]: https://github.com/cloudevents/spec/blob/main/cloudevents/formats/xml-format.md
+[xml-format-latest]: https://github.com/cloudevents/spec/blob/main/cloudevents/formats/xml-format.md
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ The following documents are available ([Release Notes](misc/RELEASE_NOTES.md)):
 | NATS Protocol Binding         | [v1.0.2](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/bindings/nats-protocol-binding.md)  |    [WIP](https://github.com/cloudevents/spec/blob/main/cloudevents/bindings/nats-protocol-binding.md)    |
 | WebSockets Protocol Binding   |                                        -                                        | [WIP](https://github.com/cloudevents/spec/blob/main/cloudevents/bindings/websockets-protocol-binding.md) |
 | Protobuf Event Format         |                                                                                 | [v1.0-rc1](https://github.com/cloudevents/spec/blob/main/cloudevents/formats/protobuf-format.md)                                  |
+| XML Event Format              | - | [WIP](https://github.com/cloudevents/spec/blob/main/cloudevents/working-drafts/xml-foramt.md) |
 | Web hook                      |      [v1.0.2](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/http-webhook.md)      |        [WIP](https://github.com/cloudevents/spec/blob/main/cloudevents/http-webhook.md)         |
 |                               |
 | **Additional Documentation:** |
--- a/cloudevents/working-drafts/xml-format.md
+++ b/cloudevents/working-drafts/xml-format.md
@@ -31,11 +31,11 @@ data type mappings for CloudEvents context attributes.
 * The [Data](#3-data) section defines the container for the data portion of a
 CloudEvent.
 
-* The [Envelope](#4-envelope) section defines an XML element for CloudEvents
-attributes and an associated media type.
+* The [Envelope](#4-envelope) section defines an XML element to contain CloudEvent
+context attributes and data.
 
 * The [Batch](#5-xml-batch-format) section describes how multiple CloudEvents
-can be packaged into a single XML document.
+can be packaged into a single XML element.
 
 ### 1.1. Conformance
 
--- a/cloudevents/working-drafts/xml-format.md
+++ b/cloudevents/working-drafts/xml-format.md
@@ -97,8 +97,9 @@ discriminator SHOULD be interpreted with an implied type of `xs:string`
 ## 3. Data
 
 The `Data` portion of a CloudEvent follows a similar model to that employed by
-the [JSON Format specification][json-format]. The element names are capitalized
-to avoid collision with curent or future CloudEvent attribute names.
+the [JSON Format specification][json-format]. A `data` element is used to
+encapsulate the payload and an `xsi:type` used to discrimate the payload
+type.
 
 ### 3.1 Binary Data
 
--- a/cloudevents/working-drafts/xml-format.md
+++ b/cloudevents/working-drafts/xml-format.md
@@ -20,7 +20,7 @@ This document is a working draft.
 
 ## 1. Introduction
 
-[CloudEvents][ce] is a standardized and protocol-agnostic definition of the
+[CloudEvents][ce-spec] is a standardized and protocol-agnostic definition of the
 structure and metadata description of events. This specification defines how
 elements defined in the CloudEvents specification are to be represented using
 [Extensible Markup Language (XML)](xml-spec) documents.
@@ -35,7 +35,7 @@ CloudEvent.
 attributes and an associated media type.
 
 * The [Batch](#5-xml-batch-format) section describes how multiple CloudEvents
-may be packaged into a single XML document.
+can be packaged into a single XML document.
 
 ### 1.1. Conformance
 
@@ -58,8 +58,8 @@ structure.
 
 The CloudEvents type system is mapped to the XML schema types as follows :
 
-| CloudEvents   |  XML Schema Type| Notes |
-| :-------------| :---------------------------------------------------------- |
+| CloudEvents   |  XML Schema Type | Notes |
+| :-------------| :--------------- | :---- |
 | Boolean       | [xs:boolean][xml-primitives] | |
 | Integer       | [xs:int][xml-primitives] | |
 | String        | [xs:string][xml-primitives] | |
@@ -68,15 +68,15 @@ The CloudEvents type system is mapped to the XML schema types as follows :
 | URI-reference | [xs:anyURI][xml-primitives] following [RFC 3986][rfc3986] | |
 | Timestamp     | [xs:dateTime][xml-primitives] following [RFC 3339][rfc3339] | |
 
-Each context attribute is represented as an XML element whose name exactly matches that of
-a required or optional CloudEvent attribute, extension attribute names `MUST` adhere and
-align with the conventions in defined in the [cloud event specification][ce-spec].
+Each context attribute is represented as an XML element whose name MUST exactly match
+that of a REQUIRED or OPTIONAL CloudEvent attribute, extension attribute names MUST
+adhere to, and align with, the conventions in defined in the [cloud event specification][ce-spec].
 
 See the [envelope](#4-envelope) for special handing of the `specversion` context attribute.
 
-Extension attributes `SHOULD` be expressed with the appropriate `xsi:type` to allow them
+Extension attributes SHOULD be expressed with the appropriate `xsi:type` to allow them
 to be exchanged without loss of type information. Extension attributes with no `xsi:type`
-discriminator `SHOULD` be interpreted with an implied type of `xs:string`
+discriminator SHOULD be interpreted with an implied type of `xs:string`
 
 ``` xml
     ...
@@ -91,12 +91,12 @@ discriminator `SHOULD` be interpreted with an implied type of `xs:string`
 ## 3. Data
 
 The `Data` portion of a CloudEvent follows a similar model to that employed by
-the [JSON Format specification][json-format]. The elmenent names used are capitalized to avoid
-collision with CloudEvent attribute names.
+the [JSON Format specification][json-format]. The element names MUST be capitalized to avoid
+collision with curent or future CloudEvent attribute names.
 
 ### 3.1 Binary Data
 
-Carried in an element with an defined type of `xs:base64Binary`
+MUST be carried in an element with an defined type of `xs:base64Binary`
 
 ``` xml
 <Data xsi:type="xs:base64Binary">.........</Data>
@@ -104,7 +104,7 @@ Carried in an element with an defined type of `xs:base64Binary`
 
 ### 3.2 Text Data
 
-Carried in an element with an defined type of `xs:string`
+MUST be carried in an element with an defined type of `xs:string`
 
 ``` xml
 <Data xsi:type="xs:string">This is text</Data>
@@ -112,7 +112,7 @@ Carried in an element with an defined type of `xs:string`
 
 ### 3.3 XML Data
 
-XML data is carried in an explicit element:
+XML data MUST be carried in an explicit element `<DataXml>`:
 
 ``` xml
 <DataXml>
@@ -125,7 +125,7 @@ XML data is carried in an explicit element:
 ## 4. Envelope
 
 Each CloudEvent is wholly represented as an XML element, the root xml
-document element is `<CloudEvent>`. This element carries the `speversion`
+document element MUST be `<CloudEvent>`. This element carries the `specversion`
 as an XML attribute value.
 
 Such a representation MUST use the media type `application/cloudevents+xml`.
@@ -145,7 +145,7 @@ eg (Namespace definitions omitted for brevity):
     <source>urn:uuid:123e4567-e89b-12d3-a456-426614174000</source>
     <type>SOME.EVENT.TYPE</type>
     <myboolean xsi:type="xs:boolean">false</myboolean>
-    <Data xsi:type="xs_string">Now is the winter of our discount tents...</Data
+    <Data xsi:type="xs_string">Now is the winter of our discount tents...</Data>
 </CloudEvent>
 ```
 
@@ -166,7 +166,7 @@ Example:
 
 ``` xml
 <CloudEventBatch>
-    <CloudEvent specverison="1.0">
+    <CloudEvent specversion="1.0">
        ....
     </CloudEvent>
     <CloudEvent specversion="1.0">
--- a/cloudevents/working-drafts/xml-format.md
+++ b/cloudevents/working-drafts/xml-format.md
@@ -63,25 +63,27 @@ An XML document preamble SHOULD be included to ensure deterministic processing.
 
 ## 2. Attributes
 
-The CloudEvents type system is mapped to the XML schema types as follows :
+The CloudEvents type system is mapped to a set of CloudEvent specific
+type designators as follows :
 
-| CloudEvents   |  XML Schema Type | Notes |
-| :-------------| :--------------- | :---- |
-| Boolean       | [xs:boolean][xml-primitives] | |
-| Integer       | [xs:int][xml-primitives] | |
-| String        | [xs:string][xml-primitives] | |
-| Binary        | [xs:base64Binary][xml-primitives] | |
-| URI           | [xs:anyURI][xml-primitives] following [RFC 3986][rfc3986]| |
-| URI-reference | [xs:anyURI][xml-primitives] following [RFC 3986][rfc3986] | |
-| Timestamp     | [xs:dateTime][xml-primitives] following [RFC 3339][rfc3339] | |
+| CloudEvents Type  |  XML Format Type | XML Schema Type | Notes |
+| :-----------------| :----------------| :--------------- | :---- |
+| Boolean       | ce:boolean | [xs:boolean][xml-primitives] | `true` or `false` |
+| Integer       | ce:int | [xs:int][xml-primitives] | |
+| String        | ce:string | [xs:string][xml-primitives] | |
+| Binary        | ce:binary | [xs:base64Binary][xml-primitives] |  |
+| URI           | ce:uri | [xs:anyURI][xml-primitives] following [RFC 3986][rfc3986]| |
+| URI-reference | ce:uriRef | [xs:anyURI][xml-primitives] following [RFC 3986][rfc3986] | |
+| Timestamp     | ce:timestamp | [xs:dateTime][xml-primitives] following [RFC 3339][rfc3339] | |
 
 Each CloudEvent context attribute is represented as an XML element whose local
 name exactly matches that of the attribute.
 
 See the [envelope](#4-envelope) for special handing of the `specversion` context attribute.
 
-Extension attributes MUST be decorated with the appropriate `xsi:type` to allow them
-to be exchanged without loss of type information.
+Extension attributes MUST be decorated with the appropriate CloudEvent format
+designators using using the `xsi:type` XML attribute, this allows them to be exchanged
+without loss of type information.
 
 REQUIRED and OPTIONAL context attributes SHOULD NOT be decorated with an `xsi:type`.
 
@@ -94,8 +96,8 @@ processing.
     ..
     <time>2021-08-14T14:30:22-08:00</time>
     ..
-    <myextension xsi:type="xs:string">my extension value</myextension>
-    <myboolean xsi:type="xs:boolean">false</myboolean>
+    <myextension xsi:type="ce:string">my extension value</myextension>
+    <myboolean xsi:type="ce:boolean">false</myboolean>
 ```
 
 ## 3. Data
@@ -110,22 +112,22 @@ The following data representations are supported:
 
 ### 3.1 Binary Data
 
-Binary data MUST be carried in an element with an defined type of `xs:base64Binary`.
+Binary data MUST be carried in an element with an defined type of `ce:binary`.
 
 Example:
 
 ``` xml
-<data xsi:type="xs:base64Binary">.........</data>
+<data xsi:type="ce:binary">.........</data>
 ```
 
 ### 3.2 Text Data
 
-Text MUST be carried in an element with an defined type of `xs:string`.
+Text MUST be carried in an element with an defined type of `ce:string`.
 
 Example:
 
 ``` xml
-<data xsi:type="xs:string">This is text</data>
+<data xsi:type="ce:string">This is text</data>
 ```
 
 ### 3.3 XML Data
@@ -167,8 +169,8 @@ Example _(XML preamble and namespace definitions omitted for brevity)_:
     <id>000-1111-2222</id>
     <source>urn:uuid:123e4567-e89b-12d3-a456-426614174000</source>
     <type>SOME.EVENT.TYPE</type>
-    <myboolean xsi:type="xs:boolean">false</myboolean>
-    <data xsi:type="xs:string">Now is the winter of our discount tents...</data>
+    <myboolean xsi:type="ce:boolean">false</myboolean>
+    <data xsi:type="ce:string">Now is the winter of our discount tents...</data>
 </event>
 ```
 
@@ -233,7 +235,7 @@ An example of an empty batch of CloudEvents (typically used in a response, but a
     <id>000-1111-2222</id>
     <source>urn:uuid:123e4567-e89b-12d3-a456-426614174000</source>
     <type>SOME.EVENT.TYPE</type>
-    <data xsi:type="xs:string">{ "salutation": "Good Morning", "text": "hello world" }</data>
+    <data xsi:type="ce:string">{ "salutation": "Good Morning", "text": "hello world" }</data>
 </event>
 ```
 
--- a/cloudevents/working-drafts/xml-format.md
+++ b/cloudevents/working-drafts/xml-format.md
@@ -54,7 +54,11 @@ custom extensions to be communicated without type loss.
 A schema-less approach has been taken favoring convention over rigid document
 structure.
 
-The namespace `http://cloudevents.io/xmlformat/v1` SHOULD be used.
+The namespace `http://cloudevents.io/xmlformat/v1` SHOULD be used, when
+namespace prefixes are used a prefix of `ce` is preferred but MUST NOT be expected
+from an XML document processing perspective.
+
+An XML document preamble SHOULD be included to ensure deterministic processing.
 
 ## 2. Attributes
 
@@ -128,8 +132,8 @@ to be represented.
 
 ## 4. Envelope
 
-Each CloudEvent is wholly represented as an XML element `<CloudEvent>` that
-carries the `specversion`as an XML attribute value.
+Each CloudEvent is wholly represented as an XML element `<Event>` that
+carries the `specversion` as an XML attribute value.
 
 Such a representation MUST use the media type `application/cloudevents+xml`.
 
@@ -138,10 +142,10 @@ The enveloping element contains:
 * A set of context attribute elements.
 * An optional `data` element.
 
-eg (Namespace definitions omitted for brevity):
+eg _(XML preamble and namespace definitions omitted for brevity)_:
 
 ``` xml
-<CloudEvent specversion="1.0">
+<Event specversion="1.0">
     <time>2020-03-19T12:54:00-07:00</time>
     <datacontenttype>text/plain</datacontenttype>
     <id>000-1111-2222</id>
@@ -149,7 +153,7 @@ eg (Namespace definitions omitted for brevity):
     <type>SOME.EVENT.TYPE</type>
     <myboolean xsi:type="xs:boolean">false</myboolean>
     <data xsi:type="xs:string">Now is the winter of our discount tents...</data>
-</CloudEvent>
+</Event>
 ```
 
 ## 5. XML Batch Format
@@ -165,24 +169,24 @@ support for the _XML Format_ is indicated.
 An XML Batch of CloudEvents MUST use the media type
 `application/cloudevents-batch+xml`.
 
-Example:
+Example _(XML preamble and namespace definitions omitted for brevity)_:
 
 ``` xml
-<CloudEventBatch>
-    <CloudEvent specversion="1.0">
+<Batch>
+    <Event specversion="1.0">
        ....
-    </CloudEvent>
-    <CloudEvent specversion="1.0">
+    </Event>
+    <Event specversion="1.0">
        ....
-    </CloudEvent>
+    </Event>
     ....
-</CloudEventBatch>
+</Batch>
 ```
 
 An example of an empty batch of CloudEvents (typically used in a response, but also valid in a request):
 
 ```xml
-<CloudEventBatch />
+<Batch />
 ```
 
 ## 6. Examples
@@ -191,47 +195,73 @@ An example of an empty batch of CloudEvents (typically used in a response, but a
 
 ```xml
 <?xml version="1.0" encoding="UTF-8"?>
-<CloudEvent xmlns="http://cloudevents.io/xmlformat/V1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xs="http://www.w3.org/2001/XMLSchema" specversion="1.0" >
+<Event xmlns="http://cloudevents.io/xmlformat/V1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
+       xmlns:xs="http://www.w3.org/2001/XMLSchema" specversion="1.0" >
     <time>2020-03-19T12:54:00-07:00</time>
     <datacontenttype>image/png</datacontenttype>
     <id>000-1111-2222</id>
     <source>urn:uuid:123e4567-e89b-12d3-a456-426614174000</source>
     <type>SOME.EVENT.TYPE</type>
     <data xsi:type="xs:base64Binary">... Base64 encoded data...</data>
-</CloudEvent>
+</Event>
 ```
 
 ### 6.2 CloudEvent with JSON event data
 
 ```xml
 <?xml version="1.0" encoding="UTF-8"?>
-<CloudEvent xmlns="http://cloudevents.io/xmlformat/V1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xs="http://www.w3.org/2001/XMLSchema" specversion="1.0" >
+<Event xmlns="http://cloudevents.io/xmlformat/V1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
+       xmlns:xs="http://www.w3.org/2001/XMLSchema" specversion="1.0" >
     <time>2020-03-19T12:54:00-07:00</time>
     <datacontenttype>application/json</datacontenttype>
     <id>000-1111-2222</id>
     <source>urn:uuid:123e4567-e89b-12d3-a456-426614174000</source>
     <type>SOME.EVENT.TYPE</type>
     <data xsi:type="xs:string">{ "salutation": "Good Morning", "text": "hello world" }</data>
-</CloudEvent>
+</Event>
 ```
 
 ### 6.3 CloudEvent with XML event data
 
+#### 6.3.1 Locally namespaced event data
+
 ```xml
 <?xml version="1.0" encoding="UTF-8"?>
-<CloudEvent xmlns="http://cloudevents.io/xmlformat/V1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xs="http://www.w3.org/2001/XMLSchema" specversion="1.0" >
+<Event xmlns="http://cloudevents.io/xmlformat/V1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
+       xmlns:xs="http://www.w3.org/2001/XMLSchema" specversion="1.0" >
     <time>2020-03-19T12:54:00-07:00</time>
     <datacontenttype>application/xml</datacontenttype>
     <id>000-1111-2222</id>
     <source>urn:uuid:123e4567-e89b-12d3-a456-426614174000</source>
     <type>SOME.EVENT.TYPE</type>
     <data xsi:type="xs:any">
-        <Location>
-            <Latitude>51.509865</Latitude>
-            <Longitude>-0.118092</Longitude>
-        </Location>
+        <geo:Location xmlns:geo="http://someauthority.example/">
+            <geo:Latitude>51.509865</geo:Latitude>
+            <geo:Longitude>-0.118092</geo:Longitude>
+        </geo:Location>
     </data>
-</CloudEvent>
+</Event>
+```
+
+#### 6.3.2 Explicit namespacing
+
+```xml
+<?xml version="1.0" encoding="UTF-8"?>
+<ce:Event xmlns:ce="http://cloudevents.io/xmlformat/V1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
+          xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:geo="http://someauthority.example/"
+          specversion="1.0" >
+    <ce:time>2020-03-19T12:54:00-07:00</ce:time>
+    <ce:datacontenttype>application/xml</ce:datacontenttype>
+    <ce:id>000-1111-2222</ce:id>
+    <ce:source>urn:uuid:123e4567-e89b-12d3-a456-426614174000</ce:source>
+    <ce:type>SOME.EVENT.TYPE</ce:type>
+    <ce:data xsi:type="xs:any">
+        <geo:Location>
+            <geo:Latitude>51.509865</geo:Latitude>
+            <geo:Longitude>-0.118092</geo:Longitude>
+        </geo:Location>
+    </ce:data>
+</ce:Event>
 ```
 
 [ce-spec]: ../spec.md
--- a/cloudevents/working-drafts/xml-format.md
+++ b/cloudevents/working-drafts/xml-format.md
@@ -45,7 +45,7 @@ interpreted as described in [RFC2119][rfc2119].
 
 ### 1.2 Approach
 
-This XML representation used is deliberately verbose in favor of readability
+The XML representation used is deliberately verbose in favor of readability
 and deterministic processing.
 
 Preservation of type information for CloudEvent attributes is supported allowing
@@ -130,15 +130,17 @@ Example:
 
 ### 3.3 XML Data
 
-XML data MUST be carried in an element with a defined type of `xs:any`
-allowing a single child XML element (with any required namespace definitions)
-to be represented.
+XML data MUST be carried in an element with a defined type of `xs:any` with
+a single child XML element (with any required namespace definitions).
+
+The child business data XML elements MUST NOT reside in the CloudEvent namespace,
+an empty namespace MAY be used as appropriate.
 
 Example:
 
 ``` xml
 <data xsi:type="xs:any">
-    <myData>
+    <myData xmlns="http://my.org/namespace">
         ....
     </myData>  
 </data>
@@ -185,7 +187,7 @@ An XML Batch of CloudEvents MUST use the media type
 
 Example _(XML preamble and namespace definitions omitted for brevity)_:
 
-``` xml
+```xml
 <batch>
     <event specversion="1.0">
        ....
@@ -278,9 +280,81 @@ An example of an empty batch of CloudEvents (typically used in a response, but a
 </ce:event>
 ```
 
+#### 6.3.3 ISO 20022 Usage Example
+
+[ISO 20022][iso-20022] is an XML based messaging standard used in the financial industry; this is a
+non-normative example.
+
+```xml
+<?xml version="1.0" encoding="UTF-8"?>
+<event xmlns="http://cloudevents.io/xmlformat/V1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
+       xmlns:xs="http://www.w3.org/2001/XMLSchema" specversion="1.0" >
+    <time>2022-02-22T15:12:00-08:00</time>
+    <datacontenttype>application/xml</datacontenttype>
+    <id>000-1111-2222</id>
+    <source>urn:uuid:123e4567-e89b-12d3-a456-426614174000</source>
+    <type>com.mybank.pain.001.001.03</type>
+    <data xsi:type="xs:any">
+        <Document xmlns="urn:iso:std:iso:20022:tech:xsd:pain.001.001.03">
+            <CstmrCdtTrfInitn>
+            <GrpHdr>
+                <MsgId>ABC/060928/CCT001</MsgId>
+                <CreDtTm>2022-02-22T14:07:00</CreDtTm>
+                <NbOfTxs>3</NbOfTxs>
+                <CtrlSum>2400.56</CtrlSum>
+                <InitgPty>
+                    <Nm>Cobelfac</Nm>
+                    <Id>
+                        <OrgId>
+                            <Othr>
+                                <Id>0468651441</Id>
+                                <Issr>KBO-BCE</Issr>
+                            </Othr>
+                        </OrgId>
+                    </Id>
+                </InitgPty>
+            </GrpHdr>
+            <PmtInf>
+                <PmtInfId> ABC/4560/2008-09-25</PmtInfId>
+                <PmtMtd>TRF</PmtMtd>
+                <BtchBookg>false</BtchBookg>
+
+            <!-- Content ommited for brevity -->
+
+            </PmtInf>
+        </Document>
+    </data>
+</event>
+```
+
+#### 6.3.4 Batch Example
+
+```xml
+<?xml version="1.0" encoding="UTF-8"?>
+<batch xmlns="http://cloudevents.io/xmlformat/V1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
+       xmlns:xs="http://www.w3.org/2001/XMLSchema">
+    <event specversion="1.0" >
+        <time>2020-03-19T12:54:00-07:00</time>
+        <datacontenttype>image/png</datacontenttype>
+        <id>000-1111-2222</id>
+        <source>urn:uuid:123e4567-e89b-12d3-a456-426614174000</source>
+        <type>SOME.EVENT.TYPE</type>
+        <data xsi:type="xs:base64Binary">... Base64 encoded data...</data>
+    </event>
+    <event specversion="1.0" >
+        <time>2020-03-19T12:59:00-07:00</time>
+        <datacontenttype>image/png</datacontenttype>
+        <id>000-1111-3333</id>
+        <source>urn:uuid:123e4567-e89b-12d3-a456-426614174000</source>
+        <type>SOME.EVENT.TYPE</type>
+        <data xsi:type="xs:base64Binary">... Base64 encoded data...</data>
+    </event>
+    .....
+</batch>
+```
+
 [ce-spec]: ../spec.md
 [ce-types]: ../spec.md#type-system
-[xml-schema]: ./cloudevents.xsd
 [xml-format]: ./xml-format.md
 [json-format]: ../formats/json-format.md
 [xml-spec]: https://www.w3.org/TR/2008/REC-xml-20081126/
@@ -288,3 +362,4 @@ An example of an empty batch of CloudEvents (typically used in a response, but a
 [rfc2119]: https://tools.ietf.org/html/rfc2119
 [rfc3339]: https://www.ietf.org/rfc/rfc3339.txt
 [rfc3986]: https://tools.ietf.org/html/rfc3986
+[iso-20022]: https://www.iso20022.org
--- a/cloudevents/working-drafts/xml-format.md
+++ b/cloudevents/working-drafts/xml-format.md
@@ -81,13 +81,12 @@ name exactly matches that of the attribute.
 
 See the [envelope](#4-envelope) for special handing of the `specversion` context attribute.
 
-Extension attributes MUST be decorated with the appropriate CloudEvent format
-designators using the `xsi:type` XML attribute, this allows them to be exchanged
+Extension context attributes MUST be decorated with the appropriate CloudEvent type format
+designators using an `xsi:type` XML attribute, this allows them to be exchanged
 without loss of type information.
 
-REQUIRED and OPTIONAL context attributes SHOULD NOT be decorated with an `xsi:type` to
-avoid potential inconsistencies. Implementors MUST process these values according to their
-types as defined by the [CloudEvent type specification][ce-types].
+REQUIRED and OPTIONAL context atrtribute MAY be decorated with `xsi:type`, if present this
+value MUST match that of the type specified by the [CloudEvent type specification][ce-types].
 
 No other XML element attributes are expected, if present they MUST be ignored during
 processing.
@@ -114,22 +113,23 @@ The following data representations are supported:
 
 ### 3.1 Binary Data
 
-Binary data MUST be carried in an element with an defined type of `ce:binary`.
+Binary data MUST be carried in an element with an defined type of `xs:base64Binary`
+and encoded appropriately..
 
 Example:
 
 ``` xml
-<data xsi:type="ce:binary">.........</data>
+<data xsi:type="xs:base64Binary">.........</data>
 ```
 
 ### 3.2 Text Data
 
-Text MUST be carried in an element with an defined type of `ce:string`.
+Text MUST be carried in an element with an defined type of `xs:string`.
 
 Example:
 
 ``` xml
-<data xsi:type="ce:string">This is text</data>
+<data xsi:type="xs:string">This is text</data>
 ```
 
 ### 3.3 XML Data
@@ -169,7 +169,7 @@ Example _(XML preamble and namespace definitions omitted for brevity)_:
     <source>urn:uuid:123e4567-e89b-12d3-a456-426614174000</source>
     <type>SOME.EVENT.TYPE</type>
     <myboolean xsi:type="ce:boolean">false</myboolean>
-    <data xsi:type="ce:string">Now is the winter of our discount tents...</data>
+    <data xsi:type="xs:string">Now is the winter of our discount tents...</data>
 </event>
 ```
 
@@ -228,7 +228,7 @@ Example _(XML preamble and namespace definitions omitted for brevity)_:
     <id>000-1111-2222</id>
     <source>urn:uuid:123e4567-e89b-12d3-a456-426614174000</source>
     <type>SOME.EVENT.TYPE</type>
-    <data xsi:type="ce:string">{ "salutation": "Good Morning", "text": "hello world" }</data>
+    <data xsi:type="xs:string">{ "salutation": "Good Morning", "text": "hello world" }</data>
 </event>
 ```
 
--- a/cloudevents/working-drafts/xml-format.md
+++ b/cloudevents/working-drafts/xml-format.md
@@ -69,23 +69,25 @@ type designators as follows :
 | CloudEvents Type  |  XML Format Type | XML Schema Type | Notes |
 | :-----------------| :----------------| :--------------- | :---- |
 | Boolean       | ce:boolean | [xs:boolean][xml-primitives] | `true` or `false` |
-| Integer       | ce:int | [xs:int][xml-primitives] | |
+| Integer       | ce:integer | [xs:int][xml-primitives] | |
 | String        | ce:string | [xs:string][xml-primitives] | |
 | Binary        | ce:binary | [xs:base64Binary][xml-primitives] |  |
 | URI           | ce:uri | [xs:anyURI][xml-primitives] following [RFC 3986][rfc3986]| |
 | URI-reference | ce:uriRef | [xs:anyURI][xml-primitives] following [RFC 3986][rfc3986] | |
 | Timestamp     | ce:timestamp | [xs:dateTime][xml-primitives] following [RFC 3339][rfc3339] | |
 
-Each CloudEvent context attribute is represented as an XML element whose local
+Each CloudEvent context attribute MUST be represented as an XML element whose local
 name exactly matches that of the attribute.
 
 See the [envelope](#4-envelope) for special handing of the `specversion` context attribute.
 
 Extension attributes MUST be decorated with the appropriate CloudEvent format
-designators using using the `xsi:type` XML attribute, this allows them to be exchanged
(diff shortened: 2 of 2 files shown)


JemDay:
Previous push addresses the comments made at todays WG meeting, namely moving the specification version statement into an attribute of the envelope and removing some blank lines ;-)

xml-format.md
```diff
@@ -0,0 +1,207 @@
...
+    <Txt>Hello World</Txt>
+</Data>
+```
+
+### XML Data
+
+Carried in an element with an expected type of `xs:any`
+
+``` xml
+<Data>
```
jskeet, line 130: Out of interest, would this be valid?

```xml
<Data>
  <Xml question="Can I add an attribute here?">
    Prefix
    <Element1>Content</Element1>
    <Element2>Content</Element2>
    Suffix
  </Xml>
</Data>
```

jskeet: (I can see this being a bit of a pain to represent in SDKs to be honest. It might be simpler to require that the content of `Xml` is exactly one element.)

xml-format.md
```diff
@@ -0,0 +1,207 @@
...
+Each CloudEvent is wholly represented as an XML element, the root xml
+document element is `<CloudEvent ver="n.n">` with the `ver` attribute
+used to define the specification version being followed.
+
+Such a representation MUST use the media type `application/cloudevents+xml`.
+
+The enveloping element contains:
+
+* A set of context attributes.
+* An optional data element.
```
jskeet, line 150: Any restriction on whether the data element is permitted/required based on the presence of a `<Attr n="datacontenttype" ...>` attribute? (That's something I've found myself wondering in a few places.)

xml-format.md
```diff
@@ -0,0 +1,207 @@
...
+</Data>
+```
+
+### Text Data
+
+Carried in an element with an expected type of `xs:string`
+
+``` xml
+<Data>
+    <Txt>Hello World</Txt>
```
jskeet, line 121: I'm a bit concerned that XML whitespace rules could lose data unless SDK authors are really careful. I suspect we should give a warning here with an example.

xml-format.md
```diff
@@ -0,0 +1,207 @@
...
+    </Data>
+</CloudEvent>
+```
+
+## 5. XML Batch Format
+
+In the _XML Batch Format_ several CloudEvents are batched into a single XML
+document. The document comprises a list of elements in the XML Format.
+
+An empty batch is unexpected should and trigger an exception during processing.
```
jskeet, line 171: Why is that the case here, but not in JSON?

xml-format.md
```diff
@@ -0,0 +1,207 @@
...
+
+Although the _XML Batch Format_ builds ontop of the _XML Format_, it is
```
jskeet, line 173:
```suggestion
Although the _XML Batch Format_ builds on top of the _XML Format_, it is
```

spec.xsd
```diff
@@ -0,0 +1,69 @@
...
+        </xs:complexType>
+    </xs:element>
+
+    <!-- Batch Container -->
+
+    <xs:element name="CloudEventBatch">
+        
+        <xs:complexType>
+            <xs:sequence>
+                <xs:element ref="CloudEvent" minOccurs="1" maxOccurs="unbounded" />
```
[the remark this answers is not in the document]
jskeet: (I've highlighted this in another comment as well on the md file.)

jskeet (commented):
Quite a lot of nit-picking - please don't read anything hostile into the number of comments :)

spec.xsd
```diff
@@ -0,0 +1,69 @@
...
+    <xs:complexType name="CloudEventContextAttribute">
+        <xs:attribute name="n" type="xs:string" />
+        <xs:attribute name="v" type="xs:anySimpleType" />
+        <xs:attribute name="t" type="CloudEventContextAttributeValueType" />
+    </xs:complexType>
+
+    <xs:complexType name="CloudEventData">
+        <xs:choice>
+            <xs:element name="Bin" type="xs:base64Binary" />
+            <xs:element name="Txt" type="xs:string" />
```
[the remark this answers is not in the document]
lance: And it's XML, so who's worried about byte count! :rofl:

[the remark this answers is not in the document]
JemDay: I have a (very personal) view on terseness - maybe a better discussion during the WG or offline...

spec.xsd
```diff
@@ -0,0 +1,69 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
+
+    <!-- Core Types -->
+
+    <xs:simpleType name="CloudEventContextAttributeValueType">
```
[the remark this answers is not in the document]
JemDay: Understood - see previous comment.

spec.xsd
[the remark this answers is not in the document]
JemDay: Good Catch .. will change appropriately.

spec.xsd
```diff
@@ -0,0 +1,69 @@
...
+                    <xs:sequence >
+                        <xs:any minOccurs="1" maxOccurs="1" />
+                    </xs:sequence>
+                </xs:complexType>
+            </xs:element>
+        </xs:choice>
+    </xs:complexType>
+
+    <!-- CloudEvent Envelope -->
+
```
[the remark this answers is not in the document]
JemDay: Ack

xml-format.md
```diff
@@ -0,0 +1,207 @@
...
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in [RFC2119][rfc2119].
+
+### 1.2 Approach
+
+This specification promotes an attribute heavy XML document style, such a model
+presents advantages in terms of parsing/processing along with a more compact
+XML document representation.
+
+Attribute and Element names are deliberately terse for performance reasons.
```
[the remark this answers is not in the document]
JemDay: Ack

xml-format.md
```diff
@@ -0,0 +1,207 @@
...
+
+Preservation of attribute value type information is supported allowing
```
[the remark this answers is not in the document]
JemDay: My understanding is that the Java SDK is intending to be strict on what it receives, so malformed events (or attribute values) will trigger exceptions.

It does sound like a cross-cutting SDK concern.

xml-format.md
```diff
@@ -0,0 +1,207 @@
...
+
+### 1.1. Conformance
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in [RFC2119][rfc2119].
+
+### 1.2 Approach
+
+This specification promotes an attribute heavy XML document style, such a model
```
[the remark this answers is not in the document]
JemDay: Ack ...

xml-format.md
```diff
@@ -0,0 +1,207 @@
...
+
+[CloudEvents][ce] is a standardized and protocol-agnostic definition of the
+structure and metadata description of events. This specification defines how
+elements defined in the CloudEvents specification are to be represented using
+[Extensible Markup Language (XML)](xml-spec) documents.
+
+* The [associated schema][xml-schema] formalizes the structure of the XML
+document.
+
+* The [Attributes](#2-attributes) section describes the naming conventions and
```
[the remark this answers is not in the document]
JemDay: you're probably right .. to much cut-and-past on my part ;-)

spec.xsd
```diff
@@ -0,0 +1,69 @@
...
+          
+            <xs:sequence>
+                <!-- The Context Attributes -->               
+                <xs:element name="Attr" type="CloudEventContextAttribute" minOccurs="3" maxOccurs="unbounded" />
+                <!-- Data -->
+                <xs:element name="Data" type="CloudEventData" minOccurs="0" maxOccurs="1" />
+            </xs:sequence>
+
+            <!-- The specification version definition -->
+            <xs:attribute name="ver" type="xs:string" use="required"/>
```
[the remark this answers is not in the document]
JemDay: Changed the name. As to the ordering, I'm not sure - certainly the VSCode plugin I'm using doesn't think it's valid to put it first.

xml-format.md
```diff
@@ -0,0 +1,207 @@
...
+To better support streaming style processing the specification version
+(`specversion` context attribute) statement is communicated in an explicit element.
+
+``` xml
+<Ver>1.0</Ver>
+```
+
+## 3. Data
+
+The `Data` portion of a CloudEvent is contained is a `<Data></Data>` element and supports
```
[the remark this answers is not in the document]
JemDay: I'm flexible .. it seemed a somewhat easier way to detect the switch from attributes to data, let me noodle on this one.

xml-format.md
JemDay: This shows how long it's been since I did XML Schema's - it was not my intention to support what you described.

I'll do some more digging on the correct schema definition.

xml-format.md
JemDay: I don't think the other formats make this type of connection especially since that is an optional context attribute, I think it's more of a usage expectation - I'll try and come up with some language.

xml-format.md
JemDay: Could you clarify - sorry, it really has been that long ;-)

xml-format.md
JemDay: Changed

xml-format.md
JemDay: Ack

JemDay:
Thanks for the comments, I've addressed most of the nits in the previous push, I will look independently at the <Data> representation concerns.

xml-format.md
jskeet: I suspect that string data with a value of "First line\n\n\nFourth line" might end up as "First line Fourth line" after XML whitespace handling. I may be wrong, and it would certainly depend on the XML API being used. Happy to look into it further.

xml-format.md
```diff
@@ -0,0 +1,213 @@
...
+See section below on special handling for `specversion`
+
+## 2.1 Specification Version
+
+To better support streaming style processing the specification version
+(`specversion` context attribute) is communicated as an attribute of the enclosing CloudEvent
+enveloping element.
+
+``` xml
+<CloudEvent specver="1.0">
```
slinkydeveloper, line 98: Maybe `specversion` to be consistent with other formats?

slinkydeveloper (commented):
I'm no expert of XML, but this generally looks good to me, I left just a minor comment

xml-format.md
```diff
@@ -0,0 +1,213 @@
...
+``` xml
+<CloudEvent ver="1.0" >
+    <Attr n="id" v="000-1111-2222" t="S">
+    <Attr n="time" v="2020-03-19T12:54:00-07:00" t="T">
+    <Attr n="type" v="SOME.EVENT.TYPE" t="S"/>
+    ....
+    <Data>
+        <Txt>Now is the winter of our discount tents...</Txt>
+    </Data>
+</CloudEvent>
```
duglin, line 165:
```
<CloudEvent specversion="1.0">
  <id>..</id>
  <time>...</time>
  <type>...</type>
  <data> .. </data>
  ...
</CloudEvent>
```

jskeet: And to capture the conversation a bit more:

```xml
<cloudevent specversion="1.0">
  <id>...</id>
  <time>...</time>
  <type>...</type>
  <mytimeextension xsi:type=""xs:dateTime>...</mytimeextension>
</cloudevent>
```

JemDay:
I ~wasted~ spent time on Friday trying to construct an XMLSchema to support the representation that we discussed during the WG call (and echo'd in the comments above) - I'm coming to the conclusion that that might not be possible.

My memories of XMLSchema may have faded greatly but I don't believe it's a vehicle for expressing a document format where the "types" are known but the element names aren't .. use of s:any in that scenario doesn't seem to add much value.

If anybody has latent XSD skills and could help I'd appreciate it, otherwise I'm tempted to ditch the notion of a schema and describe the document as a "convention" to enable something like...

``` xml
<cloudevent specversion="1.0">
  <id>...</id>
  <time>...</time>
  <type>...</type>
  <mytimeextension xsi:type="ce:timestamp">...</mytimeextension>
  <mybinaryextn xsi:type="ce:binary">..<mybinaryextn>
  <data xsi:type="xs:string">my event text</data>
</cloudevent>
```

jskeet:
I don't mind there not being a schema, but I don't think that means it has to be relegated to only being a convention. For example, I'd be fine with saying that this is *not* a valid CloudEvent:

```xml
<cloudevent specversion="1.0">
  <id>...</id>
  <time>...</time>
  <type>...</type>
  <BRIAN_BLESSED>SHOUTY EXTENSION ATTRIBUTE VALUE</BRIAN_BLESSED>
</cloudevent>
```

I also think it's reasonable to infer a type of "string" for any extension attribute that doesn't specify it.

I wonder whether we should still use XML attributes for attributes though:

```csharp
<cloudevent specversion="1.0">
  <id value="..." />
  <time value="..." />
  <type value="..." />
  <data xsi:type="xs:string">my event text</data>
</cloudevent>
```

One option to *consider* allowing would be "CE attributes that are string-valued can be expressed as attributes of the `cloudevent` element":

```xml
<cloudevent specversion="1.0" id="my-id" type="my-type">
  <time xsi:type="ce:timestamp" value="..." />
  <source xsi:type="ce:uri-ref" value="..." />
</cloudevent>
```

Or would two different representations be too confusing?

duglin:
@JemDay any update on this?

xml-format.md
JemDay: you've probably forgotten you commented on this ... I believe my latest changes address these concerns.

JemDay:
Having played with code-gen I do need to tweak this a little more as the schema isn't yet deterministic, apologies to anybody who might have been excited by an update to this PR.

duglin:
@JemDay can you rebase?

duglin:
@JemDay were you going to put these into a "working-drafts" folder per a previous call?

JemDay:
@duglin Yes I can, and yes I will..

README.md
```diff
@@ -139,3 +140,6 @@ Periodically, the group may have in-person meetings that coincide with a major
 conference. Please see the
 [meeting minutes doc](https://docs.google.com/document/d/1OVF68rpuPK5shIHILK9JOqlZBbfe91RNzQ7u_P7YCDE/edit#)
 for any future plans.
+
+[xml-format-working]: https://github.com/cloudevents/spec/blob/master/cloudevents/formats/xml-format.md
+[xml-format-latest]: https://github.com/cloudevents/spec/blob/master/cloudevents/formats/xml-format.md
```
duglin, line 145: These need to use the "main" branch

JemDay: Done

README.md
```diff
@@ -139,3 +140,6 @@ Periodically, the group may have in-person meetings that coincide with a major
 conference. Please see the
 [meeting minutes doc](https://docs.google.com/document/d/1OVF68rpuPK5shIHILK9JOqlZBbfe91RNzQ7u_P7YCDE/edit#)
 for any future plans.
+
+[xml-format-working]: https://github.com/cloudevents/spec/blob/main/cloudevents/formats/xml-format.md
```
duglin, line 144: I think this link needs to be updated.
Is the next link used?

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,243 @@
...
+data type mappings for CloudEvents context attributes.
+
+* The [Data](#3-data) section defines the container for the data portion of a
+CloudEvent.
+
+* The [Envelope](#4-envelope) section defines an XML container for CloudEvents
+attributes and an associated media type.
+
+* The [Batch](#5-xml-batch-format) section describes how multiple CloudEvents
+may be packaged into a single XML document.
```
duglin, line 38: s/may/can/

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,243 @@
...
+| Boolean       | [xs:boolean][xml-primitives] | |
+| Integer       | [xs:int][xml-primitives] | |
+| String        | [xs:string][xml-primitives] | |
+| Binary        | [xs:base64Binary][xml-primitives] | |
+| URI           | [xs:anyURI][xml-primitives] following [RFC 3986][rfc3986]| |
+| URI-reference | [xs:anyURI][xml-primitives] following [RFC 3986][rfc3986] | |
+| Timestamp     | [xs:dateTime][xml-primitives] following [RFC 3339][rfc3339] | |
+
+Each context attribute is represented as an XML element whose name exactly matches that of
+a required or optional CloudEvent attribute, extension attribute names `MUST` adhere and
```
duglin, line 72: I suspect REQUIRED and OPTIONAL need to be upper-case here, even though it's a bit of a gray area for this usage. But it'll keep the bot happy.
Also, you don't need to add back-ticks around the RFC keywords, we don't do that in other specs.

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,243 @@
...
+| :-------------| :---------------------------------------------------------- |
+| Boolean       | [xs:boolean][xml-primitives] | |
+| Integer       | [xs:int][xml-primitives] | |
+| String        | [xs:string][xml-primitives] | |
+| Binary        | [xs:base64Binary][xml-primitives] | |
+| URI           | [xs:anyURI][xml-primitives] following [RFC 3986][rfc3986]| |
+| URI-reference | [xs:anyURI][xml-primitives] following [RFC 3986][rfc3986] | |
+| Timestamp     | [xs:dateTime][xml-primitives] following [RFC 3339][rfc3339] | |
+
+Each context attribute is represented as an XML element whose name exactly matches that of
```
duglin, line 71: I think you may need a MUST in here.... whose name MUST exactly match....
or something like that

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,243 @@
...
+a required or optional CloudEvent attribute, extension attribute names `MUST` adhere and
+align with the conventions in defined in the [cloud event specification][ce-spec].
+
+See the [envelope](#4-envelope) for special handing of the `specversion` context attribute.
+
+Extension attributes `SHOULD` be expressed with the appropriate `xsi:type` to allow them
+to be exchanged without loss of type information. Extension attributes with no `xsi:type`
+discriminator `SHOULD` be interpreted with an implied type of `xs:string`
```
duglin, line 79: For well-known attributes shouldn't they follow the type defined in the CE spec? As of today I think they're all 'strings' but if that changes it seems like this spec should allow for non-strings.

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,243 @@
...
+    <time>2021-08-14T-08:00</time>
+    ..
+    <myextension xsi:type="xs:string">my extension value</myextension>
+    <myboolean xsi:type="xs:boolean">false</myboolean>
+```
+
+## 3. Data
+
+The `Data` portion of a CloudEvent follows a similar model to that employed by
+the [JSON Format specification][json-format]. The elmenent names used are capitalized to avoid
```
duglin, line 94: s/elmenent/element/

duglin, line 94: MUST be capitalized

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,243 @@
...
+collision with CloudEvent attribute names.
+
+### 3.1 Binary Data
+
+Carried in an element with an defined type of `xs:base64Binary`
+
+``` xml
+<Data xsi:type="xs:base64Binary">.........</Data>
```
duglin, line 102: should it be "data" instead of "Data" ?

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,243 @@
...
+``` xml
+<Data xsi:type="xs:string">This is text</Data>
+```
+
+### 3.3 XML Data
+
+XML data is carried in an explicit element:
+
+``` xml
+<DataXml>
```
duglin, line 118: isn't there an xsi type we can use instead of a new element name?

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,243 @@
...
+
+Carried in an element with an defined type of `xs:base64Binary`
+
+``` xml
+<Data xsi:type="xs:base64Binary">.........</Data>
+```
+
+### 3.2 Text Data
+
+Carried in an element with an defined type of `xs:string`
```
duglin, line 107: Need a MUST in here

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,243 @@
...
+
+``` xml
+<Data xsi:type="xs:string">This is text</Data>
+```
+
+### 3.3 XML Data
+
+XML data is carried in an explicit element:
```
duglin, line 115: Needs a MUST

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,243 @@
...
+        <myData>
+            ....
+        </myData>  
+</DataXml>
+```
+
+## 4. Envelope
+
+Each CloudEvent is wholly represented as an XML element, the root xml
+document element is `<CloudEvent>`. This element carries the `speversion`
```
duglin, line 128: MUST be `<CloudEvent>`.
s/speversion/specversion/  - need a MUST too.
Should it be "cloudevent" to be consistent (case-wise) with all other elements?

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,243 @@
...
+
+``` xml
+<CloudEvent specversion="1.0">
+    <time>2020-03-19T12:54:00-07:00</time>
+    <datacontenttype>text/plain</datacontenttype>
+    <id>000-1111-2222</id>
+    <source>urn:uuid:123e4567-e89b-12d3-a456-426614174000</source>
+    <type>SOME.EVENT.TYPE</type>
+    <myboolean xsi:type="xs:boolean">false</myboolean>
+    <Data xsi:type="xs_string">Now is the winter of our discount tents...</Data
```
duglin, line 148: missing > at the end

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,243 @@
...
+doesn't need to support it. The _XML Batch Format_ MUST NOT be used when only
+support for the _XML Format_ is indicated.
+
+An XML Batch of CloudEvents MUST use the media type
+`application/cloudevents-batch+xml`.
+
+Example:
+
+``` xml
+<CloudEventBatch>
```
duglin, line 168: should this just be `<cloudevents>` ? Kind of reminds me when people name things like `nameString` instead of just `name`  :-)

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,243 @@
...
+
+```xml
+<?xml version="1.0" encoding="UTF-8"?>
+<CloudEvent xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xs="http://www.w3.org/2001/XMLSchema" specversion="1.0" >
+    <time>2020-03-19T12:54:00-07:00</time>
+    <datacontenttype>application/json</datacontenttype>
+    <id>000-1111-2222</id>
+    <source>urn:uuid:123e4567-e89b-12d3-a456-426614174000</source>
+    <type>SOME.EVENT.TYPE</type>
+    <Data xsi:type="xs:string">{ "salutation" : "Good Morning", "text" : "hello world" }</Data>
```
duglin, line 211: what happens if the JSON has a `<` in it? Is there any escaping needed?

duglin:
overall it looks good - just some minor things. Thanks for doing this

README.md
JemDay: Correct

cloudevents/working-drafts/xml-format.md
JemDay: Ack

cloudevents/working-drafts/xml-format.md
JemDay: Ack

cloudevents/working-drafts/xml-format.md
JemDay: The spec defines those 7 core types, all of which *may* be represented as string types .. but using the appropriate xsi:type makes everything explicit.

cloudevents/working-drafts/xml-format.md
JemDay: Ack

JemDay: Ack

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,243 @@
...
+
+## 3. Data
+
+The `Data` portion of a CloudEvent follows a similar model to that employed by
+the [JSON Format specification][json-format]. The elmenent names used are capitalized to avoid
+collision with CloudEvent attribute names.
+
+### 3.1 Binary Data
+
+Carried in an element with an defined type of `xs:base64Binary`
```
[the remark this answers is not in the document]
JemDay: Ack

cloudevents/working-drafts/xml-format.md
JemDay: ack

cloudevents/working-drafts/xml-format.md
JemDay: Ack

cloudevents/working-drafts/xml-format.md
JemDay: ack

cloudevents/working-drafts/xml-format.md
JemDay: We can discuss ... I'd prefer the name to be most explicit but I'm not hung-up on it.

cloudevents/working-drafts/xml-format.md
JemDay: Yes ... content does need to be appropriately escaped to ensure the XML document is well formed.

cloudevents/working-drafts/xml-format.md
JemDay: ack

cloudevents/working-drafts/xml-format.md
duglin: Sorry,I think I misread things... or I can't remember what I was thinking. Ignore my comment.

cloudevents/working-drafts/xml-format.md
JemDay: doh

cloudevents/working-drafts/xml-format.md
JemDay: I don't believe there's an out-of-the-box type that will work in this situation.

cloudevents/working-drafts/xml-format.md
JemDay: Uppercase to avoid collision with context attributes

duglin: `data` is (sort of) a context attribute though. That's why the example [here](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#example) makes it look like "data" is just like all other attributes

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,243 @@
...
+
+``` xml
+<CloudEvent specversion="1.0">
+    <time>2020-03-19T12:54:00-07:00</time>
+    <datacontenttype>text/plain</datacontenttype>
+    <id>000-1111-2222</id>
+    <source>urn:uuid:123e4567-e89b-12d3-a456-426614174000</source>
+    <type>SOME.EVENT.TYPE</type>
+    <myboolean xsi:type="xs:boolean">false</myboolean>
+    <Data xsi:type="xs_string">Now is the winter of our discount tents...</Data>
```
duglin, line 148: xs_string? maybe "xs:string" ?

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,243 @@
...
+support for the _XML Format_ is indicated.
+
+An XML Batch of CloudEvents MUST use the media type
+`application/cloudevents-batch+xml`.
+
+Example:
+
+``` xml
+<CloudEventBatch>
+    <CloudEvent specversion="1.0">
```
jskeet, line 169: I'm sure in *some* spec there's something about all events in the same batch having to use the same specversion. I can't find it now though. (I've always disagreed with it anyway...) Not sure whether it needs calling out at all - worth a 2-minute discussion in the meeting (where others will no doubt be able to find the rule I'm remembering badly...)

cloudevents/working-drafts/xml-format.md
jskeet: Can we remove the spaces before the colons please? in other words:

```xml
<Data xsi:type="xs:string">{ "salutation": "Good Morning", "text": "hello world" }</Data>
```

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,243 @@
...
+
+MUST be carried in an element with an defined type of `xs:string`
+
+``` xml
+<Data xsi:type="xs:string">This is text</Data>
+```
+
+### 3.3 XML Data
+
+XML data MUST be carried in an explicit element `<DataXml>`:
```
jskeet, line 115: I think a bit more detail would be useful here. Is the intention that the content of `DataXml` is a single element? Could it just be multiple elements? Can it be empty? If it's always a single element, is the whitespace around the element considered significant? It might be worth having examples of everything that's possible.

(Oh, and clarify that attributes are not permitted on the `DataXml` element itself, assuming that's what we want?)

cloudevents/working-drafts/xml-format.md
jskeet: Ignore this - it's in the HTTP transport. No need to put anything in here.

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,243 @@
...
+* The [Attributes](#2-attributes) section describes the representation and
+data type mappings for CloudEvents context attributes.
+
+* The [Data](#3-data) section defines the container for the data portion of a
+CloudEvent.
+
+* The [Envelope](#4-envelope) section defines an XML container for CloudEvents
+attributes and an associated media type.
+
+* The [Batch](#5-xml-batch-format) section describes how multiple CloudEvents
```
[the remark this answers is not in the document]
JemDay: Ack ... tweaked in next push

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,243 @@
...
+
+### 1.1. Conformance
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in [RFC2119][rfc2119].
+
+### 1.2 Approach
+
+This specification is deliberately verbose in favor of readability and
```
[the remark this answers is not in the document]
JemDay: Ack

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,243 @@
...
+deterministic processing.
+
+Preservation of attribute value type information is supported allowing
```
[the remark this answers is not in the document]
JemDay: ack

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,243 @@
...
+
+Extension attributes SHOULD be expressed with the appropriate `xsi:type` to allow them
+to be exchanged without loss of type information. Extension attributes with no `xsi:type`
+discriminator SHOULD be interpreted with an implied type of `xs:string`
+
+``` xml
+    ...
+    <id>AAABBBCCCNNN0000</id>
+    ..
+    <time>2021-08-14T-08:00</time>
```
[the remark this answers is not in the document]
JemDay: Typo .. but I believe it is a valid representation of a date .. but it should be s timestamp.. will fix

cloudevents/working-drafts/xml-format.md
JemDay: I believe this is addressed in the next commit.

cloudevents/working-drafts/xml-format.md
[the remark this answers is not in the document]
jskeet: I'd still prefer "XML element" in the line below, or change "into" to "within" (to make it clear that this specification could still be used if you wanted to create a document that contained multiple batches, for example). Basically "packaged into a single XML document" feels like there's a transformation from CloudEvent to XML document, and I think it's really a transformation from CloudEvent to XML element (which is itself a valid document, of course).

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,277 @@
...
+This XML representation used is deliberately verbose in favor of readability
+and deterministic processing.
+
+Preservation of type information for CloudEvent attributes is supported allowing
+custom extensions to be communicated without type loss.
+
+A schema-less approach has been taken favoring convention over rigid document
+structure.
+
+The namespace `http://cloudevents.io/xmlformat/v1` SHOULD be used, when
```
jskeet, line 57: Three points:

1) This namespace isn't quite the same as the one used everywhere else - it's got "v1" where all the examples have "V1". I don't know which we should use, but we should make them consistent.
2) I'm interested in why this is only a SHOULD. That feels like anything processing XML representations has to deal with namespaced and non-namespaced elements, which is painful. Could we make this "MUST"?
3) I'd put the namespace prefixes part into a new sentence.

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,277 @@
...
+``` xml
+<data xsi:type="xs:base64Binary">.........</data>
+```
+
+### 3.2 Text Data
+
+MUST be carried in an element with an defined type of `xs:string`
+
+``` xml
+<data xsi:type="xs:string">This is text</data>
```
jskeet, line 117: I still want to check what's likely to happen with whitespace in text, and whether that's likely to affect roundtrips. I'm happy to prototype an implementation of this in C# for testing purposes if that would help. (It won't say what other XML APIs will do, of course.)

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,277 @@
...
+    ..
+    <time>2021-08-14T14:30:22-08:00</time>
+    ..
+    <myextension xsi:type="xs:string">my extension value</myextension>
+    <myboolean xsi:type="xs:boolean">false</myboolean>
+```
+
+## 3. Data
+
+The `Data` portion of a CloudEvent follows a similar model to that employed by
```
jskeet, line 99: Not sure why `Data` is a) capitalized; b) in code font here.

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,277 @@
...
+A schema-less approach has been taken favoring convention over rigid document
+structure.
+
+The namespace `http://cloudevents.io/xmlformat/v1` SHOULD be used, when
+namespace prefixes are used a prefix of `ce` is preferred but MUST NOT be expected
+from an XML document processing perspective.
+
+An XML document preamble SHOULD be included to ensure deterministic processing.
+
+## 2. Attributes
```
jskeet, line 63: Just in terms of clarity, I wonder whether we should put the envelope first. It feels a little odd to start with attributes without knowing what they're going to be part of.
Only something to consider, and can definitely be ignored.

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,277 @@
...
+    <type>SOME.EVENT.TYPE</type>
+    <myboolean xsi:type="xs:boolean">false</myboolean>
+    <data xsi:type="xs:string">Now is the winter of our discount tents...</data>
+</Event>
+```
+
+## 5. XML Batch Format
+
+In the _XML Batch Format_ several CloudEvents are batched into a single XML
+document. The document comprises a list of elements in the XML Format.
```
jskeet, line 163: document => element (unless this is really trying to say that you shouldn't create a document that includes multiple batches)

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,277 @@
...
+<data xsi:type="xs:any">
+    <myData>
+        ....
+    </myData>  
+</data>
+```
+
+## 4. Envelope
+
+Each CloudEvent is wholly represented as an XML element `<Event>` that
```
jskeet, line 136: I don't *think* we discussed the casing part here yesterday. It looks a little odd for all elements other than `Event` and `Batch` to be lowercase. (The casing part got lost in other aspects of the naming discussion.)

Given recent discussions around other specs, I'd suggest that using `<event>` and `<batch>` would cause less surprise to users.

jskeet (commented):
Just a few suggestions remaining. Jem, let me know if you'd like to have a quick chat about these separately to the regular meeting, so that my bike-shedding/nit-picking doesn't take too much time from other matters...

duglin:
@JemDay can you rebase? there's a conflict in the README

cloudevents/working-drafts/xml-format.md
JemDay: ack

cloudevents/working-drafts/xml-format.md
JemDay: I basically followed the same structure as the JSON format ... so I get your point, but do we wave the "consistency flag" here ?

cloudevents/working-drafts/xml-format.md
JemDay: point taken.

cloudevents/working-drafts/xml-format.md
JemDay: Makes sense ... now that namespaces are back it definitely makes sense to move to a MUST.

Will address the other points as well.

cloudevents/working-drafts/xml-format.md
JemDay: I'll ping you on this one...

cloudevents/working-drafts/xml-format.md
jskeet: Ah, interesting. In the JSON format there isn't really anything "to" the envelope - it's just an object. I'm happy for you to stick with this as it is. We could always re-evaluate wholesale for a later release.

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,278 @@
...
+    ..
+    <time>2021-08-14T14:30:22-08:00</time>
+    ..
+    <myextension xsi:type="xs:string">my extension value</myextension>
+    <myboolean xsi:type="xs:boolean">false</myboolean>
+```
+
+## 3. Data
+
+The data portion of a CloudEvent follows a similar model to that employed by
```
jskeet, line 100: Is every xsi:type other than xs:string, xs:base64Binary and xs:any prohibited? (And should we say that?)
And can I infer from the subsections below that the xsi:type MUST be specified? (I think I can - just checking that we didn't want any defaulting.)

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,278 @@
...
+
+Each context attribute is represented as an XML element whose name MUST exactly match
+that of a REQUIRED or OPTIONAL CloudEvent attribute, extension attribute names MUST
+adhere to, and align with, the conventions in defined in the [cloud event specification][ce-spec].
+
+See the [envelope](#4-envelope) for special handing of the `specversion` context attribute.
+
+Extension attributes SHOULD be expressed with the appropriate `xsi:type` to allow them
+to be exchanged without loss of type information. Extension attributes with no `xsi:type`
+discriminator SHOULD be interpreted with an implied type of `xs:string`
```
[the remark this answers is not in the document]
jskeet: More broadly I guess: do we want to state that any XML attributes not specified here are prohibited?

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,278 @@
...
+This XML representation used is deliberately verbose in favor of readability
+and deterministic processing.
+
+Preservation of type information for CloudEvent attributes is supported allowing
+custom extensions to be communicated without type loss.
+
+A schema-less approach has been taken favoring convention over rigid document
+structure.
+
+The namespace `http://cloudevents.io/xmlformat/V1` MUST be used.
```
jskeet, line 57: That's just for elements, not attributes, yes? I think the only attribute we use is the specversion attribute, but am I right in saying that's un-namespaced? (From what I remember, using xmnls="..." sets the default for elements but not attributes. I believe that namespaced attributes are relatively uncommon in XML.

cloudevents/working-drafts/xml-format.md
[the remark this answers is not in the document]
jskeet: Not sure why this was resolved - in general, I suspect that "document" can be replaced with "element" in a few more places. Let's discuss that if you disagree :) I think it's been replaced everywhere after section 1 already. (And there are a few places referring to document processing etc which make sense.)

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,278 @@
...
+    </myData>  
+</data>
+```
+
+## 4. Envelope
+
+Each CloudEvent is wholly represented as an XML element `<event>` that
+carries the `specversion` as an XML attribute value.
+
+Such a representation MUST use the media type `application/cloudevents+xml`.
```
jskeet, line 140: Presumably this is in the context of transports etc? Maybe clarify that - it's not relevant if you're loading an XML file, for example.

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,278 @@
...
+## 3. Data
+
+The data portion of a CloudEvent follows a similar model to that employed by
+the [JSON Format specification][json-format]. A `data` element is used to
+encapsulate the payload and an `xsi:type` used to discrimate the payload
+type.
+
+### 3.1 Binary Data
+
+MUST be carried in an element with an defined type of `xs:base64Binary`
```
jskeet, line 107: Possibly start this sentence with "Binary data" and in 3.2 "Text data" to match the 3.3 "XML data"? (At the moment there's no subject for this sentence unless you include the section heading.)

jskeet (approved):
All very nitty at this point - I don't think any of this should block submission, although if anything is left out for now just for convenience, I'm happy to create a new PR later.

I strongly suspect that when implementing this for C#, I'll find ambiguities to be pinned down further :)

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,278 @@
...
+| Boolean       | [xs:boolean][xml-primitives] | |
+| Integer       | [xs:int][xml-primitives] | |
+| String        | [xs:string][xml-primitives] | |
+| Binary        | [xs:base64Binary][xml-primitives] | |
+| URI           | [xs:anyURI][xml-primitives] following [RFC 3986][rfc3986]| |
+| URI-reference | [xs:anyURI][xml-primitives] following [RFC 3986][rfc3986] | |
+| Timestamp     | [xs:dateTime][xml-primitives] following [RFC 3339][rfc3339] | |
+
+Each context attribute is represented as an XML element whose name MUST exactly match
+that of a REQUIRED or OPTIONAL CloudEvent attribute, extension attribute names MUST
```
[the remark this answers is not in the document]
duglin: Perhaps just:
- - -
Each context attribute is represented as an XML element whose name MUST exactly match that of the CloudEvent attribute being serialized.
- - -
That should cover all attributes (required, optional and extensions).

cloudevents/working-drafts/xml-format.md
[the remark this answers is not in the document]
duglin: Should there a period at the end of the sentence and then a line like:
```
Examples:
```

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,278 @@
...
+    <time>2021-08-14T14:30:22-08:00</time>
+    ..
+    <myextension xsi:type="xs:string">my extension value</myextension>
+    <myboolean xsi:type="xs:boolean">false</myboolean>
+```
+
+## 3. Data
+
+The data portion of a CloudEvent follows a similar model to that employed by
+the [JSON Format specification][json-format]. A `data` element is used to
```
duglin, line 101: s/is used/MUST be used/

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,278 @@
...
+encapsulate the payload and an `xsi:type` used to discrimate the payload
```
duglin, line 102: a/used/MAY be used/

cloudevents/working-drafts/xml-format.md
duglin: s/with an/with a/

And end with a period and `Example:` before the sample below

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,278 @@
...
+
+MUST be carried in an element with an defined type of `xs:base64Binary`
+
+``` xml
+<data xsi:type="xs:base64Binary">.........</data>
+```
+
+### 3.2 Text Data
+
+MUST be carried in an element with an defined type of `xs:string`
```
duglin, line 115: I agree with @jskeet about starting these with "Text data ..." just so it reads better, especially for non-native english speakers

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,278 @@
...
+
+``` xml
+<data xsi:type="xs:string">This is text</data>
+```
+
+### 3.3 XML Data
+
+XML data MUST be carried in an element with a defined type of `xs:any`
+allowing a single XML element (with any required namespace definitions)
+to be represented.
```
duglin, line 125: I think we talked about dropping the default NS for elements in `<data>`, right?

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,278 @@
...
+    <myData>
+        ....
+    </myData>  
+</data>
+```
+
+## 4. Envelope
+
+Each CloudEvent is wholly represented as an XML element `<event>` that
+carries the `specversion` as an XML attribute value.
```
duglin, line 138: s/carries/MUST carry/

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,278 @@
...
+
+Such a representation MUST use the media type `application/cloudevents+xml`.
+
+The enveloping element contains:
+
+* A set of context attribute elements.
+* An optional `data` element.
```
duglin, line 145: s/optional/OPTIONAL/

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,278 @@
...
+
+eg _(XML preamble and namespace definitions omitted for brevity)_:
```
duglin, line 147: s/eg/Example:/

cloudevents/working-drafts/xml-format.md
duglin: Does the attribute need to be namespaced? If not perhaps we should mandate it either way?

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,278 @@
...
+## 4. Envelope
+
+Each CloudEvent is wholly represented as an XML element `<event>` that
+carries the `specversion` as an XML attribute value.
+
+Such a representation MUST use the media type `application/cloudevents+xml`.
+
+The enveloping element contains:
+
+* A set of context attribute elements.
```
duglin, line 144: Since you talk about "optional" in the next bullet, perhaps:
- - -
* A set of XML elements CloudEvent context attributes.
- - -

cloudevents/working-drafts/xml-format.md
duglin: This isn't true for batch, so perhaps we need to qualify the above sentence to limit it to an envelope with a single CloudEvent.

cloudevents/working-drafts/xml-format.md
JemDay: ack

cloudevents/working-drafts/xml-format.md
JemDay: ack ..

Now that namspacing is required it should be automatic ... I'll see if I can tweak the language a bit

cloudevents/working-drafts/xml-format.md
JemDay: ack

cloudevents/working-drafts/xml-format.md
JemDay: hmm.. In this context (the section related to `data`) it's an instruction, there is no option. So I'm not convinced 'MAY' is appropriate.

cloudevents/working-drafts/xml-format.md
JemDay: ack

cloudevents/working-drafts/xml-format.md
JemDay: I changed the language to a MUST :-)

cloudevents/working-drafts/xml-format.md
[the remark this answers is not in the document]
JemDay: reworded ... see next commit

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,278 @@
...
+| :-------------| :--------------- | :---- |
+| Boolean       | [xs:boolean][xml-primitives] | |
+| Integer       | [xs:int][xml-primitives] | |
+| String        | [xs:string][xml-primitives] | |
+| Binary        | [xs:base64Binary][xml-primitives] | |
+| URI           | [xs:anyURI][xml-primitives] following [RFC 3986][rfc3986]| |
+| URI-reference | [xs:anyURI][xml-primitives] following [RFC 3986][rfc3986] | |
+| Timestamp     | [xs:dateTime][xml-primitives] following [RFC 3339][rfc3339] | |
+
+Each context attribute is represented as an XML element whose name MUST exactly match
```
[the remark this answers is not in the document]
JemDay: ack

cloudevents/working-drafts/xml-format.md
JemDay: It's true it relates to the transport ... it's in a bit of an odd place, in think the JSON format spec does it here - rightly or wrongly.

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,290 @@
...
+The CloudEvents type system is mapped to the XML schema types as follows :
+
+| CloudEvents   |  XML Schema Type | Notes |
+| :-------------| :--------------- | :---- |
+| Boolean       | [xs:boolean][xml-primitives] | |
+| Integer       | [xs:int][xml-primitives] | |
+| String        | [xs:string][xml-primitives] | |
+| Binary        | [xs:base64Binary][xml-primitives] | |
+| URI           | [xs:anyURI][xml-primitives] following [RFC 3986][rfc3986]| |
+| URI-reference | [xs:anyURI][xml-primitives] following [RFC 3986][rfc3986] | |
```
jskeet, line 75: Ooh, that's a shame... this means we can't *really* preserve type information for extension attributes, as we can't tell the difference between URI and URI-reference.

Perhaps we should have an extra attribute for URI-reference of "ref=true" or something like that?

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,290 @@
...
+```
+
+## 6. Examples
+
+### 6.1 CloudEvent with PNG image data
+
+```xml
+<?xml version="1.0" encoding="UTF-8"?>
+<event xmlns="http://cloudevents.io/xmlformat/V1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
+       xmlns:xs="http://www.w3.org/2001/XMLSchema" specversion="1.0" >
```
jskeet, line 213: I don't *think* the xs alias is needed here, given that we only refer to it within the value.

jskeet (commented):
I've now got a draft implementation for C#, and I haven't discovered *many* things to update here. However:

- I think we should have a way of differentiating between URI and URI-Reference types for extension attributes
- I think we should explicitly prohibit unexpected elements and attributes, but we should probably allow `xmnls:xyz="..."` attributes anywhere
- At one point I've a sneaking memory that we wanted to require `data` to be the last element - is that still the case?

cloudevents/working-drafts/xml-format.md
JemDay: It's probably this extra isn't needed at all ... it's there more for completeness that anything else.

cloudevents/working-drafts/xml-format.md
JemDay: Yep .. I struggle with this a bit as well ...

The only really way out of this is to define our own set of 'ce' types and use those - that's the road I was going down when I was playing with a schema.

So you'd end up with something like
<myextension xsi:type="ce:string">i'm a string</myextension>

It's really no different that using the pre-defined XS types - I'm happy to make this switch, I don't think it's a major change

cloudevents/working-drafts/xml-format.md
JemDay: Does the latest version address this concern - the language has changed.

cloudevents/working-drafts/xml-format.md
jskeet: There's one benefit to using predefined xs types, which is that some editors support them - for example, for my unit tests I've got some written in code rather than as XML, because when I try to put an invalid value in an `xs:boolean` element, Visual Studio emits a warning. But I agree that in other ways it's cleaner than having an extra attribute.

So some options:

- Go with the spec as written, unable to differentiate between URI and URI-Reference
- Introduce `ce:uri` but leave the rest of the `xs:` types as they are
- Stick with just `xs:` types but add a qualifying attribute to go alongside `xs:anyURI`
- Ditch the `xs` types entirely and go full-on `ce:` types. (I don't know whether using `xsi:type` for that becomes a problem)

Smarter folks than me may well have other suggestions of course.

JemDay:
@jskeet - I don't remember an ordering discussion and I haven't made any statements about that in the spec, as it stands today for example the (unpublished) Java SDK will handle data wherever it occurs in the set of XML elements.

w.r.t addressing potential type loss - I think it's probably a good idea to address that now, I will do through and establish a set of CloudEvent specific types as opposed to simply relying on the XML Schema set.

I will endeavor to get that done by the end-of-the-week, but is unlikely to be there for Thursday

cloudevents/working-drafts/xml-format.md
```diff
@@ -130,15 +130,17 @@ Example:
 
 ### 3.3 XML Data
 
-XML data MUST be carried in an element with a defined type of `xs:any`
-allowing a single child XML element (with any required namespace definitions)
-to be represented.
+XML data MUST be carried in an element with a defined type of `xs:any` with
+a single child XML element (with any required namespace definitions).
+
+The child business data XML elements MUST NOT reside in the CloudEvent namespace,
```
jskeet, line 138: I'd like us to discuss this in the next WG meeting where both Jem and I are present.

Pros:
- Avoids *accidentally* using the CloudEvent namespace

Cons:
- May prevent genuinely useful functionality, e.g. "wrap a batch as a payload in a single CloudEvent for transports that don't understand batches".

README.md
```diff
@@ -40,8 +40,14 @@ The following documents are available ([Release Notes](misc/RELEASE_NOTES.md)):
 | MQTT Protocol Binding         | [v1.0.2](https://github.com/cloudevents/spec/blob/v1.0.2/mqtt-protocol-binding.md)  |    [WIP](https://github.com/cloudevents/spec/blob/main/cloudevents/bindings/mqtt-protocol-binding.md)    |
 | NATS Protocol Binding         | [v1.0.2](https://github.com/cloudevents/spec/blob/v1.0.2/nats-protocol-binding.md)  |    [WIP](https://github.com/cloudevents/spec/blob/main/cloudevents/bindings/nats-protocol-binding.md)    |
 | WebSockets Protocol Binding   |                                        -                                        | [WIP](https://github.com/cloudevents/spec/blob/main/cloudevents/bindings/websockets-protocol-binding.md) |
+<<<<<<< HEAD
```
duglin, line 43: I think you push while in the middle of a rebase

cloudevents/working-drafts/xml-format.md
duglin: should it be "v1" or "V2"? Feels a bit odd to have everything else be lowercase

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,367 @@
...
+| :-----------------| :----------------| :--------------- | :---- |
+| Boolean       | ce:boolean | [xs:boolean][xml-primitives] | `true` or `false` |
+| Integer       | ce:int | [xs:int][xml-primitives] | |
+| String        | ce:string | [xs:string][xml-primitives] | |
+| Binary        | ce:binary | [xs:base64Binary][xml-primitives] |  |
+| URI           | ce:uri | [xs:anyURI][xml-primitives] following [RFC 3986][rfc3986]| |
+| URI-reference | ce:uriRef | [xs:anyURI][xml-primitives] following [RFC 3986][rfc3986] | |
+| Timestamp     | ce:timestamp | [xs:dateTime][xml-primitives] following [RFC 3339][rfc3339] | |
+
+Each CloudEvent context attribute is represented as an XML element whose local
```
duglin, line 79: s/is/MUST be/

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,367 @@
...
+name exactly matches that of the attribute.
+
+See the [envelope](#4-envelope) for special handing of the `specversion` context attribute.
+
+Extension attributes MUST be decorated with the appropriate CloudEvent format
+designators using using the `xsi:type` XML attribute, this allows them to be exchanged
```
duglin, line 85: s/using using/using/

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,367 @@
...
+without loss of type information.
+
+REQUIRED and OPTIONAL context attributes SHOULD NOT be decorated with an `xsi:type`.
```
duglin, line 88: perhaps add ", as to avoid potential type-conflict inconsistencies" to the end? I think we should say something about why we're making this rule when I think most people might add the xsi:type for consistency.

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,367 @@
...
+
+No other XML element attributes are expected, if present they MUST be ignored during
```
duglin, line 90: I'm confused by the phrase "no other". We have spec-defined and we have "extensions", what else is there? Anything not defined by the spec must be an extension and ignoring it doesn't seem right.

jskeet: Spec-defined and extensions are *CloudEvent* attributes. These are *XML* attributes.

So an attribute like this:

```xml
<ce:event>
  <ce:id inquisition="Spanish" />
</ce:event>
```

No-one expects that. Personally I think they should cause a parsing error though.

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,367 @@
...
+    <myboolean xsi:type="ce:boolean">false</myboolean>
+```
+
+## 3. Data
+
+The data portion of a CloudEvent follows a similar model to that employed by
+the [JSON Format specification][json-format]. A `data` element MUST be used to
+encapsulate the payload.
+
+An `xsi:type` is used to discrimate the payload type and MUST be present.
```
duglin, line 109: I don't have a strong feeling about this but I thought I'd ask... in JSON we allow for a receiver to infer that a non-type-specified Data is JSON. Should we allow for a non-type-specified data default to XML too?

If we're going to require xsi:type on "data", which is a spec defined attribute, I'm struggling with the inconsistency above where we ban xsi:type on the other spec defined ones. I understand that "data" is special, but the inconsistency is hard to justify. Couldn't we just say that using an xsi:type that is inconsistent with the spec's definition of those attributes is an error?

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,367 @@
...
+
+The following data representations are supported:
+
+### 3.1 Binary Data
+
+Binary data MUST be carried in an element with an defined type of `ce:binary`.
```
duglin, line 115: While the "types" table above says it needs to be base64 encoded, it might be good to remind people of this:
```
Note, the data MUST be base64 encoded per the ce:binary definition in the #attributes section. 
```

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,367 @@
...
+
+Example:
+
+``` xml
+<data xsi:type="ce:binary">.........</data>
+```
+
+### 3.2 Text Data
+
+Text MUST be carried in an element with an defined type of `ce:string`.
```
duglin, line 125: Do you think we need to mention that it needs to be xml-escaped?

cloudevents/working-drafts/xml-format.md
duglin: the phrase "child business data" might be hard for some people to parse. Perhaps just "The child XML elements MUST NOT...".

However, I actually do think the CE namespace is ok if the data is a nested structured CloudEvent, no?

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,367 @@
...
+       ....
+    </event>
+    <event specversion="1.0">
+       ....
+    </event>
+    ....
+</batch>
+```
+
+An example of an empty batch of CloudEvents (typically used in a response, but also valid in a request):
```
duglin, line 204: I'd drop the comment about responses since events don't have responses and I don't think we want to get into that rat-hole.

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,367 @@
...
+
+```xml
+<batch />
+```
+
+## 6. Examples
```
duglin, line 210: I really like all of the examples!

cloudevents/working-drafts/xml-format.md
duglin: Ah, i totally misread that as just "XML elements" - thanks

JemDay: @jskeet - I got the Monty Python reference there .. well played sir.

duglin:
rebase is needed

cloudevents/working-drafts/xml-format.md
JemDay: I would hope not, but i can add that clarification - it does sound a bit redundant though as if the escaping hasn't happened then the overall XML document wouldn't be valid and wouldn't parse.

cloudevents/working-drafts/xml-format.md
JemDay: on the fence as to the need for that.

cloudevents/working-drafts/xml-format.md
jskeet: I agree with Jem on this one. I think "it should be valid XML" can be assumed. (And indeed I've taken that approach to testing my implementation: the tests should be of the formatter code, not the XML parser.)

cloudevents/working-drafts/xml-format.md
JemDay: If the user wishes to infer that it's fine, but i don't see how that impacts the format specification given the purpose is to support the reliable exchange of a CE.

'data' in this situation is-not a CloudEvent attribute, it's a format specific  construct.

In the early dsays of this spec i had separate XML element names for the different data-content types, that was changed to this more XML idiomatic way of specifying type information.

cloudevents/working-drafts/xml-format.md
JemDay: Ack

cloudevents/working-drafts/xml-format.md
JemDay: ack

README.md
JemDay: hmm

JemDay: i'll work on this one ..

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,367 @@
...
+
+## 2. Attributes
+
+The CloudEvents type system is mapped to a set of CloudEvent specific
+type designators as follows :
+
+| CloudEvents Type  |  XML Format Type | XML Schema Type | Notes |
+| :-----------------| :----------------| :--------------- | :---- |
+| Boolean       | ce:boolean | [xs:boolean][xml-primitives] | `true` or `false` |
+| Integer       | ce:int | [xs:int][xml-primitives] | |
```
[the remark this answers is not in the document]
JemDay: ack

cloudevents/working-drafts/xml-format.md
jskeet: I think what @duglin was potentially suggesting was *something* like:

> If a CloudEvent doesn't specify a `datacontenttype` attribute, but has a `ce:data` element with an `xsi:type` of `xs:any`, the event should be processed as if it had a `datacontenttype` of `text/xml`

Or it's possible that I misunderstood entirely :)

jskeet: I think what @duglin was potentially suggesting was *something* like:

> If a CloudEvent doesn't specify a `datacontenttype` attribute, but has a `ce:data` element with an `xsi:type` of `xs:any`, the event should be processed as if it had a `datacontenttype` of `text/xml`

Or it's possible that I misunderstood entirely :)

cloudevents/working-drafts/xml-format.md
JemDay: I *will* relax this language - i'm not sure how we ended up here but having this restriction means that you can't (very easily) wrap an XML CloudEvent inside another XML CloudEvent.

This "Russian Doll" model is a useful construct to support.

cloudevents/working-drafts/xml-format.md
duglin: yes that was part of what I was trying to say.

the other part is that I'm not sure why we're banning xsi:type on spec defined attributes. Seems like an unnecessary restriction, especially for folks who like to be explicit about things

cloudevents/working-drafts/xml-format.md
duglin: Just didn't want people to miss that link between the two spots in the spec.

duglin:
If you rebase the checker should be happy now

cloudevents/working-drafts/xml-format.md
duglin: Is there a reason we don't want to allow for extensions (e.g. other attributes on the xml elements)? There's no guarantee that the receiver will understand them, but extensions are really useful in general

jskeet: I think it depends on what the expectations are. Personally I *tend* to favour really strict specs, resulting in a relatively low but very-common denominator, rather than a more open-ended spec which leaves a lot of room for some implementations to do one thing and others to do another, resulting in a feature which no-one can use confidently.

duglin: Just seems like a shame for someone to not be allowed to use a feature of the format just because they're sending a CloudEvent. I view this along the lines of a Kube based project saying "we're not going to allow people to add new annotations and labels to the CR for the CRD that we define". It makes it harder for third parties to build on top of what's there w/o a lot of extra work and sync'ing. 

All I'm looking for is to replace the "MUST" here with something like:
```
This specification allows for additional non-CloudEvents namespaced XML elements and XML element
attributes to appear. However, this specification makes no requirement on receivers to understand
those extensions.
```
Note: I have the same feelings about extension elements. This allows people to build on top of our stuff w/o a lot of pain - and, what's interesting to me, is that this actually gives people an incentive to use XML since I'm not sure what other formats give them this level of flexibility.

First we resurrect XML, then we bring back SOAP! :-)

jskeet: Without extension attributes existing, I'd agree. But they *do* exist for this purpose - they're the way we have extensibility within some guardrails. I'd suggest the incentive to use XML should be more about fitting in with other bits of infrastructure (e.g. including CloudEvents within a document that might contain other information) rather than relying on additional capabilities.

We could make this strictness opt-out, in terms of "processors MUST fail by default if unexpected attributes, elements or text nodes are present, but MAY provide mechanisms of parsing less strictly".

cloudevents/working-drafts/xml-format.md
jskeet: > the other part is that I'm not sure why we're banning xsi:type on spec defined attributes. Seems like an unnecessary restriction, especially for folks who like to be explicit about things

Agreed - my implementation currently allows this, but requires that the xsi:type matches the one in the spec. (You can't have an event with an ID that happens to be "true" and declare it with `<ce:id xsi:type="ce:boolean">true</ce:id>` for example.)

cloudevents/working-drafts/xml-format.md
duglin: I probably wasn't clear. I meant there's no easy way to add extra metadata to each CE extension property in formats like structured JSON w/o doing something funky like add another CE extension. With XML you can add that metadata directly on the new CE extension xml element as an xml attributes. And that's pretty cool/handy.

Likewise, structured JSON doesn't allow you to add extra properties w/o them being interpreted as CE extension properties. But you can with XML via namespaces. Again, cool. Worth pointing out that HTTP CE binary mode does let you do this via non CE- prefixed HTTP headers.

w.r.t. text, what about:
```
Receivers MAY reject messages with extra XML element attributes or XML elements
if they choose to do so.
```
I wouldn't want to default to failing since that's a high bar to force everything look for extra things rather than blindly ignore them.

cloudevents/working-drafts/xml-format.md
duglin: Agreed

cloudevents/working-drafts/xml-format.md
jskeet: > I wouldn't want to default to failing since that's a high bar to force everything look for extra things rather than blindly ignore them.

I'd argue it gives the best developer experience though - if your CloudEvent is parseable with one SDK, it should be parseable by all, on the default settings.

I've never been comfortable with Postel's Law - my experience is that that's the route to everyone having to work around laxness in other systems, because forcing people to actually do the right thing is too hard. ("But it works on my machine.")

Something to discuss in the next WG meeting?

cloudevents/working-drafts/xml-format.md
JemDay: I an going to tweak this to be xs:base64Binary, the types of 'data' use the xs types; where as the types of attributes use the ce types.

Why i hear you ask ... well, i just (re)discovered that xs:string is not the same as a ce:string ;-)

cloudevents/working-drafts/xml-format.md
JemDay: OK ... 

Will change the context-attribute commentary to reflect the fact that xsi:type MAY be present for non-extension attributes, if present the xi:type MUST match that of the expected type for the CE attribute.

I'm not a fan of coercing the data on reception, imho datacontenttype is only used during serialization, or deserialization you either get (1) Binary, (2) Text, or (3) XML; anything other than that is for a higher-order function to decide.

cloudevents/working-drafts/xml-format.md
JemDay: This sounds like a discussion item.

cloudevents/working-drafts/xml-format.md
JemDay: Update text clarifies this I believe

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,360 @@
...
+data type mappings for CloudEvents context attributes.
+
+* The [Data](#3-data) section defines the container for the data portion of a
+CloudEvent.
+
+* The [Envelope](#4-envelope) section defines an XML element for CloudEvents
+attributes and an associated media type.
+
+* The [Batch](#5-xml-batch-format) section describes how multiple CloudEvents
+can be packaged into a single XML document.
```
duglin, line 38: @JemDay your comment in the previous (resolved) comment seemed to indicate you wanted to switch to 'element'. Did you change your mind?

cloudevents/working-drafts/xml-format.md
duglin: @JemDay ^^

cloudevents/working-drafts/xml-format.md
JemDay: I believe this is addressed now.

cloudevents/working-drafts/xml-format.md
JemDay: Spec has been updated to reflect `ce` types for the context attributes and `xs` for the data content.

spec.xsd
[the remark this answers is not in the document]
JemDay: resolving old & outdated comments.

xml-format.md
JemDay: I believe this should be resolved ... the comment is over a year old so if still a concern we can re-create it.

cloudevents/working-drafts/xml-format.md
JemDay: That change was recently committed to the specification.

cloudevents/working-drafts/xml-format.md
JemDay: I probably missed that one .. I'll flip it to `element`.

This has the potential to be a little "trippy" .. you can create a `<batch>..</batch>` and put it in the `data` of another CloudEvent .. in fact you could nest them many times, just like a nesting doll ;-)

cloudevents/working-drafts/xml-format.md
```diff
@@ -0,0 +1,360 @@
...
+name exactly matches that of the attribute.
+
+See the [envelope](#4-envelope) for special handing of the `specversion` context attribute.
+
+Extension context attributes MUST be decorated with the appropriate CloudEvent type format
+designators using an `xsi:type` XML attribute, this allows them to be exchanged
+without loss of type information.
+
+REQUIRED and OPTIONAL context atrtribute MAY be decorated with `xsi:type`, if present this
+value MUST match that of the type specified by the [CloudEvent type specification][ce-types].
```
duglin, line 89: I think you meant to reference the type definition per attribute, not the types table. As currently worded, someone could interpret it as "it must be one of our types", but not necessarily "the type defined for THAT attribute per the spec". Perhaps:
- - -
if present this value MUST match that of the type specified for the specific context attribute in the [CloudEvents content attributes](#context-attributes) section of the specification.
- - -

duglin: @JemDay ^

JemDay: Ack

PR #801, opened 2021-04-01
Add support for protobuf batch format

### LINKED ISSUE #794: Should we introduce a protobuf batch representation?
It feels like it would be trivial to add a batch representation to https://github.com/cloudevents/spec/blob/v1.0.1/protobuf-format.md, and if we get that done *before* the event format is widely implemented, it's more likely to end up with widespread support.

I'd expect this to literally be a matter of adding a batch content type, and then a message like this:

```proto
message CloudEventBatch {
  repeated CloudEvent events = 1;
}
```

I'm happy to create a PR if that would be useful. (How does that affect versioning of the document within GitHub? It's not clear to me how event formats are versioned...)

JemDay:
feat: Add specification for protobuf batch format.

Extend protobuf format specification to include a batch format for protobuf. 

Closes #794 

### DIFF
--- a/protobuf-format.md
+++ b/protobuf-format.md
@@ -19,7 +19,8 @@ This document is a working draft.
 2. [Attributes](#2-attributes)
 3. [Data](#3-data)
 4. [Transport](#4-transport)
-5. [Examples](#5-examples)
+5. [Batch Format](#5-batch-format)
+6. [Examples](#6-examples)
 
 ## 1. Introduction
 
@@ -128,8 +129,6 @@ oneof data {
 * When the type of the data is binary the value MUST be stored in the `binary_data` property.
   * `datacontenttype` SHOULD be populated with the appropriate media-type.
 
-
-
 ## 4. Transport
 
 Transports that support content identification MUST use the following designation:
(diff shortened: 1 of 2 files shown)


protobuf-format.md
```diff
@@ -138,12 +137,39 @@ Transports that support content identification MUST use the following designatio
...
+of the batch representation. The batch format MUST only be used where supported.
+
+### 5.1 Envelope
+
+The enveloping container is a _CloudEventBatch_ protobuf message containing a
+repeating set of _CloudEvent_ message(s):
+
+```proto
+message CloudEventBatch {
+  repeated CloudEvent event = 1;
```
jskeet, line 155:
```suggestion
  repeated CloudEvent events = 1;
```

Explanation: while Google style used to suggest using a singular name even for repeated fields, these days (certainly for public protos) we use plural names.

spec.proto
```diff
@@ -56,4 +56,12 @@ message CloudEvent {
...
     }
   }
+}
+
+/**
+ * CloudEvent Protobuf Batch Format
+ */
+
+message CloudEventBatch {
+  repeated CloudEvent event = 1;
```
jskeet, line 66:
```suggestion
  repeated CloudEvent events = 1;
```

jskeet (approved):
One minor nit in two places, but otherwise this is exactly as I'd expected - thanks :)

protobuf-format.md
```diff
@@ -138,12 +137,40 @@ Transports that support content identification MUST use the following designatio
...
 
-## 5. Examples
+## 5. Batch Format
+
+Batch format allows for a set of CloudEvents to be represented, no relationship
+between those events is implied.
+
+Although the _protobuf batch format_ builds on the _protobuf format_ it is considered
+seperate, that is to say that support of _protobuf format_ does not indicate support
+of the batch representation. The batch format MUST only be used where supported.
```
duglin, line 147: I'm not sure how to interpret that MUST. Is that a statement for the sender or receiver? How does the sender know?

JemDay: There's similar language in the JSON format specification .. the essence is that just because somebody claims to support _cloudevents+json_ does NOT imply that they also support _cloudevents-batch+json_.

I can tweak the language a bit more if desired, or copy-paste form the JSON spec ;-)

duglin:
Approved on the 4/8 call

PR #803, opened 2021-04-06
[Subscriptions API] Multiple dialects in the filter object

slinkydeveloper:
This PR modifies the shape of the `filter` subscription request attribute to an array of filter expressions:

This:

```yaml
filter:
  exact: ...
```

Becomes this:

```yaml
filters:
  - exact: ...
```

The reason to introduce this change is to improve ergonomics of the filter usage in most use cases, where at the top level the user wants to do an and of several filters. In other words this:

```yaml
filter:
  all:
    - exact: ...
    - prefix: ...
```

Can be written as:

```yaml
filters:
  - exact: ...
  - prefix: ...
```

It also renames `filter` to `filters`, in order to be more consistent with the change itself.

The openapi schema change is going to be introduced in another PR.

### DIFF
--- a/subscriptions-api.md
+++ b/subscriptions-api.md
@@ -358,7 +358,7 @@ Each subscription is represented by an object that has the following properties:
 - Description: An array of filter expressions that evaluates to true or false. 
   If any filter expression in the array evaluates to false, the event
   MUST NOT be sent to the sink. If all the filter expressions in the array evaluates to true, the event
-  MUST be attempted to be delivered. Absence of a filter implies a value
+  MUST be attempted to be delivered. Absence of a filter or empty array implies a value
   of true.
 
   Each filter dialect MUST have a name that is unique within the scope of the
@@ -372,7 +372,6 @@ Each subscription is represented by an object that has the following properties:
 
 - Constraints:
   - OPTIONAL for both subscription managers and subscribers to support
-  - If present, the array MUST have at least one filter expression
 - Examples:
   - "prefix": [ "type": "com.github.issue" ]
 
--- a/subscriptions-api.md
+++ b/subscriptions-api.md
@@ -356,7 +356,7 @@ Each subscription is represented by an object that has the following properties:
 ##### filters
 - Type: `Array of Objects`
 - Description: An array of filter expressions that evaluates to true or false. 
-  If a filter expression in the array evaluates to false, the event
+  If any filter expression in the array evaluates to false, the event
   MUST NOT be sent to the sink. If all the filter expressions in the array evaluates to true, the event
   MUST be attempted to be delivered. Absence of a filter implies a value
   of true.
--- a/subscriptions-api.md
+++ b/subscriptions-api.md
@@ -277,9 +277,9 @@ Subscription:
     "[key]": [subscription manager specific value], *
   },
 
-  "filter": { ?
-    "[dialect name]": [dialect specific object] ?
-  },
+  "filters": [ ?
+    { "[dialect name]": [dialect specific object] } +
+  ],
 
   "sink": "[URI to where events are delivered]",
   "protocol": "[delivery protocol]",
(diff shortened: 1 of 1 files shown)


subscriptions-api.md
```diff
@@ -608,20 +608,25 @@ settings. All other settings SHOULD be supported.
...
-defines the type of filter and the set of additional properties that are
-allowed within the filter expression. If a filter dialect is specified in a
-subscription that is unsupported by the subscription manager, creation or
-update of the subscription MUST be rejected with an error.
+Filters allow for subscriptions to specify that only a subset of events are to
+be delivered to the sink based on a set of criteria. The `filter` property in a
+subscription is a filter expressions that evaluates to either true or false for
+each event generated.
+
+If the filter expression evalutes to true, the event MUST be attempted to be
```
duglin, line 616: s/evalutes/evaluates/

duglin:
No strong opinion but I do suspect AND will be one of the more popular dialects.

duglin:
rebase needed

duglin:
we'd need to change the schema too, from:
```
  "filter": { ?
    "[dialect name]": [dialect specific object] ?
  },
```
to
```
  "filter": { ?
    "[dialect name]": [dialect specific object] *
  },
```
right?

slinkydeveloper:
Schema fixed

duglin:
change here too: https://github.com/cloudevents/spec/blob/master/subscriptions-api.md#321-subscription-object

slinkydeveloper:
I've removed the openapi schema change as discussed during the meeting

slinkydeveloper:
After discussing with @clemensv, I've changed this PR completely. I've updated the description of the PR accordingly

subscriptions-api.md
```diff
@@ -353,12 +353,11 @@ Each subscription is represented by an object that has the following properties:
...
-##### filter
-- Type: `Object`
-- Description: A filter expression that evaluates to true or false and that
-  determines whether an instance of a CloudEvent will be delivered to the
-  subscription's sink. If a filter expression evaluates to false, the event
-  MUST NOT be sent to the sink. If the expression evaluates to ttue, the event
+##### filters
+- Type: `Array of Objects`
+- Description: An array of filter expressions that evaluates to true or false. 
+  If a filter expression in the array evaluates to false, the event
```
duglin, line 359: s/If a/If any/

subscriptions-api.md
```diff
@@ -373,6 +372,7 @@ Each subscription is represented by an object that has the following properties:
 
 - Constraints:
   - OPTIONAL for both subscription managers and subscribers to support
+  - If present, the array MUST have at least one filter expression
```
duglin, line 375: I can't recall the exact topic, but didn't we have a concern where the diff between "empty" and "missing" was one that some implementations couldn't handle? I bring this up because I'm wondering if we really should ban a zero-length filter array?

slinkydeveloper: Maybe for this special case, we can have an empty one? I have no strong arguments about it.

duglin: I'd prefer to allow for an empty array since someone will do it and I'm not sure why we would ban it. It seems unnecessarily strict.

subscriptions-api.md
```diff
@@ -453,11 +453,9 @@ Below is an example JSON serialization of a subscription resource:
     "interval": 5
   },
 
-  "filter": {
-    "prefix": {
-      "type": "com.example."
-    }
-  },
+  "filters": [
+    { "prefix": { "type": "com.example." } }
```
duglin, line 457: Do we anticipate more than just `"type": { filter object }` in there? If not, would it make sense to make it:
```
"filters": [
  "prefix": { "type": "com.example." },
  "suffix": { "type": "foo" }
]
```
IOW, do we need the extra {}'s ?

duglin:
Just a few minor syntax questions, otherwise SGTM

subscriptions-api.md
slinkydeveloper: @duglin now it should be good to go, check the last commit

duglin:
Approved on the 5/20 call

slinkydeveloper:
I'm going to push the followup change to the openapi spec soon

PR #804, opened 2021-04-07
[Subscriptions API] Remove strong constraints on filter combinators

slinkydeveloper:
The `all` and `any` combinators enforces too strict constraints on the actual evaluation of the filter. I propose to remove them, since they may impede any eventual optimization in the evaluation of filters

### DIFF
--- a/subscriptions-api.md
+++ b/subscriptions-api.md
@@ -693,8 +693,7 @@ For example:
 
 Use of this MUST include a nested array of filter expressions, where all
 nested filter expressions MUST evaluate to true in order for the `all`
-filter expression to be true. The order in which the nested expressions are
-are listed is the order in which they MUST be evaluated.
+filter expression to be true.
 
 Note: there MUST be at least one filter expression in the array.
 
@@ -712,9 +711,7 @@ For example:
 
 Use of this MUST include one nested array of filter expressions, where at least
 one nested filter expressions MUST evaluate to true in order for the `any`
-filter expression to be true. The order in which the nested expressions are
-are listed is the order in which they MUST be evaluated, and processing MUST
-stop on the first successful match.
+filter expression to be true.
 
 Note: there MUST be at least one filter expression in the array.
 


duglin:
Can there ever be a case in where evaluating the 2nd part of an AND could result in an error if the first part of the AND was 'false' ? In other words, we would need to ensure we don't evaluate the 2nd before the first.  E.g. `if a != nil && a.foo ...` to avoid a NPE.  Perhaps something like:  `type != "" AND type.substring(5) == "foo"`   ?

slinkydeveloper:
> Can there ever be a case in where evaluating the 2nd part of an AND could result in an error if the first part of the AND was 'false' ? 

@duglin the definition of the filters in this document is that they're infallible, they either return true or false, but never a third "fail" state.

I also don't see how evaluating one part of the expression should influence the other. The filters shouldn't generate side effects. If they do, either they're not filters or the implementation is somehow broken.

> E.g. if a != nil && a.foo ... to avoid a NPE

I think this is not a good example, because of the above comment: filter dialects are infallible in this definition, they either return true or false. So the `a.foo.fn()` cannot throw NPE in this world

duglin:
All - Please add examples that would fail if we adopted this PR

linuxbasic:
One situation that could fail, is when you have a side-effect (e.g. network call) in your filter that changes the state of an external system and you rely on the shortcutting mentioned in the call (filter does not get executed if the previous filter returns false).

But I have a few reasons why we should optimize for this case:
1. In my opinion is causing side effects (in the sense of changing external state) during a filter evaluation a big no-no. We should discourage this pattern.
2. I don't see the advantage of taking the optimization of the filter evaluation out of the hands of the implementation and putting it into the hands of the API-User.
3. If we want the shortcutting to be a thing, we must specify it.

slinkydeveloper:
> In my opinion is causing side effects (in the sense of changing external state) during a filter evaluation a big no-no. We should discourage this pattern.

+1, that's the point of this PR. side effects, if necessary, should live inside the dialects itself, where you can express complex predicates using sql or other expression languages

duglin:
Approved on the 4/29 call

PR #806, opened 2021-04-13
[CESQL] TCK

slinkydeveloper:
I've extracted from the reference implementation I'm working on (https://github.com/cloudevents/sdk-java/pull/363) the test cases and I've built this TCK, testing all the basic features of the language (~200 test cases).

The README contains the details on how to use it.


duglin:
If we adopt #808 then I think this should go under the cesql dir and renamed to just "tck"

duglin:
approved on the 4/22 call

PR #808, opened 2021-04-15
chore: first pass at dir restructure

duglin:
I think the README needs more work to talk about the various specs but this at least this moves things into the new locations.


duglin:
The dir structure becomes this:
```
./schemaregistry
./share
./subscriptions
./discovery
./artwork
./pagination
./cesql
./community
./tools
./cloudevents
./cloudevents/extensions
./cloudevents/adapters
```

Should we create:
```
./cloudevents/bindings
./cloudevents/formats
```
too?

slinkydeveloper:
I think you didn't rebased, can you move `cesql_tck` under `cesql/tck`?

README.md
```diff
@@ -49,8 +49,9 @@ The following documents are available:
 | Proprietary Specifications    |                                        -                                        |      [master](https://github.com/cloudevents/spec/blob/master/proprietary-specs.md)      |
```
slinkydeveloper, line 49: All the links in this file seems broken

jskeet (commented):
It looks like I can't add comments to files that are just moved, so...

- Pagination: it's even clearer under this new structure that pagination is a bit of an orphan. I'd expect pagination to be specified by any specific API (e.g. a subscriptions API) rather than being on its own like this. I'd like to talk about this one more.

- It feels like we have several non-spec things... I wonder whether they could either go in a separate repository, or be bundled into a single directory:
  - Share
  - Artwork
  - Tools
  - (.github - but that's pretty much enforced in its location)

At the last meeting we started considering the idea of versioning being applied to bundles of specs, and it looks like the direction of this PR is each top-level directory being a bundle - is that correct? If so, that's really useful for my still-to-be-done work on proposing a specific versioning scheme. As I think you (Doug) have said before, we could potentially *consider* splitting each bundle into its own repo.

JemDay:
+1 to having a separate directory for `bindings` & `formats`, I will admit it feels a bit weird to have a `cloudevents` directory but I can live with that ..

duglin:
closing in favor of:  #893

PR #816, opened 2021-04-22
Clarify HTTP header value encoding and decoding requirements (#793)

### LINKED ISSUE #777: Clarify HTTP header encoding/decoding rules

jskeet:
* fix: Clarify HTTP header value encoding and decoding requirements

Fixes #777

Once this is agreed, we should also document conformance tests, but
it's unclear at the moment where they should live.

### DIFF
--- a/http-protocol-binding.md
+++ b/http-protocol-binding.md
@@ -213,16 +213,75 @@ and per [RFC7230, section 3][rfc7230-section-3], HTTP headers MUST only use
 printable characters from the US-ASCII character set, and are terminated by a
 CRLF sequence with OPTIONAL whitespace around the header value.
 
-String values MUST be percent-encoded as described in [RFC3986, section
-2.4][rfc3986-section-2-4] before applying the header encoding rules described in
-[RFC7230, section 3.2.6][rfc7230-section-3-2-6].
+When encoding a CloudEvent as an HTTP message, string values
+represented as HTTP header values MUST be percent-encoded as
+described below. This is compatible with [RFC3986, section
+2.1][rfc3986-section-2-1] but is more specific about what needs
+encoding. The resulting string SHOULD NOT be further encoded.
+(Rationale: quoted string escaping is unnecessary when every space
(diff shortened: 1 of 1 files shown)


jskeet:
Note: this is #793, but against the `master` branch.

duglin:
approved on 4/22

PR #820, opened 2021-04-29
[CESQL] Expand like test cases

slinkydeveloper:
Added 2 test cases that I've found out to fail while testing the reference implementation in sdk-java

### DIFF
--- a/cesql_tck/like_expression.yaml
+++ b/cesql_tck/like_expression.yaml
@@ -1,7 +1,10 @@
 name: Like expression
 tests:
-  - name: Exact match
+  - name: Exact match (1)
     expression: "'abc' LIKE 'abc'"
+    result: true
+  - name: Exact match (2)
+    expression: "'ab\\c' LIKE 'ab\\c'"
     result: true
   - name: Exact match (negate)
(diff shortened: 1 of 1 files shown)


cesql_tck/like_expression.yaml
```diff
@@ -25,6 +28,12 @@ tests:
   - name: Percentage operator (6)
     expression: "'' LIKE 'abc'"
     result: false
+  - name: Percentage operator (7)
+    expression: "'.ab.cde.' LIKE '.%.%.'"
+    result: true
+  - name: Percentage operator (8)
+    expression: "'ab.cde' LIKE '.%.%.'"
```
duglin, line 35: I'm curious, are you afraid that someone will treat "." as a special char?

duglin:
LGTM

cesql_tck/like_expression.yaml
slinkydeveloper: I am, because I was converting this to a regular expression in the ref implementation without escaping the dot :laughing:

duglin:
Approved on the 5/20 call

PR #821, opened 2021-04-29
Adding operation ids to discovery contract

clemensv:
Several tools are unhappy with missing operation ids in OpenAPI and will generate rather unpleasant names for proxies.


deissnerk:
LGTM

duglin:
Approved on the 5/20 call

PR #827, opened 2021-06-01
[Subscriptions API] CESQL filter

slinkydeveloper:
Added the `sql` filter dialect to subscriptions api filters.

### DIFF
--- a/subscriptions-api.md
+++ b/subscriptions-api.md
@@ -739,8 +739,9 @@ For example:
 ###### `sql` filter dialect
 
 Use of this MUST have a string value, representing a [CloudEvents SQL Expression](./expression-language.md).
-The filter result MUST be false if an error occurred while evaluating the expression or if the result value, 
-coerced to boolean, equals to the `FALSE` boolean value, otherwise the filter result MUST be true.
+The filter result MUST be true if the result value of the expression, coerced to boolean, equals to the `TRUE` boolean value,
+otherwise MUST be false if an error occurred while evaluating the expression or if the result value, 
+coerced to boolean, equals to the `FALSE` boolean value.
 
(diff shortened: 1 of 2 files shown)


subscriptions-api.md
```diff
@@ -736,6 +736,19 @@ For example:
 }
 ```
 
+###### `sql` filter dialect
+
+Use of this MUST have a string value, representing a [CloudEvents SQL Expression](./expression-language.md).
+The filter result MUST be false if an error occurred while evaluating the expression or if the result value, 
+coerced to boolean, equals to the `FALSE` boolean value, otherwise the filter result MUST be true.
```
duglin, line 743: This is ok, but it just felt weird to call out the cases where it's "false" and then say "everything else is 'true' ".
Normally I would expected it to say "it's true if the expression evaluates to true, and all other cases evaluate to false".  But I guess either way works. :-)

duglin:
LGTM

subscriptions-api.md
slinkydeveloper: I wanted to be explicit about the fact that if there's an error, then the result is false. But I can swap the sentence if you want to.

duglin: we chatted about this on the call today and overall people are ok with the PR, but would prefer if you turned it around and made it more like:   if it valuates to 'true' then the result is 'true', other all other cases (including an error) results in 'false' (in your own words).

slinkydeveloper: done, check it out now

duglin:
@slinkydeveloper can you rebase?

duglin:
Approved on the June 3 call - will merge after the rebase

slinkydeveloper:
@duglin done, ready for merging

duglin:
Thanks - merging!

Issue #811, opened 2021-04-17
Webhooks link is broken


raminorujov:
Introduction section of [HTTP Webhooks spec document](https://github.com/cloudevents/spec/blob/v1.0.1/http-webhook.md) contains [a broken link http://progrium.com/blog/2007/05/03/web-hooks-to-revolutionize-the-web/](http://progrium.com/blog/2007/05/03/web-hooks-to-revolutionize-the-web/)

duglin:
thanks!
@clemensv do you know if it's been moved? I wasn't able to find it doing a quick search.
Would pointing to: https://en.wikipedia.org/wiki/Webhook  be sufficient?

duglin:
Lu Dang found it: https://progrium.github.io/blog/2007/05/03/web-hooks-to-revolutionize-the-web/index.html
I'll PR it - thanks for the issue

raminorujov:
Thanks

PR #831, opened 2021-06-16
Fix bad link

duglin:
Fixes: #811


duglin:
Approved on the 6/17 call

PR #832, opened 2021-06-17
Add missing property name

### LINKED ISSUE #667: What's the attribute key for sampled-rate extension?
https://github.com/cloudevents/spec/blob/v1.0/extensions/sampled-rate.md

I see from prior histories of this document that more information was specified at one point in the past - including a "sampledRate" (changed to "sampledrate") key, for which the value was as the current document specified.

However a recent commit deleted all the information about this, leaving the document rather confusing as it specifies the concept of a sampled field with the requisite Value definition, but no key used to define it.

duglin:
Fixes #667

### DIFF
--- a/documented-extensions.md
+++ b/documented-extensions.md
@@ -41,5 +41,5 @@ information.
 - [Dataref (Claim Check Pattern)](extensions/dataref.md)
 - [Distributed Tracing](extensions/distributed-tracing.md)
 - [Partitioning](extensions/partitioning.md)
-- [Sampling](extensions/sampled-rate.md)
+- [Sampling](extensions/sampledrate.md)
 - [Sequence](extensions/sequence.md)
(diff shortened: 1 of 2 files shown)


extensions/sampled-rate.md
```diff
@@ -17,7 +17,9 @@ emitted event represents a given number of other similar events. It also
 provides a place for intermediary transport systems to modify the event when
 they impose additional sampling.
 
-## Value
+## Attributes
+
+### samplerate
```
duglin, line 22: sampledrate here and on line 25 and title of doc

duglin:
Made updates per our call last week. Can I get some eyes on it to make sure I did it right?
@markpeek ??

duglin:
re-approved on 7/1 call

PR #833, opened 2021-06-17
remove commit linter

duglin:


duglin:
executive decision since we're using something new from now on....

PR #836, opened 2021-06-25
Fix for #673 - WebHook-Allowed-Origin

### LINKED ISSUE #673: Webhook-Allowed-Origin instead of Webhook-Request-Origin
For the webook, http-webhook.md - in the example of using a wildcard for the validation response, do we mean Webhook-Allowed-Origin instead of Webhook-Request-Origin?

deissnerk:
As @schabungbam pointed out, there was an inconsistency in one example that is fixed with this PR.
@duglin Should this also be cherry-picked into 1.0.1, as it is just about a typo?

Signed-off-by: Klaus Deißner <klaus.deissner@sap.com>

Fixes #673

### DIFF
--- a/http-webhook.md
+++ b/http-webhook.md
@@ -310,7 +310,7 @@ WebHook-Allowed-Origin: eventemitter.example.com
 or
 
 ```text
-WebHook-Request-Origin: *
+WebHook-Allowed-Origin: *
 ```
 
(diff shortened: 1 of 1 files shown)


duglin:
LGTM

duglin:
Approved on 7/1 call

Issue #699, opened 2020-09-17
[Discovery API]  Consider adding history


erikerikson:
Resulting from discussion 9/17/2020 about #690 during weekly call.

The notion of soft versus hard deletes of service records was being discussed in the context of #690.  The question came up of how consumers of the API react to either.  In a larger context, if we offered a log of the actions taken to define the service (a stream of CloudEvents) there would be no ambiguity and adding support for stream processing would benefit users of such architectures.  That log could be compacted (or not) as Clemens noted.  Ensuring the log is strictly ordered according to observation by the node would support consistency.  Access to the materialization of that log would remain in the specification and could be based on the events, although that would likely be an implementation detail.

duglin:
@erikerikson, or any one else, is this something we'd like to more strongly consider? Should someone take an AI to open a PR? Or is this still in a "thinking about it" state?

erikerikson:
Certainly that would be my preference but being involved rather than committed I'm reticent to exert my preference.  I wouldn't say that there isn't a significant community that shares the preference on the other hand.

erikerikson:
via #868

PR #837, opened 2021-07-06
Clean API spec/sync OpenAPI doc

erikerikson:
Prepares for #699

## Proposed Changes

- clean up the API specification
  - remove duplication
  - follow REST best practices
    - use filters over collections
    - avoid using query params to alter HTTP semantics
- synchronize OpenAPI declaration with Specification

**Release Note**

```release-note
refactor(discovery)!: Clean API spec/sync OpenAPI doc
```

### DIFF
--- a/ce_discovery.yaml
+++ b/ce_discovery.yaml
@@ -12,77 +12,11 @@ paths:
       description: Get an optionally filtered collection of the Services known by this discovery end point
       parameters:
        - in: query
-         name: id
-         description: The id of the service to be returned
-         required: false
-         schema:
-           $ref: '#/components/schemas/service/properties/id'
-       - in: query
-         name: epoch
-         description: The epoch of the service to be returned
-         required: false
-         schema:
-           $ref: '#/components/schemas/service/properties/epoch'
-       - in: query
          name: name
          description: The name of the service to be returned
          required: false
          schema:
            $ref: '#/components/schemas/service/properties/name'
-       - in: query
-         name: url
-         description: The url of the service to be returned
-         required: false
-         schema:
-           $ref: '#/components/schemas/service/properties/url'
-       - in: query
-         name: description
-         description: The description of the service to be returned
-         required: false
-         schema:
-           $ref: '#/components/schemas/service/properties/description'
-       - in: query
-         name: docsurl
-         description: The docsurl of the service to be returned
-         required: false
-         schema:
-           $ref: '#/components/schemas/service/properties/docsurl'
-       - in: query
-         name: specversions
-         description: The specversions of the service to be returned
-         required: false
-         schema:
-           $ref: '#/components/schemas/service/properties/specversions'
-       - in: query
-         name: subscriptionurl
-         description: The subscriptionurl of the service to be returned
-         required: false
-         schema:
-           $ref: '#/components/schemas/service/properties/subscriptionurl'
-       - in: query
-         name: subscriptionconfig
-         description: The subscriptionconfig of the service to be returned
-         required: false
-         schema:
-           $ref: '#/components/schemas/service/properties/subscriptionconfig'
-       - in: query
-         name: subscriptiondialects
-         description: The subscriptiondialects of the service to be returned
-         required: false
-         schema:
-           $ref: '#/components/schemas/service/properties/subscriptiondialects'
-       - in: query
-         name: authscope
-         description: The authscope of the service to be returned
-         required: false
-         schema:
-           $ref: '#/components/schemas/service/properties/authscope'
-       - in: query
-         name: protocols
-         description: The protocols of the service to be returned
-         required: false
-         schema:
-           $ref: '#/components/schemas/service/properties/protocols'
       responses:
         '200':
           description: A list of services (optionally matching the query parameter)
--- a/ce_discovery.yaml
+++ b/ce_discovery.yaml
@@ -34,11 +34,11 @@ paths:
               $ref: '#/components/schemas/services'
       responses:
         '200':
-          description: A list of Service Instances that reference the updated Services
+          description: A list of the resulting Service values resulting from processing the request, in the same order as in the request
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/serviceinstances'
+                $ref: '#/components/schemas/services'
         '400':
           description: Bad Request - constraint failure
         '409':
@@ -104,7 +104,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/serviceinstance'
+                $ref: '#/components/schemas/service'
         '400':
           description: Bad Request - constraint failure
         '404':
--- a/ce_discovery.yaml
+++ b/ce_discovery.yaml
@@ -9,81 +9,296 @@ paths:
   /services:
     get:
       operationId: getServices
+      description: Get an optionally filtered collection of the Services known by this discovery end point
       parameters:
        - in: query
-         name: matching
-         description: The search term.
+         name: id
+         description: The id of the service to be returned
          required: false
          schema:
-           type: string
+           $ref: '#/components/schemas/service/properties/id'
+       - in: query
+         name: epoch
+         description: The epoch of the service to be returned
+         required: false
+         schema:
+           $ref: '#/components/schemas/service/properties/epoch'
+       - in: query
+         name: name
+         description: The name of the service to be returned
+         required: false
+         schema:
+           $ref: '#/components/schemas/service/properties/name'
+       - in: query
+         name: url
+         description: The url of the service to be returned
+         required: false
+         schema:
+           $ref: '#/components/schemas/service/properties/url'
+       - in: query
+         name: description
+         description: The description of the service to be returned
+         required: false
+         schema:
+           $ref: '#/components/schemas/service/properties/description'
+       - in: query
+         name: docsurl
+         description: The docsurl of the service to be returned
+         required: false
+         schema:
+           $ref: '#/components/schemas/service/properties/docsurl'
+       - in: query
+         name: specversions
+         description: The specversions of the service to be returned
+         required: false
+         schema:
+           $ref: '#/components/schemas/service/properties/specversions'
+       - in: query
+         name: subscriptionurl
+         description: The subscriptionurl of the service to be returned
+         required: false
+         schema:
+           $ref: '#/components/schemas/service/properties/subscriptionurl'
+       - in: query
+         name: subscriptionconfig
+         description: The subscriptionconfig of the service to be returned
+         required: false
+         schema:
+           $ref: '#/components/schemas/service/properties/subscriptionconfig'
+       - in: query
+         name: subscriptiondialects
+         description: The subscriptiondialects of the service to be returned
+         required: false
+         schema:
+           $ref: '#/components/schemas/service/properties/subscriptiondialects'
+       - in: query
+         name: authscope
+         description: The authscope of the service to be returned
+         required: false
+         schema:
+           $ref: '#/components/schemas/service/properties/authscope'
+       - in: query
+         name: protocols
+         description: The protocols of the service to be returned
+         required: false
+         schema:
+           $ref: '#/components/schemas/service/properties/protocols'
       responses:
         '200':
-          description: A list of services (optionally matching the query parameter).
+          description: A list of services (optionally matching the query parameter)
           content:
             application/json:
               schema:
-                type: array
-                items:
-                  $ref: '#/components/schemas/service'
-  /services/{name}:
+                $ref: '#/components/schemas/services'
+    post:
+      operationId: postServices
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/services'
+      responses:
+        '200':
+          description: A list of Service Instances that reference the updated Services
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/serviceinstances'
+        '400':
+          description: Bad Request - constraint failure
+        '409':
+          description: Conflict - epoch not greater
+    delete:
+      operationId: deleteServices
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/serviceinstances'
+      responses:
+        '200':
+          description: A list of the Services that were deleted
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/services'
+        '400':
+          description: Bad Request - constraint failure
+        '404':
+          description: Not Found
+        '409':
+          description: Conflict - epoch not greater
+  /services/{id}:
     get:
       operationId: getService
       parameters:
        - in: path
-         name: name
-         description: The name of the service to be returned.
+         name: id
+         description: The id of the service to be returned
          required: true
          schema:
-           type: string
+           $ref: '#/components/schemas/service/properties/id'
       responses:
         '200':
-          description: The corresponding service.
+          description: The corresponding service
           content:
             application/json:
               schema:
-                type: array
-                items:
-                  $ref: '#/components/schemas/service'
+                $ref: '#/components/schemas/service'
+        '404':
+          description: Not Found
+    post:
+      operationId: postService
+      parameters:
+       - in: path
+         name: id
+         description: The id of the service to be returned
+         required: true
+         schema:
+           $ref: '#/components/schemas/service/properties/id'
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/service'
+      responses:
+        '200':
+          description: A Service Instance referencing the updated Service
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/serviceinstance'
+        '400':
+          description: Bad Request - constraint failure
+        '404':
+          description: Not Found
+        '409':
+          description: Conflict - epoch not greater
+    delete:
+      operationId: deleteService
+      parameters:
+       - in: path
+         name: id
+         description: The id of the service to be returned
+         required: true
+         schema:
+           $ref: '#/components/schemas/service/properties/id'
+      requestBody:
+        required: false
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/serviceinstance'
+      responses:
+        '200':
+          description: The Service that was deleted
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/service'
+        '400':
+          description: Bad Request - constraint failure
+        '404':
+          description: Not Found
+        '409':
+          description: Conflict - epoch not greater
 components:
   schemas:
-    service:
+    eventtype:
+      type: object
+      properties:
+        type:
+          type: string
+          description: CloudEvents type attribute
+          example: com.github.pull.create, com.example.object.delete.v2
+        description:
+          type: string
+          description: Human readable description
+        dataschema:
+          type: string
+          description: CloudEvents datacontenttype attribute. Indicating how the data attribute of subscribed events will be encoded
+        dataschematype:
+          type: string
+          description: If using dataschemacontent for inline schema storage, the dataschematype indicates the type of schema represented there
+        dataschemacontent:
+          type: string
+          description: An inline representation of the schema of the data attribute encoding mechanism. This is an alternative to using the dataschema attribute
+        sourcetemplate:
+          type: string
+          description: A URI Template according to RFC 6570 that defines how the source attribute will be generated
+          example: "http://blob.store/{bucket}"
+        extensions:
+          type: array
+          description: An array or CloudEvents Extension Context Attributes that are used for this event type
+          items:
+            type: object
+            properties:
+              name:
+                type: string
+                description: The CloudEvents context attribute name used by this extension
+              type:
+                type: string
+                description: The data type of the extension attribute
+              specurl:
+                type: string
+                description: An attribute pointing to the specification that defines the extension
+          example: '{ "name": "dataref", "type": "URI-reference", "specurl": "https://github.com/cloudevents/spec/blob/master/extensions/dataref.md" }'
+    eventtypes:
+      type: array
+      items:
+        $ref: '#/components/schemas/eventtype'
+    serviceinstance:
       type: object
       properties:
         id:
           type: string
-          description: A unique identifier for this Service. This value MUST be globally unique.
+          description: A unique identifier for this Service. This value MUST be globally unique
           format: uuid
           example: bf5ff5cc-d059-4c79-a89a-2513e45a1340
         epoch:
           type: integer
-          description: The Discovery Endpoint's epoch value for this Service Entry.
+          description: The Discovery Endpoint's epoch value for this Service Entry
           format: int64
           example: 42
+    serviceinstances:
+      type: array
+      items:
+        $ref: '#/components/schemas/serviceinstance'
+    service:
+      type: object
+      properties:
+        id:
+          $ref: '#/components/schemas/serviceinstance/properties/id'
+        epoch:
+          $ref: '#/components/schemas/serviceinstance/properties/epoch'
         name:
           type: string
-          description: A unique identifier for this Service.
+          description: A unique identifier for this Service
           example: my storage service
         url:
           type: string
-          description: A URL that references this Service. This value MUST be usable in subsequent requests, by authorized clients, to retrieve this Service entity.
+          description: A URL that references this Service. This value MUST be usable in subsequent requests, by authorized clients, to retrieve this Service entity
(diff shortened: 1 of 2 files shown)


discovery.md
```diff
@@ -523,9 +523,10 @@ as if that Service does not exist. For example, it would be excluded from
 any array of Services returned and it would result in a `404 Not Found`
 error for a request to that Service directly.
 
-Unless there is some out-of-band agreement, all APIs MUST use JSON encoding,
-which means when there is an HTTP Body the HTTP `Content-Type` Header MUST
-be `application/json`.
+All APIs MUST support JSON encoding, which means that an HTTP requests
+including an HTTP `Content-Type` Header of `application/json` and body in that
+format MUST be supported.  Likewise requests with `Accept` header of
+`application\json` MUST be supported.
```
erikerikson, line 529: Of course everyone can make agreements to violate the specification and so long as they keep the arrangement that will be fine.

discovery.md
```diff
@@ -538,6 +539,10 @@ This MUST return an array of zero or more Services. The array MUST contain all
 Services available via this Discovery Enpoint. Any Service previously returned
 to a client that does not appear in this result can be assumed to be deleted.
 
+The collection of services MAY be filtered by supplying an attribute as a query
+parameter.  Doing so will cause only Services with an exact match for that
+attribute to be included in the result.
```
erikerikson, line 544: This replaces the `GET /services?name={name}` path with a more general capability that is consistent with RESTful practices.

discovery.md
```diff
@@ -650,30 +630,49 @@ upon return of the `202 Accepted` from the original request.
...
+possible to rename that Service to `YourService` and then to add, or update,
+another Service to use `MyService` as its `name`. But this is only true if
+the rename of the first Service happens first the request message.
 
-If the Discovery Endpoint is unable to successully add all of the Services
-in the incoming request then an error MUST be generated and none of the
-specified Services are to be added to the Discovery Endpoint.
+Likewise, while it might be less than optimal, it is technically possible for
+a request to update a Service multiple times if the same `id` is used more
+than once within a request.
```
erikerikson, line 667: this condenses the version with `?import` into this single endpoint definition to eliminate the multifaced behavior.

discovery.md
```diff
@@ -650,30 +630,49 @@ upon return of the `202 Accepted` from the original request.
...
-  - If there was only one Service in the incoming request, then the response
-    MUST include an HTTP `Location` Header with a value that points to the
-    correspdonding Service.
-  - If the request contained more than one Service, then the HTTP `Location`
-    Header MUST NOT appear in the response.
-  - The HTTP Response Body MUST include an array of `id` values whose order
-    and values MUST match the list of Services in the request.
+- `200 OK` if all the specified Services were processed successfully
+  - The HTTP Response Body MUST include an array of `id` and `epoch` values
+    whose order and values MUST match the list of Services in the request
```
erikerikson, line 672: Perhaps a missing sophistication is that `id` and `epoch` would be included for newly created Services (i.e. those without a preset `id`.  For those which were replacements the old values could be returned.  This would be useful and consistent with the DELETE behavior but imply increased implementation work.

discovery.md
```diff
@@ -650,30 +630,49 @@ upon return of the `202 Accepted` from the original request.
...
+- `400 Bad Request` if the first error encountered is a constraint failure
+- `409 Conflict` if the first error encountered is that a given epoch was less
+  than or equal to the existing epoch of the Service with a given `id`
```
erikerikson, line 675: here, as elsewhere, making the behavior associated with handling errors specified by this document explicit and standard.

erikerikson (commented):
This review is intended to call out the changes and opinions expressed herein explicitly.

Please note that these changes alter the API and as such are breaking.

This review is intended to call out the changes and opinions expressed herein explicitly.  Some changes applied to multiple locations but I only mention the first occurrence.

Spec changes without commentary are nits.

discovery.md
duglin: I think the intent was that we didn't want to force people to support queries across all possible attributes. I may have missed it, but did you define the list of attributes that MUST be supported?

erikerikson: I did not specify a list.  Given the beginning state it seems `name` is the minimum supported set of filters.  I'll clarify, thank you for the gentle reminder of that intention.

erikerikson: Please let me know if [69359c3](https://github.com/cloudevents/spec/pull/837/commits/69359c303c7733345fdde4da1fb78b42a1478058) doesn't fully resolve this.

discovery.md
```diff
@@ -523,9 +523,10 @@ as if that Service does not exist. For example, it would be excluded from
 any array of Services returned and it would result in a `404 Not Found`
 error for a request to that Service directly.
 
-Unless there is some out-of-band agreement, all APIs MUST use JSON encoding,
-which means when there is an HTTP Body the HTTP `Content-Type` Header MUST
-be `application/json`.
+All APIs MUST support JSON encoding, which means that an HTTP requests
```
duglin, line 526: I think the "an" is unneeded

discovery.md
```diff
@@ -523,9 +523,10 @@ as if that Service does not exist. For example, it would be excluded from
...
+including an HTTP `Content-Type` Header of `application/json` and body in that
+format MUST be supported.  Likewise requests with `Accept` header of
```
duglin, line 528: use of the phrase "MUST be supported" here (both of them) is a bit awkward - I think. To me, I'm reading it as "you MUST be able to accept them", but I think you're trying to say "when present, a Content-Type header MUST have a value of application/json when the Body is json", or something like that, no? Same for "Accept".
And it's app/json not app\\json

discovery.md
```diff
@@ -538,6 +539,15 @@ This MUST return an array of zero or more Services. The array MUST contain all
 Services available via this Discovery Enpoint. Any Service previously returned
 to a client that does not appear in this result can be assumed to be deleted.
 
+The collection of services MAY be filtered by supplying an attribute as a query
+parameter.  Doing so will cause only Services with an exact match for that
```
duglin, line 543: I think there needs to be a MUST in here on the DE side so they don't just ignore the filter. But maybe you cover that later on....

discovery.md
```diff
@@ -538,6 +539,15 @@ This MUST return an array of zero or more Services. The array MUST contain all
...
+attribute to be included in the result.
+
+Discovery endpoints MUST support filtering with the following attributes and
+MAY ignore other filters.
```
duglin, line 547: I think I'd prefer to generate an error if they don't support the specified filter - otherwise there's no way for the client to know if the result set is empty, or the filter was ignored.

discovery.md
```diff
@@ -562,12 +572,11 @@ prepared to support paginated responses.
...
 #### `GET /services/{id}`
 
-If this refers to a valid Service, then this MUST return that single Service
-entity.
+This MUST return the Service if it exists.
 
 The following responses are defined by this specification:
 - `200 OK` and the Service representation in the HTTP Response Body.
-- `404 Not Found` if there is no Service with the specified `id`.
+- `404 Not Found` if the Service does not exist.
```
duglin, line 579: it's kind of pedantic, but as it stands now this doesn't actually say which service to return. Meaning, there's no MUST to link the "id" to the service. It is certainly implied though. Perhaps we need to make it explicit

discovery.md
duglin: I think we need to be more explicit about the shape of the query params... eg. its name (and case) related to the attributes? Do we support nested attributes?

discovery.md
erikerikson: No.  I was trying to maintain the original assertion.  The service can accept other types but it must accept JSON so when a request that declares it is JSON and actually contains valid JSON shows up, it must handle the request according to the specification.

erikerikson: I'm apparently blind to the backwards slash/escape.  Where?

discovery.md
```diff
@@ -650,30 +635,49 @@ upon return of the `202 Accepted` from the original request.
...
-array MAY include the `id` or `epoch` attributes but they MUST be ignored by
-the Discovery Endpoint. Both of these attributes' values MUST be defined by
-the Discovery Endpoint.
+This MUST update the collection of Services in the Discovery Endpoint to
+contain all Services in the request or fail.  Any existing Service identified
+in the request will be completely replaced by the Service definition in the
+request.  The body of the request message MUST contain an array of zero or more
+Service entries.
+
+The `id` and `epoch` attributes of the Services MAY be omited. In such cases
```
duglin, line 644: s/omited/omitted/ I think

discovery.md
```diff
@@ -691,108 +695,102 @@ Content-Type: application/json
...
-they appear. While any error in the overall process MUST rollback any changes
-made due to the request, the processing of each Service in the request MUST be
-applied as if all previous Services in the request were already completed.
-This means that attribute uniqueness checking, such as on the `name`
-attribute, MUST be done on the new state of all Services as if all previously
-specified Services in the incoming request were already successfully processed.
+This MUST delete all of the Services in the Discovery Endpoint that are
+contained in the request.  If any of those Services cannot be deleted for any
+reason (e.g. an `id` is not given or a Service with the given `id` does not
+exist) then the entire request MUST fail with no effect.  The deletion MAY be a
```
duglin, line 716: I think we may need to define what 'soft delete" means w.r.t. the GET - meaning should a soft deleted one appear in there? I guess not. Can someone resurrect one? If not, I wonder if we should just leave this as an impl detail since if people can't see it but it lives for a while then who cares?

discovery.md
```diff
@@ -691,108 +695,102 @@ Content-Type: application/json
...
 [
-  "id": "{id}"
+  {
+    id: "{id}",
+    epoch: "{epoch}"
+  },
   ...
 ]
 ```
+#### `DELETE /services`
```
duglin, line 711: Should we support a DELETE w/o a body too - meaning put the id/epoch in the URL ?  For the singleton case it might be easier.

discovery.md
```diff
@@ -691,108 +695,102 @@ Content-Type: application/json
...
+If an `id` is not specified in any of the Services the entire request MUST
+fail.  If no Service can be found that corresponds to a specified `id`, the
+entire request MUST fail.  An `epoch` value MAY be omitted.  If a Service
+corresponding to a specified `id` has a higher recorded `epoch` value than the
+one given the entire request MUST fail.
 
-Likewise, while it might be less than optimal, it is technically possible for
-a request to update a Service multiple times if the same `id` is used more
-than once within a request.
+If other service attributes are included in the request, those SHOULD be
```
duglin, line 725: not MUST?

discovery.md
```diff
@@ -691,108 +695,102 @@ Content-Type: application/json
...
 ```
+200 OK
 
 [
-  "id": "{id}"
+  {
+    "url": "{url}",
+    "name": "{name}",
+    ... remainder of Service attributes ...
```
duglin, line 764: Are they expected to include the full service here? Should we just do IDs?  Just wondering, no strong opinion.

discovery.md
```diff
@@ -691,108 +695,102 @@ Content-Type: application/json
...
-contain an `id` attribute that matches the `{id}` in the `PUT` URL.
+#### `POST /services/{id}`
 
-The request MAY include the `epoch` attribute. If present then the Discovery
-Endpoint MUST ensure that the value on the incoming request matches the
-current value on the Service. If they do not match then an error MUST be
-generated and the Service MUST NOT be updated. However, if the incoming
-request does not include this attribute then this verification MUST NOT
-be done.
+This MUST update the collection of Services in the Discovery Endpoint to
```
duglin, line 772: collection? isn't it a singleton?

discovery.md
```diff
@@ -808,79 +806,73 @@ Content-Type: application/json
...
 
-The format of the HTTP Response MUST adhere to the following:
+The format of the `200 OK` HTTP Response MUST adhere to the following:
 ```
+200 OK
 
 {
-  "url": "{url}",
+  "id": "{id}",
```
duglin, line 815: If id/epoch are good enough for POST responses, it seems like it should be good enough for DELETE too

discovery.md
```diff
@@ -808,79 +806,73 @@ Content-Type: application/json
...
 ```
-#### `PUT /services/{id}?import`
 
-This MUST create or update a Service in a Discovery Endpoint. The Body of the
-request message MUST contain a single Service definition, however the presence
-of the `epoch` attribute is OPTIONAL. The Service MUST contain an `id`
-attribute that matches the `{id}` in the `PUT` URL.
+#### `DELETE /services/{id}`
+
+This MUST delete the Service at the referenced URL.  The deletion MAY be a soft
```
duglin, line 822: same comment here about soft deletes

discovery.md
```diff
@@ -808,79 +806,73 @@ Content-Type: application/json
...
+deletion and MAY have a defined time to live (a.k.a. `ttl`).
 
-If `epoch` attribute is present then the resulting value of this attribute
-MUST result in a value that is larger than both the incoming `epoch` value and
-any existing Service's `epoch` value.  If it is not present then the Discovery
-Endpoint MUST assign a value that is larger than any existing Service's
-`epoch` value.
+The request MAY include a body.  Any such body MUST be an object containing an
+`id` matching the `{id}` in the path.  That object MAY include an `epoch`
+value.  If the Service corresponding with the specified `id` has a higher
```
duglin, line 827: I like this safety check

discovery.md
```diff
@@ -808,79 +806,73 @@ Content-Type: application/json
...
-  "url": "{url}",
-  "epoch": "{int}",
-  "name": "{name}",
-  ... remainder of Service attributes ...
-}
+[
+  {
+    "url": "{url}",
+    "name": "{name}",
+    ... remainder of Service attributes ...
```
duglin, line 867: same comment as above about whether we really need the full service or not

duglin:
I didn't check the openapi doc too closely, but overall I like these edits. I think it really helps to clarify/simplify things, especially the import stuff.

discovery.md
duglin, line 529: backslash is here - on line 529

erikerikson: TYVM

discovery.md
erikerikson: Before resolving this because I do think it expresses a desired target (though, please do discuss if you disagree), I think there is a lot of room for better phrasing.  Basically I am attempting to declare that all compliant APIs must give JSON when requested and accept JSON when it is given.  However, I was trying to be more concrete than that in following suit with the previous declaration and being explicit about what "supporting JSON encoding" means.

discovery.md
erikerikson: I'd like to understand a little better the benefit adding these characters brings.  I thought it fairly clear that the names as given earlier in the document (e.g. 249 for `name`) would apply here and I could add something but is "...an attribute `(from the Service attributes above)` as a query..." sufficient to satsify?  Is there better language that you would recommend?

discovery.md
erikerikson: My first instinct is to agree with that being reasonable.  And then I go down an OPTIONS like rabbit hole for tools to identify which filters are supported.  I suppose that's not the end of the world and services could do binary discoveries on which filters are supported ... but do you have a thought on that?  Others welcome to express opinions too of course :D

discovery.md
erikerikson: Haha.  That really is kind of pedantic.  But, I love it and agree.  Implied, sure, but I can improve the language.

Does `s/the Service if/the latest version of the Service with the given `id` if/ suffice in your opinion?

Also adding the `latest version` language to the `GET /services` endpoint.

discovery.md
erikerikson: that is correct, thank you for catching the spelling error

discovery.md
erikerikson: > should a soft deleted one appear in there?
I would think so by default but I'd be happy to leave it unspecified (I like making sure things are in user's faces when they might be important).  Otherwise I would probably need to be quite explicit about the optional filter and its behavior.  It feels a little weird since the soft delete behavior is optional.  Happy to do whatever is preferred, I have only a default instinct and a weak opinion.

discovery.md
erikerikson: That makes a stronger statement clearly.  I would read the difference as saying that there is no use case where an implementer will want to validate every field as a "is this really the thing you want to delete?" check.  As I think about it a record with the same `id` and `epoch` should always be identical but I am not certain that every use case agrees with me.  Maybe there's a risk to mitigate by switching to MUST?  I'm entirely open to changing this, just trying to understand the reasoning and trade offs.

discovery.md
erikerikson: The service is.  Yet the set of services known by the DE, while itself a kind of singleton, seems reasonably called a collection.  The next line is where the singleton nature of the service is reflected but perhaps the language is clumsy.  I'm open to alternative phrasings.  Something that I was intending to promote to the reader's consciousness is that the API is based on the collection of resources model of REST.  By upserting a singleton Service version here the set of all services is modified to contain that version replacing the old version if it exists.

discovery.md
erikerikson: I made [a comment](https://github.com/cloudevents/spec/pull/837/files#r664990150) meaning to call out this choice and ask if the two should be more aligned.

discovery.md
erikerikson: As [noted](https://github.com/cloudevents/spec/pull/837/files#r670146051), I [asked about](https://github.com/cloudevents/spec/pull/837/files#r664990150) whether these should be consistent.  The thought basis for returning the entire service in response to a DELETE is that one can do that only providing the `id` and optionally `epoch` but may not have the rest of the information in the service.  I'm not feeling super strong about this supporting vital use cases.  However, if we went that route, as commented, it may be appropriate to respond with the old values of a service in cases where it is overwritten as opposed to created.  I would prefer that one of these strategies should be selected and that we should be consistent about it.  Sorry to ask the ether but I know I'm not holding all the history and every context so I aimed for a middle ground.

discovery.md
```diff
@@ -534,9 +535,19 @@ compliant Discovery Endpoint implementations.
 
 #### `GET /services`
 
-This MUST return an array of zero or more Services. The array MUST contain all
-Services available via this Discovery Enpoint. Any Service previously returned
-to a client that does not appear in this result can be assumed to be deleted.
+This MUST return an array of zero or more Services. The array MUST contain the
+latest version of all Services available via this Discovery Enpoint. Any
+Service previously returned to a client that does not appear in this result can
+be assumed to be deleted.
```
deissnerk, line 541: I wonder, if the assumption that a service is _deleted_ can be made. There might simply be a change of access rules. Technically the service in that case would still exist, it would just not be visible to me any more.

deissnerk (commented):
@erikerikson Thanks for taking a close look at this spec! So far, I just have a minor thought that touches the spec in general and not specifically your proposal. I hope I find more time to take a closer look later today.

discovery.md
erikerikson: This is an interesting point.

I unthinkingly kept this but I would expect a loss of access rights to a service to have the same result.  Since it is commentary as opposed to normative we could just remove the comment.  Yet, when I think of how I might change it I wonder if, in a practical regard, it is as if the service is deleted and the requestor should act accordingly.   The statement "... assumed to be removed from the set of Services the requestor has available" is more correct but the requestor has to decide what to do about that regardless of the reason that the service is missing from the result.  Certainly I wouldn't expect to be required to give indication of the reason as a discovery endpoint especially if I don't have any record because it was deleted.

Any suggestions on how to change this?

discovery.md
erikerikson: A motivational note: I hoped that the phrasing given left space for implementers to provide richer capabilities (e.g. [LIKE semantics](https://docs.microsoft.com/en-us/sql/t-sql/language-elements/like-transact-sql?view=sql-server-ver15)).  To say it differently, I had hoped that exact match would be a dependable minimum bar for supported filters.

duglin:
All - reminder please review in prep for tomorrow's call. Leave any comments ASAP so @erikerikson can address them before the call if necessary

ce_discovery.yaml
```diff
@@ -9,81 +9,230 @@ paths:
...
+                description: The data type of the extension attribute
+              specurl:
+                type: string
+                description: An attribute pointing to the specification that defines the extension
+          example: '{ "name": "dataref", "type": "URI-reference", "specurl": "https://github.com/cloudevents/spec/blob/master/extensions/dataref.md" }'
+    eventtypes:
+      type: array
+      items:
+        $ref: '#/components/schemas/eventtype'
+    serviceinstance:
```
deissnerk, line 187: If I understand this correctly, this is essentially a service _version_? I know it is hard to name these things. The way I usually understand _service instance_ is that there is something like a service broker or a provisioning API that lets you create an instance of a service.

erikerikson: You're correct, it [I think] MUST uniquely identify a version of the service record.  `serviceidentifiers` or `serviceversion` may have been a better choice (I'm not attached to this one if you have a preference).  I needed a name with which to reference the schema for a service but only requiring the subset of attributes that uniquely identify the service that was created/updated or is to be deleted.  This is used only here in a the response to `POST /services` and the request body of `DELETE /services`.

discovery.md
duglin: I"m ok with it - I think I was just misreading it before.

discovery.md
duglin: maybe instead of "deleted" we say that the old version is no longer part of the result set. And whether "result set" is influence by ACL or something else is out of scope

discovery.md
duglin: Even if we think it should be obvious that the url should be `?name=foo` instead of `?Name=foo` I think it would be good to be explicit since parts of URLs are case insensitive.   However, I think also need to be clear on the "exact match" part. Yes it's pedantic, but I think being clear that we also mean case matching too would be good.

As for unknown query parameters, we could just say that the server MUST generate an error if there are unknown query parameters, but that through an out-of-band agreement servers MAY extend the list of query parameters.

discovery.md
duglin: > Does s/the Service if/the latest version of the Service with the given id` if/ suffice in your opinion?

yes

discovery.md
duglin: I'd prefer that at the API level "delete == delete and it's no longer seen". Whether it lives on in the server is an impl choice as long as the user doesn't see it any more. It keeps the model simple.

discovery.md
duglin: From an interop perspective I'd think we'd want to be clear on what each impl will do w.r.t. finding and accepting the DELETE operation. If one rejects a request while another accepts it based on these extra attributes, then I think we'll have an interop issue.

discovery.md
duglin: hmmm at first I didn't see the value in returning the old service (and on delete too), but your comments reminded me that I think I've seen some service (like github's events I think) that will return the old data (at least for some fields) so you can do a diff if you really wanted. And I have found that useful.

Let's see what other people think - I can go either way right now

discovery.md
duglin: ok - rereading it, I can live with it.

discovery.md
duglin: As I commented above.. I'd refer delete==delete

discovery.md
erikerikson: As per the call, both POST and DELETE will (in case of success) return the entire Service value resulting from / just prior to (respectively) the successful processing of the request.

discovery.md
erikerikson: Two changes to call out:

> Doing so MUST cause only Services with at least an exact value match for those filters to be
included in the result.

`at least` was inserted.  This allows for fuzzy matching and other above-and-beyond capabilities (the intention is to set a minimum bar rather than limit implementations to the minimum bar)

Later, after requiring `name` filtering support and the explicit rejection of unsupported filters...

>   Endpoints MAY add additional query parameters or filter processing to improve upon this minimum provision.

Calling these out since it risks going slightly beyond the approved changes but I think captures the intent more thoroughly.  Clearly this can be altered by follow on PRs.

discovery.md
erikerikson: As discussed we will remove mention of soft deletes from the specification, at least for now.

discovery.md
erikerikson: I will remove this out of rabbit hole avoidance.

erikerikson: Actually, I think this serves a purpose.  The significance of a service being returned (or not) is that it is "on offer" or "available" for use (or not) to the requester.  Rather than "deleted" we can write "no longer available" which addresses the consequence while not making the tautological statement that not being in the result set is not being in the result set.

duglin:
Approved on the 7/22 call

PR #849, opened 2021-07-22
bug (discovery): update example response

erikerikson:
Follow up to #837 

Example `POST /services` and `POST /services/{id}` response must
reflect changes too.

### DIFF
--- a/discovery.md
+++ b/discovery.md
@@ -710,7 +710,8 @@ Content-Type: application/json
 [
   {
     id: "{id}",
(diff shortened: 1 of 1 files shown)


erikerikson:
I believe the intent of approving #837 included this change.  It was my oversight not to align the example responses with the specification changes, sorry.

duglin:
LGTM

duglin:
Approved on the 7/29 call

PR #851, opened 2021-07-26
Update primer.md

abdelouahabb:
Added Pub/Sub to the docs


duglin:
@abdelouahabb can you sign your commit?

jskeet:
A more helpful title and description would be useful too :)

duglin:
@abdelouahabb thanks for the PR but the group decided to remove references to all companies rather than worry about adding more.

PR #852, opened 2021-08-03
add ppts from spec reviews

duglin:


duglin:
Added the changes to the README last week - so based on the pre-approval from the group on the call last week I'm going to merge

PR #856, opened 2021-08-05
remove offering from list of contributors

duglin:
Per the call today, we decided to remove the projects/offerings from the list of contributors. If people want to list their projects then they should do so on the cloudevents.io page


duglin:
Approved on the 8/12 call

PR #865, opened 2021-08-12
Fixed erroneous header name

### LINKED ISSUE #781: WebHook "Origin" header concept clashes with RFC6454
https://github.com/cloudevents/spec/blob/v1.0.1/http-webhook.md#21-delivery-request 

We mandate that push deliveries made after the abuse protection handshake use the "Origin" header.

While our Web Hook delivery notion of "Origin" is somewhat similar in intent to the Origin concept defined in RFC6454 and as used, for instance in CORS (W3C/WHATWG, https://fetch.spec.whatwg.org/#origin-header) it is not the same. The RFC6454 concept refers to the relationship of sequences of HTTP requests amongst each other (referrer, but without path), and the Origin points to the server to which the prior request was made whose result is now motivating this request. Our notion of Origin describes the originating server of the request and it purely an identifier.

To make matters worse the RFC6454 Origin header holds a URI and our header is a DNS style expression.

We have two choices here:

a) We are not at all referencing RFC6454 in our spec, which means we can just take the stance that our "Origin" header is a different thing. The problem is that existing HTTP clients do indeed know about RFC6454. Arguably, WebHook push requests are always issued independent from other requests, meaning that the RFC6454 concept does not apply.
b) We change our header name for delivery requests to 'WebHook-Request-Origin', e.g. the same as for the validation handshake.

clemensv:

Fixes #781

### DIFF
--- a/http-webhook.md
+++ b/http-webhook.md
@@ -70,10 +70,10 @@ This specification does not further constrain the content of the notification,
 and it also does not prescribe the [HTTP target resource][rfc7230-section-5-1]
 that is used for delivery.
 
-If the delivery target supports and requires
-[Abuse Protection](#4-abuse-protection), the delivery request MUST include the
-`Origin` header. The `Origin` header value is a DNS name expression that
-identifies the sending system.
+If the delivery target supports and requires [Abuse
+Protection](#4-abuse-protection), the delivery request MUST include the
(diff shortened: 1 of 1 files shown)


duglin:
Approved on the 8/19 call
@clemens to make a PR for 'master' branch

PR #866, opened 2021-08-18
Minor/misc discovery changes

### LINKED ISSUE #863: Issues with discovery
WIP

https://github.com/cloudevents/spec/blob/master/discovery.md#get-services
- [ ] `Any Service previously returned to a client that does not appear in this result can be assumed to be no longer available.` I think this needs to be limited to just unfiltered queries since a missing entry from a filtered set could mean that it was deleted or that the filtered field changed.
- [ ] `Discovery endpoints MUST support filtering with the following attributes and MUST reject any unsupported filters. ` we should probably say what the HTTP response code is.
- [x] is Epoch allowed to be a negative int?
- [x] "name" must be non-empty string
- [x] typo: dataschama

duglin:
Fixes #863

- `Any Service previously returned to a client that does not appear in this result can be assumed to be no longer available.` I think this needs to be limited to just unfiltered queries since a missing entry from a filtered set could mean that it was deleted or that the filtered field changed.
- Made Epoch an unsigned 32-bit int
- "name" must be non-empty string
- typo: dataschama

### DIFF
--- a/discovery.md
+++ b/discovery.md
@@ -230,7 +230,7 @@ The following sections define the attributes that appear in a Service entity.
 
 ##### epoch
 
-- Type: `Integer`
+- Type: `Unsigned 32-bit Integer`
 - Description: The Discovery Endpoint's epoch value for this Service Entry.
   This specification does not mandate any particular semantic meaning to
   the value used. For example, implementations are free to use a value that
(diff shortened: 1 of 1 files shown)


duglin:
Approved on the 8/28 call

PR #868, opened 2021-08-18
Discovery add history

### LINKED ISSUE #699: [Discovery API]  Consider adding history

erikerikson:
Fixes #699 

## Proposed Changes

- Add discovery endpoint as a Service entity by which it publishes changes to itself
- Add historical querying capability (perhaps better delivered by allowing subscriptions to specify delivery of events starting from those available as of a time in the past)

**Release Note**

```release-note
Add Discovery Endpoint as a "self" Service Entity
Add historical service change query endpoint 
```

### DIFF
--- a/ce_discovery.yaml
+++ b/ce_discovery.yaml
@@ -27,18 +27,10 @@ paths:
     post:
       operationId: postServices
       requestBody:
-        required: true
-        content:
-          application/json:
-            schema:
-              $ref: '#/components/schemas/services'
+        $ref: '#/components/requestBodies/servicespostrequest'
       responses:
         '200':
-          description: A list of the resulting Service values resulting from processing the request, in the same order as in the request
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/services'
+          $ref: '#/components/responses/servicespostresponse'
         '400':
           description: Bad Request - constraint failure
         '409':
@@ -46,18 +38,10 @@ paths:
     delete:
       operationId: deleteServices
       requestBody:
-        required: true
-        content:
-          application/json:
-            schema:
-              $ref: '#/components/schemas/serviceinstances'
+        $ref: '#/components/requestBodies/servicesdeleterequest'
       responses:
         '200':
-          description: A list of the Services that were deleted
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/services'
+          $ref: '#/components/responses/servicesdeleteresponse'
         '400':
           description: Bad Request - constraint failure
         '404':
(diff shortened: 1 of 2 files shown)


ce_discovery.yaml
```diff
@@ -114,32 +80,91 @@ paths:
...
-         schema:
-           $ref: '#/components/schemas/service/properties/id'
+        - $ref: '#/components/parameters/id'
       requestBody:
-        required: false
-        content:
-          application/json:
-            schema:
-              $ref: '#/components/schemas/serviceinstance'
+        $ref: '#/components/requestBodies/servicedeleterequest'
```
erikerikson, line 85: Having a requestBody on a delete is discouraged by OpenAPI for reasons discussed [here](https://github.com/OAI/OpenAPI-Specification/issues/1801).  To abusively summarize the most useful piece, there is no standard about the interaction of HTTP DELETEs on to resource collections and the caching layers which may cause undesirable lack of caching on one hand and/or a use of cached resources inappropriately on the other.

e.g.

```
DELETE /services

[
  { id: {id_0} },
  ...
  { id: {id_N} },
]
```

Does not clear `/services/{id_0}` from HTTP response caches which, unless configured not to cache, are allowed to cache and return that value following the resource's deletion using a DELETE with a request body.

ce_discovery.yaml
```diff
@@ -114,32 +80,91 @@ paths:
...
-              schema:
-                $ref: '#/components/schemas/service'
+          $ref: '#/components/responses/servicedeleteresponse'
         '400':
           description: Bad Request - constraint failure
         '404':
           description: Not Found
         '409':
           description: Conflict - epoch not greater
+  /changes:
```
erikerikson, line 95: We may want to remove this and everything in the commit that adds it.  See discussion in the spec.

discovery.md
```diff
@@ -876,6 +876,75 @@ Content-Type: application/json
...
+          "request": {request},       // as per path specification
+          "response": {response}      // as per path specification
+        }
+      }
+    }
+  ]
+}
+```
+
+#### `GET /changes`
```
erikerikson, line 916: We may want to remove this and the OpenAPI changes the reflect it.  A potentially preferable alternative would allow subscribers to declare an offset or other "time" (including "NOW") from which to start receiving events of the subscribed service.  By specifying a time in the past, a historical record could be retrieved using CloudEvent supporting protocols and avoiding the use of this HTTP only interface.  The guarantees of the requested/negotiated protocols and scaling/factoring elements (such as partitioning) would thereby be a part of the delivery of the historical information.  As such it would provide for consensus over replication results and avoid some of the system pitfalls that can result from some implementation choices (e.g. imagine a partitioned data store across time where query results changed to alter the order of processing change events).

erikerikson:
@duglin this has been pending the AIs from Clemens and Klaus but I am wondering if we expect those to alter this in some way?  If not, we might be able to merge this and proceed by altering it as necessary?

duglin:
Approved on the 10/14 call

PR #869, opened 2021-08-24
Reject queries with unknown query params (WIP)

duglin:
Split this one from #866 so we can discuss it w/o blocking the other changes.


duglin:
@ryanhorn you mentioned that some clients/tooling might add query parameters that the user can't control/stop. Do you have any examples of that? I'd be kind of annoyed if my client didn't have total control over those.

I see a few options:
1. leave it as is where we silently ignore unknown query params and then the user/client is unaware of whether their request to filter based on a well-defined discovery property was adhered to, or not
2. require that the discovery endpoint reject unsupported, but known, service property names - or spec defined queries
3. reject all unknown query params - PR as it is today

My 2 cents:
- all of these might require us to add a mechanism by which the discovery endpoint can advertise which query filters it supports.  Might be a good thing to do no matter which way we go.
- option 1 scares me because it could be interpreted as the endpoint lying about what it returned. But it's the least amount of work for the server. Less clarity for the client tho.
- option 2 requires work for the server since it would need to check each query parameter to determine if it's "known" or not - even if it's not going to filter over it. Still leave the client with less clarity if it **thinks** the server supports a query parameter that it really doesn't
- I still like option 3 as it reduces the chances of incorrectly interpreting the results. But it is pretty strict.

Another option would be to prefix the filter query parameter with something - e.g. `?filter=name=myservice`. Then it's clear that the query parameter isn't some random thing, but a filter and it would then be ok to reject unsupported keys. Might be a good thing to do no matter which way we go... less worry about service attributes overlapping with possible (future) query params we may want to use.

What about pagination query parameters? Should we reject a request with those parameters if the server doesn't support it? I would think so if we don't want ambiguity in understanding the results.

ryanhorn:
I don't have an example off-hand, my comment was just being cautious about outright rejecting requests based on query parameters that might or might not have been intended to be used as filters. It feels a bit draconian to me. I think having query parameters that are specific to filtering and validating those is a find middle ground.

erikerikson:
1 and 3 seem similarly plausible but problematic as they are (2 just seems like trouble) but maybe...

4. the service must include in the response a list of the unsupported parameters that were nonetheless submitted in the request  (perhaps as a header but it would be easy to ignore there)

I second that it is likely very helpful, if also a lot of work, to offer a supported parameters name and type discovery endpoint so that tooling providing a Discovery UI can generate an controls for the user.

IMO adding a `filter=[...]` doesn't seem to improve the circumstance

RE: pagination parameters - making the handling of unsupported parameters standard across the APIs sounds useful but far more difficult to describe en mass

duglin:
@erikerikson couple of questions:
> IMO adding a filter=[...] doesn't seem to improve the circumstance

I guess I thought it helped because it reduced the surface area of concern that @ryanhorn mentioned... that being there could be some client side tooling that adds query parameters w/o the client app having control over it. By scoping things under the "filter" query parameter I was hoping we'd have the latitude to be as strict as we wanted since it's pretty much totally under our control.

> RE: pagination parameters - making the handling of unsupported parameters standard across the APIs sounds useful but far more difficult to describe en mass

See what you think of the current version of the PR. It removes the uncertainty for filters I was looking for, but leaves it unspecified for unknown query parameters in general. I didn't add this, but I was wondering if we should recommend that extension query parameters include some kind of metadata in the response (even if to just say "yes I understood the request") so the client can know it was adhered to even if it didn't end up changing the result set.

erikerikson:
@duglin

My default thought was that the same dynamic that would have them put filters on the URL would have them put them on the filter parameter.  I could be wrong but my default take is that they are equivalent.  If they are equivalent the more idiomatic placement of filters directly into the URL seems more idiomatic (as opposed to the use of `filter`).  I suppose that there could be default parameters tools put in every URL but that seems like an oddball choice to make in general.

I don't have a very strong opinion of what to do WRT filter rejection but while I can appreciate the notion that potentially letting the user believe filters were used when they were ignored instead seems problematic, it also seems problematic to leave users that are not in control of their tool to be stuck with a tool that either doesn't work or doesn't support some of the filtering supported by a compliant API specification that advertises filter support.  I think this leaves a supported filters query endpoint a best option but I am not excited about that either.  Perhaps a HEAD or OPTIONS request could be used to describe those as an alternative but I am oblivious to how that is generally accomplished.  I'd be surprised if we were the first ones to think of this in the API specifications space.

erikerikson:
Just for the record, the thrust of the approaches we've been discussing might be superficially in opposition to [Postel's law](https://en.wikipedia.org/wiki/Robustness_principle).  The problem with that simplification is what it means to conservatively respond to the liberally accepted but unsupported filter parameters.  However, I want to note that in practice I have observed most HTTP APIs ignoring unknown parameters, instead selecting for known parameters and validating they exist if required and their content if supplied.

An example parameter that one might add to a request would be a ["cash buster"](https://www.keycdn.com/support/what-is-cache-busting).  As duglin pointed out well on the call, the use of `filter` as suggested would allow for that to be accepted while rejecting unknown filters.  I had been thinking too narrowly about equivalence - thanks for unbreaking that.

duglin:
>  However, I want to note that in practice I have observed most HTTP APIs ignoring unknown parameters

Yup, exactly why I think scoping it to a "filters" query parameter give us the ability to be more strict and choose (what I consider) the "safer" approach and reject unknown filters so the client doesn't make the mistake of thinking they got what they asked for. I view it as the client saying "I only want services that match this criteria", so to me the first "very boolean" decision here is "does the server support that criteria?". If not, they must stop, otherwise they're going to lie about the results - or force the client to then do its own filtering to remove incorrect entries. That's not the UX I'd want as a client.

discovery.md
```diff
@@ -541,16 +541,28 @@ This MUST return an array of zero or more Services. The array MUST contain the
...
-cause only Services with at least an exact value match for those filters to be
-included in the result.  Parameter names and values are case sensitive.
+[service attributes](#service-attributes) under the `filter` query parameter.
+The general format for the `filter` query parameter is:
 
-Discovery endpoints MUST support filtering with the following attributes and
-MUST reject any unsupported filters.  Endpoints MAY add additional query
-parameters or filter processing to improve upon this minimum provision.
+```
+?filter=ATTRIBUTE=VALUE[,ATTRIBUTE=VALUE]
```
duglin, line 548: Open question: how do people specify a comma in a value? Should we support an escape, or should we not allow multiple per "filter" and require multiple "filter" query params instead? I'm leaning toward forcing multiple query parameters.

discovery.md
```diff
@@ -530,6 +530,8 @@ including an HTTP `Content-Type` Header of `application/json` and body in that
 format MUST be supported.  Likewise requests with `Accept` header of
 `application/json` MUST be supported.
 
+Unkonwn query parameters on the APIs MUST be ignored.
```
jskeet, line 533:
```suggestion
Unknown query parameters on the APIs MUST be ignored.
```

(If I can't make substantive contributions, I can at least help with typos...)

discovery.md
```diff
@@ -541,16 +543,30 @@ This MUST return an array of zero or more Services. The array MUST contain the
...
+Multiple attribute/value combinations MAY be specified as separate `filter`
+query parameters. When there are multiple filters, the resulting set of
+Services MUST only include ones that match all of the filters specified.
+Matching of attribute names and values MUST be case sensitive.
+
+Discovery endpoints MUST support filtering by the following attributes:
 
 - `name`
 
+Requests with unsupported filter attributes, MUST reject the request with a
```
jskeet, line 565:
```suggestion
Requests with unsupported filter attributes MUST be rejected with a
```

Or "Endpoints MUST reject requests with unsupported filter attributes, returning a `400 Bad Request` response."

duglin:
Approved on the 9/30 call

PR #870, opened 2021-08-26
Fix for issue 781

clemensv:

Follow-up PR equivalent to errata merged into 1.0.1


duglin:
Approved on the 8/28 call

PR #872, opened 2021-09-09
fix: filters subscription example

loopingz:
Fixes #

The filters example was not matching the latest changes


duglin:
LGTM

duglin:
Approved on the 9/16 call

PR #876, opened 2021-09-16
Make it clear that empty query is still a 200

duglin:
During interop testing someone was returning a 404 when the query yielded zero results. It shouldn't be an error.

### DIFF
--- a/discovery.md
+++ b/discovery.md
@@ -551,6 +551,9 @@ parameters or filter processing to improve upon this minimum provision.
 
 - `name`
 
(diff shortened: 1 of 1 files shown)


jskeet (approved):
Completely agree, so long as the reason for the empty result set wasn't an unsupported filter, as covered a few lines above. (I'm not suggesting a change, just checking that this wasn't the case in the interop test.)

duglin:
Approved on the 9/23 call

PR #877, opened 2021-09-22
Use code formatting for example URLs

Oberon00:
The RefinedGitHub plugin strips off the https:// otherwise, and they aren't meant to be clickable anyway.


Oberon00:
I don't think I can sign a DCO, but this change is so minor that it should not be copyrightable anyway, so feel free to apply the changes outside this PR or close it.

duglin:
thanks!
/lgtm

PR #878, opened 2021-09-23
feat(cesql): remove explicit casting from the spec

loopingz:

## Proposed Changes

- Remove explicit casting from the spec in favor of implicit casting

The functions like IS_BOOL were not usable as is, we have no IF within the language.
Discussed last week on the call, but I'm happy to not merge if we have use case of explicit casting, I did not find one

### DIFF
--- a/expression-language.md
+++ b/expression-language.md
@@ -199,7 +199,7 @@ Each CloudEvent context attribute and extension MUST be addressable from an expr
 by the spec. For example, using `id` in an expression will address the CloudEvent [id attribute][ce-id-attribute].
 
 Unless otherwise specified, every attribute and extension MUST be represented by the _String_ type as its initial type.
-Through explicit and implicit type casting, the user can convert the addressed value instances to _Integer_ and
+Through implicit type casting, the user can convert the addressed value instances to _Integer_ and
 _Boolean_.
 
 When addressed an attribute not included in the input event, an empty _String_ MUST be assumed as value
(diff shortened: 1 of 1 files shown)


duglin:
@devguyio  any thoughts?

expression-language.md
```diff
@@ -361,14 +351,7 @@ operators.
...
 
-CESQL supports both implicit and explicit type casting among the _primitive_ types. Users can perform explicit type
-casting through the functions defined in the [Casting and type checking](#351-casting-and-type-checking) sub-paragraph.
-
-When input parameters' types of operator/function don't match the signatures, the CESQL engine MUST try to perform an
-implicit cast.
-
-Implicit casts must follow the same semantics of their equivalent explicit cast functions, as defined in
-[Casting and type checking](#351-casting-and-type-checking) sub-paragraph.
+When input parameters' types of operator/function don't match the signatures, the CESQL engine MUST try to perform an implicit cast.
```
jskeet, line 354: It's possible that it's just because I'm not really familiar with CESQL, but I find this sentence hard to read. I suspect that making it longer may help.

(Additionally, we may wish to consider using "argument" and "parameter" terminology in the same way as it's used in various programming languages, where the parameter is what gets declared by the function/operator/whatever, and the argument is what is passed in by the caller. I haven't checked whether that's been used elsewhere already. If folks think it would be valuable for me to look over the CESQL spec and *maybe* make a bigger proposal on this, I'd be happy to do so. No promises when!)

loopingz: I did not add the sentence, it is just a diff glitch, the PR is about removing reference to explicit cast, but I'm fine with a review and update of the spec like you proposed, probably in another PR.

loopingz:
@slinkydeveloper any thoughts also on your side?

duglin:
approved on the 9/30 call

PR #879, opened 2021-09-23
feat(discovery): replace id by authority

loopingz:

## Proposed Changes

- Adopt the authority concept from Schema Registry within Discovery Services
- Remove id as it is no use anymore as authority + name become de facto global unique id

### DIFF
--- a/discovery.md
+++ b/discovery.md
@@ -196,37 +196,20 @@ the wire format/API defined by this specification.
 
 The following sections define the attributes that appear in a Service entity.
 
-##### id
+#### `authority` (service authority)
 
-- Type: `String`
-- Description: A unique identifier for this Service. This value MUST be globally
-  unique. While other metadata about this Service MAY change, this value MUST
-  NOT so that clients can use this attribute to know whether a Service returned
-  by a query is the same Service returned by a previous query.
-
-  Whether a change to a Service would result in changing of the Service's
-  metadata (except `id`) and thus be just an update of an existing Service, or
-  whether the change would result in a brand new Service (with a new `id`) is
-  not defined by this specification.
-
-  For example, if a Service's implementation is upgraded to a new version then
-  whether this would result in a new Service (and `id`) or is an update to the
-  existing Service's metadata, is an implementation choice. Likewise, this
-  specification makes no statement, or guarantees, as to the forwards or
-  backwards compatibility of a Service as it changes over time.
-
-  Note, unlike the `name` attribute which only needs to be unique within the
-  scope of a single Discovery Endpoint, the global uniqueness aspect of this
-  attribute allows for the discovery of the same Service across multiple
-  Discovery Endpoints.
-
-  See the Primer for more information.
-
+- Type: `URI`
+- Description: Identifies the authority for this service. Similar to schemas authority.
 - Constraints:
-  - REQUIRED in responses from the Discovery Endpoint.
-  - MUST be a valid UUID per RFC4122.
+  - OPTIONAL. If the attribute is absent or empty, its implied default value is the base
+    URI of the API endpoint.
+  - MUST be a valid URI.
+  - For services imported from other discovery endpoints in replication scenarios, the
+    attribute is REQUIRED to be not empty. If the value is empty or absent
+    during import, it MUST be explicitly set to its implied default value.
 - Examples:
-  - `bf5ff5cc-d059-4c79-a89a-2513e45a1340`
+  - `urn:com-example-schemas`
+  - `https://schemas.example.com`
 
 ##### epoch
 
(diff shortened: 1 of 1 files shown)


jskeet:
I don't know much about Discovery, so this comment may be off-base - but I tend to find that human-readable descriptions tend to make poor unique identifiers.

If we were to keep id + authority as the uniqueness constraint, that would allow for a service entity to be renamed (e.g. due to rebranding or a typo) without changing its identity.

Is there any disadvantage to keeping "identity" and "human-readable name" separate? Note that if we do so, I'd suggest a minor edit to the description of "name" (as it would no longer be an identifier). I think it's still reasonable to require that names are unique at any point in time, mind you.

loopingz:
I think this is a valid point, I did remove id because with their current definition I found it was kind of a duplicate.

I believe we can do the update as you stated, without the UUID constraint on id, which let more freedom of implementation without compromising with the overall intend

discovery.md
```diff
@@ -195,12 +195,11 @@ the wire format/API defined by this specification.
 #### Service Attributes
 
 The following sections define the attributes that appear in a Service entity.
-
 ##### id
 
 - Type: `String`
-- Description: A unique identifier for this Service. This value MUST be globally
-  unique. While other metadata about this Service MAY change, this value MUST
+- Description: A unique identifier for this Service. This value MUST be unique.
```
duglin, line 201: I think we need to qualify the scope of the uniqueness. Even just saying withing the scope of the DE might be good enough

discovery.md
```diff
@@ -215,19 +214,29 @@ The following sections define the attributes that appear in a Service entity.
...
   - REQUIRED in responses from the Discovery Endpoint.
-  - MUST be a valid UUID per RFC4122.
 - Examples:
   - `bf5ff5cc-d059-4c79-a89a-2513e45a1340`
 
+#### authority (service authority)
+
+- Type: `URI`
+- Description: Identifies the authority for this service. Similar to schemas authority.
+  Note, the pair `authority` and `id` are globally unique
```
duglin, line 228: maybe...
When `authority` and `id` are combined they MUST be globally unique.

discovery.md
```diff
@@ -215,19 +214,29 @@ The following sections define the attributes that appear in a Service entity.
...
+- Constraints:
+  - OPTIONAL. If the attribute is absent or empty, its implied default value is the base
```
duglin, line 230: The OPTIONAL and REQUIRED aspects here are a bit awkward. Perhaps just say that when created/imported, a missing values MUST be assigned the base URI. Then we never need to worry about it being "empty".

I do wonder though, is it possible the same DE can have more than one URL associated with it? Which base would be used? Does it matter?

clemensv:
We need to ensure that the "id" string is a valid path segment for using within URIs, e.g. in `/services/{id}`. 

It MUST conform to the "segment" syntax here: https://datatracker.ietf.org/doc/html/rfc3986#section-3.3

discovery.md
deissnerk: I agree. I would like to see a concept like the authority and schema version URI defined for the schema registry. As the versioning requirements for discovery are different, it could just be a service URI instead of a service version URI.

discovery.md
deissnerk: I think this was copied from [here](https://github.com/cloudevents/spec/blob/master/schemaregistry/schemaregistry.md#authority-schema-authority). We would need something corresponding to `getSchemaVersionByURI` for the discovery API.

discovery.md
```diff
@@ -215,19 +214,29 @@ The following sections define the attributes that appear in a Service entity.
...
+    URI of the API endpoint.
+  - MUST be a valid URI.
+  - For services imported from other discovery endpoints in replication scenarios, the
+    attribute is REQUIRED to be not empty. If the value is empty or absent
+    during import, it MUST be explicitly set to its implied default value.
+- Examples:
+  - `urn:com-example-schemas`
+  - `https://schemas.example.com`
+
 ##### epoch
```
deissnerk, line 240: If we can agree on the concept of `authority`, we may revisit the definition of `epoch`. I think it should only be increased by the DE that represents the authority and then be used for replicating discovery data downstream.

duglin: Does this mean that only the authority can edit it? Even to make local changes like update the URLs? Or does that mean it needs to be a new ID at that point and lose it's history/linkage?

duglin:
@loopingz did you want to add @clemensv's concern to this PR to should we do it in a different one?

discovery.md
loopingz: It can probably happen if vhost are used in front, but I do not think it matters as they will be treated as different services

loopingz:
> We need to ensure that the "id" string is a valid path segment for using within URIs, e.g. in `/services/{id}`. 
> 
> It MUST conform to the "segment" syntax here: https://datatracker.ietf.org/doc/html/rfc3986#section-3.3
> 
> (Just ran into this during prototyping)

@duglin i will address it later today

loopingz:
> We need to ensure that the "id" string is a valid path segment for using within URIs, e.g. in `/services/{id}`.
> 
> It MUST conform to the "segment" syntax here: https://datatracker.ietf.org/doc/html/rfc3986#section-3.3
> 
> (Just ran into this during prototyping)

Should be good now

loopingz:
@JemDay thanks for keeping me consistent

duglin:
approved on the 10/14 call

PR #883, opened 2021-10-14
First pass at differentiating between parameters, operands and arguments

jskeet:
In somewhat woolly terms, my expectations from other language specs are:

- A *parameter* is used in the declaration/specification of a function or operator
- An *operand* specifies the value of an operator parameter
- An *argument* specifies the value of a function parameter

I personally find this distinction useful when talking about "the thing being invoked" vs "the values provided at the point of invocation".


jskeet:
> Should we define these terms in the doc?

Maybe. I think we could at least defer that - there don't seem to be many other terms defined in the spec that aren't CE-specific.

PR #895, opened 2021-11-04
move more files into "formats" dir

duglin:
@JemDay noticed I missed a few files


duglin:
Can someone do a quick double-check before I merge?

JemDay:
Litmus

PR #911, opened 2021-11-19
More filter constraints

duglin:
It seemed like the filtering stuff could use more clarity and MUSTs to ensure interop.

### DIFF
--- a/discovery/spec.md
+++ b/discovery/spec.md
@@ -596,7 +596,8 @@ The following additional constraints apply:
 - The feature names are case sensitive.
 - The `servicefilterattributes` property MUST be present, and MUST have at
   least the `name` attribute in its list. The filter attribute names are case
-  sensitive.
+  sensitive. See the [`/services` API](#get-services) for more information
+  about the format of nested attribute names.
 - The `pagination` attribute is OPTIONAL with an implied default value of
   `false`.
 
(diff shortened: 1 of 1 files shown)


jskeet (approved):

discovery/spec.md
```diff
@@ -610,25 +610,28 @@ The format for the `filter` query parameter MUST be:
...
+- a `VALUE` of "" (empty string) is valid and a filter expression of
+  `?filter=ATTRIBUTE=` MUST only match Services that contain the specified
+  attribute with an empty string value.
+- Multiple attribute/value combinations MAY be specified as separate `filter`
+  query parameters. When there are multiple filters, the resulting set of
+  Services MUST only include ones that match all of the filters specified.
+- Matching of attribute names and values MUST be case sensitive.
+- Requests with unsupported filter attributes, MUST be rejected with a
+  `400 Bad Request` response. Endpoints SHOULD return an error message that
+  indicates which filter attributes were not supported.
```
duglin, line 627: Add a note about how:  ?filter=foo=a,b=c    means "foo" == "a,b=c"

duglin:
Per last week's call I was looking at how OData and GraphQL deal with this type of filtering, and they are bit more expressive/robust. Do we want basic or robust filtering?

I see a few categories of options:
1:
```
?filter=name=foo     # exact match
?filter=name         # name is set (not null)
?filter=~name        # name is not set (is null)    ??
?filter=name=foo1 foo2&filter=attr=bar   # exact match for both
?filter=epoch=45
?filter=epoch>45
```
Too simplistic? No quoting for strings. Note the spaces. No data type considerations as part of the syntax, it's implicit

2:
```
?filter=name="foo"
?filter=name!=null      # would ?filter=name   be the same thing or a syntax error?
?filter=name=null
?filter=name="foo1 foo2"&filter=attr="bar"    # would ?filter=name=foo    be an error?
?filter=epoch=45       # ?filer=epoch="45"    error ?
?filter=epoch>45
```
Add quoting for strings

3:
```
?filter=name eq "foo"       # or use "=" instead of "eq" ?
?filter=name ne null
?filter=name eq null
?filter=name eq "foo1 foo2" AND attr eq "bar"
?filter=epoch eq 45
?filter=epoch gt 45         # use ">"  ?
```
A more expressive language

duglin:
Approved on the 1/20 call

PR #914, opened 2021-11-23
First pass at camel-casing Discovery attributes

### LINKED ISSUE #885: Discovery spec: avoid "all lower case without punctuation"?
The Discovery spec uses JSON with all-lower-case property keys (e.g. "subscriptiondialects" and "subscriptionurl") which are quite tricky to read.

I'd expect more idiomatic JSON keys to be "subscriptionDialects" and "subscriptionUrl" for example - or if we really wanted to stick to lower case ASCII, we could use "subscription_dialects" and "subscription_url".

The current JSON is consistent with CloudEvent attribute requirements, but I don't think it absolutely has to be - and I would suggest we might want to revisit those requirements for v2 anyway.

This would be a big change, of course - but better now than after 1.0...

jskeet:
Some areas for discussion:

- Under events, I've left *everything* all lower case. There's an
  argument for making attributes in `extensions` (e.g. `specurl`)
  camel-cased as those aren't (in themselves) CloudEvents attributes
- I've camel-cased `specVersions` for consistency with peers, and
  because `specversions` itself isn't an attribute name. I'm happy
  to be persuaded (or overruled) on this one.

Fixes #885

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

### DIFF
--- a/discovery/discovery.yaml
+++ b/discovery/discovery.yaml
@@ -234,34 +234,34 @@ components:
         description:
           type: string
           description: Human readable description
-        docsurl:
+        docsUrl:
           type: string
           description: Absolute URL that provides a link to additional documentation about the service. This is intended for a developer to use in order to learn more about this service's events produced
           example: http://cloud.example.com/docs/blobstorage
-        specversions:
+        specVersions:
           type: array
-          description: CloudEvents specversions that can be used for events published for this service
+          description: CloudEvents specversion values that can be used for events published for this service
           items:
             type: string
             description: version
           example: '[ "1.0" ]'
-        subscriptionurl:
+        subscriptionUrl:
           type: string
           description: An absolute URL indicating where CloudSubscriptions subscribe API calls MUST be sent to
           format: url
-        subscriptionconfig:
+        subscriptionConfig:
           type: object
           description: "A map indicating supported options for the config parameter for the CloudSubscriptions subscribe() API call. Keys are the name of keys in the allowed config map, the values indicate the type of that parameter, confirming to the CloudEvents type system. TODO: Needs resolution with CloudSubscriptions API"
           additionalProperties:
             type: string
-        subscriptiondialects:
+        subscriptionDialects:
           type: array
           description: Filter dialects that can be used in subscriptions for this service
           items:
             type: string
             description: filter dialect
           example: '[ "basic" ]'
-        authscope:
+        authScope:
           type: string
           description: Authorization scope needed for creating subscriptions. The actual meaning of this field is determined on a per-service basis
           example: storage.read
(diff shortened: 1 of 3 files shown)


discovery/discovery.yaml
```diff
@@ -234,34 +234,34 @@ components:
         description:
           type: string
           description: Human readable description
-        docsurl:
+        docsUrl:
```
duglin, line 237: Not a big deal, but given a choice I'd vote for "URL" over "Url".

discovery/primer.md
```diff
@@ -150,7 +150,7 @@ Response:
 ```
 - services:
   - service: Github
-    subscriptionuri: https://api.github.com/subscribe
+    subscriptionUri: https://api.github.com/subscribe
```
duglin, line 153: URI?

discovery/spec.md
```diff
@@ -390,6 +390,9 @@ The following sections define the attributes that appear in a Service entity.
 
 The following sections define the attributes that appear in EventType definitions of a Service object.
 
+*Note:* while attributes elsewhere are camelCased, the attributes within `EventType` are all lower-case,
+as they correspond directly to CloudEvent attributes (with the exception of `extensions`).
+
```
duglin, line 395: Makes sense.

duglin:
Aside from my question about URL and URI, LGTM
thanks!

discovery/discovery.yaml
jskeet: Let's discuss in the meeting. I definitely prefer using Pascal case for abbreviations, unless they're only two characters. It makes it clearer when there's a word after it. For example, `docsUrlPage` splits more clearly (IMO) into "docs", "URL", "page" than "docsURLPage" would.

When it comes to two-letter abbreviations (e.g. IO) it gets trickier.

duglin: you lost me at the word "Pascal" :-)

discovery/spec.md
duglin: Been thinking about this since our last call... while it's less pretty, I'm very worried about the pain our users will experience in our perceived inconsistency here. While having everything as lowercase isn't ideal, at least it doesn't require people to think about which set of attributes are camel-cased and which are not.

jskeet: Right. Let's take a straw poll at the meeting tomorrow.

A thoroughly-considered decision to stay as we are is absolutely fine by me - I just didn't want to get there accidentally.

jskeet:
Closing as we've decided not to do this after all - it got a bit too hairy in terms of trying to be consistent to multiple conventions at the same time with subtle reasons about when to go with which.

PR #926, opened 2022-01-04
Add support for "update" in GET /features

duglin:
Would be nice to know if POST/PUT are supported w/o "discovery by error".


duglin:
maybe "management" would be a better word?

duglin:
Approved on the 1/13 call

PR #927, opened 2022-01-05
Just some wordsmithing

duglin:
Found some of the text hard to follow


duglin:
Approved on the 1/13 call

PR #928, opened 2022-01-05
Use PUT instead of POST

duglin:
Doesn't seem right to use POST to update (add) a new Service when the ID is already known.

### DIFF
--- a/discovery/spec.md
+++ b/discovery/spec.md
@@ -570,7 +570,7 @@ error for a request to that Service directly.
 
 All APIs MUST support JSON encoding, which means that HTTP requests
 including an HTTP `Content-Type` Header of `application/json` and body in that
-format MUST be supported.  Likewise requests with `Accept` header of
+format MUST be supported. Likewise requests with `Accept` header of
 `application/json` MUST be supported.
 
 Unknown query parameters on the APIs MUST be ignored.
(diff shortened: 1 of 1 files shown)


erikerikson:
This hasn't updated the associated OpenAPI spec.  That should be changed before this is merged.

erikerikson:
Quick note: this change makes the collection and item methods inconsistent.  I believe that part of the original reason for the POST is that a new epoch of the service with the given ID is created.  It may be worth keeping as is.

duglin:
Approved on the 1/20 call

erikerikson:
I would have expected that updating the OpenAPI spec should have delayed this merge

duglin:
Crap - sorry I forgot about that. That's what I get for doing too many things at the same time. Let me create a 2nd PR...

discovery/spec.md
```diff
@@ -734,14 +734,14 @@ upon return of the `202 Accepted` from the original request.
...
 in the request will be completely replaced by the Service definition in the
-request.  The body of the request message MUST contain an array of zero or more
+request. The body of the request message MUST contain an array of zero or more
 Service entries.
 
 The `id` and `epoch` attributes of the Services MAY be omitted. In such cases
 the discovery endpoint MUST assign an appropriate value to the omitted
-field(s).  An appropriate `id` MUST be globally unique.  An appropriate `epoch`
+field(s). An appropriate `id` MUST be globally unique. An appropriate `epoch`
 MUST be greater than the any existing `epoch` for a service with the same `id`.
```
erikerikson, line 745: fix `the any` to be `the` or `any`?

duglin: fixed in #942

PR #930, opened 2022-01-05
Update processing of Epoch on all operations

### LINKED ISSUE #925: [discovery] DELETE /services/ID should support ?epoch=... as an optional field
We support ID+epoch on the multi-DELETE request but we don't for the single DELETE request, but it seems like we should. And since there's no body on the request, adding "epoch" to the URL as an optional query parameter seems like a way to do it.

duglin:
Based on last week's call I updated our "epoch" processing for the APIs to be more consistent:
- epoch is used for version checking. If present on the request then the current value in the DE must match
- PUT is used to update - not create
- POST is used for create or update (mainly for bulk update)
- import/sync'ing is no supported (and I removed any text that implies it)
- I removed the ability to specify the same Service more than once in a POST or DELETE

We can add import/sync support in another PR.

OLD:
Also:
- reject requests if the passed-in epoch value doesn't exactly match current value
- make DELETE idempotent by returning a 200 (instead of 404) of the Service is already deleted
- removed the Body on the `DELETE /services/ID` since we don't really need it if we have the query param
- reply should be singleton, not an array

Fixes #925

### DIFF
--- a/discovery/discovery.yaml
+++ b/discovery/discovery.yaml
@@ -34,7 +34,7 @@ paths:
         '400':
           description: Bad Request - constraint failure
         '409':
-          description: Conflict - epoch not greater
+          description: Conflict - epoch does not match
     delete:
       operationId: deleteServices
       requestBody:
@@ -42,12 +42,8 @@ paths:
       responses:
         '200':
           $ref: '#/components/responses/servicesdeleteresponse'
-        '400':
-          description: Bad Request - constraint failure
-        '404':
-          description: Not Found
         '409':
-          description: Conflict - epoch not greater
+          description: Conflict - epoch does not match
   /services/{id}:
     get:
       operationId: getService
@@ -76,9 +72,16 @@ paths:
         '404':
           description: Not Found
         '409':
-          description: Conflict - epoch not greater
+          description: Conflict - epoch does not match
     delete:
       operationId: deleteService
+      parameters:
+       - in: query
+         name: epoch
+         description: The epoch of the service to be deleted
+         required: false
+         schema:
+           $ref: '#/components/schemas/service/properties/epoch'
       parameters:
         - $ref: '#/components/parameters/id'
       requestBody:
@@ -86,12 +89,8 @@ paths:
       responses:
         '200':
           $ref: '#/components/responses/servicedeleteresponse'
-        '400':
-          description: Bad Request - constraint failure
-        '404':
-          description: Not Found
         '409':
-          description: Conflict - epoch not greater
+          description: Conflict - epoch does not match
 components:
   parameters:
     id:
--- a/discovery/discovery.yaml
+++ b/discovery/discovery.yaml
@@ -34,7 +34,7 @@ paths:
         '400':
           description: Bad Request - constraint failure
         '409':
-          description: Conflict - epoch does not match
+          description: Conflict - invalid epoch value
     delete:
       operationId: deleteServices
       requestBody:
@@ -43,7 +43,7 @@ paths:
         '200':
           $ref: '#/components/responses/servicesdeleteresponse'
         '409':
-          description: Conflict - epoch does not match
+          description: Conflict - invalid epoch value
   /services/{id}:
     get:
       operationId: getService
@@ -72,7 +72,7 @@ paths:
         '404':
           description: Not Found
         '409':
-          description: Conflict - epoch does not match
+          description: Conflict - invalid epoch value
     delete:
       operationId: deleteService
       parameters:
@@ -90,7 +90,7 @@ paths:
         '200':
           $ref: '#/components/responses/servicedeleteresponse'
         '409':
-          description: Conflict - epoch does not match
+          description: Conflict - invalid epoch value
 components:
   parameters:
     id:
--- a/discovery/discovery.yaml
+++ b/discovery/discovery.yaml
@@ -34,7 +34,7 @@ paths:
         '400':
           description: Bad Request - constraint failure
         '409':
-          description: Conflict - invalid epoch value
+          description: Conflict - epoch not greater
     delete:
       operationId: deleteServices
       requestBody:
@@ -42,8 +42,10 @@ paths:
       responses:
         '200':
           $ref: '#/components/responses/servicesdeleteresponse'
+        '400':
+          description: Bad Request - constraint failure
         '409':
-          description: Conflict - invalid epoch value
+          description: Conflict - epoch not greater
   /services/{id}:
     get:
       operationId: getService
@@ -56,6 +58,8 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/service'
+        '400':
+          description: Bad Request - constraint failure
         '404':
           description: Not Found
     put:
@@ -72,7 +76,7 @@ paths:
         '404':
           description: Not Found
         '409':
-          description: Conflict - invalid epoch value
+          description: Conflict - epoch not greater
     delete:
       operationId: deleteService
       parameters:
@@ -84,13 +88,13 @@ paths:
            $ref: '#/components/schemas/service/properties/epoch'
       parameters:
         - $ref: '#/components/parameters/id'
-      requestBody:
-        $ref: '#/components/requestBodies/servicedeleterequest'
       responses:
         '200':
           $ref: '#/components/responses/servicedeleteresponse'
+        '400':
+          description: Bad Request - constraint failure
         '409':
-          description: Conflict - invalid epoch value
+          description: Conflict - epoch not greater
 components:
   parameters:
     id:
@@ -122,13 +126,6 @@ components:
         application/json:
           schema:
             $ref: '#/components/schemas/service'
-    servicedeleterequest:
-      description: A request to remove the given service from the discovery endpoint's collection of services
-      required: false
-      content:
-        application/json:
-          schema:
-            $ref: '#/components/schemas/serviceinstance'
   responses:
     servicespostresponse:
       description: A list of the resulting Service values resulting from processing the request, in the same order as in the request
@@ -350,8 +347,6 @@ components:
           $ref: '#/components/schemas/changedelete'
         path:
           $ref: '#/components/schemas/changeservicepath'
-        request:
-          $ref: '#/components/requestBodies/servicedeleterequest'
         response:
           $ref: '#/components/responses/servicedeleteresponse'
       required: [operation, path, request, response]
--- a/discovery/spec.md
+++ b/discovery/spec.md
@@ -279,7 +279,7 @@ The following sections define the attributes that appear in a Service entity.
   Note: this value will most likely be Discovery Endpoint defined and therefore
   will most likely be ignored when provided by clients.
 - Constraints:
-  - MUST be read-only, Discovery Endpoint, manageg
+  - MUST be specified by the Discovery Endpoint and otherwise read-only
   - REQUIRED in responses from the Discovery Endpoint.
   - MUST be a non-empty URL
   - MUST end with `fsegments` (per RFC1738) of: `/services/{id}` where `id` is
@@ -813,7 +813,7 @@ The following rules apply to processing the Services specified in the request:
 
 There is no requirement that the incoming Services be processed in the
 order in which they appear. Constraints apply only to the result of the
-entire request. For eaxmple, if a Service with a name of `dog` exists and
+entire request. For example, if a Service with a name of `dog` exists and
 the incoming request is creating a new Service called `dog` as well as
 renaming the existing Service to `cat` then the request is expected to succeed
 even if it would cause two Services to temporarily have the same name.
@@ -870,11 +870,11 @@ This MUST delete all of the Services in the Discovery Endpoint that are
 contained in the request. If any of those Services cannot be deleted for any
 reason then the entire request MUST fail with no effect.
 
-If an `id` is not specified in any of the Services the entire request MUST
-fail. If no Service can be found that corresponds to a specified `id`, the
-endpoint MUST behave as though the Service was deleted. An `epoch` value MAY
-be omitted in the incoming Service.  If present and the request value is not
-greater than existing value, then the request MUST fail.
+If any of the Services in the incoming request are missing an `id` then the
+entire request MUST fail. If no Service can be found that corresponds to a
+specified `id`, the endpoint MUST behave as though the Service was deleted.
+An `epoch` value MAY be omitted in the incoming Service.  If present and the
+request value is not greater than existing value, then the request MUST fail.
 
 If other service attributes are included in the request, those SHOULD be
 ignored for the purposes of processing the request.
@@ -999,7 +999,9 @@ is no Service with the specified `id` then it is considered to have already
 been deleted and the request will have no change to the Discovery Endpoint.
 A missing Service is not considered an error condition.
 
-This specification does not define a Body for the request message.
+This specification does not define a Body for the request message and the
+presence of a Body (even with a Service definition, valid or not) MUST NOT
+result in an error.
 
 The request URL MAY include an OPTIONAL `epoch` query parameter, and if
 present and is not greater than the Service's current `epoch` value, then the
--- a/discovery/spec.md
+++ b/discovery/spec.md
@@ -214,6 +214,7 @@ The following sections define the attributes that appear in a Service entity.
   See the Primer for more information.
 
 - Constraints:
+  - MUST be immutable
   - REQUIRED in responses from the Discovery Endpoint.
   - MUST be a non-empty string
   - MUST conform with [RFC3986/3.3](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3) `segment-nz-nc` syntax.
@@ -275,8 +276,11 @@ The following sections define the attributes that appear in a Service entity.
   Discovery Endpoint. This value MUST be usable in subsequent requests, by
   authorized clients, to retrieve this Service
   entity.
+  Note: this value will most likely be Discovery Endpoint defined and therefore
+  will most likely be ignored when provided by clients.
 - Constraints:
-  - REQUIRED
+  - MUST be read-only, Discovery Endpoint, manageg
+  - REQUIRED in responses from the Discovery Endpoint.
   - MUST be a non-empty URL
   - MUST end with `fsegments` (per RFC1738) of: `/services/{id}` where `id` is
     the Service's `id` attribute.
@@ -558,7 +562,7 @@ The endpoints defined by this specification are broken into two categories:
 
 The relative paths specified below are NOT REQUIRED to be at the root of
 the `fpath` (per RFC1738). However, they are REQUIRED to match the end of it.
-For example, the follow are valid URLs/paths:
+For example, the following are valid URLs/paths:
 
 ```
 https://example.com/services
@@ -717,7 +721,7 @@ This MUST return the latest version of the Service with the given `{id}` if it
 exists.
 
 The following responses are defined by this specification:
-- `200 OK` and the Service representation in the HTTP Response Body.
+- `200 OK` and the Service representation in the HTTP response body.
 - `404 Not Found` if the Service does not exist.
 
 Other responses are allowed, but not defined by this specification.
@@ -750,7 +754,7 @@ to process the incoming request asynchronous then the following rules apply:
   request has been accepted but not processed yet.
 - The `202 Accepted` response MUST include a `Location` HTTP Header that
   points to a "status endpoint", which can be used to determine the status
-  of the original request. The HTTP Response body MAY be empty.
+  of the original request. The HTTP response body MAY be empty.
 - The `202 Accepted` response MAY include a `Retry-After` HTTP Header
   indicating the time, in seconds, that the sender SHOULD wait before
   querying the status endpoint.
@@ -760,7 +764,7 @@ original request. The following responses are defined:
 
 - `200 OK` indicates that the original request was successfully processed.
   - Unless otherwise noted for a specific operation, the response MUST also
-    include the HTTP Headers and HTTP Response Body that is defined for the
+    include the HTTP Headers and HTTP response body that is defined for the
     original operation as if the response were sent synchronously.
 - `202 Accepted` indicates that the original request is still being processed.
   The response body MAY be empty.
@@ -782,8 +786,8 @@ contain all Services in the request or fail. The body of the request message
 MUST contain an array of zero or more Service entries.
 
 The following rules apply to processing the Services specified in the request:
-- any error processing the request MUST result in the entire operation failing
-  and no updates to the Discovery Endpoint are to happen.
+- any error processing the request MUST result in the request having no
+  effect on the Discovery Endpoint.
 - a Service (as identified by its `id`) MUST NOT appear more than once in the
   request.
 - any existing Service (as identified by its `id`) MUST be completely
@@ -791,42 +795,46 @@ The following rules apply to processing the Services specified in the request:
   This means that any missing property in the incoming request MUST be
   treated as a request to delete the property from the existing Service
   definition.
-- if the incoming Service has an `epoch` value then the request MUST fail if
+- if an incoming Service does not contain an `id` then the Discovery Endpoint
+  MUST assign an appropriate value. The value MUST be globally unique.
+- if there is no existing Service with the given `id` then a new Service MUST
+  be created with that `id`.
+- if an incoming Service has an `epoch` value then the request MUST fail if
   that value is not greater than the current Service's `epoch` value.
-- if the incoming Service does not have an `epoch` value then the Service's
+- if an incoming Service does not have an `epoch` value then the Service's
   `epoch` value MUST be set to a valid value of the Discovery Endpoint's
   choosing. This means that if there is an existing Service with the same
   `id` then the value MUST be set to a value greater than the current value.
-- if the incoming Service does not contain an `id` then the Discovery Endpoint
-  MUST assign an appropriate value. The value MUST be globally unique.
-- aside from `id`, `epoch` and `url`, unless there is an out of band agreement,
+- aside from `id` and `epoch`, unless there is an out of band agreement,
   all mandatory attributes MUST be present in the request, and a Discovery
   Endpoint MUST reject requests that are missing such attributes.
-- the Discovery Endpoint MUST ensure that the `url` of the resulting Service
-  is correct even if that means overriding the value from the request.
+- the Discovery Endpoint MUST ignore attempts to modify read-only or immutable
+  attributes - such as `id` and `url`.
 
 There is no requirement that the incoming Services be processed in the
-order in which they appear. However, any constraint/consistency checks
-(e.g. `name` uniqueness checking) MUST only be performed after all of the
-incoming Services have been processed.
+order in which they appear. Constraints apply only to the result of the
+entire request. For eaxmple, if a Service with a name of `dog` exists and
+the incoming request is creating a new Service called `dog` as well as
+renaming the existing Service to `cat` then the request is expected to succeed
+even if it would cause two Services to temporarily have the same name.
 
-Upon successfully processing the request, the HTTP Response Body MUST include
-an array of the Services that were created, or updated, as a result of
+Upon successfully processing the request, the HTTP response body MUST be
+an array of the Services that were created or updated as a result of
 processing the request. The order of the array MUST match the order of the
 incoming Service array. Each Service in the response SHOULD include the
 current value for all attributes, even if the Discovery Endpoint added or
-modified some values during the time of processing of the request.
+modified some values during processing of the request.
 
-The follow responses are defined by this specification:
+The following responses are defined by this specification:
 - `200 OK` if all the specified Services were processed successfully.
 - `400 Bad Request` if the first error encountered is a constraint failure
 - `409 Conflict` if the first error encountered is that a given `epoch` was
-  invalid. The Response Body SHOULD include information about which Service
+  invalid. The response body SHOULD include information about which Service
   caused the error.
 
 Other responses are allowed, but not defined by this specification.
 
-The format of the HTTP Request MUST adhere to the following:
+The format of the HTTP request MUST adhere to the following:
 ```
 Content-Type: application/json
 
@@ -839,7 +847,7 @@ Content-Type: application/json
 ]
 ```
 
-The format of the `200 OK` HTTP Response MUST adhere to the following:
+The format of the `200 OK` HTTP response MUST adhere to the following:
 ```
 200 OK
 Content-Type: application/json
@@ -848,6 +856,7 @@ Content-Type: application/json
   {
     "id": "{id}",
     "epoch": "{epoch}",
+    "url": "{url}",
     "name": "{name},
     ... remainder of Service attributes ...
   },
@@ -863,10 +872,9 @@ reason then the entire request MUST fail with no effect.
 
 If an `id` is not specified in any of the Services the entire request MUST
 fail. If no Service can be found that corresponds to a specified `id`, the
-processing MUST ignore that Service and not fail as a result of the Service
-already being deleted. An `epoch` value MAY be omitted in the incoming Service.
-If present, and is not greater than the Service's current `epoch` value, then
-the request MUST fail.
+endpoint MUST behave as though the Service was deleted. An `epoch` value MAY
+be omitted in the incoming Service.  If present and the request value is not
+greater than existing value, then the request MUST fail.
 
 If other service attributes are included in the request, those SHOULD be
 ignored for the purposes of processing the request.
@@ -874,7 +882,7 @@ ignored for the purposes of processing the request.
 There is no requirement that the Services in the request are processed in
 the order in which they appear.
 
-Upon successfully processing the request, the HTTP Response Body MUST include
+Upon successfully processing the request, the HTTP response body MUST be
 an array of Services that match the order of the array of Services specified
 in the request. The Services in the response MUST include at least the `id`
 attribute and SHOULD include the remaining Service attributes as they existed
@@ -882,16 +890,16 @@ immediately prior to the Service being deleted, if possible. If the Service
 was deleted prior the processing of this request the Service values might not
 be available to be returned.
 
-If the response Services include an `epoch` value then each MUST be greater
-than the Service's value prior to the delete operation.
+If the response Services include `epoch` values then they MUST be the values
+present in each Service at the time they were deleted.
 
-The follow responses are defined by this specification:
+The following responses are defined by this specification:
 - `200 OK` if the request was successfully processed.
 - `409 Conflict` if the first error encountered is that a given `epoch` was
-  invalid.  The Response Body SHOULD include information about which Service
+  invalid.  The response body SHOULD include information about which Service
   caused the error.
 
-The format of the HTTP Request MUST adhere to the following:
+The format of the HTTP request MUST adhere to the following:
 ```
 Content-Type: application/json
 
@@ -904,7 +912,7 @@ Content-Type: application/json
 ]
 ```
 
-The format of the `200 OK` HTTP Response MUST adhere to the following:
+The format of the `200 OK` HTTP response MUST adhere to the following:
 ```
 200 OK
 Content-Type: application/json
@@ -930,6 +938,9 @@ completely replaced by the Service definition, except where noted below.
 The body of the request message MUST contain a Service with an `id`
 attribute matching the `{id}` in the path.
 
+If there is no existing Service with the given `id` then a new Service MUST
+be created with that `id`.
+
 The `epoch` attribute of the Service MAY be omitted. In such cases the
 discovery endpoint MUST assign a value that is greater than the existing
 `epoch` value for the Service.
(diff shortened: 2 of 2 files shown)


discovery/spec.md
```diff
@@ -820,8 +820,8 @@ exist) then the entire request MUST fail with no effect.
 If an `id` is not specified in any of the Services the entire request MUST
 fail.  If no Service can be found that corresponds to a specified `id`, the
 entire request MUST fail.  An `epoch` value MAY be omitted.  If a Service
-corresponding to a specified `id` has a higher recorded `epoch` value than the
-one given the entire request MUST fail.
+corresponding to a specified `id` has a different recorded `epoch` value than
```
erikerikson, line 823: This changes the behavior that the client cannot supply the value of an epoch.  This may be fine, just being explicit.  Consider the case where the epoch is a sequence ID or monotonic offset from a log.

discovery/spec.md
```diff
@@ -830,15 +830,14 @@ Service deletions MUST be processed in the order in which they are given. Thus,
 if a Service is specified multiple times the entire request MUST fail.
 
 The follow responses are defined by this specification:
-- `200 OK` if all the specified Services were deleted successfully.
+- `200 OK` if all the specified Services were deleted successfully or are not
```
erikerikson, line 833: perhaps "`200 OK` if none of the specified Services remain in the Discovery Endpoint."

discovery/spec.md
```diff
@@ -924,52 +923,29 @@ Content-Type: application/json
...
 This MUST delete the Service at the referenced URL.
 
-The request MAY include a body.  Any such body MUST be an object containing an
-`id` matching the `{id}` in the path.  That object MAY include an `epoch`
-value.  If the Service corresponding with the specified `id` has a higher
-recorded `epoch` value than given, the request MUST fail.
-
-If other service attributes are included in the request, those MUST be
-ignored for the purposes of processing the request.
+The request URL MAY include an OPTIONAL `epoch` query parameter, and if
```
erikerikson, line 926: This change reduces the consistency with bulk delete.  Not inherently a problem but the original form may be cleaner for clients.

discovery/spec.md
```diff
@@ -924,52 +923,29 @@ Content-Type: application/json
...
-- `409 Conflict` if the given epoch was less than or equal to the existing
-  epoch of the Service with the given `id`
+- `409 Conflict` if the given epoch was not equal to the existing epoch of the
+  Service with the given `id`
 
 Other responses are allowed, but not defined by this specification.
 
-The format of the OPTIONAL HTTP Request body MUST adhere to the following:
-```
-Content-Type: application/json
```
erikerikson, line 949: We have consistently supplied an example body, why remove it?

discovery/spec.md
```diff
@@ -924,52 +923,29 @@ Content-Type: application/json
...
 
-[
-  {
-    "url": "{url}",
-    "name": "{name}",
-    ... remainder of Service attributes ...
-  }
-  ...
-]
+{
```
erikerikson, line 944: this too reduces consistency.

erikerikson (commented):
Minor comments, I am willing for them to be ignored.  Sorry for timing, this was cached in the browser.

discovery/spec.md
duglin: does the value need to be higher to retain the semantics we intended? I was assuming that the purpose of including the epoch was to ensure that the client was issuing the delete() on the version they think is the latest. And if the latest.epoch != client.epoch then they wanted the delete to fail - regardless of whether the client's is old (or even newer). Is that not your understanding?

discovery/spec.md
duglin: which consistency are you thinking of?

discovery/spec.md
duglin: done

discovery/spec.md
erikerikson: Bulk deletes (i.e. on /services) have callers specifying `id`s in the body.  This change means this endpoint doesn't explicitly support the same convention.

discovery/spec.md
erikerikson: Epoch is defined [earlier in the doc](https://github.com/cloudevents/spec/blob/b6f972af0502cd3d38291e63ee3ab9ae995e6a22/discovery/spec.md#epoch):

> For example, implementations are free to use a value that represents a timestamp or could choose to simply use a monotonically increasing number. The only requirement is that the value MUST always increase each time the Service Entry is updated.

The positive monotonicity of the epoch has been part of the definition for some time.  The question is something like whether the semantic of the epoch given is "here's the new epoch to use" or "here's the epoch I am targeting".  #928 seems to also have changed this semantic.  If you look under POST the convention that has been previously followed across these management APIs was that the caller was supplying the value for the epoch the record update was occurring under.

erikerikson: If we are going to change it, perhaps we do this as  part of a separate PR and across the endpoints?

duglin: > The question is something like whether the semantic of the epoch given is "here's the new epoch to use" or "here's the epoch I am targeting"

but we're talking about delete() here, not update.

Let me ask this... what do you consider to be the purpose behind providing an epoch on delete() ? By providing it the client is asking for some kind of check, otherwise they wouldn't include it at all and the service would always be deleted.
So, It seems like there are 3 possible conditions:
1 - provided epoch is less than the DE service's epoch's value. Which means the client had an old version. And the only possible result here is to reject the request because it's stale. If the user wanted some other result to happen then it shouldn't have included the epoch at all.
2 - provided epoch is the same as the DE service's epoch value. The result here must be to delete it.
3 - provided epoch is higher than the DE service's epoch value. Technically this should never happen unless they just made up a value, which is probably leaving the scope of "proper usage of the spec". But if we had to choose an action, I'd say we should reject the delete() to be on the safe side.

So, if we really only need to worry about 1 and 2, then we're really only talking about "are the epoch values different" - aren't we?

discovery/spec.md
duglin: True but remember, this variant of the delete() in today's spec doesn't have a body either and the service in question is referenced in the URL - which means we're already using a different convention.

In general, while having a body on a delete() is REST legal, it is a bit unconventional IMO. And I thought that's one of the reasons we have this variant in the spec - ie. so people could do things as easily as `curl .../services/id` and nothing more. In that mode it's very natural to include more info in the URL. Having to add a body for just a single field adds quite a bit of complexity to an otherwise simple request.  And, TBH, I don't think people are going to be looking for consistency between these two. If they want consistency between a singleton delete() and a multi-service delete() then they're still free to use the other variant of delete() with a single service in the body.

discovery/spec.md
duglin: See the comment above. In this variant of delete() where the ID is in the URL, the body serves no purpose other than to carry the epoch since the URL's ID and the body always the same.

discovery/spec.md
duglin: See comment above. Consistency can be achieved (if people want it) by using the "DELETE /services" with a single Service in the body.

duglin: The client is deleting a single resource, so it doesn't make sense to return an array

discovery/spec.md
erikerikson: > but we're talking about delete() here, not update.

Yes.  Have we stopped supporting soft deletes?

erikerikson: > 3 - provided epoch is higher than the DE service's epoch value. Technically this should never happen unless they just made up a value, which is probably leaving the scope of "proper usage of the spec"

It's not clear to me that using the offset of an event on a particular partition of a distributed log as the epoch is "made up" or an improper use of the service.  In fact, I can imagine a fair bit of utility to using that value.

discovery/spec.md
erikerikson: That is accurate.

I don't mean to try and block this by noting the reduced consistency, I am merely noting a cost.

discovery/spec.md
duglin: > Have we stopped supporting soft deletes?

We don't have soft deletes in the spec right now. We've talked about them but I don't think we ever added them.

discovery/spec.md
erikerikson: Adding the param seems like a win, especially for use cases involving curl or just a browser.  It's not clear that it is as natural to use one calling convention for one case and another for the other when writing code.

WRT consistency, I can accept that as a judgement call that is not clearly superior either way.

discovery/spec.md
erikerikson: You are right that we don't discuss soft deletes in the spec.  I [originally included but then removed it](https://github.com/cloudevents/spec/pull/837/commits/3846c460bba31f64e33098f296df026bc75bed69) when I was cleaning up this API in preparation for the history PR.  As I recall the discussion we decided that by not mentioning it we were not prohibiting (i.e. allowing) the practice without having to get into the complexity of its ramifications of specifying the behavior.

discovery/spec.md
erikerikson: I understand your intention better now.  I still think that at times when coding that it's easier to use the body but I don't care much.  As you note, so long as the semantics of PUT and POST are mostly the same using POST would be fine to submit single records.  If the semantics of PUT will change, this is more relevant even if not terribly important since its just a form factor matter.

erikerikson:
Additional point - this PR is missing modifications to the OpenAPI spec even though it changes the defintiion.  Please synchronize once the definition is discussed and clarified.

duglin:
We approved this on the 1/20 call but after the call @erikerikson raised some concerns so we should discuss those before we merge this. He and I chatted off-line and produced a gdoc to facilitate the future discussions around how to deal with epoch:  https://docs.google.com/document/d/14kKnAzV15-QWlL7iRXcUVu5MydqYhh1zIJ608H-x2qk/edit#

discovery/spec.md
n3wscott: I prefer the old wording.

Exact epoch match seems like a larger burden for the client, I think there is a valid case for the client having a more up to date catalog than the server and wanting to delete the server's copy to update? This way requires a refetch and is racy.

duglin: @n3wscott can you elaborate on the scenario? When would a client have a higher value? And even if they did, and if they didn't care, wouldn't they just leave off the epoch all together?

erikerikson: Scott can speak for himself but replication seems like a scenario where the client replicating the later version of the record necessarily holds a higher epoch (because the replication target has last version before receiving the update).

duglin: yup that scenario crossed my mind too. However, it seems to me that if we want to ensure that we support replication of Discovery Endpoints then we have a lot more stuff to do beyond DELETE. As of now the spec doesn't address it at all so I'm more focused on the "this DE is the single source of truth" scenario, and in that case a client with a higher epoch is more like an error.

I will say however that during our other epoch discussion we might quickly get into the replication scenario - I just didn't want to jump the gun or do a partial solution.

duglin:
I updated this PR based on last week's call. Please see the first comment in the PR for the latest info.

I opened https://github.com/cloudevents/spec/issues/951 so we can discuss import/sync support next.

erikerikson:
Please make sure to reflect these changes in the OpenAPI spec.

duglin:
Sorry - I forgot to push the commit.
Tried to update the openapi spec too

erikerikson:
Thanks, I'll take a look through

discovery/spec.md
```diff
@@ -783,54 +783,47 @@ upon return of the `202 Accepted` from the original request.
...
-specified Services in the incoming request were already successfully processed.
-
-For example, if a Discovery Endpoint has a Service called `MyService`, it is
-possible to rename that Service to `YourService` and then to add, or update,
-another Service to use `MyService` as its `name`.
-
-Likewise, while it might be less than optimal, it is possible for
-a request to update a Service multiple times if the same `id` is used more
-than once within a request.
+contain all Services in the request or fail. The body of the request message
```
erikerikson, line 786: This seems like a pretty big shift in the behavior of this endpoint.  More specifically, it seems, by 798 and 803 that import is blocked and I don't think we discussed or found agreement on eliminating that use case.

discovery/spec.md
```diff
@@ -855,98 +847,48 @@ Content-Type: application/json
...
-```
 
 #### `PUT /services/{id}`
 
-This MUST update the collection of Services in the Discovery Endpoint to
-contain the Service in the request. If the Service identified in the
-request exists it MUST be completely replaced by the Service definition in the
-request. If the Service does not exist then a new Service MUST be created.
+This MUST update the Service referenced by the specified URL.
+If the Service identified by the specified URL does not exist then the
```
erikerikson, line 862: I don't recall discussing that only updates would be supported by PUT

discovery/spec.md
```diff
@@ -974,31 +916,71 @@ Content-Type: application/json
...
+Content-Type: application/json
+
+{
+  "url": "{url}",
+  "name": "{name}",
+  ... remainder of Service attributes ...
+}
+```
+
+#### `DELETE /services`
```
erikerikson, line 951: Usually interactions with the collection are grouped together.  I would prefer this move this back up between POST and PUT.

erikerikson (changes requested):
There are some rather big changes here that seem inconsistent with [your comment](https://github.com/cloudevents/spec/pull/930#issuecomment-1033276178).  During the call we discussed the semantics of `epoch` but in this latest version of the PR you are changing the semantics and handling of `id` as well.  I would suggest reducing the scope of the changes to what was discussed and agreed.

discovery/spec.md
duglin: I think I mentioned on the call that I removed import logic because I wanted to do it in a follow-on PR. I wanted to focus on the simple cases first.  Is there another aspect that's part of the "big shift" you're thinking of?

duglin:
I've update https://docs.google.com/document/d/14kKnAzV15-QWlL7iRXcUVu5MydqYhh1zIJ608H-x2qk/edit with a new table at the bottom to show my later proposal. Cells in yellow should be very carefully reviewed.

FYI @erikerikson

discovery/spec.md
```diff
@@ -964,61 +980,51 @@ Content-Type: application/json
...
+  "epoch": "{epoch}",
   ... remainder of Service attributes ...
 }
 ```
 
-#### `DELETE /services/{id}`
+#### `DELETE /services/{id}[?epoch={epoch}]`
 
-This MUST delete the Service at the referenced URL.
+This MUST delete the Service as indetified by the `id` in the URL. If there
```
erikerikson, line 990:
```suggestion
This MUST delete the Service as identified by the `id` in the URL. If there
```

discovery/spec.md
```diff
@@ -964,61 +980,51 @@ Content-Type: application/json
...
+to be returned.
 
-Other responses are allowed, but not defined by this specification.
+If the response Services include an `epoch` value then each MUST be greater
+than the Service's value prior to the delete operation.
 
-The format of the OPTIONAL HTTP Request body MUST adhere to the following:
-```
-Content-Type: application/json
+The follow responses are defined by this specification:
```
erikerikson, line 1012:
```suggestion
The following responses are defined by this specification:
```
probably my fault...

erikerikson (commented):
Mostly nits (sorry) but it feels like we're getting close.

A few questions that could lead to changes and a few minor change suggestions

discovery/spec.md
```diff
@@ -964,61 +980,51 @@ Content-Type: application/json
...
+been deleted and the request will have no change to the Discovery Endpoint.
+A missing Service is not considered an error condition.
 
-The request MAY include a body. Any such body MUST be an object containing an
-`id` matching the `{id}` in the path. That object MAY include an `epoch`
-value. If the Service corresponding with the specified `id` has a higher
-recorded `epoch` value than given, the request MUST fail.
+The request URL MAY include an OPTIONAL `epoch` query parameter, and if
+present and is not greater than the Service's current `epoch` value, then the
+request MUST fail.
```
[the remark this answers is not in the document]
duglin: hmmm but there's no body defined for this operation. Added a sentence to make that clear.

discovery/spec.md
```diff
@@ -778,54 +778,51 @@ upon return of the `202 Accepted` from the original request.
...
+  definition.
+- if the incoming Service has an `epoch` value then the request MUST fail if
+  that value is not greater than the current Service's `epoch` value.
+- if the incoming Service does not have an `epoch` value then the Service's
+  `epoch` value MUST be set to a valid value of the Discovery Endpoint's
+  choosing. This means that if there is an existing Service with the same
+  `id` then the value MUST be set to a value greater than the current value.
+- if the incoming Service does not contain an `id` then the Discovery Endpoint
+  MUST assign an appropriate value. The value MUST be globally unique.
+- aside from `id`, `epoch` and `url`, unless there is an out of band agreement,
```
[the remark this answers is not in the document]
duglin: The `url` is the URL of the service we're updating so I'm expecting the Discovery Endpoint to set it appropriately. If the client sets it to "ibm.com" I don't think the DE should accept it since we know that's not correct.  And in cases where the DE will pick the ID, there's no way for the client to pass it in. Just seemed easier to have the DE ignore it.

discovery/spec.md
```diff
@@ -778,54 +778,51 @@ upon return of the `202 Accepted` from the original request.
...
+  all mandatory attributes MUST be present in the request, and a Discovery
+  Endpoint MUST reject requests that are missing such attributes.
+- the Discovery Endpoint MUST ensure that the `url` of the resulting Service
+  is correct even if that means overriding the value from the request.
```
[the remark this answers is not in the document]
duglin: It needs to put in there a URL that will point to the Service being touched. It should know it's own URL... if nothing else it's the related to the base URL from this request.

discovery/spec.md
```diff
@@ -778,54 +778,51 @@ upon return of the `202 Accepted` from the original request.
...
+
+There is no requirement that the incoming Services be processed in the
+order in which they appear. However, any constraint/consistency checks
+(e.g. `name` uniqueness checking) MUST only be performed after all of the
+incoming Services have been processed.
```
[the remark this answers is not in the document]
duglin: true - is there a better wording?

discovery/discovery.yaml
```diff
@@ -76,22 +72,25 @@ paths:
         '404':
           description: Not Found
         '409':
-          description: Conflict - epoch not greater
+          description: Conflict - invalid epoch value
     delete:
       operationId: deleteService
+      parameters:
```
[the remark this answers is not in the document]
duglin: nope - I removed the body definition

discovery/spec.md
```diff
@@ -850,51 +846,59 @@ Content-Type: application/json
...
+Upon successfully processing the request, the HTTP Response Body MUST include
+an array of Services that match the order of the array of Services specified
+in the request. The Services in the response MUST include at least the `id`
+attribute and SHOULD include the remaining Service attributes as they existed
+immediately prior to the Service being deleted, if possible. If the Service
+was deleted prior the processing of this request the Service values might not
+be available to be returned.
+
+If the response Services include an `epoch` value then each MUST be greater
+than the Service's value prior to the delete operation.
```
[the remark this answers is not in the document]
duglin: I struggled with this because in the PUT/POST cases the returned Service contains the new epoch value, so from that perspective it seemed like this one should too. But it did feel weird to write that sentence :-)
Let's go with your semantics and see who complains. I don't like my new working... see what you think

duglin:
@erikerikson i think I got all of your edits... see what you think

discovery/spec.md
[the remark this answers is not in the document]
erikerikson: There used to be and I would have thought there still would be.  This wouldn't block use but would constrain the DX to the curl side rather than code side.

discovery/spec.md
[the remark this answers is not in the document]
erikerikson: I have no objection to the language I just read and wouldn't block it.  However, just above that is ``The Services in the response MUST include at least the `id` attribute and SHOULD include the remaining Service attributes as they existed immediately prior to the Service being deleted, if possible. `` which I think already creates clarity.  I would recommend removal.

discovery/spec.md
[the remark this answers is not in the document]
erikerikson: I see.  I had not noticed those semantics and filled in with an incorrect understanding.  I understand why you've put this in here (thanks).  I think we need to clarify this in [the definition of the `url` attribute](https://github.com/cloudevents/spec/blob/653aa6424f62c55ae3b00c0a589ed8435381169c/discovery/spec.md#url).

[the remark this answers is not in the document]
erikerikson: To clarify slightly, this optionality of the attribute seems proper but also in contradiction with the attribute definition.  This isn't your fault but it currently says "REQUIRED" as opposed to what `id` says "REQUIRED in responses from the Discovery Endpoint."

discovery/spec.md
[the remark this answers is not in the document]
erikerikson: As noted, I lost the semantic of `url`.  Thanks for noticing and reminding everyone of the right semantic here.

discovery/spec.md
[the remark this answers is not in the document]
erikerikson: always a good question... maybe:
```suggestion
There is no requirement that the incoming Services be processed in the
order in which they appear. Constraints apply only to the result of 
the request.  For example, if a Service with `name: "dog"` exists
and a request creates a new Service with `name: "dog"` and later in the
request updates the existing Service to have `name: "cat"` the request
is expected to succeed even if the Discovery Endpoint implementation
would cause two services to temporarily have `name: "dog"`.
```
The example is long so I would be fine if it were just omitted.

discovery/discovery.yaml
```diff
@@ -42,10 +42,6 @@ paths:
       responses:
         '200':
           $ref: '#/components/responses/servicesdeleteresponse'
-        '400':
-          description: Bad Request - constraint failure
```
erikerikson, line 46: I missed this.  I think we still want to reject any `id`s with bad formatting or any other request schema validation error.  Same below too.

discovery/spec.md
```diff
@@ -778,135 +778,143 @@ upon return of the `202 Accepted` from the original request.
...
-exist) then the entire request MUST fail with no effect.
+reason then the entire request MUST fail with no effect.
 
 If an `id` is not specified in any of the Services the entire request MUST
 fail. If no Service can be found that corresponds to a specified `id`, the
-entire request MUST fail. An `epoch` value MAY be omitted. If a Service
-corresponding to a specified `id` has a higher recorded `epoch` value than the
-one given the entire request MUST fail.
+endpoint MUST behave as though the Service was deleted. An `epoch` value MAY
+be omitted in the incoming Service.  If present and request value is not
```
erikerikson, line 870:
```suggestion
be omitted in the incoming Service.  If present and the request value is not
```

erikerikson:
Looking great Doug.  I've gone through and resolved most of my comments.  There are a few where you'll note I've commented further.  Did you have a thought on [this comment](https://github.com/cloudevents/spec/pull/930#discussion_r833909634)?

discovery/spec.md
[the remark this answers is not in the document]
duglin: I decided to attack this from a different angle. I made `id` immutable and `url` read-only. Then I could talk about how these types of attributes can not be modified by clients w/o calling these two out as special in the PUT/POST. See what you think.

discovery/spec.md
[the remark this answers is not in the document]
duglin: I tweaked it slightly - see watcha think

discovery/spec.md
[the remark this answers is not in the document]
duglin: Can you elaborate? Are you saying that not having a Body is problematic? Right now the spec allows for a Body but we basically ignore it as it's totally unneeded.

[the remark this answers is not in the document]
erikerikson: As currently written, the body is potentially the exclusive source of the epoch.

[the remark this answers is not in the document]
erikerikson: Not to ignore that you want to support curl and the like.  That seems reasonable as it gives people writing curl statements a better experience.

If you're writing code and you've already written code against the POST endpoint then you have the code to write the JSON into the body and could reuse that.  Often that feels like the more appropriate way to do things: "here, here's the thing I want deleted".

discovery/spec.md
```diff
@@ -275,8 +276,11 @@ The following sections define the attributes that appear in a Service entity.
   Discovery Endpoint. This value MUST be usable in subsequent requests, by
   authorized clients, to retrieve this Service
   entity.
+  Note: this value will most likely be Discovery Endpoint defined and therefore
+  will most likely be ignored when provided by clients.
 - Constraints:
-  - REQUIRED
+  - MUST be read-only, Discovery Endpoint, manageg
```
erikerikson, line 282:
```suggestion
  - MUST be specified by the Discovery Endpoint and otherwise read-only
```

discovery/spec.md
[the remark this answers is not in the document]
erikerikson: Interesting.  It does attack the problem from another angle.  I like the new language.  I also think the API specification would benefit from calling out how this behavior shows up in POST and PUT.  I won't push hard on that point.

discovery/spec.md
```diff
@@ -778,135 +782,145 @@ upon return of the `202 Accepted` from the original request.
...
+  `id` then the value MUST be set to a value greater than the current value.
+- aside from `id` and `epoch`, unless there is an out of band agreement,
+  all mandatory attributes MUST be present in the request, and a Discovery
+  Endpoint MUST reject requests that are missing such attributes.
+- the Discovery Endpoint MUST ignore attempts to modify read-only or immutable
+  attributes - such as `id` and `url`.
+
+There is no requirement that the incoming Services be processed in the
+order in which they appear. Constraints apply only to the result of the
+entire request. For eaxmple, if a Service with a name of `dog` exists and
```
erikerikson, line 816:
```suggestion
entire request. For example, if a Service with a name of `dog` exists and
```

discovery/spec.md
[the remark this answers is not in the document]
erikerikson: The new wording is great.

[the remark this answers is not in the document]
erikerikson: The new wording is great.

discovery/spec.md
erikerikson, line 816:
```suggestion
entire request. For example, if a Service with a name of `dog` exists and
```

erikerikson, line 816:
```suggestion
entire request. For example, if a Service with a name of `dog` exists and
```

discovery/spec.md
duglin: much nicer-thanks

discovery/spec.md
```diff
@@ -778,54 +778,51 @@ upon return of the `202 Accepted` from the original request.
...
+  replaced by the Service definition in the request, except as noted below.
+  This means that any missing property in the incoming request MUST be
+  treated as a request to delete the property from the existing Service
+  definition.
+- if the incoming Service has an `epoch` value then the request MUST fail if
+  that value is not greater than the current Service's `epoch` value.
+- if the incoming Service does not have an `epoch` value then the Service's
+  `epoch` value MUST be set to a valid value of the Discovery Endpoint's
+  choosing. This means that if there is an existing Service with the same
+  `id` then the value MUST be set to a value greater than the current value.
```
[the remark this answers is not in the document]
erikerikson: I think so.  You're discussing the endpoint behavior when an epoch isn't defined.  You specify its behavior when there is an existing service with the same ID but not what happens when there isn't an existing service with that ID.  It'll be fine without.

discovery/spec.md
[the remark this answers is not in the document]
erikerikson: My request was only syntactic sugar.  I think the code reuse goes differently than described but I'm okay with not supporting a body.  It's simpler.  If we allow it I could see people trying to supply an epoch there (and not in the URL params) so ignoring it comes at some potential impact.

discovery/spec.md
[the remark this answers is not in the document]
erikerikson: It does.  It also covered the case of the existing service but no supplied epoch.  As noted already, I'm fine not adding it but it seemed in first read that if we were going to call out the existing service case that it might be appropriate to also call out the handling when there isn't.

discovery/spec.md
```diff
@@ -778,135 +782,145 @@ upon return of the `202 Accepted` from the original request.
...
+- if an incoming Service has an `epoch` value then the request MUST fail if
+  that value is not greater than the current Service's `epoch` value.
+- if an incoming Service does not have an `epoch` value then the Service's
+  `epoch` value MUST be set to a valid value of the Discovery Endpoint's
+  choosing. This means that if there is an existing Service with the same
+  `id` then the value MUST be set to a value greater than the current value.
+- aside from `id` and `epoch`, unless there is an out of band agreement,
+  all mandatory attributes MUST be present in the request, and a Discovery
+  Endpoint MUST reject requests that are missing such attributes.
+- the Discovery Endpoint MUST ignore attempts to modify read-only or immutable
```
jskeet, line 811: I'm not sure we need to say that you can't modify an `id`. How could you, given that the `id` is used to identify the service you're trying to update?

discovery/spec.md
```diff
@@ -778,135 +782,145 @@ upon return of the `202 Accepted` from the original request.
...
-- `200 OK` if all the specified Services were processed successfully
-  - The HTTP Response Body MUST include an array of the Service values
-    resulting from the successful processing of the request. The order of
-    those values MUST match the order of the Services in the request. The
-    response MAY include attributes and values which were added by the
-    Discovery Endpoint.
+contain all Services in the request or fail. The body of the request message
+MUST contain an array of zero or more Service entries.
+
+The following rules apply to processing the Services specified in the request:
```
jskeet, line 788: Do we want a constraint saying that any given `id` can't occur more than once in a request?

discovery/spec.md
```diff
@@ -778,135 +782,145 @@ upon return of the `202 Accepted` from the original request.
...
+  This means that any missing property in the incoming request MUST be
+  treated as a request to delete the property from the existing Service
+  definition.
+- if an incoming Service does not contain an `id` then the Discovery Endpoint
+  MUST assign an appropriate value. The value MUST be globally unique.
+- if there is no existing Service with the given `id` then a new Service MUST
+  be created with that `id`.
+- if an incoming Service has an `epoch` value then the request MUST fail if
+  that value is not greater than the current Service's `epoch` value.
+- if an incoming Service does not have an `epoch` value then the Service's
```
jskeet, line 804: Am I right in saying that if an incoming service *does* have an `epoch` value, and a service with the same `id` already exists, the `epoch` in the request must be strictly greater than the existing one? I'm pretty sure that's the case, and it would be a repetition of earlier rules, but it might be worth restating.

discovery/spec.md
```diff
@@ -778,135 +782,145 @@ upon return of the `202 Accepted` from the original request.
...
+
 #### `DELETE /services`
 
 This MUST delete all of the Services in the Discovery Endpoint that are
 contained in the request. If any of those Services cannot be deleted for any
-reason (e.g. an `id` is not given or a Service with the given `id` does not
-exist) then the entire request MUST fail with no effect.
+reason then the entire request MUST fail with no effect.
 
 If an `id` is not specified in any of the Services the entire request MUST
```
jskeet, line 873: This wording is ambiguous. I suspect it's intended to mean "Every service in the request must have an id" - but at the moment it could also be read as "At least one service in the request must have an id".

discovery/spec.md
```diff
@@ -917,108 +931,111 @@ Content-Type: application/json
 
 #### `PUT /services/{id}`
 
-This MUST update the collection of Services in the Discovery Endpoint to
-contain the Service in the request. If the Service identified in the
-request exists it MUST be completely replaced by the Service definition in the
-request. If the Service does not exist then a new Service MUST be created.
+This MUST update, or create, the Service referenced by the specified URL.
```
jskeet, line 934: Given that a service has a URL as well, I think this would be clearer as:

> This MUST update, or create, the Service identified by the `{id}` in the path.

(Or "as identified by the `id` in the URL" which is the wording for DELETE.)

discovery/spec.md
```diff
@@ -917,108 +931,111 @@ Content-Type: application/json
...
+`epoch` value for the Service.
+
+If an `epoch` is given in the request and it is not greater than the current
+Service's value, then the request MUST fail.
+
+Aside from `id` and `epoch`, unless there is an out of band agreement,
+all mandatory attributes MUST be present in the request, and a Discovery
+Endpoint MUST reject requests that are missing such attributes.
+
+The Discovery Endpoint MUST ignore attempts to modify read-only or immutable
```
jskeet, line 955: Again, there's no way of expressing an intention to modify `id`, so I wouldn't mention that here.

discovery/spec.md
```diff
@@ -917,108 +931,111 @@ Content-Type: application/json
...
-#### `DELETE /services/{id}`
+#### `DELETE /services/{id}[?epoch={epoch}]`
+
+This MUST delete the Service as identified by the `id` in the URL. If there
+is no Service with the specified `id` then it is considered to have already
+been deleted and the request will have no change to the Discovery Endpoint.
+A missing Service is not considered an error condition.
 
-This MUST delete the Service at the referenced URL.
+This specification does not define a Body for the request message and the
```
jskeet, line 1002: This PR lower-cases Response and Request - maybe do the same here for Body?

discovery/spec.md
```diff
@@ -917,108 +931,111 @@ Content-Type: application/json
...
-The request MAY include a body. Any such body MUST be an object containing an
-`id` matching the `{id}` in the path. That object MAY include an `epoch`
-value. If the Service corresponding with the specified `id` has a higher
-recorded `epoch` value than given, the request MUST fail.
+The request URL MAY include an OPTIONAL `epoch` query parameter, and if
+present and is not greater than the Service's current `epoch` value, then the
+request MUST fail.
 
-If other service attributes are included in the request, those MUST be
+If other service attributes are included in the request, those SHOULD be
```
jskeet, line 1010: They would have to be as extra query parameters, right?

discovery/spec.md
```diff
@@ -917,108 +931,111 @@ Content-Type: application/json
...
-  epoch of the Service with the given `id`
+Upon successfully processing the request, the HTTP response body MUST be
+a Service definition that includes at least the `id` attribute and SHOULD
+include the remaining Service attributes as they existed immediately prior
+to the Service being deleted, if possible. If the Service was deleted prior
+to the processing of this request the Service value might not be available
+to be returned.
 
-Other responses are allowed, but not defined by this specification.
+If the response Services include an `epoch` value then each MUST be greater
```
jskeet, line 1020: There's only a single response Service in this case, so I think these lines need a bit of rewording.

jskeet (approved):
I'm fine with the epoch handling, but I'm sure everyone would be disappointed if I didn't find some nits to pick...

discovery/spec.md
erikerikson: I don't agree about the ambiguity, please re-read.  I would agree that rewording might make it easier to understand.

discovery/spec.md
erikerikson: +1

discovery/spec.md
jskeet: Nope, I've re-read it several times and I still think this sentence is ambiguous. I think you can work out what's intended based on the rest of the document, but I bet if you gave *just* that sentence to 100 people and asked them to spell out exactly what was meant, they wouldn't all get it right...

erikerikson:
```
If an `id` is not specified in any of the Services the entire request MUST
fail.
```
I believe the statement is `Arx in(r, x) ^ not_id(x) ⊃ fail(r)`.  Perhaps you can identify what the source of ambiguity is?

jskeet: I'm afraid I don't understand your notation, but I can easily see this being read as either:

- "If there exists any Service in which an id is absent, the request must fail" (I believe this is the intended meaning)
- "If there isn't any Service in which an id is present, the request must fail"

I suspect we really don't need to agree on whether or not it's actually ambiguous, if we *do* agree that it would be better reworded to be clearer. For example:

> Every Service in the request must contain an `id`. If any Service in the request does not have an `id`, the entire request MUST fail with no effect.

erikerikson: > I suspect we really don't need to agree on whether or not it's actually ambiguous

Absolutely.  I was mostly trying to admit our contradictory opinions and drop into things around which are more decidable.  In particular, to ask you to help me see the possible interpretation I was missing in my feeble mindedness :D

discovery/spec.md
erikerikson: I believe 791-792 are intended to add that.

jskeet: Aha - I hadn't spotted that. Thanks!

discovery/spec.md
erikerikson: I believe that 802-803 are intended to specify this but it could be more clear.

jskeet: Nope, it's perfectly clear - I just didn't read it properly. Not sure why, but it's all good :)

discovery/spec.md
erikerikson: Given [this](https://github.com/cloudevents/spec/pull/930/files/aee0e42152993400fa5e32af78d9d1256d958c95#r835604807), I believe so.  Query parameters are a part of the request but your statement is correct as written.  Are you suggesting that the language be more specific?

jskeet: It was more making sure I'd understood correctly. I'm not sure whether I'd agree with ignoring them at all... I tend to think that if a client puts a query parameter on a URL, they expect it to be used. But I suspect I'm alone in that :)

discovery/spec.md
duglin: To me it's not so much about repeating what "immutable" means, but rather what to do in the event that someone tries... ignore vs fail. And in this case I'm suggesting it ignore it. I think that's easier to think about for URL, but for ID... might be unnecessary for POST, but I think it applies for PUT.

discovery/spec.md
duglin: I'm definitely ok with rewording but to keep beating the dead horse....
```
If an `id` is not specified in any of the Services...
```
I would think the "in any" means... well, IF ANY of them satisfy the concern about a missing ID, then it must fail.

I went with a slight variant of Erik's wording:
```
If any of the Services in the incoming request are missing an `id` then the entire request MUST fail.
```
Just because I wasn't sure if "If even one" was hard to parse for a non-English native person.  Hope that's ok.

discovery/spec.md
jskeet: Ah... because the id is present in both the service object and in the URL? That's already required in the PUT request via

> The body of the request message MUST contain a Service with an id attribute matching the {id} in the path.

... so I don't think it needs repeating, and in the case of the POST, such an attempt can't even be expressed as far as I can tell.

duglin: ok I removed it.

discovery/spec.md
duglin: ok done

discovery/spec.md
duglin: ok - removed reference to id

discovery/spec.md
duglin: done

discovery/spec.md
duglin: hmm I think that sentence may have been left over from a previous version where the Body meant something. 
I'm going to remove it because we have this sentence:
```
Unknown query parameters on the APIs MUST be ignored.
```
in another part of the doc.

discovery/spec.md
duglin: FIxed

duglin:
I think I addressed all comments and it's ready for review again on this week's call.
@cloudevents/notify

duglin:
Approved on the 4/28 call

PR #935, opened 2022-01-17
Message catalog spec

clemensv:

Work in progress. Comments welcome.


messagecatalog/spec.md
```diff
@@ -0,0 +1,1042 @@
...
+
+### 2.2.2. Message definition version URI
+
+To refer to a specific message definition object, unambiguous references to
+specific [definition versions](#22-schema-and-schema-version) are needed. These
+references can be resolved into definition documents with the help of a catalog.
+
+The message definition version URI is composed from the
+[authority](#authority-definition-authority) (as the base URI) and the
+[message definition version id](#id-definition-version-id).
```
deissnerk, line 256: I'm not able to follow these links. What are they pointing to?

messagecatalog/spec.md
```diff
@@ -0,0 +1,1042 @@
...
+authority URI in the previous section. A consumer SHOULD then be configured with
+a fixed message catalog endpoint for it to use, and obtain definitions
+identified by a URI using the
+["getMessage definitionVersionByURI"](#342-get-a-specific-schema-version) API
+operation on that endpoint, rather than trying to resolve the URI directly.
+
+The following table shows some examples for a catalog hosted at
+`https://example.com/messagecatalog`:
+
+| [authority](#authority-definition-authority) | [id](#id-definition-version-id)                               | Message definition URI                                                                         |
```
deissnerk, line 273: Is `Message definition URI` and `Message definition version URI` the same?

deissnerk: What is a `definition version id`? Below an `id` is defined for a _message definition_, and a `version` is defined for a _message definition version_.

messagecatalog/spec.md
```diff
@@ -0,0 +1,1042 @@
...
+
+#### `id` (message definition)
+
+- Type: `String`
+- Description: Identifies the message definition.
+- Constraints:
+  - REQUIRED. IMMUTABLE.
+  - MUST be a non-empty string
+  - MUST conform with RFC3986/3.3 `segment-nz-nc` syntax
+  - MUST be unique within the scope of the message definition group
```
deissnerk, line 294: How does this work in replication scenarios? Are only complete _message definition groups_ replicated? Otherwise there might be collisions during replication, if message message definitions from multiple catalogs or groups are aggregated into a single group.

messagecatalog/spec.md
```diff
@@ -0,0 +1,1042 @@
...
+The `attributes` object MAY contain any number of uniquely named attribute
+fields, whereby any name MUST conform with CloudEvents naming rules.
+
+Whenever the name of an attribute field matches that of a well-known CloudEvents
+attribute, its value or the constraints for the value expressed with the regular
+expression MUST match the constraint rules of the CloudEvents attribute.
+
+- A definition for the `type` field is REQUIRED and its value MUST follow all
+  constraints of the `type` attribute of CloudEvents. The `id` of a message
+  definition version using this metaschema MUST match the `type` field value.
```
deissnerk, line 437: How would this work, if event definitions from multiple groups that contain definitions for the same event type are aggregated into a single group?

deissnerk (commented):
Thanks for writing this up @clemensv. My remarks are mostly around identifiers and cardinalities, especially with replication scenarios in mind.

messagecatalog/messagecatalog-data.json
```diff
@@ -0,0 +1,491 @@
...
+    "properties": {
+      "id": {
+        "type": "string"
+      },
+      "authority": {
+        "type": "string"
+      },
+      "format": {
+        "type": "string",
+        "format": "uri"
```
deissnerk, line 58: I think `authority` and not `format` is supposed to be a URI.

messagecatalog/messagecatalog-data.json
```diff
@@ -0,0 +1,491 @@
...
+  "MessageDefinitionVersion": {
+    "allOf": [
+      {
+        "$ref": "#MessageDefinitionBase"
+      },
+      {
+        "type": "object",
+        "schema": {
+          "version": {
+            "type": "string"
```
deissnerk, line 106: Integer or String?

messagecatalog/spec.md
```diff
@@ -0,0 +1,1059 @@
...
+| (empty, therefore implied)                   | `342`                                                         | <https://example.com/messagecatalog/342>                                                       |
+| <https://example.com/messagecatalog>         | `342`                                                         | `https://example.com/messagecatelog/342`                                                       |
+| `http://messagecatalog.example.com/`         | `/mygroup/messagedefinitions/myMessage definition/versions/2` | <http://messagecatalog.example.com/mygroup/messagedefinitions/myMessage definition/versions/2> |
+| `http://messagecatalog.example.org/`         | `http://messagedefinitions.example.org/41751`                 | <http://messagecatalog.example.org/41751>                                                      |
+| `urn:example:messagecatalog`                 | `a1b2c3d4`                                                    | `urn:example:messagecatalog:a1b2c3d4`                                                          |
```
deissnerk, line 295: The URN example might be tricky for a generic player, as to my understanding the `urn` scheme is not hierarchical, if I understand [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986/#section-3.3) correctly.

messagecatalog/spec.md
```diff
@@ -0,0 +1,1059 @@
...
+#### `version` (message definition version)
+
+- Type: `Integer`
+- Description: The version of the Message definition. This is a simple counter
+  and tracks the version in the scope of this Message definition within the
+  message definition group. The Message definition document content MAY indicate
+  a different versioning scheme.
+- Constraints:
+  - REQUIRED
+  - Assigned by server.
```
deissnerk, line 410: What about push replication? Could a replication potentially result in a different version on the target?

messagecatalog/spec.md
```diff
@@ -0,0 +1,1475 @@
...
+data type definitions for payloads MAY be held inline inside the Message
+Catalog.
+
+The message definition entries in the catalog contain some fixed metadata
+information. That fixed information MUST uniquely identify the type of message
+and depends on the particular message information model. For example, for CNCF
+CloudEvents message definitions, the "type" attribute is REQUIRED as fixed
+information for each entry.
+
+- Message definition group: A named collection of message definitions. Each
```
n3wscott, line 69: my first thought on groups is you might want to have an event in many groups, and an easy way do to that would be to make group hierarchy? like trees or some such thing.

messagecatalog/spec.md
```diff
@@ -0,0 +1,1475 @@
...
+    notation", e.g. `org.example.myapp.module` for logical organization of
+    message definition groups, even though the namespace is flat.
+  - MUST be unique within the scope of the message catalog
+- Examples:
+  - mygroup
+  - my-group
+  - org.example.myapp.module
+  - events\@example.com
+
+#### `specversion` (message definition group)
```
n3wscott, line 143: I feel like this should be in the URI and not the payload.

messagecatalog/spec.md
```diff
@@ -0,0 +1,1475 @@
...
+
+- Type: `String`
+- Description: Version of this specification that this document complies with.
+- Constraints:
+  - REQUIRED.
+  - MUST be "0.1" for this version of the specification.
+
+#### `format` (message definition group)
```
n3wscott, line 151: to support several formats for the same group set, I would have to make several message groups?

messagecatalog/spec.md
```diff
@@ -0,0 +1,1475 @@
...
+
+- Type: `String`
+- Description: Defines the message definition format managed by this message
+  definition group. The format MAY be one of formats defined in this
+  specification (CloudEvents, NATS, AMQP, MQTT) or a custom format. All
+  definitions in this group MUST conform with the rules of the chosen format.
+- Constraints:
+  - REQUIRED. IMMUTABLE.
+  - MUST NOT be modified if at least one definition exists in the group.
+  - "CloudEvents", "NATS", "AMQP", "MQTT" for the definition formats defined in
```
n3wscott, line 161: I feel apprehensive about extending the reach of this into NATS and AMQP, etc. Might it be an overstep?

messagecatalog/spec.md
```diff
@@ -0,0 +1,1475 @@
...
+  - OPTIONAL
+- Examples:
+  - "This group holds Message definitions for the fabulous example app."
+
+#### `documentation` (message definition group)
+
+- Type: `String`
+- Description: Documents the purpose of the message definition group in greater
+  detail than the `description`. The text MAY use [CommonMark](https://commonmark.org/) markdown formatting.
+- Constraints:
```
n3wscott, line 184: is there a length limit? or redirect to the full documentation? seems like I might want to point to a url that holds a mardown file vs embedding it in this payload.

messagecatalog/spec.md
```diff
@@ -0,0 +1,1475 @@
...
+
+#### `updatedtimeutc` (message definition group)
+
+- Type: `Timestamp`
+- Description: Instant when the message definition group was last updated
+- Constraints:
+  - OPTIONAL
+- Assigned by the server.
+
+#### `definitions' (message definition group)
```
n3wscott, line 214:
```suggestion
#### `definitions` (message definition group)
```

messagecatalog/spec.md
```diff
@@ -0,0 +1,1475 @@
...
+    "description" : "Raised when an object has been deleted from the catalog",
+    "format" : "CloudEvents",
+    "versions" : [
+      {
+        "version" : 1,
+        "definition" : {
+          "attributes" : {
+            "specversion" : {
+              "value" :"1.0",
+              "required" : true,
```
n3wscott, line 259: this seems fairly redundant if the format is already defined as cloudevents, we should just need to list core attributes that are expected and possibly extensions that are likely?

messagecatalog/spec.md
```diff
@@ -0,0 +1,1475 @@
...
+              "description" : "CloudEvents specification version"
+            },
+            "type" : {
+              "value" : "io.cloudevents.messagecatalog.deleted",
+              "required" : true,
```
n3wscott, line 264: also, who is the "required" field for? the consumer? the producer likely created this document and so it is not `required` but perhaps `present` or `always`? idk, feels weird on this one

messagecatalog/spec.md
```diff
@@ -0,0 +1,1475 @@
...
+              "description" : "type of this event"
+            }
+          }
+        }
+    },
+    {
+        "version" : 2,
```
n3wscott, line 271: We might want version to be semver, not just a number to allow us to signal when something is an addition vs a message change.

messagecatalog/spec.md
```diff
@@ -0,0 +1,1475 @@
...
+}
+```
+
+The API is created such that a message definition MUST be created by posting an
+initial version and that the message definition (including any common parts like
+the documentation) MUST only be changed by posting a new version. Existing
+versions can be individually retrieved and deleted, but not modified.
+
+A newer message definition version might introduce breaking changes or it might
+only introduce careful changes that preserve compatibility. These strategies are
```
n3wscott, line 332: semver passes a tool to the implementors to make smarter choices on how to upgrade and represent versions.

n3wscott: or you can do what proto does and deprecate  fields and never delete them

messagecatalog/spec.md
```diff
@@ -0,0 +1,1475 @@
...
+transparent. Whenever a message definition is updated, a new version number is
+assigned and prior definition versions are retained. This specification does not
+mandate a retention policy, but implementations MAY retire and remove outdated
+definition versions.
+
+The latest available definition MUST always the default version that is
+retrieved when the URI to the Message definition is used without the version
+having been specified.
+
+### 2.2.1 Authority
```
n3wscott, line 347: I suggest adding audience as well. Then you have symmetry on `who` and `for whom`

messagecatalog/spec.md
```diff
@@ -0,0 +1,1475 @@
...
+#### `createdtimeutc` (message definition version)
+
+- Type: `Timestamp`
+- Description: Instant when the Message definition version was added to the
+  catalog.
+- Constraints:
+  - OPTIONAL
+  - Assigned by the server.
+
+#### `definition' (message definition version)
```
n3wscott, line 579:
```suggestion
#### `definition` (message definition version)
```

clemensv:
@n3wscott 

Trying to address your top-level concern:

What we are finding is that having message/event definitions embedded in the "services" in discovery is not sufficient since the exact same event definitions turn up in several places in a system, at a minimum when you have several instances of the same service in multiple tenant contexts. 

That means we need a separate place to manage the definitions.

The groups exist because we need a way to add/remove and replicate related sets of events, and we also need to have a way to create a "bracket" around events that we can reference with a single link. 

If you have a business system with 20 modules, each of them will emit events and since not all customers may buy all 20 modules, it will be useful to have some structure around them if they all end up the same catalog.

Also, a customer may be using a module of the "standard" SKU which emits 5 events and another customer may be using a module of the "enterprise" SKU which emits 7 events, and it will be good to be able to link to those different compositions ("scopes" as @deissnerk calls them) with a single link.

The definitions are means to completely replace the definitions in the discovery spec, so this is not meant to overlap. The definitions are also meant to be embeddable into a "service" object for the simplest case. 

@duglin also posed the question of whether it makes sense to fold discovery/catalog/schemas together into one spec.

messagecatalog/spec.md
duglin: We’ve talked about how there might be a single doc with all data from all our specs inlined, I wonder if it would be useful to show an example of that doc in an upcoming call so that people can see whether there really is duplication of data.

clemensv:
For illustration, I put the Azure Storage events into this format. The "dataschema" definitions reference the official payload types.

https://gist.github.com/clemensv/d75e9bc05767f9624e2a3244df351d6f

n3wscott:
Ok, I like this idea in general, but I wonder if we could take a smaller step first and do the following:

- define a message format that could be an offline file.
- introduce a event-ref concept in the discovery spec that can point to a document that defines the event (could be that offline file format hosted in github)
- introduce a catalog endpoint in addition to the /services endpoint which can host the defined message format, but could also be `-ref` type pointers to the other offline catalog of events. 
  - /catalog would be to find the services attached to events, /services is a way to find events attached to services. Both could choose to defer to the `-ref` url to define the real event shape.

The addition of the groups feels like an overstep at the moment.

messagecatalog/spec.md
```diff
@@ -0,0 +1,1475 @@
...
+A producer produces messages and publishes them. A subscriber subscribes to
+published messages based on some criteria. A consumer is the delivery
+destination for such subscriptions and handles the delivered information.
+
+## 1. Introduction
+
+Message and event catalogs are useful tools at design and development time
+because they allow interactive discovery of the kinds of messages that are
+available for subscription from a system and to find out, for instance, which
+event sources raise them. A chosen message definition catalog entry can serve as
```
duglin, line 30: Couple of questions:
1 - are people going to be looking for "messages" or "events" ? I would guess it's probably "events" and then they would use the related "message" to understand the extra metadata around it.
2 - does this spec assume "structured cloudEvents"? I'm asking because (unless I missed it) it seems like the message definitions assume all properties are in the Body. And just to complete things... in the "binary CE" cases, the schema of the Body would be the dataschema, right?

messagecatalog/spec.md
```diff
@@ -0,0 +1,1475 @@
...
+- Message definition: A document describing the metadata and structure of
+  messages and events.
+
+- Message definition version: A specific version of such a definition. All
+  documents that are stored and retrieved through the API are such versions. The
+  identifying criteria of a definition MUST be identical across versions.
+
+A message definition group, including all definitions and its versions, can be
+stored and shared in a single document. A folder in a git repository containing
+such documents is also a catalog in the sense of this specification.
```
duglin, line 83: Do the catalog entries need to be reference-able via a URL?

messagecatalog/spec.md
```diff
@@ -0,0 +1,1475 @@
...
+
+For interoperability and sharing design-time definitions across system
+boundaries, conformance to a common data model and document format is most
+important and it is assumes that the authoritative source for many definitions
+will into be a document under source control that is then shared at runtime via
+the common API defined here.
```
duglin, line 89: s/assumes/assumed/
s/will into/will be/

messagecatalog/spec.md
```diff
@@ -0,0 +1,1475 @@
...
+
+## 2. Message Catalog Elements
+
+This section describes the elements enumerated in the introduction.
+
+All data types used in this section are defined in the CNCF CloudEvents
```
duglin, line 95: we should add a link to the CE spec so people know exactly which section to look at.

messagecatalog/spec.md
```diff
@@ -0,0 +1,1475 @@
...
+#### `format` (message definition group)
+
+- Type: `String`
+- Description: Defines the message definition format managed by this message
+  definition group. The format MAY be one of formats defined in this
+  specification (CloudEvents, NATS, AMQP, MQTT) or a custom format. All
+  definitions in this group MUST conform with the rules of the chosen format.
+- Constraints:
+  - REQUIRED. IMMUTABLE.
+  - MUST NOT be modified if at least one definition exists in the group.
```
duglin, line 160: feels a bit odd to allow them to change this value if the list drops down to zero. Granted, they could delete the group and recreate it to yield the same results, but that feels more "proper" to me than to allow for this special case.

messagecatalog/spec.md
```diff
@@ -0,0 +1,1475 @@
...
+
+#### `documentation` (message definition group)
+
+- Type: `String`
+- Description: Documents the purpose of the message definition group in greater
+  detail than the `description`. The text MAY use [CommonMark](https://commonmark.org/) markdown formatting.
+- Constraints:
+  - OPTIONAL
+
+#### `tags` (message definition group)
```
duglin, line 187: My initial reaction was to make these name/value pairs, but then I realized we could get the same results via single-string-tags IF we also restrict the character set. For example, don't allow `,` in name or value (at least not w/o an escape), but do allow `=`.

duglin:
@clemensv should we close this?

duglin:
Closing in favor of the new Discovery spec work

PR #937, opened 2022-01-18
Add C# namespace option to proto

jskeet:
The naming choice is based on the C# SDK core being
"CloudNative.CloudEvents". This creates a bias towards using the C#
SDK instead of "roll-your-own" - but I think that's okay,
personally. Happy to consider other options.

## Proposed Changes

- Add a C# namespace option to cloudevents.proto

**Release Note**

```release-note
Adds a C# namespace option of `CloudNative.CloudEvents.V1` to the proto representation.
```

### DIFF
--- a/cloudevents/formats/cloudevents.proto
+++ b/cloudevents/formats/cloudevents.proto
@@ -13,6 +13,7 @@ package io.cloudevents.v1;
 import "google/protobuf/any.proto";
 import "google/protobuf/timestamp.proto";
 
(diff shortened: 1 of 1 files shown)


jskeet:
Note: this is against the `main` branch. I've lost the plot in terms of what we're doing in which branch at the moment. I believe I've got an action item to propose an alternative, but...

duglin:
Approved on the 1/20 call

PR #938, opened 2022-01-20
openapi changes for #928

duglin:
@erikerikson wanna take a look? I don't know openapi so I'm not sure if this is correct

### DIFF
--- a/discovery/discovery.yaml
+++ b/discovery/discovery.yaml
@@ -62,15 +62,15 @@ paths:
                 $ref: '#/components/schemas/service'
         '404':
           description: Not Found
-    post:
-      operationId: postService
+    put:
+      operationId: putService
       parameters:
         - $ref: '#/components/parameters/id'
       requestBody:
-        $ref: '#/components/requestBodies/servicepostrequest'
+        $ref: '#/components/requestBodies/serviceputrequest'
       responses:
         '200':
-          $ref: '#/components/responses/servicepostresponse'
+          $ref: '#/components/responses/serviceputresponse'
         '400':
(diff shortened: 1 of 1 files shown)


discovery/discovery.yaml
```diff
@@ -131,7 +131,7 @@ components:
           schema:
             $ref: '#/components/schemas/serviceinstance'
   responses:
-    servicespostresponse:
+    servicesputresponse:
```
erikerikson, line 134: I think this one needs to remain `servicespostresponse`

erikerikson (changes requested):
Thank you for synchronizing.  Lots of details here so well done.

There appears to be a missing `changeput` definition [between 296 and 297](https://github.com/cloudevents/spec/blob/3765d7ce87e7c92583bad2b68ae569c6a94aad51/discovery/discovery.yaml#L297).  It is referenced on [335](https://github.com/cloudevents/spec/blob/3765d7ce87e7c92583bad2b68ae569c6a94aad51/discovery/discovery.yaml#L335).  Suggested:

```
    changeput:
      type: string
      enum:
        - PUT
```

If we will be cracking the semantics of the API, it's not clear that this was the no-op presented and I think we may want to review the POST => PUT change again.

discovery/discovery.yaml
duglin: fixed

duglin:
Add the "changeput" - thanks for noticing

erikerikson:
LGTM

duglin:
Approved on the 1/27 call

PR #947, opened 2022-01-28
remove old code

duglin:

left over code from debugging


duglin:
executive decision... merging

PR #949, opened 2022-02-08
Fix subscriptions OpenAPI

n3wscott:
I noticed the OpenAPI for subscription was never updated after the spec change a year ago.  Fixing that.

**Release Note**

```release-note
updated the subscription OpenAPI to the latest defined by the subscription text.
```

### DIFF
--- a/subscriptions/subscriptions-openapi.yaml
+++ b/subscriptions/subscriptions-openapi.yaml
@@ -180,7 +180,6 @@ components:
               - $ref: "#/components/schemas/ExactFilter"
               - $ref: "#/components/schemas/PrefixFilter"
               - $ref: "#/components/schemas/SuffixFilter"
-              - $ref: "#/components/schemas/DialectFilter"
               - $ref: "#/components/schemas/SqlFilter"
         config:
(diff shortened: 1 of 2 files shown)


subscriptions/subscriptions-openapi.yaml
```diff
@@ -235,7 +234,6 @@ components:
                 - $ref: "#/components/schemas/ExactFilter"
                 - $ref: "#/components/schemas/PrefixFilter"
                 - $ref: "#/components/schemas/SuffixFilter"
-                - $ref: "#/components/schemas/DialectFilter"
                 - $ref: "#/components/schemas/SqlFilter"
           additionalProperties: false
     AnyFilter:
```
n3wscott, line 239: Needs to be an array.

subscriptions/subscriptions-openapi.yaml
```diff
@@ -214,7 +213,7 @@ components:
       title: "Filter"
       description: "A filter from a selection of multiple filter types and dialects"
       type: object
-      additionalProperties: false
+      additionalProperties: true
     AllFilter:
```
n3wscott, line 217: Needs to be an array.

duglin:
Approved on the 2/10 call

Issue #934, opened 2022-01-17
Revisiting JSON format's data member


jskeet:
I'm back to implementing #861, and I've run into this:

> If the `datacontenttype` declares the data to contain JSON-formatted content, a
JSON serializer MUST translate the data value to a [JSON value][json-value], and
use the member name `data` to store it inside the JSON representation. The data
value MUST be stored directly as a JSON value, rather than as an encoded JSON
document represented as a string.

I have a *memory* that the value of `data` must be a JSON string or object, *not* an array, number or Boolean value. However, I don't see any indication of that within the spec. The [JSON value](https://datatracker.ietf.org/doc/html/rfc7159#section-3) reference includes numbers, Boolean values, arrays and indeed null.

So if an SDK is provided CloudEvent data of "the number 1.5" and a content type of application/json (explicitly or implicitly), is it reasonable for the structured event representation to be:

```json
{
  "id": "...",
  "type": "...",
  "datacontenttype": "application/json",
  "data": 1.5
}
```

? If that's valid, then my life will be easier but my memory is incorrect. If it's invalid, then is that already covered somewhere in the spec and I just haven't seen it, or do we need to change the spec?

duglin:
From the spec, I see this example:
```
{
    "specversion" : "1.0",
    "type" : "com.example.someevent",
    "source" : "/mycontext",
    "subject": null,
    "id" : "D234-1234-1234",
    "time" : "2018-04-05T17:31:00Z",
    "comexampleextension1" : "value",
    "comexampleothervalue" : 5,
    "data" : "I'm just a string"
}
```
And from that I would conclude that if `data` was a number then it would look like: `"data": 1.5` since the switch from string to number is `"I'm just a string"` to `1.5`. No extra json magic/wrapping should be needed. Also, note that `comexampleothervalue` does the same thing.

jskeet:
comexampleothervalue is just an extension attribute - that's fine. If we could talk about this on Thursday, someone else at remember the same bit as me about "only a string or object for data". If I'm just misremembering, that would make my life easier :)

duglin:
Another way I look at it is this: any valid/stand-alone JSON should be able to be a child of `data`. If we require some kind of wrapper, how can we tell if the wrapper is there just for CE or there because it's part of the original JSON. That would then mean that: `"data": 1.5` must be valid because there would be ambiguity if the `1.5` had to be wrapped/modified.

JemDay:
I still think the spec is pretty clear ...

[...]

PR #950, opened 2022-02-08
Clarifications to JSON format

### LINKED ISSUE #933: Serialization of binary data with no content type in JSON format
I'm trying to adjust the C# JSON serializers to take account of #861 and #881. I'm struggling on a couple of corner cases though, around binary data.

The JSON event format says that a serializer has to distinguish between binary data and non-binary data (to know which data property to populate) and also that the default content type (if otherwise unspecified) should be application/json. The placement of the defaulting (which was my own sentence!) is alongside the "not binary" case.

What does this mean for the structured mode encoded form of a CloudEvent with binary data but no data content type? Should a data content type of `application/json` be used anyway? Do we expect *deserialization* to cope with a JSON content type but binary data? (Just assume that the binary data is UTF-8-encoded text?)

It's possible that this kind of corner case doesn't need to be handled the same way by every SDK, of course...

jskeet:
Fixes #933 and #934

## Proposed Changes

- Clarify that the "default to JSON" handling of datacontenttype
  does not apply to binary data
- Give an example of a JSON number as the `data` property value

**Release Note**

```release-note
Clarifications to JSON format data/data_base64 handling
```

### DIFF
--- a/cloudevents/formats/json-format.md
+++ b/cloudevents/formats/json-format.md
@@ -143,8 +143,10 @@ If the implementation determines that the type of data is `Binary`, the value
 MUST be represented as a [JSON string][json-string] expression containing the
 [Base64][base64] encoded binary value, and use the member name `data_base64` to
 store it inside the JSON representation. If present, the `datacontenttype` MUST
-reflect the format of the original binary data. If the `datacontenttype` is
-unspecified, processing MUST NOT assume a default of `application/json`.
+reflect the format of the original binary data. If a `datacontenttype` value is
+not provided, no assumptions can be made as to the format of the data and
+therefore the `datacontenttype` attribute MUST not be present in the resulting
+CloudEvent.
 
 If the type of data is not `Binary`, the implementation will next determine
 whether the value of the `datacontenttype` attribute declares the `data` to
@@ -184,7 +186,7 @@ When a CloudEvents is deserialized from JSON, the presence of the `data_base64`
 member clearly indicates that the value is a Base64 encoded binary data, which
 the deserializer MUST decode into a binary runtime data type. The deserializer
 MAY further interpret this binary data according to the `datacontenttype`. If
-the `datacontenttype` attribute is absent, the decoding should MUST NOT make an
+the `datacontenttype` attribute is absent, the decoding MUST NOT make an
 assumption of JSON-formatted data (as described below for the `data` member).
 
 When a `data` member is present, the decoding behavior is dependent on the value
@@ -387,6 +389,32 @@ ce-comexampleothervalue: 5
 content-type: application/json
 
 "I'm just a string"
+```
+
+Example event with a `Binary`-valued `data_base64` but no `datacontenttype`.
+Even though the data happens to be a valid JSON document when interpreted as
+text, no content type is inferred.
+
+```JSON
+{
+    "specversion" : "1.0",
+    "type" : "com.example.someevent",
+    "source" : "/mycontext",
+    "id" : "D234-1234-1234",
+    "data_base64" : "eyAieHl6IjogMTIzIH0="
+}
+```
+
+The above example re-encoded using [HTTP Binary Content Mode][http-binary].
+Note that there is no `content-type` header present.
+
+```
+ce-specversion: 1.0
+ce-type: com.example.someevent
+ce-source: /mycontext
+ce-id: D234-1234-1234
+
+{ "xyz": 123 }
 ```
 
 ## 4. JSON Batch Format
--- a/cloudevents/formats/json-format.md
+++ b/cloudevents/formats/json-format.md
@@ -143,7 +143,8 @@ If the implementation determines that the type of data is `Binary`, the value
 MUST be represented as a [JSON string][json-string] expression containing the
 [Base64][base64] encoded binary value, and use the member name `data_base64` to
 store it inside the JSON representation. If present, the `datacontenttype` MUST
-reflect the format of the original binary data.
+reflect the format of the original binary data. If the `datacontenttype` is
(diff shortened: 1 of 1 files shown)


cloudevents/formats/json-format.md
```diff
@@ -143,7 +143,8 @@ If the implementation determines that the type of data is `Binary`, the value
 MUST be represented as a [JSON string][json-string] expression containing the
 [Base64][base64] encoded binary value, and use the member name `data_base64` to
 store it inside the JSON representation. If present, the `datacontenttype` MUST
-reflect the format of the original binary data.
+reflect the format of the original binary data. If the `datacontenttype` is
+unspecified, processing MUST NOT assume a default of `application/json`.
```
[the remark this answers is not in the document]
duglin: I guess you mention it there already - I'm not sure it needs to go here too since this section is about serializing not "processing". right?

cloudevents/formats/json-format.md
```diff
@@ -315,6 +318,39 @@ content-type: application/json
...
+ce-source: /mycontext
+ce-id: C234-1234-1234
+ce-time: 2018-04-05T17:31:00Z
+ce-comexampleextension1: value
+ce-comexampleothervalue: 5
+content-type: application/json
+
+1.5
+```
+
```
duglin, line 353: Should we also include a binary data example to go along with your new text to re-enforce the idea of not making assumptions?

cloudevents/formats/json-format.md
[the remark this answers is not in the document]
jskeet: I used "processing" as that's what's used in the paragraph below. "If the datacontenttype is unspecified, processing SHOULD proceed as if the datacontenttype had been specified explicitly as application/json." Happy to use another word if we can find a suitable one though.

cloudevents/formats/json-format.md
jskeet: Could do - I'm not 100% sure it's worth it though. Let's discuss in the meeting later, and I can add it in if desired.

cloudevents/formats/json-format.md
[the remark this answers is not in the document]
duglin: interesting. Seems like perhaps both should just be in the deserialize section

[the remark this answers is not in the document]
jskeet: We definitely need it in the serialization section as well - because serializing it *does* (or at least can) add the implicit datacontenttype in the HTTP response (or whatever).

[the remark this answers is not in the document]
jskeet: https://github.com/cloudevents/sdk-csharp/issues/200 is the C# issue of "whoops, we messed up here".

Here's the C# code which is broken, by unconditionally adding in `"datacontenttype": "application/json"` into the document if the event has data but no `datacontenttype`.

https://github.com/cloudevents/sdk-csharp/blob/409982e9e97a8a450b1bfbbbc2a16a8c08c78f62/src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs#L449-L453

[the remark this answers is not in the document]
jskeet: (I'd expected that to actually show the code, but it doesn't for some reason. Ah well - follow the link to see my failure...)

cloudevents/formats/json-format.md
```diff
@@ -182,7 +183,9 @@ payload as distinct from the absence of the `data` member.
 When a CloudEvents is deserialized from JSON, the presence of the `data_base64`
 member clearly indicates that the value is a Base64 encoded binary data, which
 the deserializer MUST decode into a binary runtime data type. The deserializer
-MAY further interpret this binary data according to the `datacontenttype`.
+MAY further interpret this binary data according to the `datacontenttype`. If
+the `datacontenttype` attribute is absent, the decoding should MUST NOT make an
```
duglin, line 187: s/should MUST NOT/MUST NOT/

cloudevents/formats/json-format.md
duglin, line 147: Based on the call last week, what about:
- - -
If a `datacontenttype` value is not provided, a default value of `application/json` MUST NOT be assumed nor used as the value of the `datacontenttype` attribute in the serialized CloudeEvent.
- - -
Without explicitly stating that it shouldn't appear in the CE, it wasn't clear to me what "processing" we were referring to.

TBH I wasn't sure if the "be assumed" is even needed but perhaps you wanted to give guidance about some internal processing (which we probably can't control anyway). If not, then perhaps:
- - -
If a `datacontenttype` value is not provided, no assumptions can be made as to the format of the data and therefore the `datacontenttype` attribute MUST not be present in the resulting CloudEvent.
- - -

cloudevents/formats/json-format.md
jskeet: Whoops, done - thanks.

cloudevents/formats/json-format.md
jskeet: Going with the latter - I like that.

cloudevents/formats/json-format.md
jskeet: Yes, it's merely my inefficiency that has failed here.

jskeet: Added a new example - we probably want to haggle over the wording a little.

jskeet (commented):
Pushed a new commit now - PTAL.

duglin:
@jskeet aside from the edits we talked about last week, can you sign all of your commits to make the DCO checker happy?

jskeet:
Doh, sorry. Yes, will fix.

jskeet:
Second commit signed.

duglin:
LGTM

PR #952, opened 2022-02-10
Introduce reserved attribute names.

JemDay:
Added a section to define a set of context attribute names
that cannot be used.

## Proposed Changes

- Introduces the concept of reserved attribute names that cannot be used.

**Release Note**

```release-note
`data` is now reserved and MUST NOT be used as an extension attribute name.
```

### DIFF
--- a/cloudevents/spec.md
+++ b/cloudevents/spec.md
@@ -163,7 +163,7 @@ These attributes, while descriptive of the event, are designed such that they
 can be serialized independent of the event data. This allows for them to be
 inspected at the destination without having to deserialize the event data.
 
-### Attribute Naming Convention
+### Naming
 
 The CloudEvents specifications define mappings to various protocols and
 encodings, and the accompanying CloudEvents SDK targets various runtimes and
@@ -178,11 +178,8 @@ CloudEvents attribute names MUST consist of lower-case letters ('a' to 'z') or
 digits ('0' to '9') from the ASCII character set. Attribute names SHOULD be
 descriptive and terse and SHOULD NOT exceed 20 characters in length.
 
-#### Reserved Names
-
-The following attribute names are reserved and MAY NOT be used:
-
-- `data`
+CloudEvent attributes MUST NOT have the name `data`; this name is reserved as it
+is used in some event formats.
 
 ### Type System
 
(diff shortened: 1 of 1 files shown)


cloudevents/spec.md
```diff
@@ -178,6 +178,12 @@ CloudEvents attribute names MUST consist of lower-case letters ('a' to 'z') or
 digits ('0' to '9') from the ASCII character set. Attribute names SHOULD be
 descriptive and terse and SHOULD NOT exceed 20 characters in length.
 
+#### Reserved Names
+
+The following attribute names are reserved and MAY NOT be used:
```
duglin, line 183: without additional text I can imagine someone being confused because in https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#event-data it talks about `data` as if it were a context attribute, even though (I guess) technically it is not. But I suspect most casual readers of the spec won't pick up on the distinction.

Perhaps:
- - -
- `data` : while technically not a "context attribute" this name is reserved for cases where the event payload needs to appear along-side the context attributes. For example, in JSON structured mode.
- - -

Also, feels odd to have an entire section for just one and I wonder if we'll ever have more?? If we have any more I suspect we'll add them as optional attributes won't we? Maybe just add a sentence to the "Event Data" section explaining that "data" is special?

jskeet, line 183: Should "MAY NOT" be "MUST NOT"? (RFC 2119 doesn't include MAY NOT.)

Possibly "The following names are reserved and MUST NOT be used for attributes"? (They're not attribute names, arguably...)

jskeet: @duglin: I think all we need to do is remove "When present, this information will be encapsulated within `data`." in the event-data section - and it would be good to do that in this PR to make it a consistent change.

I do think we need to prevent `data` from being used *within the context attributes section* but I agree it probably doesn't need a new subsection.

I would propose:

- Renaming "Attribute Naming Convention" to "Attribute Naming" (the section contains normative requirements)
- At the end of the current second paragraph in that section, append: "Context attributes MUST NOT have the name `data`; this name is reserved as it is used for CloudEvent data in some event formats in places which would otherwise be ambiguous."

jskeet: If it's easiest for me to come up with a counter-PR, let me know :)

JemDay: Let me take a crack at this..

JemDay: ack

duglin:
Approved on the 2/17 call

PR #958, opened 2022-02-24
SDK should validate context attribute values

duglin:
Per a chat we had in a previous SDK call, the SDK SHOULD validate attributes
so users can be notified if they gave an invalid value.

### DIFF
--- a/cloudevents/SDK.md
+++ b/cloudevents/SDK.md
@@ -131,6 +131,9 @@ Validation MUST be possible on an individual Event. Validation MUST take into
 account the spec version, and all the requirements put in-place by the spec at
 each version.
 
+SDKs SHOULD perform validation on context attribute values provided to it by
+the SDK user. This will help ensure that only valid CloudEvents are generated.
+
 ## Documentation
 
 Each SDK must provide examples using at least HTTP transport of:


duglin:
@n3wscott @lance

lance (approved):

duglin:
> I wasn’t on the SDK call but is the implication here that an SDK actively reject attempts to create context attributes with invalid content?

Yup. The original usecase was someone was trying to set "source" to a value that wasn't a valid URI-reference and they were wondering if the SDK should have flagged it.

lance:
FWIW - in the JavaScript SDK, by default, users must create valid events.  We allow users to create invalid events if they explicitly indicate that they want to forego validation. However, when receiving an event over the wire and decoding it, the reverse is true. We default to no validation since we want to avoid failing when receiving bad data. The user can, in this case, turn validation on if they want it.

deissnerk:
@embano1 I think this was triggered by your request on Slack. What do you think?

jskeet:
> We default to no validation since we want to avoid failing when receiving bad data.

Do JavaScript JSON/XML/etc parsers typically follow the same approach? My experience is that if some tools are lax in validation, users come to believe that their invalid data is actually valid, and blame *stricter* tools for rejecting things. I'm a big believer in being strict, at least by default. (I'm okay with there being an opt-out for extreme circumstances.)

embano1:
> @embano1 I think this was triggered by your request on Slack. What do you think?

I think this is a very reasonable approach!

jskeet:
Discussion during working group meeting:

- Java, C# and Go are all strict by default (both parsing and programmatic creation of CloudEvents)
- There are performance reasons to be lax in some cases
- "MUST be strict by default, but MAY provide options to remove validation (for various reasons)" could potentially work - but might make existing SDKs non-conformant, so needs wider consultation

lance:
> Do JavaScript JSON/XML/etc parsers typically follow the same approach?

When parsing JSON, if it is malformed, the builtin JavaScript `JSON.parse()` function, used by the SDK, will throw an exception.

The JavaScript SDK only attempts to parse JSON. Even then, that's only when encoding/decoding to a JSON representation of the event. There is no XML parsing at all.

jskeet:
> The JavaScript SDK only attempts to parse JSON. Even then, that's only when encoding/decoding to a JSON representation of the event. There is no XML parsing at all.

Sorry, I was unclear - I wasn't talking about the use of JSON and XML parsers within the CloudEvents SDK, but in terms of general expectations from JavaScript developers when they're asking a library to perform parsing.

lance:
> Sorry, I was unclear - I wasn't talking about the use of JSON and XML parsers within the CloudEvents SDK, but in terms of general expectations from JavaScript developers when they're asking a library to perform parsing.

Gotcha - with regard to JSON, the vast majority of developers will simply use the builtin `JSON.parse()`, which as I mentioned, will throw with bad input. Other JSON parsers I have seen behave similarly. I am honestly not sure about other text formats such as XML, although I imagine developers doing any kind of text parsing in JavaScript are likely accustomed to the behavior of `JSON.parse()` and would find similar behavior with other parsers acceptable.

jskeet:
Right - in that case, it sounds like "throw when parsing invalid data" is reasonably common - so would it be feasible to change the CloudEvents SDK to do that by default as well, and make validation opt-out instead of opt-in? I don't know what your versioning policy is, of course - I completely understand if that's a breaking change that would be very hard to do any time soon. (But maybe it's worth considering for the next major version?)

lance:
> Right - in that case, it sounds like "throw when parsing invalid data" is reasonably common - so would it be feasible to change the CloudEvents SDK to do that by default as well, and make validation opt-out instead of opt-in? I don't know what your versioning policy is, of course - I completely understand if that's a breaking change that would be very hard to do any time soon. (But maybe it's worth considering for the next major version?)

Yes, it's possible. But I'm not sure I agree with this approach. We made the decision to be liberal in what we accept, but strict about user creation based on a discussion in the CE working group a couple of years ago. I knew this was all sounding familiar! :smile: If I recall correctly, @duglin weighed in on this at the time. I'm curious what your thoughts are, Doug.

We discussed it on the call (though I don't see the topic in meeting minutes from that time), and in the JavaScript SDK PR here: https://github.com/cloudevents/sdk-javascript/pull/328#issuecomment-682059664

My apologies for arriving late and leaving early on today's call. I am squeezed with overlapping meetings on each end.

duglin:
Geez, this was supposed to be an easy one :-)

Reviewing https://github.com/cloudevents/sdk-javascript/pull/328#issuecomment-682059664 I like that approach (strict on user creation, lax from the wire - by default) mainly because in the case of getting a bad CE from a remote host will (in many/most?) cases mean the sender may not detect the validation error quickly - if at all. Especially if the some middleware returns a 202 before any checks are done. And that means systems will break. And yes in a perfect world it's nice to say "we're going to force everyone to do what's right", but reality is that it's not always that easy. So to me as long as people can twiddle the config to get what they want then I think that's good enough.

But in the end, I'm not an SDK maintainer and I'll modify the PR based on what you guys decide

jskeet (approved):

jskeet:
> So to me as long as people can twiddle the config to get what they want then I think that's good enough.

That sounds reasonable - and the PR as it stands is already an improvement. Given that this is an area where SDKs may well be inconsistent with each other (but hopefully internally consistent), I think it would be good to require the per-language SDK policy to be clearly documented. We can come to that another time though :)

duglin:
Approved on the 3/10 call

PR #960, opened 2022-02-25
fix checker for macs

duglin:


duglin:
executive decision - merging

PR #961, opened 2022-02-25
Skip problematic file during link check

duglin:
I think github thinks we're ddos-ing it with all of these GETs so let's just
skip the file since it's not that important anyway.


duglin:
executive decision - merging

PR #962, opened 2022-02-25
Add some reasoning to why we limit things.

### LINKED ISSUE #955: Reason behind 20 character limit suggestion for attributes?
Hello, I was wondering why 20 was chosen as the recommended character limit for attributes? I see its not a hard and fast rule but was curious how this number came to be. Currently looking at my implementation I see I've exceeded the limit for some attributes but I don't understand the pros/cons of this beyond just trying to keep the whole message small vs more descriptive.

duglin:
Fixes #955

**Release Note**

```release-note
Add some text to explain why we're so restrictive on attribute names
```

### DIFF
--- a/cloudevents/spec.md
+++ b/cloudevents/spec.md
@@ -170,9 +170,11 @@ the available character set of all attributes such that case-sensitivity issues
 or clashes with the permissible character set for identifiers in common
 languages are prevented.
 
-CloudEvents attribute names MUST consist of lower-case letters ('a' to 'z') or
-digits ('0' to '9') from the ASCII character set. Attribute names SHOULD be
-descriptive and terse and SHOULD NOT exceed 20 characters in length.
+In order to maximize the likelihood of interoperability and portability across
+transport protocols and messaging formats, CloudEvents attribute names MUST
+consist of lower-case letters ('a' to 'z') or digits ('0' to '9') from the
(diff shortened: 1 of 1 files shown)


pierDipi (approved):
LGTM

cloudevents/spec.md
```diff
@@ -170,9 +170,11 @@ the available character set of all attributes such that case-sensitivity issues
 or clashes with the permissible character set for identifiers in common
 languages are prevented.
 
-CloudEvents attribute names MUST consist of lower-case letters ('a' to 'z') or
-digits ('0' to '9') from the ASCII character set. Attribute names SHOULD be
-descriptive and terse and SHOULD NOT exceed 20 characters in length.
+In order to maximize the likelihood of interoperability and portability across
+as many transport protocols and messaging formats, CloudEvents attribute names
```
jskeet, line 174: Add "as possible" before the comma? Otherwise "as many transport protocols and messaging formats" doesn't really work as a phrase :)

(Of course, having the requirement that the attributes can only consist of a-z and 0-9 *reduces* interoperability in other ways - if you're trying to convert another messaging format metadata *into* CloudEvents, and that other messaging format metadata has capital letters etc, then you're in a tricky position.)

duglin: went with option 2 - thanks

jskeet (approved):

duglin:
Approved on the 3/10 call

PR #972, opened 2022-03-16
Proposed clarifications to the XML format spec

jskeet:
The main intention of this is to be more specific about where we're flexible:

- No unexpected attributes in CE-specified elements
- No unexpected elements in CE attributes or data elements other than for xsi:any
- No unexpected text nodes
- Elements not in the CE namespace are allowed as direct children of event or batch

### DIFF
--- a/cloudevents/working-drafts/xml-format.md
+++ b/cloudevents/working-drafts/xml-format.md
@@ -23,7 +23,7 @@ This document is a working draft.
 [CloudEvents][ce-spec] is a standardized and protocol-agnostic definition of the
 structure and metadata description of events. This specification defines how
 elements defined in the CloudEvents specification are to be represented using
-[Extensible Markup Language (XML)](xml-spec) documents.
+[Extensible Markup Language (XML)](xml-spec) elements.
 
 * The [Attributes](#2-attributes) section describes the representation and
 data type mappings for CloudEvents context attributes.
@@ -59,7 +59,8 @@ The namespace `http://cloudevents.io/xmlformat/V1` MUST be used.
 When namespace prefixes are used a prefix of `ce` is preferred but MUST NOT
 be expected from an XML document processing perspective.
 
-An XML document preamble SHOULD be included to ensure deterministic processing.
+Where an event (or batch of events) is represented as a complete XML document,
+an XML document preamble SHOULD be included to ensure deterministic processing.
 
 ## 2. Attributes
 
@@ -85,11 +86,13 @@ Extension context attributes MUST be decorated with the appropriate CloudEvent t
 designators using an `xsi:type` XML attribute, this allows them to be exchanged
 without loss of type information.
 
-REQUIRED and OPTIONAL context attribute MAY be decorated with an`xsi:type`, if present this
-designator MUST match that of the type specified by the [CloudEvent context attributes][ce-attrs].
+An XML element representing a REQUIRED or OPTIONAL context attribute MAY be decorated with
+an `xsi:type` XML attribute. If present, this designator MUST match that of the type specified
+by the [CloudEvent context attributes][ce-attrs].
 
-No other XML element attributes are expected, if present they MUST be ignored during
-processing.
+An XML element representing a CloudEvent attribute MUST NOT contain any XML attributes
+other than `xsi:type` and attributes in the `xmlns` namespace. The XML element MUST NOT
+contain any child elements.
 
 ``` xml
     ...
@@ -109,12 +112,15 @@ encapsulate the payload.
 
 An `xsi:type` is used to discrimate the payload type and MUST be present.
 
+The `data` element MUST NOT contain any XML attributes
+other than `xsi:type` and attributes in the `xmlns` namespace.
+
 The following data representations are supported:
 
 ### 3.1 Binary Data
 
 Binary data MUST be carried in an element with an defined type of `xs:base64Binary`
-and encoded appropriately..
+and encoded appropriately. The element MUST NOT contain any child elements.
 
 Example:
 
@@ -125,6 +131,7 @@ Example:
 ### 3.2 Text Data
 
 Text MUST be carried in an element with an defined type of `xs:string`.
+The element MUST NOT contain any child elements.
 
 Example:
 
@@ -135,7 +142,9 @@ Example:
 ### 3.3 XML Data
 
 XML data MUST be carried in an element with a defined type of `xs:any` with
-a single child XML element (with any mandatory namespace definitions).
+a exactly one child XML element (with any mandatory namespace definitions).
+The XML element MUST NOT contain any direct child text nodes with non-whitespace
+content.
 
 Example:
 
@@ -173,6 +182,13 @@ Example _(XML preamble and namespace definitions omitted for brevity)_:
 </event>
 ```
 
+The `event` element MUST NOT contain any direct child text nodes with non-whitespace
+content, and it MUST NOT contain any XML attributes other than `specversion` and attributes
+in the `xmlns` namespace.
+
+The `event` element MAY contain child elements which are not in the namespace of
+`http://cloudevents.io/xmlformat/V1`. Any such elements MUST be ignored during processing.
+
 ## 5. XML Batch Format
 
 In the _XML Batch Format_ several CloudEvents are batched into a single XML
@@ -199,6 +215,12 @@ Example _(XML preamble and namespace definitions omitted for brevity)_:
     ....
 </batch>
 ```
+
+The `batch` element MUST NOT contain any direct child text nodes with non-whitespace
+content, and it MUST NOT contain any XML attributes other than ones in the `xmlns` namespace.
+
+The `batch` element MAY contain child elements which are not in the namespace of
+`http://cloudevents.io/xmlformat/V1`. Any such elements MUST be ignored during processing.
 
 ## 6. Examples
 
--- a/cloudevents/working-drafts/xml-format.md
+++ b/cloudevents/working-drafts/xml-format.md
@@ -94,6 +94,9 @@ An XML element representing a CloudEvent attribute MUST NOT contain any XML attr
 other than `xsi:type` and attributes in the `xmlns` namespace. The XML element MUST NOT
 contain any child elements.
 
+XML elements representing CloudEvent attributes MUST have distinct names within the same
+CloudEvent representation. This applies to both context attributes and extension attributes.
+
 ``` xml
     ...
     <id>AAABBBCCCNNN0000</id>
@@ -114,6 +117,8 @@ An `xsi:type` is used to discrimate the payload type and MUST be present.
 
 The `data` element MUST NOT contain any XML attributes
 other than `xsi:type` and attributes in the `xmlns` namespace.
+
+The `data` element MUST NOT occur more than once within an `<event>` element.
 
 The following data representations are supported:
 
--- a/cloudevents/working-drafts/xml-format.md
+++ b/cloudevents/working-drafts/xml-format.md
@@ -143,8 +143,8 @@ Example:
 
 XML data MUST be carried in an element with a defined type of `xs:any` with
 a exactly one child XML element (with any mandatory namespace definitions).
-The XML element MUST NOT contain any direct child text nodes with non-whitespace
-content.
+The `data` XML element MUST NOT contain any direct child text nodes with non-whitespace
+content. There are no restrictions on the content of the child element.
 
 Example:
 


jskeet:
(If some aspects of this are controversial and others aren't, I'm very happy to create separate PRs to get the uncontroversial bits merged first, so we end up with a diminishing set of bits still under discussion.)

cloudevents/working-drafts/xml-format.md
```diff
@@ -135,7 +142,9 @@ Example:
 ### 3.3 XML Data
 
 XML data MUST be carried in an element with a defined type of `xs:any` with
-a single child XML element (with any mandatory namespace definitions).
+a exactly one child XML element (with any mandatory namespace definitions).
+The XML element MUST NOT contain any direct child text nodes with non-whitespace
```
duglin, line 146: So, `<myevent>hello world</myevent>` is invalid?

jskeet: Whoops, I should clarify. (That's bad C&P.) It's meant to be that the `data` element must not contain any direct child text nodes.

cloudevents/working-drafts/xml-format.md
```diff
@@ -173,6 +182,13 @@ Example _(XML preamble and namespace definitions omitted for brevity)_:
 </event>
 ```
 
+The `event` element MUST NOT contain any direct child text nodes with non-whitespace
+content, and it MUST NOT contain any XML attributes other than `specversion` and attributes
+in the `xmlns` namespace.
+
+The `event` element MAY contain child elements which are not in the namespace of
+`http://cloudevents.io/xmlformat/V1`. Any such elements MUST be ignored during processing.
```
duglin, line 190: I don't think we want the MUST there, otherwise we're saying "yes you can have extensions but you can't actually use them". I think we want something more like "Unknown extension elements MUST NOT result in an error during receive processing" or something like that.

cloudevents/working-drafts/xml-format.md
```diff
@@ -200,6 +216,12 @@ Example _(XML preamble and namespace definitions omitted for brevity)_:
 </batch>
 ```
 
+The `batch` element MUST NOT contain any direct child text nodes with non-whitespace
+content, and it MUST NOT contain any XML attributes other than ones in the `xmlns` namespace.
+
+The `batch` element MAY contain child elements which are not in the namespace of
+`http://cloudevents.io/xmlformat/V1`. Any such elements MUST be ignored during processing.
```
duglin, line 223: same comment here about ignoring unknown rather than ignoring all.

cloudevents/working-drafts/xml-format.md
jskeet: Updated in an additional commit.

cloudevents/working-drafts/xml-format.md
```diff
@@ -173,6 +182,13 @@ Example _(XML preamble and namespace definitions omitted for brevity)_:
 </event>
```
duglin, line 182: This comment is probably more appropriate for the main spec, and I'll open a PR if people think we need it, but in formats like XML where the same element name can appear more than once do we need to add something along the lines of "Each CE attribute MUST only appear at most once"? I'm not sure we ever actually come right out and say each is a singleton - though it certainly is implied.

This popped in my head because I was trying to think of a case where someone may want to extend a CE attribute with an XML attribute and duplicating the CE attribute came to mind, e.g.:
```
<id foo:from="src1">123</id>
<id foo:from="src2">321</id>
```
Thoughts?

jskeet: Definitely! (I think I've got that in my test suite... I should really make sure that everything in there is covered at some point.) Coming up in another commit...

cloudevents/working-drafts/xml-format.md
```diff
@@ -85,11 +86,13 @@ Extension context attributes MUST be decorated with the appropriate CloudEvent t
...
-REQUIRED and OPTIONAL context attribute MAY be decorated with an`xsi:type`, if present this
-designator MUST match that of the type specified by the [CloudEvent context attributes][ce-attrs].
+An XML element representing a REQUIRED or OPTIONAL context attribute MAY be decorated with
+an `xsi:type` XML attribute. If present, this designator MUST match that of the type specified
+by the [CloudEvent context attributes][ce-attrs].
 
-No other XML element attributes are expected, if present they MUST be ignored during
-processing.
+An XML element representing a CloudEvent attribute MUST NOT contain any XML attributes
+other than `xsi:type` and attributes in the `xmlns` namespace. The XML element MUST NOT
```
jskeet, line 94: Note for discussion: what about the xml namespace, e.g. for `xml:space`?

cloudevents/working-drafts/xml-format.md
jskeet: Done. What about multiple elements in the same batch having the same ID+source? That might be overkill.

cloudevents/working-drafts/xml-format.md
duglin: I'm not sure what the "`xmlns` namespace" is. Do you mean "the CloudEvents XML namespace" ?

jskeet: I mean as in it's okay to have:

```xml
<ce:xyz xsi:type="string" xmlns:xsi=xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" />
```

(Or even introduce the `ce` namespace alias in the same element.)

duglin: True - and I would like to allow people to add extensions, even on the CE defined data. Just haven't found the right time to mention it....but I guess I just did :-)

duglin:
@jskeet can you sign your commits?

jskeet:
@duglin: Will wait until we're happy with the changes, then squash everything into a single signed commit, if that's okay with you.

cloudevents/working-drafts/xml-format.md
duglin: I'm leaning towards not touching the ID+source issue. The main spec already says "Producers MUST ensure that source + id is unique for each distinct event." so I think that should cover it. Since a replayed event could happen across two different messages, that spec text should be enough to cover those cases. I don't view the same event in a single batch as much different - just a less verbose case.

duglin:
of course!

cloudevents/working-drafts/xml-format.md
JemDay: The original statement was aimed to SDK writers, I don't believe there's a way we can accommodate "random stuff" from a predictable processing perspective.

By that what I meant was..  _"sure, go-ahead and sprinkle your random stuff into the XML document, but if you do that don't expect any of it to be preserved across different CE Formats and transports as it doesn't form a part of the CE information model and thus is not understood by CE aware s/w and infrastructure"_

cloudevents/working-drafts/xml-format.md
JemDay: See comment above ... if it's not "ignored" by an SDK,  what are we meant to do with it ...

None of the other Format specifications support this type of ad-hoc (format specific) extension model to the best of my knowledge.

cloudevents/working-drafts/xml-format.md
JemDay: +1 on this .. nice catch ...

cloudevents/working-drafts/xml-format.md
JemDay: From w3.org:

> The prefix xmlns is used only to declare namespace bindings and is by definition bound to the namespace name http://www.w3.org/2000/xmlns/. It MUST NOT be declared . Other prefixes MUST NOT be bound to this namespace name, and it MUST NOT be declared as the default namespace. Element names MUST NOT have the prefix xmlns

cloudevents/working-drafts/xml-format.md
JemDay: Could you post an example of invalid content ...what is the danger in the child xml element having such content ??

cloudevents/working-drafts/xml-format.md
duglin: @JemDay yes, and that's the way I'd like to see this spec go... "sure add extensions but there's no guarantees about it". In fact, that's kind of what we already do in general.... you can add new "ce-*" attributes w/o any guarantee that it'll be understood.

cloudevents/working-drafts/xml-format.md
jskeet: > See comment above ... if it's not "ignored" by an SDK, what are we meant to do with it ...

Raise an error, basically. This PR saying they should be ignored is a compromise - I would personally rather prohibit them entirely...

> None of the other Format specifications support this type of ad-hoc (format specific) extension model to the best of my knowledge.

Indeed - so I'm not sure why we should allow them here, really.

cloudevents/working-drafts/xml-format.md
jskeet: There's a difference between "allowing unknown attributes that still follow our data scheme" (we're not allowing anyone to include CloudExtension attributes with unsupported xsi types here, for example) and "we allow extra data completely outside the CloudEvent data model to be intermingled with the rest of it". Anyway, let's chat about it in the meeting later - if I'm the only one who wants to be restrictive about what's valid, I'm happy to write another version of this PR which is even more liberal in what's accepted. I *do* want to be absolutely explicit about it either way though.

cloudevents/working-drafts/xml-format.md
duglin: Raising an error would mean that a newer producer could never talk to an older consumer and I think that would be a big issue - especially if a consumer can not be upgraded at all, or at least not very often. 

As for not mentioned in other formats, if not then I think that's an oversight. It jumped out at me in this spec because XML allows for extensions WITHIN spec defined metadata. Meaning, something like:
```
<ce:id myns:foo="bar"> 123 </ce:id>
```
and as long as "foo" isn't in the "ce" namespace this is perfectly legal per the extensibility of XML - and is actually one of it's benefits.

jskeet: > Raising an error would mean that a newer producer could never talk to an older consumer and I think that would be a big issue - especially if a consumer can not be upgraded at all, or at least not very often.

But that's always going to be the case with a change to a data model that allows more than it did before. I'd expect that to count as XML format v2.

> As for not mentioned in other formats, if not then I think that's an oversight.

It would be good to explore this a bit more. For example, should an SDK processing a JSON format event ignore any properties starting with "_" as they can't possibly be CE attributes?

Protobuf is interesting here as we could definitely add more fields to the .proto file and they'd be parsed as "unknown field" by most proto implementations, so the data could be passed along without interpretation... but I'd personally only want to see people use that in terms of *us* extending the protobuf schema rather than users adding arbitrary fields.

> as long as "foo" isn't in the "ce" namespace this is perfectly legal per the extensibility of XML - and is actually one of it's benefits.

It's valid XML, but that doesn't mean it has to be a valid CloudEvent, any more than this is:

```xml
<ce:this_isnt_a_valid_extension_attribute_name>oops</ce:this_isnt_a_valid_extension_attribute_name>
```

I suspect this is a philosophical disagreement in terms of how beneficial extensibility is vs strictness, where I tend to favour strictness with only *very* well-defined extensibility and I believe you tend to favour more unrestricted extensibility. It's probably not worth trying to persuade each other, but I'd be interested to hear more voices than just the three of us involved in this discussion so far. I'm definitely happy to go along with the majority.

duglin: Using an underscore in a CE attribute name (core or extension) isn't the same thing as a new CE extension attribute since the spec mandates that all attributes (core and extensions) MUST adhere to our limited character set. So I view that as a different situation.

> I suspect this is a philosophical disagreement

Fur-sure! Whether it's specs or code, I get very nervous with a "we know what people will need" design pattern that doesn't allow for people to go beyond what the authors originally thought of because inevitably people will want more. And if they can't do any additions then they'll go find some other tech to use instead (or re-invent the wheel). I won't name names, but I can think a few projects that I view as "dead" due to this - they just may not know it yet :-)

While we joke about SOAP, their "mustUnderstand" attribute was brilliant!

duglin: Been thinking about the case @JemDay mentioned on the call today (json binary + invalid ce json top-level attribute), and I'm wondering if in this case we basically own the schema for the body since it's a CE mime type, which means we can decide what's legal. And in this case since all top level json properties are considered CE extension attributes they must adhere to our naming rules. This feels similar to us owning the CE XML namespace. However, in JSON there's only one namespace and no real way to specify something outside of that namespace.... which is why it's a different situation.  Thoughts?

jskeet: I agree that it's a different situation in that XML namespaces give us the *option* of allowing more extensibility. It doesn't *require* us to do so, however - having a single format that allows mingling of data while none of the rest do feels slightly odd, but it's not infeasible by any means.

cloudevents/working-drafts/xml-format.md
jskeet: Note: in my follow-up PR I haven't included this part, as @duglin's change to the main spec covers it.

jskeet:
Closing this for now as it's been somewhat-superseded, e.g. by #988. There's more work to do, but it'll be in a new PR.

PR #973, opened 2022-03-18
Make it clear each CE attribute is a singleton

duglin:
Per [this chat](https://github.com/cloudevents/spec/pull/972/files#r829061428), we should be very clear that each CE attribute can only appear at most once in an Event. While this introduces a MUST, it doesn't change the semantics of our intent and therefore could be seen as a clarification (or bug to not have had this in the past).

**Release Note**

```release-note
Make it clear each CE attribute can only appear at most once per event
```

/cc @jskeet

### DIFF
--- a/cloudevents/primer.md
+++ b/cloudevents/primer.md
@@ -18,7 +18,7 @@ itself to focus on the normative technical details.
 - [Architecture](#architecture)
 - [Versioning of CloudEvents](#versioning-of-cloudevents)
 - [CloudEvent Attributes](#cloudevent-attributes)
-- [CloudEvent Attribute Extensions](#cloudevent-attribute-extensions)
+- [CloudEvent Extension Attributes](#cloudevent-extension-attributes)
 - [Creating CloudEvents](#creating-cloudevents)
 - [Qualifying Protocols and Encodings](#qualifying-protocols-and-encodings)
 - [Proprietary Protocols and Encodings](#proprietary-protocols-and-encodings)
@@ -192,12 +192,12 @@ Confidentiality, as the current intention behind this spec is not to define
 Security principles with regards to CloudEvents. Every implementor has a
 different principle for enhancing their security model. We leave it up to the
 implementor of the spec to provide additional details for hardening their
-security model as an extension field, which can be furthermore interpreted by
-the components implemented by the implementor of the specification themselves.
-However, if the community observes a pattern in usage of certain extension
-fields, as a standard way to deal with the topic of data integrity. In that
-case, such extension fields can be declared as official extension to the
-CloudEvent specification.
+security model as an extension attribute, which can be furthermore interpreted
+by the components implemented by the implementor of the specification
+themselves. However, if the community observes a pattern in usage of certain
+extension attributes, as a standard way to deal with the topic of data
+integrity. In that case, such extension attributes can be declared as official
+extensions to the CloudEvent specification.
 
 ## Architecture
 
@@ -348,7 +348,7 @@ context, for the purposes of this CloudEvent attribute those meanings are not
 relevant and therefore using `id` for some other purpose beyond uniqueness
 checking is out of scope of the specification and not recommended.
 
-## CloudEvent Attribute Extensions
+## CloudEvent Extension Attributes
 
 In order to achieve the stated goals, the specification authors will attempt to
 constrain the number of metadata attributes they define in CloudEvents. To that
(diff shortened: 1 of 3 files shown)


cloudevents/spec.md
```diff
@@ -153,7 +153,8 @@ together into a transport envelope body.
 
 Every CloudEvent conforming to this specification MUST include context
 attributes designated as REQUIRED, MAY include one or more OPTIONAL context
-attributes and MAY include one or more extension attributes.
+attributes and MAY include one or more extension attributes. Each context
+attribute MUST only appear at most once in a CloudEvent.
```
jskeet, line 157: Shouldn't we require that for extension attributes as well?

jskeet: If not, I think we may have different mental models that would be worth discussing. My world view is:

A CloudEvent has attribute. Each attribute has a name, a type, and may have some validation constraints. Each attribute is either:

- Required by the declared spec version
- Optional but specified by the declared spec version
- An extension attribute, i.e. unspecified in the declared spec version

Typically "context attribute" is used to mean the first two but does not include extension attributes. Does that match your view? Maybe we should call this out and have a single term ("event attribute"? just "attribute") which can be applied whenever we mean all three. Or maybe "context attribute" should be specified to include them all?

duglin: I think we have the same mental model, and the spec does say:
- - -
A CloudEvent MAY include any number of additional context attributes with distinct names, known as "extension attributes". Extension attributes MUST follow the same [naming convention](https://github.com/cloudevents/spec/blob/44c9579dbe09d1b8f25a7655fc480e7f6d6b8796/cloudevents/spec.md#naming-conventions) and use the same [type system](https://github.com/cloudevents/spec/blob/44c9579dbe09d1b8f25a7655fc480e7f6d6b8796/cloudevents/spec.md#type-system) as standard attributes. 
- - -
which to me means extensions attributes are context attributes. But since it wasn't clear to you in this new sentence, there it might not be clear to others too. I modified the new sentence and tweaked the previous sentence to it even more clear that extensions attributes ARE context attributes too. See if that helps.

duglin: Actually, re-reading your comment:
> Typically "context attribute" is used to mean the first two but does not include extension attributes.

I disagree with this part :-)  I think context attributes include all 3 - as the spec implies in the text I quoted.

jskeet: Great. This all sounds good - I'll see if I can find where in the spec I got that impression from, or whether it was just a misreading.

jskeet (approved):

cloudevents/spec.md
jskeet: At a brief look, I think this sentence in spec.md could be changed very slightly:

> Every CloudEvent conforming to this specification MUST include context attributes designated as REQUIRED, MAY include one or more OPTIONAL context attributes and MAY include one or more extension attributes.

Just changing "extension attributes" to "extension context attributes" would help a lot.

The primer could do with a bit more work in terms of consistency - it talks about "CloudEvent Attribute Extensions" but only uses "context attribute" as a term in the versioning part. Obviously the primer isn't intended to be as formal as the main spec, but some consistency would be good.

I did find a term that I think would be good to use more widely when we mean "the attributes already in the spec" - "standard attributes".

Let me know if you want me to have a pass at clarifying things.

duglin: > Just changing "extension attributes" to "extension context attributes" would help a lot.

Yup - that was part of my last commit. I just push a tweak to the Primer to make the wording a bit more consistent

cloudevents/primer.md
```diff
@@ -192,12 +192,12 @@ Confidentiality, as the current intention behind this spec is not to define
...
-However, if the community observes a pattern in usage of certain extension
-fields, as a standard way to deal with the topic of data integrity. In that
-case, such extension fields can be declared as official extension to the
-CloudEvent specification.
+security model as an extension attribute, which can be furthermore interpreted
+by the components implemented by the implementor of the specification
+themselves. However, if the community observes a pattern in usage of certain
+extension attributes, as a standard way to deal with the topic of data
+integrity. In that case, such extension attributes can be declared as official
+extension to the CloudEvent specification.
```
jskeet, line 200: Nit: extension => extensions

jskeet (approved):

cloudevents/primer.md
duglin: fixed!

cloudevents/spec.md
```diff
@@ -153,7 +153,9 @@ together into a transport envelope body.
 
 Every CloudEvent conforming to this specification MUST include context
 attributes designated as REQUIRED, MAY include one or more OPTIONAL context
-attributes and MAY include one or more extension attributes.
+attributes and MAY include one or more
+[extension context attributes](#extension-context-attributes). Each
+context attribute MUST only appear at most once in a CloudEvent.
```
duglin, line 158: FYI - this new sentence at the end is the real change. The other stuff in this PR is either reformatting or s/field/attribute/ for consistency.

jskeet (approved):

duglin:
Approved on the 3/24 call

Issue #984, opened 2022-03-22
[Docs] Translate cloudevents/spec/blob/main/README.md to zh-CN


JieDing:
Document to translate:

https://github.com/cloudevents/spec/blob/main/README.md

Translation target:

https://github.com/cloudevents/spec/blob/main/languages/zh-CN/README.md

Translators who accept the task SHOULD follow [instructions for CloudEvents translation](https://github.com/cloudevents/spec/blob/main/languages/languages-spec.md).

JieDing:
Plz assign this to me. BTY, can you create a new label `{language-code}-translation` for all zh-CN translation issues? @duglin

JieDing:
The related-PR is merged.

PR #968, opened 2022-03-09
Proposed governance changes from the branching discussions

jskeet:
The intention is that this covers the *content* changes required as part of #954.

(There may be other files I'm not aware of... happy to be educated and take another pass.)

PR #985, opened 2022-03-23
docs: update languages/zh-CN/README.md

JieDing:
- add new translation languages/zh-CN/README.md
- add `README.md` and `RELEASE_NOTES.md` to `languages` sub-dirs of every spec directory
- update languages-spec

Signed-off-by: Jie Ding <jie.ding@linkall.cloud>

#984


languages/zh-CN/README.md
```diff
@@ -0,0 +1,142 @@
+# CloudEvents 规范
```
duglin, line 1: Should this be under the "cloudevents" dir? Or should there be a "cloudevents" subdir under "languages" ?
I'm leaning more towards "cloudevents/languages/..."

duglin, line 1: I guess we have a decision to make... should all CE docs (for all languages) be under `/cloudevents` or should we group all docs for each languages under `/languages/LANG` ? 

ie.
```
/README.md
/languages/README.md
/cloudevents/README.md
/cloudevents/languages/zn-cn/README.md
/cloudevents/languages/zn-cn/spec.md
/discovery/README.md
/discovery/languages/zn-cn/README.md
/discovery/languages/zn-cn/spec.md
```
vs
```
/README.md
/cloudevents/README.md
/discovery/README.md
/languages/zh-cn/README.md
/languages/zh-cn/cloudevents/README.md
/languages/zh-cn/cloudevents/spec.md
/languages/zh-cn/discovery/README.md
/languages/zh-cn/discovery/spec.md
```

I'm wondering if our new release process, which seems to be focused on releasing "directories" (sort of), would influence this decision.

JieDing: For the first situation, each directory has it's own sub-dir /languages/LANG, while the other grouping all docs for each languages under /languages/LANG.  The difference between these two situations reflects two different specification structure.
The first structure seems to treat each directory as an independent spec.
i.e. 
```
/cloudevents  dir represents the CloudEvents spec
/discovery dir represents the Discovery spec
```
If that's the structure we want to release, then it makes sense that each dir holds its language sub-dir.

The second structure seems to treat all directories as the whole CloudEvents spec.
```
/cloudevents  dir represents the CloudEvents core spec
/discovery dir represents the CloudEvents additional spec about discovery
```
For that structure, it seems better if we group all docs under /languages/LANG.

So it all depends on what kind of spec structure we want to release finally.

JieDing: I personally prefer the second one coz I think that the CloudEvents spec should consist of core spec and other related specs.

JieDing: Or maybe we can re-org the structure like this:
```
/cloudevents
    /core    -> this core dir holds everything(like SDK.md, spec.md, adapters dir, etc) from the original /cloudevents dir
    /discovery   
    /subscription  
    /schemeregistry  
    /cesql 
    ...
/misc
/community
...
```

duglin: Not sure if you were aware of the decision by the group but we decided to have the other spec dirs (eg. "discovery") hold all info for that spec. Basically like a nested git repo, but w/o the overhead of the repo. Then each release would be a 'tag' off of the 'main' branch with the dir being released as part of the tag name. Yes this means we'll be deleting the other branches and just have 'main' going forward. See: https://github.com/cloudevents/spec/pull/968

So, your statement:
> The first structure seems to treat each directory as an independent spec.

is how I view the current direction of the group. But we can wait for other folks to chime in to see if they have a different interpretation.

JieDing: I went through the tag-based structure. It does make sense. I noticed it's approved on the 3/18 call. So is this direction settled or not?

duglin: yes I just need to find the time to "make it so".

JieDing: Cool, I'll update the language spec then to ensure each directory has a "languages" sub-dir.

duglin:
@JieDing let me know when you think each PR Is ready to be considered by the group for merging. I don’t want to do it prematurely

JieDing:
> @JieDing let me know when you think each PR Is ready to be considered by the group for merging. I don’t want to do it prematurely

@duglin I reviewed this PR and thought it's ready for merging.

PR #988, opened 2022-03-25
Minor tweaks to the XML format

jskeet:
This does not attempt to cover any of the controversial aspects of #972
and indeed there may be some uncontroversial aspects of #972 that
I haven't yet taken.)

### DIFF
--- a/cloudevents/working-drafts/xml-format.md
+++ b/cloudevents/working-drafts/xml-format.md
@@ -23,7 +23,7 @@ This document is a working draft.
 [CloudEvents][ce-spec] is a standardized and protocol-agnostic definition of the
 structure and metadata description of events. This specification defines how
 elements defined in the CloudEvents specification are to be represented using
-[Extensible Markup Language (XML)](xml-spec) documents.
+[Extensible Markup Language (XML)](xml-spec) elements.
 
 * The [Attributes](#2-attributes) section describes the representation and
 data type mappings for CloudEvents context attributes.
@@ -59,7 +59,8 @@ The namespace `http://cloudevents.io/xmlformat/V1` MUST be used.
 When namespace prefixes are used a prefix of `ce` is preferred but MUST NOT
 be expected from an XML document processing perspective.
 
-An XML document preamble SHOULD be included to ensure deterministic processing.
+Where an event (or batch of events) is represented as a complete XML document,
+an XML document preamble SHOULD be included to ensure deterministic processing.
 
 ## 2. Attributes
 
@@ -85,8 +86,9 @@ Extension context attributes MUST be decorated with the appropriate CloudEvent t
 designators using an `xsi:type` XML attribute, this allows them to be exchanged
 without loss of type information.
 
-REQUIRED and OPTIONAL context attribute MAY be decorated with an`xsi:type`, if present this
-designator MUST match that of the type specified by the [CloudEvent context attributes][ce-attrs].
+An XML element representing a REQUIRED or OPTIONAL context attribute MAY be decorated with
+an `xsi:type` XML attribute. If present, this designator MUST match that of the type specified
+by the [CloudEvent context attributes][ce-attrs].
 
 No other XML element attributes are expected, if present they MUST be ignored during
 processing.
@@ -104,10 +106,12 @@ processing.
 ## 3. Data
 
 The data portion of a CloudEvent follows a similar model to that employed by
-the [JSON Format specification][json-format]. A `data` element MUST be used to
+the [JSON Format specification][json-format]. A `<data>` element MUST be used to
 encapsulate the payload.
 
 An `xsi:type` is used to discrimate the payload type and MUST be present.
+
+The `<data>` element MUST NOT occur more than once within an `<event>` element.
 
 The following data representations are supported:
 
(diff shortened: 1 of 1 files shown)


cloudevents/working-drafts/xml-format.md
```diff
@@ -135,7 +139,7 @@ Example:
 ### 3.3 XML Data
 
 XML data MUST be carried in an element with a defined type of `xs:any` with
-a single child XML element (with any mandatory namespace definitions).
+exactly one child XML element (with any mandatory namespace definitions).
```
jskeet, line 142: This probably isn't strictly required, but feels like it gives more of a nudge to SDK authors to validate it.

cloudevents/working-drafts/xml-format.md
```diff
@@ -85,8 +86,9 @@ Extension context attributes MUST be decorated with the appropriate CloudEvent t
 designators using an `xsi:type` XML attribute, this allows them to be exchanged
 without loss of type information.
 
-REQUIRED and OPTIONAL context attribute MAY be decorated with an`xsi:type`, if present this
-designator MUST match that of the type specified by the [CloudEvent context attributes][ce-attrs].
+An XML element representing a REQUIRED or OPTIONAL context attribute MAY be decorated with
```
jskeet, line 89: Would "standard context attribute" be better than "a REQUIRED or OPTIONAL context attribute"? (The point is to differentiate it from extension context attributes.)

duglin: s/standard/specification defined/   perhaps?  In case "standard" wasn't clear since some people might interpret it as "normal" instead of as an alternative for "spec"

duglin:
After we get agreement on the "standard" question.... LGTM - thanks!

cloudevents/working-drafts/xml-format.md
JemDay: I think this reads OK...

JemDay (commented):
LGTM

cloudevents/working-drafts/xml-format.md
jskeet: I think it would be useful to have a term that we *can* use in format specifications, consistently. I think we've used "standard" elsewhere (possibly in the main spec itself). One tricky aspect is that "specification-defined" could be understood (incorrectly in this case) to include "extensions documented in this repository, even if they're not part of the core specification document". Happy to discuss options here.

duglin: How about something like "CloudEvents namespaced context attributes"? Take advantage of the fact we're using XML.

jskeet: A context attribute doesn't *have* a namespace though - context attribute is a CloudEvent concept, and namespace is an XML concept. The two don't go together... which means I wonder if I'm misunderstanding you.

duglin: We're in the XML format spec. Each CE defined attribute (not extensions) MUST appear in the CE XML namespace, right? Aren't those the attributes this sentence is referring to?

jskeet: No - *every* context attribute (whether it's defined in the spec or not) must be represented by an XML element in the CE XML namespace. This paragraph is trying to distinguish between spec-defined context attributes and extension attributes, in terms of `xsi:type` requirements.

duglin: oh doi - I was putting extensions into a user-defined namespace.  Hmm it feels odd not to though. Doesn't it?
grr I don't know what to type next about this because any direction I take it in my head I feel like something breaks LOL

jskeet: > oh doi - I was putting extensions into a user-defined namespace. Hmm it feels odd not to though. Doesn't it?

No, I don't think so. In particular, it fits in with the way the JSON format handles extensions - they're *not* in a separate property, because if an extension becomes part of the main standard, it's useful if an existing formatted message is still valid. (Admittedly the message already contains the version number, so...)

The way I see it, the CE namespace is intended to mean "Elements and attributes in this namespace adhere to the CE XML format" - which is still true for extension attributes. If we want to go with "you can intermingle non-CE data with CE data in XML elements" then using a different namespace is a way of indicating "this isn't CE data, and shouldn't be treated as such".

duglin: Yes that is one way to view it. But the standards geek in me says that only spec authors are allowed to define things in THEIR XML namespace and extensions MUST use a different namespace. So it feels like we're breaking the XML model a bit.

But as you said, this isn't how we do it for JSON because we wanted to allow extensions to migrate to "spec defined" attributes w/o a name change. And this is where things break for me when thinking about things from the other direction.

duglin: I'm leaning towards a position of: XML is different from JSON, etc... in that it has a well defined extensibility model so we need to use it. Even if that means the conversion of extensions to spec-defined properties means a breaking change on the wire by forcing a NS change. That's pretty much how XML was designed.

jskeet: If we're going to expect elements in other namespaces to be extension attributes, then we should expect that's the *only* thing they can be, and we shouldn't allow any other extensibility.

Personally this feels (to me) significantly more restrictive than what's currently documented, but let's discuss it.

PR #989, opened 2022-03-25
rebuild multi-language structure

JieDing:
- add new 'languages' sub-dir for every spec directory
- update languages spec

Signed-off-by: Jie Ding <jie.ding@linkall.cloud>


duglin:
I'm going to merge this since it's mainly just a re-org and I assume it would be best to get it in sooner rather than later given there are other PRs in-flight and will need to be rebased due to this change.

@JieDing thanks for all of the hard work on this!

Issue #998, opened 2022-04-07
XML format "looseness" voting issue


jskeet:
This is a somewhat experimental way of gathering opinions on this issue, but hopefully it'll be clear enough.

Each of my comments below contains:

- A description of what it's trying to convey
- A complete example of an XML CloudEvent or batch of CloudEvents. It doesn't include the XML document preamble, but does include all namespaces - so if anything is missing, its presence should *not* be inferred.

For each example, please add a 👍 response if you think it should be valid, and a 👎 if you think it should be invalid. (😕 is okay if you're on the fence!) The assumption is that any "extra data" that we deem valid would be ignored by SDKs by default, although any given SDK could have hooks to make additional information available of course.

jskeet:
**Example 1: additional attributes in the `<event>` element**

```xml
<event additional="extra" specversion="1.0" xmlns="http://cloudevents.io/xmlformat/V1">
    <id>sample-id</id>
    <source>https://github.com/cloudevents</source>
    <type>io.cncf.sample</type>
</event>
```

jskeet:
**Example 2: text node in the `<event>` element**

```xml
<event specversion="1.0" xmlns="http://cloudevents.io/xmlformat/V1">
    Additional text
    <id>sample-id</id>
    <source>https://github.com/cloudevents</source>
    <type>io.cncf.sample</type>
</event>
```

jskeet:
**Example 3: non-CE-namespace element in the `<event>` element**

```xml
<event specversion="1.0" xmlns="http://cloudevents.io/xmlformat/V1">
    <additional xmlns="http://example.com">This is not in the CE namespace</additional>
    <id>sample-id</id>
    <source>https://github.com/cloudevents</source>
    <type>io.cncf.sample</type>
</event>
```

jskeet:
**Example 4: additional attribute in an element representing a context attribute**

```xml
<event specversion="1.0" xmlns="http://cloudevents.io/xmlformat/V1">
    <id additional="extra">sample-id</id>
    <source>https://github.com/cloudevents</source>
    <type>io.cncf.sample</type>
</event>
```

jskeet:
**Example 5: element within an element representing a context attribute**

```xml
<event specversion="1.0" xmlns="http://cloudevents.io/xmlformat/V1">
    <id>sample-id
      <additional>Extra</additional>
    </id>
    <source>https://github.com/cloudevents</source>
    <type>io.cncf.sample</type>
</event>
```

[...]

PR #1002, opened 2022-04-20
Next round of restrictions in XML format

jskeet:
These represent the following clear votes from #998:

- Example 2: text node in an `<event>` element
- Example 5: element within a context attribute element
- Example 6: element within binary `<data>` element
- Example 7: element within a text `<data>` element
- Example 8: text within an "any" `<data>` element
- Example 10: text node in a `<batch>` element
- Example 12: comments are allowed anywhere

The remaining examples did not have clear-cut votes, although it
looks like we *probably* want to prohibit additional XML attributes
within context attribute elements.

### DIFF
--- a/cloudevents/working-drafts/xml-format.md
+++ b/cloudevents/working-drafts/xml-format.md
@@ -63,7 +63,9 @@ Where an event (or batch of events) is represented as a complete XML document,
 an XML document preamble SHOULD be included to ensure deterministic processing.
 
 XML comments are permitted anywhere within an XML representation of a
-CloudEvent or CloudEvent batch, and MUST be ignored during processing.
+CloudEvent or CloudEvent batch. Comments MUST be ignored during processing, except
+for the child element of a `<data>` element containing [XML data](#33-xml-data),
+where comments SHOULD be preserved.
 
 ## 2. Attributes
 
@@ -146,6 +148,9 @@ Example:
 
 XML data MUST be carried in an element with a defined type of `xs:any` with
 exactly one child XML element (with any mandatory namespace definitions).
+
+Unlike other parts of CloudEvent XML message processing, XML comments within
+the child XML element SHOULD be preserved during processing.
 
 The `<data>` XML element MUST NOT contain any direct child text nodes with
 non-whitespace content. There are no restrictions on the content of the
--- a/cloudevents/working-drafts/xml-format.md
+++ b/cloudevents/working-drafts/xml-format.md
@@ -62,6 +62,9 @@ be expected from an XML document processing perspective.
 Where an event (or batch of events) is represented as a complete XML document,
 an XML document preamble SHOULD be included to ensure deterministic processing.
 
+XML comments are permitted anywhere within an XML representation of a
+CloudEvent or CloudEvent batch, and MUST be ignored during processing.
+
 ## 2. Attributes
 
 The CloudEvents type system is mapped to a set of CloudEvent specific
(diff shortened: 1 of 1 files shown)


cloudevents/working-drafts/xml-format.md
```diff
@@ -62,6 +62,9 @@ be expected from an XML document processing perspective.
 Where an event (or batch of events) is represented as a complete XML document,
 an XML document preamble SHOULD be included to ensure deterministic processing.
 
+XML comments are permitted anywhere within an XML representation of a
+CloudEvent or CloudEvent batch, and MUST being ignored during processing.
```
duglin, line 66: s/being/be/

jskeet: Doh - fixed, thanks.

cloudevents/working-drafts/xml-format.md
```diff
@@ -62,6 +62,11 @@ be expected from an XML document processing perspective.
 Where an event (or batch of events) is represented as a complete XML document,
 an XML document preamble SHOULD be included to ensure deterministic processing.
 
+XML comments are permitted anywhere within an XML representation of a
```
jskeet, line 65: @duglin I've modified this in a second commit (only separated from the first so it's easier to spot the differences and change if necessary). I've included the behavior both here and in section 3.3 so it's harder to miss. See what you think.

cloudevents/working-drafts/xml-format.md
```diff
@@ -62,6 +62,11 @@ be expected from an XML document processing perspective.
...
+CloudEvent or CloudEvent batch. Comments MUST be ignored during processing, except
+for the child element of a `<data>` element containing [XML data](#33-xml-data),
+where comments SHOULD be preserved.
```
duglin, line 68: Perhaps this should be a MUST to ensure interop between SDKs?

cloudevents/working-drafts/xml-format.md
```diff
@@ -141,6 +149,13 @@ Example:
 XML data MUST be carried in an element with a defined type of `xs:any` with
 exactly one child XML element (with any mandatory namespace definitions).
 
+Unlike other parts of CloudEvent XML message processing, XML comments within
+the child XML element SHOULD be preserved during processing.
```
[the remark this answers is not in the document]
jskeet: If interoperability *depends* on handling comments, that would suggest something is using comments inappropriately, IMO :) (I suspect we have far greater areas of potential data loss - not least in our handling of timestamps where the UTC offset can be dropped while propagating an event, for example...)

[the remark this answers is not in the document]
duglin: It's not a question of whether they're using comments "inappropriately", it's about whether people would expect their data to be "untouched". We have no idea how, or if, people are counting on the comments to be retained, we shouldn't be removing them from the user's data w/o a really good reason.

cloudevents/working-drafts/xml-format.md
duglin: Been thinking about this one some more... and while I still believe that we should not be touching the "data" at all, and try to pass it along "as is" as best we can, perhaps it might be better to solve this at the spec level instead of at the format level. Then we don't need to say anything in this doc about data+comments at all.

If so, I then wonder about the other text related to comments since w/o the "except..." part of the sentence someone could interpret it as it being ok to remove comments from data. Perhaps rewording the "Comments MUST be ignored...." sentence as:
```
The presence of comments MUST NOT impact the CloudEvents metadata related processing.
```
This scopes it to just CE processing and not app-level processing.

Thoughts?

jskeet: Ah, but I said that about comments *with respect to XML*. I'm imagining a different text-based format that has very few options (e.g. no attributes) and where using comments for metadata is somewhat-idiomatic. That's the problem - if we move this to the spec, it needs to apply to all formats, including those we haven't thought of yet... although I guess we could always say "Unless the format says otherwise".

Anyway, let's get more input from other folks tomorrow. No head-messing was intended - just an added bonus ;)

duglin:
Approved on the 4/28 call

PR #1029, opened 2022-07-12
Add some clarity around nested CEs

### LINKED ISSUE #653: Primer: nested event details seem confusing
"Creating cloud events" contains this paragraph: https://github.com/cloudevents/spec/blame/master/primer.md#L419-L426

>There might exist special cases in which it is necessary to create a CloudEvent that contains another CloudEvent. Although the specification does not define nesting explicitly, it is possible. While the inner event will always be encoded in a stand-alone event format, the outer event can be either binary or structured mode. The datacontenttype attribute of the outer event must not be set to application/cloudevents+json or any other media type that is used to denote the usage of structured mode. A correct example of event nesting would be:

Which is... confusing to me.

1. "the inner event will always be encoded in a stand-alone event format": This seems unnecessarily prescriptive, as [stand-alone event formats are defined as being protocol-agnostic](https://github.com/cloudevents/spec/blob/master/spec.md#event-format), and seems to imply there is no other reasonable option.  Is there some advantage to be gained by choosing e.g. json?  It doesn't seem like consumers of the outer-event can generically understand any nested-event, especially as the spec defines no nesting (as it mentions in the second sentence)... so they'll necessarily have to understand _something_ about it being a custom nested format, and can do whatever is needed in that case.

2. "The datacontenttype attribute of the outer event must not be set to application/cloudevents+json ...": There don't seem to be any implied universal behaviors for specific datacontenttypes, much less `application/cloudevents+json` specifically, so I don't see why this should be avoided.  If there is some reason why it must be avoided?

3. There's no spec, but it's possible, and there's a correct way to do it: this whole paragraph seems a bit self-contradictory.  Maybe it'd be better to just get rid of it?

So I guess to summarize my questions:
1. why would protocol-agnostic inner data formats be the only option (equivalent to MUST)?
2. why MUST datacontenttype `application/cloudevents+json`(etc) be avoided?
3. should this paragraph just be removed?  if there's no spec for nesting, why mention it?

duglin:
Fixes #653

### DIFF
--- a/cloudevents/spec.md
+++ b/cloudevents/spec.md
@@ -126,8 +126,8 @@ Bindings MAY define formats that are dependent on the protocol.
 
 Events are transported from a source to a destination via messages.
 
-A "structured-mode message" is one where the event is fully encoded using
-a stand-alone event format and stored in the message body.
+A "structured-mode message" is one where the entire event (attributes and data)
+are encoded in the message body.
 
 A "binary-mode message" is one where the event data is stored in the message
 body, and event attributes are stored as part of message metadata.
--- a/cloudevents/primer.md
+++ b/cloudevents/primer.md
@@ -363,7 +363,7 @@ The core attributes are those defined within the specification itself.
 Core attributes then have subcategories of required or optional. As the category
 names imply, "required" attributes will be the ones that the group considers
 vital to all events in all use cases, while "optional" ones will be used in a
-majority of the cases. 
+majority of the cases.
 
 When the group determines that an attribute is not common enough to fall into
 those two categories but would still benefit from the level of interoperability
@@ -496,12 +496,17 @@ the incoming CloudEvent.
 
 There might exist special cases in which it is necessary to create a CloudEvent
 that contains another CloudEvent. Although the specification does not define
-nesting explicitly, it is possible. While the inner event will always be encoded
-in a [stand-alone event format](spec.md#event-format), the outer event
-can be either binary or structured mode. The `datacontenttype` attribute of
-the outer event must not be set to `application/cloudevents+json` or any other
-media type that is used to denote the usage of structured mode. A correct
-example of event nesting would be:
+nesting explicitly, it is possible. When doing so, the inner event will always
+use [structured mode](spec.md#event-format), this is because if the inner event
+was in binary mode then its metadata could be misinterpretted as metadata for
+the outer event. The outer event can use either binary or structured mode,
+however, its `datacontenttype` attribute must not be set to
+`application/cloudevents+json` or any other media type that is used to denote
+the usage of structured mode. This is because it would then be difficult to
+know if the use of `application/cloudevents+json` is due to the outer event
+using structured mode or due to the inner event being a CloudEvent.  One option
+in this case is to use a value of `application/json` instead, as seen in this
+example:
 
 ```
 Content-Type: application/json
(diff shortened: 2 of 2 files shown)


cloudevents/spec.md
```diff
@@ -118,7 +118,7 @@ information.
 #### Event Format
 
 An Event Format specifies how to serialize a CloudEvent as a sequence of bytes.
-Stand-alone event formats, such as the [JSON format](formats/json-format.md),
+Structured event formats, such as the [JSON format](formats/json-format.md),
```
deissnerk, line 121: The [terminology section](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#event-format) of the spec talks about _structured mode_ vs. _binary mode_. The expression _stand-alone format_ is introduced for event formats that do not depend on a protocol:
> Stand-alone event formats, such as the [JSON format](https://github.com/cloudevents/spec/blob/main/cloudevents/formats/json-format.md), specify serialization independent of any protocol or storage medium. Protocol Bindings MAY define formats that are dependent on the protocol.

duglin: Is this comment just an FYI, or are you saying we need to keep "stand-alone" here?

deissnerk: I'm sorry, I actually didn't realize that you are intending to change exactly the part of the spec I was quoting. I thought you were just changing the primer and wanted to maintain consistency with the spec. 

Nevertheless, I think _stand-alone event format_ is not synonymous with _structured mode_. The term was introduced because of AMQP, where you could have a _structured format_ that would NOT be a _stand-alone format_. 

Reading through the terminology again, I get the feeling that it is not fully consistent regarding the distinction of event formats, messages, and binary vs. structured mode. I could explain in next week's call what I mean. We can then decide, if we extend this PR, or if we create a separate one.

duglin: What's the diff between "structured" and "stand-alone" to you? If there really is a diff I think we need to define "stand-alone" and be more clear on when each is used because I think the spec/primer today use them interchangeably.

deissnerk: In my opinion ANY usage of an event format suggests structured mode. We intentionally introduced structured vs. binary MODE. The addition _stand-alone_ was just introduced to distinguish formats that are independent from the used protocol from others that have dependencies. As an example for a non-stand-alone format you could think of an AMQP format. The AMQP messaging layer allows to specify an AMQP sequence or value instead of a binary section. In this case the whole event would be transformed to the AMQP type system. 
For the nesting case I intentionally used the term _stand-alone_, because the inner event is supposed to be in a protocol-independent format. 
I see an event format as a function _f(context, data)_ that returns a byte stream. In binary mode you don't really have that, as _context_ and _data_ are handled separately.

duglin: I think I understand what you're saying but I think it's confusing to a reader because we use the term "stand-alone" as the generic way to talk about "structured" messages. And at an abstract level that might be accurate, but I think it adds confusion because in our cases "stand-alone" == "structured", so switching between the two (or switching between being "abstract" and "concrete") doesn't change the point in the text we're trying to make... which is, when they must use "a structured CE" or not.

In the case of a nested CE, the inner CE (I believe) MUST be a structured CE, not just a "stand-alone" message, right? If you're suggesting that for certain serializations their "stand-alone" format allows for CE attributes to be injected and therefore we don't need a separate CE structured format, then that might be true but I would think then that the stand-alone+CE attributes definition IS the CE structured format for that serialization. Which means we can still use the term "CE structure" instead of "stand-alone". But a I re-read what I wrote, maybe I'm still missing your point :-)  Let's chat off-line

duglin:
Approved on the 7/28 call

PR #1032, opened 2022-07-14
PROPOSAL: minor versions

jskeet:
This is an initial drafty proposal for how we might handle minor versions. It's intended for discussion as a pull request, although we could potentially end up with it in the working-drafts folder if we think the idea has merit but we're not *quite* ready to turn it into spec language yet.

### DIFF
--- a/cloudevents/working-drafts/minor-versions.md
+++ b/cloudevents/working-drafts/minor-versions.md
@@ -1,0 +1,337 @@
+# Allowing minor version numbers for the CloudEvents spec (and specversion attribute)
+
+## Introduction
+
+We're building up a backlog (currently not in a very organised way)
+of things we would like to do with the CloudEvents spec in the
+future. Some of these are *definitely* in the realm of a new major
+version, but there has been discussion of some changes that could
+legitimately be called a minor version - if we supported such a
+thing. This proposal proposes a change to our spec versioning to
+allow for minor versions.
+
+Note: this document uses "event" as a shorthand for "CloudEvent" throughout for
+simplicity. If any part of this document becomes part of the spec, we can
+rewrite it appropriately.
+
+## Is changing our mind about minor versions a breaking change?
+
+The current spec describes the `specversion` attributes as follows:
+
+> The version of the CloudEvents specification which the event uses.
+> This enables the interpretation of the context. Compliant event
+> producers MUST use a value of 1.0 when referring to this version of
+> the specification.
+> 
+> Currently, this attribute will only have the 'major' and 'minor'
+> version numbers included in it. This allows for 'patch' changes to
+> the specification to be made without changing this property's value
+> in the serialization. Note: for 'release candidate' releases a
+> suffix might be used for testing purposes.
+
+The GOVERNANCE.md doc states:
+
+> Since changing the CloudEvents specversion string could have a
+> significant impact on implementations, all non-breaking changes will
+> be made as "patch" version updates - this allows for the value "on
+> the wire" to remain unchanged. If a breaking change is introduced
+> the normal semver rules will apply and the "major" version number
+> will change. The net effect of this is that the "minor" version
+> number will always be zero and the specversion string will always be
+> of the form X.0.
+
+I'd view the governance doc as rather less binding than the spec -
+it's "current practice" rather than a promise.
+
+Any SDK (and potentially other consuming code) that *expects* the
+current documented will fail to parse a v1.1 (etc) event - and
+that's probably okay. For maximum compatibility, publish v1.0
+events for a while. But we don't even need a v1.1 right now, so
+there's enough time for SDKs to start implementing the below
+compatibility code *now*, and so hopefully by the time we actually
+want a v1.1, the majority of deployed software will be able to
+handle it.
+
+## Out of scope (major)
+
+There are probably many changes which could count as major, but I'd
+like to call out the following explicitly:
+
+- Any change to required attributes:
+  - Adding
+  - Removing
+  - Changing type
+  - Changing validation criteria
+- Changes to the type system:
+  - Adding types
+  - Removing types
+- Changes to attribute naming rules
+- Breaking changes to CloudEvent formats
+
+## Intended minor change
+
+Fundamentally, the purpose of this proposal is to allow the
+introduction of new *optional* attributes, which are
+expected to usually have been documented extension attributes in the
+past - in other words, a graduation process from "suggested as useful"
+to "officially specified". However, there's no requirement for an
+attribute to be a documented extension attribute before being added;
+that's just an informal part of the usual process.
+
+SemVer minor versions are intended for new features that don't break
+existing code. How SemVer applies to specifications is less clear-cut
+than how it applies to actual code. This document attempts to imagine
+the impact of a change on various stakeholders.
+
+For the purpose of this document, we'll consider a single new optional
+attribute into version 1.1:
+
+- Name: `xyz`
+- Type: integer
+- Constraints: value in the range 1-100 inclusive
+
+Obviously this is entirely artificial, but it should be enough to help
+find corner cases and problems.
+
+## Logical compatibility
+
+Every valid v1.1 event can be "rewritten" as a valid v1.0 event, but the reverse is not true.
+A v1.0 event containing an `xyz` extension attribute with value "test" is valid,
+but cannot be expressed as a valid v1.1 event. As there are no restrictions on
+v1.0 extension attributes, there's no way that the v1.1 specification can *claim*
+that all v1.0 events would be valid as v1.1 events, even if the addition
+is only a previously-documented extension attribute. (Extension attribute documentation
+is advisory, not mandatory.)
+
+## Event producer
+
+An event producer that doesn't know about v1.1 should create an event
+with v1.0. If the code producing the event uses an SDK, then it's *possible*
+that it will create a v1.1 event accidentally after an SDK upgrade - and that
+could cause problems. (The SDK handling of minor versions is relatively
+complex, and discussed later.) But if they produce a v1.0 event with an
+`xyz` attribute with value "test", that's fine: it's still a valid v1.0 event.
+
+An event producer that is aware of v1.1 should ensure that any use of the `xyz`
+attribute is valid - in other words, it can't have values of "test"
+(wrong type) or 123 (constraint violation). That would be an invalid v1.1 event
+in the same way that specifying a `time` attribute value of "lunchtime" would be
+invalid.
+
+## Event proxies
+
+Proxies which are unaware of v1.1 should be able to proxy v1.1 events without
+any issues, *unless* they attach special meaning to unspecified attributes (whether
+`xyz` or anything else). For example, a proxy which makes use of the `sequence` extension
+attribute but isn't aware of v1.0 cannot safely assume that v1.1 events will use `sequence`
+in the same way as v1.0. (`sequence` may become a core attribute with a particular meaning.)
+The proxy could make a choice of:
+
+- Expecting that the previous attribute *is* used in the same way, and treating
+  it as per v1.0 events, with a suitable failure mode.
+- Not attaching any significance to the attribute for v1.1 events, treating it as
+  any other extension attribute.
+
+There's no one-size-fits-all approach here.
+
+## Event consumers
+
+An event consumer that doesn't know about v1.1 should still be able to consume
+a v1.1 event without any knowledge of the `xyz` attribute; it should be treated as a regular
+extension attribute. However, the consumer does "know" that the event is v1.1, so can be aware
+that `xyz` (and any other extension attributes) *may* be a core attribute in v1.1.
+(In other words, they can distinguish between `xyz` appearing in a v1.0 event which is *definitely*
+an extension attribute as the consumer is aware of all core v1.0 attribtes,
+and `xyz` appearing in a v1.1 event which is *either* an extension attribute *or* a new optional
+core attribute.)
+
+## SDKs
+
+If we adopt minor versions, SDKs should change to support them, so that (for example) an SDK
+supporting v1.1 will *also* parse v1.2 events. We should keep track of which SDKs
+(and which versions of those SDKs) support parsing v1.x events.
+
+Different SDKs expose features very differently. This document is written from the perspective
+of the C# SDK, but other viewpoints would be highly valued: please add feedback, particularly
+if the approach suggested doesn't make sense for an SDK you're familiar with.
+
+### Creating events: default version handling
+
+It may sound natural to allow something like:
+
+```csharp
+var evt = new CloudEvent { Id = "..." };
+```
+
+... to just use the latest spec version that the SDK supports. However, this is a breaking change.
+Consider:
+
+```csharp
+var evt = new CloudEvent
+{
+    Id = "...",
+    ["xyz"] = "test"
+};
+```
+
+When the SDK only supports v1.0, that creates a valid event. When the SDK supports v1.1 as well,
+that would produce an invalid event or throw an exception, if the version is implicitly v1.1.
+Instead, the SDK should *implicitly* use the latest version that was present when the first stable
+version of the SDK came out, within its own current major version. In other words, if v1.1 ships
+and a new SDK is created, that can default to v1.1 - and if an SDK creates its own new major version,
+*that* can default to v1.1 as well... but in both cases that default should stay as v1.1 in new minor
+SDK versions even after v1.2 ships.
+
+An SDK *may* choose to provide a way of explicitly opting into the "latest supported" version:
+
+```csharp
+var evt = new CloudEvent(CloudEventsSpecVersion.Latest)
+{
+  
+};
+```
+
+### Accessing attributes
+
+If an SDK exposes core attributes in a distinct way from extension attributes (as shown above,
+where `Id` is a strongly-typed property, whereas `["xyz"]` uses an indexer with a string literal)
+then any attempt to use the "special" access should fail when accessing an attribute within
+an event with a version that doesn't have that attribute as a core attribute. Example:
+
+```csharp
+// Create a v1.0 event
+var evt = new CloudEvent(CloudEventsSpecVersion.V1_0);
+
+// Try to access the new Xyz property, referring to the optional attribute introduced in
+// v1.1: this should fail.
+evt.Xyz = 1;
+
+// Access the xyz attribute by name, as an extension attribute.
+// This should work, even though the value wouldn't be valid in a v1.1 event.
+evt["xyz"] = "test";
+
+// Likewise accessing the property for an event parsed from elsewhere:
+evt = httpRequest.ParseCloudEvent();
+// If evt is v1.0, this should fail *even if it has an xyz attribute value that's valid in v1.1.
+int xyz = evt.Xyz;
+```
+
+Setting a value within an event where the version indicates that the attribute *is* known should
+validate the attribute value according to that version. Example:
+
+```csharp
+// Create a v1.1 event
+var evt = new CloudEvent(CloudEventsSpecVersion.V1_1);
+
+// This should work: 1 is a valid value, and xyz is known in v1.1
+evt.Xyz = 1;
+
+// This should fail: 123 is not a valid value for xyz in v1.1
+evt.Xyz = 123;
+
+// This should fail to compile, where that's feasible: xyz should always be an integer
+evt.Xyz = "test";
+
+// This should succeed at execution time
+evt["xyz"] = 123;
+
+// This should fail at execution time due to validation
+evt["xyz"] = "test";
+
+// This should fail at execution time due to validation
+evt["xyz"] = 123;
+```
+
+### "Changing" event versions
+
+An SDK may provide the ability to change the specversion of an event,
+or (preferred) create a new object based on the old one. When "upgrading" an event's specversion,
+the SDK should perform validation as if parsing the event from scratch. Examples:
+
+```csharp
+// Successful upgrade: no xyz value
+var evt10 = new CloudEvent(CloudEventsSpecVersion.V1_0) { Id = "..." };
+var evt11 = evt10.WithSpecVersion(CloudEventsSpecVersion.V1_1);
+
+// Successful upgrade: valid value for xyz
+var evt10 = new CloudEvent(CloudEventsSpecVersion.V1_0)
+{
+    Id = "...",
+    ["xyz"] = 10
+};
+evt11 = evt10.WithSpecVersion(CloudEventsSpecVersion.V1_1);
+
+// Failed upgrade: invalid value for xyz
+var evt10 = new CloudEvent(CloudEventsSpecVersion.V1_0)
+{
+    Id = "...",
+    ["xyz"] = "test"
+};
+evt11 = evt10.WithSpecVersion(CloudEventsSpecVersion.V1_1);
+
+// Successful downgrade
+var evt11 = new CloudEvent(CloudEventsSpecVersion.V1_1)
+{
+    Id = "...",
+    Xyz = 10
+};
+evt10 = evt11.WithSpecVersion(CloudEventsSpecVersion.V1_0);
+// Access to evt10.Xyz would fail as described above;
+// access to evt10["xyz"] would return 10.
+// Setting evt10["xyz"] to "test" or 123 could fail or succeed - it's
+// up to the SDK to decide (and document!) whether downgrading retains constraints
+// for attributes that were present in the original event.
+// TODO: What about if Xyz hadn't been present before? Should evt10["xyz"] = "test" fail then?
+```
+
+Until we have a new major version, we probably don't want to specify whether
+SDKs should allow upgrade/downgrade across major versions.
+
+### Handling unknown versions
+
+An SDK which supports minor versions **MUST** be able to parse events with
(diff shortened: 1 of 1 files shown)


cloudevents/working-drafts/minor-versions.md
```diff
@@ -0,0 +1,337 @@
...
+- Constraints: value in the range 1-100 inclusive
+
+Obviously this is entirely artificial, but it should be enough to help
+find corner cases and problems.
+
+## Logical compatibility
+
+Every valid v1.1 event can be "rewritten" as a valid v1.0 event, but the reverse is not true.
+A v1.0 event containing an `xyz` extension attribute with value "test" is valid,
+but cannot be expressed as a valid v1.1 event. As there are no restrictions on
```
duglin, line 100: True but we do (sort of) warn people about picking a name that might be used in the future.
> New extension definitions SHOULD use a name that is descriptive enough to reduce the chances of name collisions with other extensions. In particular, extension authors SHOULD check the [documented extensions](https://github.com/cloudevents/spec/blob/main/cloudevents/documented-extensions.md) document for the set of known extensions - not just for possible name conflicts but for extensions that might be of interest.

Granted, this talks about other extensions but I believe it was meant to also cover possible future CE defined attributes. Perhaps we need to call that out.

My point is that adding extensions is always a risky business if you don't choose a name that's unique or has some scoping mechanism/namespace that others won't use.  Not saying we shouldn't do what you're suggesting, just saying that this is the nature of extensions and a version of this problem will always exist no matter what we do.

duglin: Note that even HTTP itself has a variant of this problem since new RFC defined headers can be created at any time that overlap with an extension header. And people just deal with it. At some point this might come down to "is it worth the effort".

cloudevents/working-drafts/minor-versions.md
```diff
@@ -0,0 +1,337 @@
...
+attribute but isn't aware of v1.0 cannot safely assume that v1.1 events will use `sequence`
+in the same way as v1.0. (`sequence` may become a core attribute with a particular meaning.)
+The proxy could make a choice of:
+
+- Expecting that the previous attribute *is* used in the same way, and treating
+  it as per v1.0 events, with a suitable failure mode.
+- Not attaching any significance to the attribute for v1.1 events, treating it as
+  any other extension attribute.
+
+There's no one-size-fits-all approach here.
```
duglin, line 135: I wonder what happens for proxies that do things like convert binary CEs to structured CEs, and in the process upgrade the version string?

cloudevents/working-drafts/minor-versions.md
```diff
@@ -0,0 +1,337 @@
...
+```
+
+... to just use the latest spec version that the SDK supports. However, this is a breaking change.
+Consider:
+
+```csharp
+var evt = new CloudEvent
+{
+    Id = "...",
+    ["xyz"] = "test"
```
duglin, line 173: What's interesting here is that if the SDK supported:
```
["Id"] = "..."
```
too (meaning also allow all attributes to be defined using the same generic syntax and the SDK would check to see if the "xxx" is a known name or not) then this problem isn't as bad any more.

cloudevents/working-drafts/minor-versions.md
```diff
@@ -0,0 +1,337 @@
...
+then any attempt to use the "special" access should fail when accessing an attribute within
+an event with a version that doesn't have that attribute as a core attribute. Example:
+
+```csharp
+// Create a v1.0 event
+var evt = new CloudEvent(CloudEventsSpecVersion.V1_0);
+
+// Try to access the new Xyz property, referring to the optional attribute introduced in
+// v1.1: this should fail.
+evt.Xyz = 1;
```
duglin, line 207: I actually disagree that it should fail. If the SDK has explicitly decided to support an extension as a first class thing then I think this should work.

cloudevents/working-drafts/minor-versions.md
```diff
@@ -0,0 +1,337 @@
...
+
+// Access the xyz attribute by name, as an extension attribute.
+// This should work, even though the value wouldn't be valid in a v1.1 event.
+evt["xyz"] = "test";
```
duglin, line 211: I think it would be nice if SDKs supported this for spec defined attributes too :-)

cloudevents/working-drafts/minor-versions.md
```diff
@@ -0,0 +1,337 @@
...
+
+// Likewise accessing the property for an event parsed from elsewhere:
+evt = httpRequest.ParseCloudEvent();
+// If evt is v1.0, this should fail *even if it has an xyz attribute value that's valid in v1.1.
+int xyz = evt.Xyz;
```
duglin, line 216: again, I disagree if the SDK chooses to know about it as a first class attribute.
Yes they're taking a risk the same way the extension author was taking a risk by using a name that the spec might use later on... but that's a risk they're willing to take. And that risk is always there w.r.t. conflicting extensions.

cloudevents/working-drafts/minor-versions.md
```diff
@@ -0,0 +1,337 @@
...
+// This should work: 1 is a valid value, and xyz is known in v1.1
+evt.Xyz = 1;
+
+// This should fail: 123 is not a valid value for xyz in v1.1
+evt.Xyz = 123;
+
+// This should fail to compile, where that's feasible: xyz should always be an integer
+evt.Xyz = "test";
+
+// This should succeed at execution time
```
duglin, line 235: I think you meant "fail" right?

cloudevents/working-drafts/minor-versions.md
```diff
@@ -0,0 +1,337 @@
...
+evt["xyz"] = 123;
+
+// This should fail at execution time due to validation
+evt["xyz"] = "test";
+
+// This should fail at execution time due to validation
+evt["xyz"] = 123;
```
duglin, line 242: dup of line 236?
However, note that this is allowing known attributes to be set via the extension mechanism, which is good.

cloudevents/working-drafts/minor-versions.md
```diff
@@ -0,0 +1,337 @@
...
+evt11 = evt10.WithSpecVersion(CloudEventsSpecVersion.V1_1);
+
+// Successful downgrade
+var evt11 = new CloudEvent(CloudEventsSpecVersion.V1_1)
+{
+    Id = "...",
+    Xyz = 10
+};
+evt10 = evt11.WithSpecVersion(CloudEventsSpecVersion.V1_0);
+// Access to evt10.Xyz would fail as described above;
```
duglin, line 279: If not an SDK defined core attribute, yes.

cloudevents/working-drafts/minor-versions.md
```diff
@@ -0,0 +1,337 @@
...
+// access to evt10["xyz"] would return 10.
+// Setting evt10["xyz"] to "test" or 123 could fail or succeed - it's
+// up to the SDK to decide (and document!) whether downgrading retains constraints
+// for attributes that were present in the original event.
```
duglin, line 283: I'd think this should always work since for a v1.0 event it can be either value. If you allow for the 1.1 constraints to live on, then your `evt10.Xyz` example should work if you want to be consistent.

cloudevents/working-drafts/minor-versions.md
```diff
@@ -0,0 +1,337 @@
...
+
+Until we have a new major version, we probably don't want to specify whether
+SDKs should allow upgrade/downgrade across major versions.
+
+### Handling unknown versions
+
+An SDK which supports minor versions **MUST** be able to parse events with
+a minor version of which it's unaware, but **MUST NOT** parse events with a
+*major* version of which it's unaware. In other words, a v1.1-aware SDK must
+be able to parse a v1.2 event, but must reject any attempt to parse a v2.0 event.
```
duglin, line 295: Would the resulting CE be a 1.1 CE or a 1.2 CE?
No matter which way it goes something feels wrong.
If it's 1.1 then are we lying to the user because the CE came in as 1.2.
If it's 1.2 then are we teasing the user because they can't access the 1.2 attributes as first class attributes and we won't do any validation on the 1.2 attributes like we would for 1.1 attributes?

cloudevents/working-drafts/minor-versions.md
```diff
@@ -0,0 +1,337 @@
...
+When parsing an event with an unknown minor version:
+
+- All attribute values must be retained, with as much type information as is available
+  (defaulting to string when type information is not available)
+- Values of unknown attributes must be readable as if they're regular extension attributes
+  (so a minor-version aware SDK that only supports v1.0 must allow access to `xyz` of
+  a v1.1 event after receiving it)
+- All attribute values must be validated according to the highest-known minor version
+  within the same major version
+- The original specversion must be retained, and available to SDK users
```
duglin, line 313: See comment above about lying/teasing.

cloudevents/working-drafts/minor-versions.md
```diff
@@ -0,0 +1,337 @@
...
+an SDK to only support v1.0 and v1.1, but it must not support *only* v1.1 or *only* v1.2.)
+
+When parsing an event with an unknown minor version:
+
+- All attribute values must be retained, with as much type information as is available
+  (defaulting to string when type information is not available)
+- Values of unknown attributes must be readable as if they're regular extension attributes
+  (so a minor-version aware SDK that only supports v1.0 must allow access to `xyz` of
+  a v1.1 event after receiving it)
+- All attribute values must be validated according to the highest-known minor version
```
duglin, line 311: All attribute values must be validated according to the CE specversion specified in the incoming CE, or if that minor version is unknown to the SDK then it must use the highest-known minor version's semantics.

cloudevents/working-drafts/minor-versions.md
```diff
@@ -0,0 +1,337 @@
...
+  within the same major version
+- The original specversion must be retained, and available to SDK users
+- If changing event versions is supported, the SDK should allow an event of an unknown minor
+  version to be downgraded to a known minor version, but should *not* allow an event
+  to be upgraded to an unknown minor version.
```
duglin, line 316: This seems like it should only "not allow" it if there's a semantic conflict.

cloudevents/working-drafts/minor-versions.md
```diff
@@ -0,0 +1,337 @@
...
+
+After an event with an unknown minor version has been parsed, an SDK *may* enforce that
+extension attributes cannot be modified. This ensures that the SDK can serialize the event
```
duglin, line 319: This feels a bit restrictive since it could break existing code that's using an extension that isn't even defined in a newer CE spec version.

cloudevents/working-drafts/minor-versions.md
jskeet: That ends up being slightly tricky for the C# SDK, as an event can be created with known extension attributes which can have constraints. So it would be feasible to create a 1.0 event with the xyz attribute with the same constraints as in 1.1. I *think* I'd expect that to be the behavior when downgrading. But this is definitely a detailed area we should only get into if we actually go ahead with the whole thing.

cloudevents/working-drafts/minor-versions.md
jskeet: It would be a 1.2 event, but I'd expect the SDK to have a way of saying "This is 1.2 and I don't understand that fully."

cloudevents/working-drafts/minor-versions.md
jskeet: I think we're in agreement here, and it's just a matter of being clearer in the wording - is that right?

cloudevents/working-drafts/minor-versions.md
jskeet: It's only lying if you're not transparent about it :)

cloudevents/working-drafts/minor-versions.md
jskeet: The SDK can't know if there's a semantic conflict or not, because it doesn't know about the target minor version - *unless* there are actually no extension attributes in the event.

I suppose we *could* go with "hey, we can upgrade it and make it clear that we don't know whether or not it's valid" as if we'd parsed it using the new (unknown) minor version.

cloudevents/working-drafts/minor-versions.md
jskeet: Yes, it's restrictive... that's the cost of being protective and avoiding producing an invalid event. I wonder whether SDKs should actually allow the user code to decide on what risks they want to take.

jskeet (commented):
Thanks for the comments, Doug. I don't yet have a read on whether you think this is *broadly* going in the right direction or whether:

- It should be much looser, with far fewer restrictions anywhere, with increased risk of invalid events being propagated
- We should just abandon the idea of minor versions and stay with the status quo

cloudevents/working-drafts/minor-versions.md
duglin: They dropped the use of X-* headers, I believe, for many of the same reasons we didn't create an "extension" bag of attributes. It's a false security blanket and eventually causes pain as things move from being extensions to core.

cloudevents/working-drafts/minor-versions.md
duglin: > But it means something in 1.1 and not in 1.0

This is a v1.0 CE so the SDK is safe, and allowed, to define their own attributes. The fact it MIGHT mean something in a future CE spec shouldn't impact what it does for v1.0.

Now, if `evt` was a 1.1 CE then `evt.Xyz = 1;` should fail.

cloudevents/working-drafts/minor-versions.md
duglin: I wonder if the SDK doc should encourage this, if it doesn't already (I couldn't find it).

cloudevents/working-drafts/minor-versions.md
duglin: Yes the SDK user is the one who feels the pain, but "taking on the risk" is an interesting statement because you seem to be assuming that they know the boundaries of CE 1.0 and therefore knows that "xyz" isn't part of CE 1.0 - they might not. To them, CE core might be whatever the SDK exposes as a core property. Much in the same way a compiler user may not bother to check to see if a certain features of their compiler/language is part of that language's spec or it's an extension that this particular compiler added.

But either way... yes there's risk/pain associated with all of this and it's true no matter what we do - even if we never create a new 1.x optional attribute.

duglin:
> Thanks for the comments, Doug. I don't yet have a read on whether you think this is broadly going in the right direction or whether:
> -    It should be much looser, with far fewer restrictions anywhere, with increased risk of invalid events being propagated
> -   We should just abandon the idea of minor versions and stay with the status quo

I'm leaning towards just keeping things as they are today. Meaning, we can do minor versions (with new optional attributes) but the version string on the wire doesn't change. I say this because then it puts extensions and optional attributes on par with each other. One interesting thing to me is that the spec specifically says this about optional core attributes:
> A consumer that does not support that feature will then silently ignore that part of the message.

To me, a consuming SDK that chooses to only know about required attributes should still expose the unsupported attributes to the user in some way - if nothing else as an extension attribute. This means that SDK supported attributes might have first class support (e.g. via `ce10.source`), but whether the SDK exposes the optional or extension attributes via `ce10.xxx` is an impl choice. And, I would argue that both optional and extensions could be supported via `ce10.xxx`.

Additionally, the concern about attribute conflicts (between new optional ones and extensions) would then apply equally between them. That symmetry is appealing to me.

But, I think it would help if you could add a short summary to your write-up that listed the short list of changes (or restrictions) an SDK, or code in general, would need to make to support your idea. That might help clarify exactly what the impact of your proposal would be on both SDKs and users.

jskeet:
> I'm leaning towards just keeping things as they are today. Meaning, we can do minor versions (with new optional attributes) but the version string on the wire doesn't change.

That means an event which is valid today might not be valid tomorrow. (If it uses a string value for an unspecified extension attribute that becomes an optional integer attribute, for example.) That makes me *really* nervous - I don't see why we wouldn't highlight that change with a new version.

I don't have time to add anything more in this PR today (I'm going on vacation for a week starting tomorrow, which means lots of loose-end-tying with the day job) but I'll be at the meeting later and we can discuss it then. I think it would be good to try to agree on some principles, and then work out what that means for SDKs. (To me, the principle of "if it's valid today, it's valid tomorrow" is a really important one.)

duglin:
> That means an event which is valid today might not be valid tomorrow

Yup and I think I'm less nervous about it than you because to me that's the nature of extensions - no matter what you do you're at risk. Your example is an interesting one because (w/o proof) I kind of wonder how realistic it is since if we decide that we need "foo" added as an optional attribute you'd think that the odds of it being a different type (or at least a different set of valid chars) than the extension would be pretty low - unless folks did a really bad job at naming it. But, again, no proof :-)

jskeet:
> Yup and I think I'm less nervous about it than you because to me that's the nature of extensions - no matter what you do you're at risk.

I don't think that's the case with the proposals in this PR. There's a risk that a valid v1.0 event wouldn't be valid as a v1.1 event, but that bothers me less - because the existing events could still flow through systems which are upgraded to be aware of 1.1.

> Your example is an interesting one because (w/o proof) I kind of wonder how realistic it is since if we decide that we need "foo" added as an optional attribute you'd think that the odds of it being a different type (or at least a different set of valid chars) than the extension would be pretty low.

I agree that it's *reasonably* unlikely - but I'd still like to eliminate that risk if possible.

clemensv:
Per discussion; regarding X- https://datatracker.ietf.org/doc/html/rfc6648

cloudevents/working-drafts/minor-versions.md
```diff
@@ -0,0 +1,337 @@
...
+```csharp
+var evt = new CloudEvent(CloudEventsSpecVersion.Latest)
+{
+  
+};
+```
+
+### Accessing attributes
+
+If an SDK exposes core attributes in a distinct way from extension attributes (as shown above,
```
n3wscott, line 196: This is how the go sdk works. It supported `1.0` to `0.3` conversions by puting the known new property into a `0.3` extension, but it is true that you would be unable to upconvert a `1.0` `xyz` string encoded string property into a `1.1` string encoded int.

cloudevents/working-drafts/minor-versions.md
n3wscott: The go sdk does this first class support and does it by having whatever is returned from `var evt = new CloudEvent(CloudEventsSpecVersion.V1_0);`, evt is strongly typed as a `v1.0` event. It would have extensions accessors but `evt.Xyz` is not possible. The member did not exist and does not exist in v1.0.

cloudevents/working-drafts/minor-versions.md
```diff
@@ -0,0 +1,337 @@
...
+```csharp
+// Create a v1.0 event
+var evt = new CloudEvent(CloudEventsSpecVersion.V1_0);
+
+// Try to access the new Xyz property, referring to the optional attribute introduced in
+// v1.1: this should fail.
+evt.Xyz = 1;
+
+// Access the xyz attribute by name, as an extension attribute.
+// This should work, even though the value wouldn't be valid in a v1.1 event.
```
n3wscott, line 210: the go sdk assumption is that it is not valid, if the evt is known to be a 1.1 object, setting an extension with a name of a known first class property is strongly typed to the type defined in the spec. This will fail for v1.1 event implementations but succeed for v1.0. It will throw an exception for conversion requests.

cloudevents/working-drafts/minor-versions.md
n3wscott: I would agree this would fail for a 1.0 event. The type `evt` is not aware of `Xyz` in the implementation, so the code should not compile in a strongly typed language.

cloudevents/working-drafts/minor-versions.md
n3wscott: I think the comment is about what is possible at compile time for a typed language. There is no hook here if you access the member directly, and you do not call a validation method. So at compile time, there is not opportunity to validate this style of direct value setting.

n3wscott: There might be a typo in the comment, this should be a compile time example.

cloudevents/working-drafts/minor-versions.md
```diff
@@ -0,0 +1,337 @@
...
+// This should fail: 123 is not a valid value for xyz in v1.1
+evt.Xyz = 123;
+
+// This should fail to compile, where that's feasible: xyz should always be an integer
+evt.Xyz = "test";
+
+// This should succeed at execution time
+evt["xyz"] = 123;
+
+// This should fail at execution time due to validation
```
n3wscott, line 238: Fail only when there is an active request to validate or translate the event

cloudevents/working-drafts/minor-versions.md
n3wscott: The point here is the "core attributes" are not defined genetically, they are versioned in the implementations.

cloudevents/working-drafts/minor-versions.md
```diff
@@ -0,0 +1,337 @@
...
+evt["xyz"] = "test";
+
+// This should fail at execution time due to validation
+evt["xyz"] = 123;
+```
+
+### "Changing" event versions
```
n3wscott, line 245: For ref, the go sdk does this: https://github.com/cloudevents/sdk-go/blob/main/v2/event/eventcontext_v1.go#L145 

Each version implementation had a method added that allowed it to move one version up or one version down. You could also implement this in a different way if you had a generic event wrapper with something like a `core` and `extensions` inner members that had versioned implementations. Then you could write a `Upgrade` and `Downgrade` methods that would do the lookup work and/or fail if the event is at the highest/lowest supported versions.

cloudevents/working-drafts/minor-versions.md
n3wscott: At the moment the go sdk fails if you present a minor version it does not understand, but I could see how we could support the idea of a typed "Next" version in the same minor family. It would have core members that match the newest known version but present newer core members as extensions to the integration. It would also possibly change some validation rules, mostly about what a valid version number is.

cloudevents/working-drafts/minor-versions.md
```diff
@@ -0,0 +1,337 @@
...
+- All attribute values must be retained, with as much type information as is available
+  (defaulting to string when type information is not available)
+- Values of unknown attributes must be readable as if they're regular extension attributes
+  (so a minor-version aware SDK that only supports v1.0 must allow access to `xyz` of
+  a v1.1 event after receiving it)
+- All attribute values must be validated according to the highest-known minor version
+  within the same major version
+- The original specversion must be retained, and available to SDK users
+- If changing event versions is supported, the SDK should allow an event of an unknown minor
+  version to be downgraded to a known minor version, but should *not* allow an event
```
n3wscott, line 315: If the sdk is able to receive a version 1.x event, it seems like a possible situation that there is a producer or consumer that is only able to process 1.x events, in this case you might want to allow the upgrade (possibly with a warning?) as a break glass integration case.

cloudevents/working-drafts/minor-versions.md
n3wscott: I would prefer to default to users being able to choose and if they mutate the event into an invalid future version, that was their choice and they must have known what they were doing. This is all assuming there was some amount of warning that they see.

The use case I am thinking about is middleware adding a fingerprint to the event as it passes, if an upstream producer starts producing v1.Next events before the middleware is upgrades, the fingerprinting would cause a new unexpected error, and it would be uncalled for in my opinion.

cloudevents/working-drafts/minor-versions.md
duglin: Right and that's fine for an official SDK - but we need to remember that anyone who creates code to deal with CEs is like an SDK and whatever rules we define apply to them as well. It also means that anyone who creates custom CE processing code could choose to support `evt.Xyz` as an extension - which means I think it would be wrong for us to mandate that `evt.Xyz = 1` MUST fail.

cloudevents/working-drafts/minor-versions.md
duglin: See comment above about extensions - not all CE processing code will come from us.

The mental model I have is similar to how some C compilers add extensions to the ISO standard C definition. Ignoring flags like a --strict type of compilation mode, the extensions appear on-par with the standard stuff. This feels like the same type of thing to me.

cloudevents/working-drafts/minor-versions.md
duglin: I think that's an impl choice

duglin:
The more I think about my C compiler comparison the more I like it :-) because for the most part there aren't versions involved - it's just "cc myapp.c" and so I think looking at how people deal with that might be a fair comparison.

If my compiler defines an extension that eventually conflicts with the standard C lang definition, then when I upgrade my compiler (assuming they change things to be compliant) then my code will most likely not compile cleanly - no promise that an old program will work with the latest (or different C) compiler if I use extensions. That's just the risk of defining and using extensions.

TBH this is partially why I sometimes wish we didn't have a version string at all - it gives a false sense of security. Kind of like how we tell people "if you stick with the Kubernetes resources your app is 100% portable" or the "v1" part of the of the K8s version strings guarantee compatibility :-)

duglin:
On the Sept 1 call we agreed to close this for now but really put it "on hold" until we have a concrete need to address. Jon will keep the branch around for people to read if they want.

PR #1041, opened 2022-08-09
discovery draft v2

duglin:
First pass at the next major rev of the discovery spec.

Tries to align with this: https://drive.google.com/file/d/1C8JrRPMlzwk9uBvHEaRVbfZU2orHmeGp/view

### DIFF
--- a/discovery/spec.md
+++ b/discovery/spec.md
@@ -1,51 +1,41 @@
-# CloudSubscriptions: Discovery - Version 0.1-wip
+# Discovery Service - Version 0.2-wip
 
 ## Abstract
 
-CloudSubscriptions Discovery API is a vendor-neutral API specification for
-determining what events are available from a particular service, as well as how
-to subscribe to those events.
+The Discovery Service specification allows for the discovery of messaging
+Endpoints and their related metadata such as Message Definitions and the
+mechanisms by which interactions with those Endpoints can be established.
 
 ## Table of Contents
 
 - [Overview](#overview)
 - [Notations and Terminology](#notations-and-terminology)
 - [Resource Model](#resource-model)
+- [Serializations](#serializations)
+- [Usage Values](#usage-values)
+- [Message Formats](#message-formats)
 - [API Specification](#api-specification)
 - [Privacy & Security](#privacy-and-security)
 
 ## Overview
 
-In order for consumers to receive events from services, they need to first
-subscribe, or ask for, events from those services. To do so, there is often a
-process necessary that involves steps such as discovering which service is of
-interest, what events it generates and how to create the subscription for those
-events.
+Message Endpoints provide a location from which Message producers and consumers
+can share Messages. Endpoints could act as either consumers or producers
+of Messages, and the mechanisms by which those Messages are transferred
+might vary.
 
-This specification defines a set of APIs to allow for consumers to perform these
-queries against a "Discovery Endpoint". A Discovery Endpoint acts as a catalog
-of [Services](#service) (event producers), that consumers can query to find the
-ones of interest. Once found, additional metadata is provided in order to
-consume and subscribe to events. The goal of this API is to be such that tooling
-can be built where all possible services and event types aren't known in
-advance.
+This specification defines a set of APIs, and related documents, that allow
+for the discovery of these Endpoints and their related metadata. This
+information can then be used to perform actions such as:
+- subscribing for Messages
+- pushing or pulling of Messages with the Endpoint
+- generating of Message validation code based on the schema associated with
+  a Message Definition
 
-The deployment relationship of a Discovery Endpoint to the Services and Event
-Producers that it advertises is out of scope of this specification. For example,
-a Discovery Endpoint could choose to be implemented as part of a Service, or
-Event Producer, or it could be acting as an independent aggregator of this
-metadata. This implementation detail will be transparent to consumers.
+While the specification is written in term of "messaging", it applies equally
+to "eventing" since it can be considered as subset of messaging.
 
-The main use-case to consider from the viewpoint of the event consumer is what
-services are available, and what event types do they generate?
-
-The CloudEvent `source` attribute is a potential cause of high fan-out. For
-example, consider a blob storage system where each directory constitutes a
-distinct `source` attribute value. For this reason, the exact CloudEvents
-`source` attribute value that might appear in a CloudEvent will not appear in
-the Discovery API query result. Instead, a higher level classifier (`service`)
-will be used to represent the abstract notion of the generic event producer of
-those events - in the example case, the blob storage service itself.
+** we need more verbage here **
 
 ## Notations and Terminology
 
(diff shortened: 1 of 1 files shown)


discovery/spec.md
```diff
@@ -55,305 +45,194 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
+  - `{ "owner": "John", "isVerified: "" }`
 
-##### docsurl
+#### docsurl
 
-- Type: `URL`
-- Description: Absolute URL that provides a link to additional documentation
-  about the service. This is intended for a developer to use in order to learn
-  more about this service's events produced.
+- Type: URL
```
n3wscott, line 186: I find it odd that this is a URL and not a URI-Ref. This choice forces the implementations to be aware of their public endpoints. It raises the bar on implementation and config distribution.

It makes me wonder if anything should be forced to produce an absolute URL, and URI-Ref is what should be used. This lets you continue using the context you are communicating over (the host + extras) or the

discovery/spec.md
```diff
@@ -55,305 +45,194 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
+- Description: A URL to additional documentation about this Endpoint. This
+  specification does not define any constraints on the data returned from this
+  URL.
 - Constraints:
   - OPTIONAL
-  - If present, MUST be a non-empty absolute URI
+  - if present, MUST be a non-empty URL
```
n3wscott, line 192: We might need to declare URL is only `HTTP/HTTPS`

discovery/spec.md
```diff
@@ -55,305 +45,194 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
+    client is expected to investigate the Endpoint to determine if it is
     appropriate.
 
   - docsurl<br>
     An OPTIONAL property specifying the URL to additional information about
-	the deprecation of the Service. This specification does not mandate any
-	particular format or information, however some possibilities include:
-	reasons for the deprecation or additional information about likely
-	alternative Services.
+    the deprecation of the Endpoint. This specification does not mandate any
```
n3wscott, line 220: Would this (/why would this) docsurl be different than the endpoint docsurl?

discovery/spec.md
```diff
@@ -55,305 +45,194 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
-      ]
-    }
-  ]
+    "id": "URI-reference",
+    "name": "STRING",
+    "self": "URI",
+    "epoch": UINT,
+    "description": "STRING", ?
+    "tags": { "STRING": "STRING", ... } ?
+    "docsurl": "URL", ?
```
n3wscott, line 86: I would rename this to `docs`

discovery/spec.md
```diff
@@ -55,305 +45,194 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
+- Type: Object container the following properties:
   - effectivetime<br>
-    An OPTIONAL property indicating the time when the Service entered, or will
-	enter, a deprecated state. The date MAY be in the past or future.
-	If present, this MUST be an [RFC3339][rfc3339] timestamp.
+    An OPTIONAL property indicating the time when the Endpoint entered, or will
+    enter, a deprecated state. The date MAY be in the past or future.
+    If present, this MUST be an [RFC3339][rfc3339] timestamp.
 
   - removaltime<br>
```
n3wscott, line 204: I would rename to `removal`

discovery/spec.md
```diff
@@ -55,305 +45,194 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
 - Examples:
-  - `http://cloud.example.com/docs/blobstorage`
+  - `https://example.com/docs/myQueue`
 
-##### deprecated
+#### deprecated
 
-- Type: Object containing 2 properties:
+- Type: Object container the following properties:
   - effectivetime<br>
```
n3wscott, line 199: I would rename to `effective`

discovery/spec.md
```diff
@@ -365,775 +244,693 @@ The following sections define the attributes that appear in a Service entity.
...
     ````
 
-##### specversions
+#### channel
 
-- Type: Array of `String` values
-- Description: CloudEvents
-  [`specversions`](../cloudevents/spec.md#specversion)
-  that can be used for events published for this service.
+- Type: String
```
n3wscott, line 249: Feels like `channel` should be an array.

discovery/spec.md
```diff
@@ -365,775 +244,693 @@ The following sections define the attributes that appear in a Service entity.
...
+- Description: A string that can be used to correlate Endpoints. Any Endpoints
+  within an instance of a Discovery Service that share the same `channel` value
+  MUST have some relationship. This specification does not define that
```
n3wscott, line 252: If we do not define what `channel` means, I would lean towards `channel` being either a documented extension that defines one way to interpret `channel`, or drop it entirely and vendors can add private extensions to the api.

discovery/spec.md
```diff
@@ -365,775 +244,693 @@ The following sections define the attributes that appear in a Service entity.
...
+  relationship or the specific values used in this property. 
+
+  This property MAY be serialized as an empty string or MAY excluded from the
```
n3wscott, line 255:
```suggestion
  This property MUST be serialized as an empty string or MAY excluded from the
```

discovery/spec.md
```diff
@@ -365,775 +244,693 @@ The following sections define the attributes that appear in a Service entity.
...
+  serialization entirely if there is no value.
 - Constraints:
-  - REQUIRED
-  - MUST be a non-empty array of non-empty strings
+  - OPTIONAL
+  - if present, MUST be a string
+- Examples:
+  - `queue1`
+
+#### authscope
```
n3wscott, line 263: I would rename to `auth` and define it as an undefined object.

discovery/spec.md
```diff
@@ -365,775 +244,693 @@ The following sections define the attributes that appear in a Service entity.
...
+
+- Type: TBD
+- Description: TBD
+- Constraints:
+  - OPTIONAL
+- Examples:
+  - TBD
 
-##### subscriptionurl
+#### usage
```
n3wscott, line 272: `usage` seems odd to me. I would call this `role`.

discovery/spec.md
```diff
@@ -365,775 +244,693 @@ The following sections define the attributes that appear in a Service entity.
...
 
-- Type: `URL`
-- Description: An absolute URL indicating where CloudSubscriptions `subscribe`
-  API calls MUST be sent to.
+- Type: String
```
n3wscott, line 274: Do I need to define a new endpoint for the same service if it provides more than one `usage` type? Example: this can be either a `consumer or a producer`.

discovery/spec.md
```diff
@@ -365,775 +244,693 @@ The following sections define the attributes that appear in a Service entity.
...
+- Description: The interaction model supported by this Endpoint. This
+  specification defines a set of possible values (see the
+  [Usage Values](#usage-values) section for more information, however,
+  implementations MAY define new values as extensions. It is RECOMMENDED
```
n3wscott, line 278: This implies there needs to be an endpoint for me to list the `usage`s available to be able to understand the system and the kinds of queries I can make. It also tells me that we should not mandate any specific usage strings

discovery/spec.md
```diff
@@ -365,775 +244,693 @@ The following sections define the attributes that appear in a Service entity.
...
+  - MUST be a non-empty string
+- Examples:
+  - `consumer`
+
+#### config
 
-##### subscriptionconfig
+- Type: Object
+- Description: Metadata the provides additional information concerning the
+  interactions with this Endpoint. Each `usage` value MUST define the set of
```
n3wscott, line 291: Can you link to the mechanism a `usage`'s `config` is defined?

discovery/spec.md
```diff
@@ -365,775 +244,693 @@ The following sections define the attributes that appear in a Service entity.
...
-  to the CloudEvents
-  [type system](../cloudevents/spec.md#type-system).
-  TODO: Needs resolution with CloudSubscriptions API
+  This property MAY be serialized as an empty object or MAY excluded from the
+  serialization entirely if there is no value.
 - Constraints:
   - OPTIONAL
 - Examples:
-  - ??
+  - `{ "protocol": "kafka" , "endpoint": "https://example.com/queue/1234" }`
```
n3wscott, line 299: I wonder if this example is misleading, as it needs the `usage` to be understood. That might tell me that usage and options should be colocated in an object together one level off of the root?

Is this example of a producer?

Yeah it feels like this would be easier to program to if we had more of this kind of structure:

```
{
   ...
   "usage": {
      "role": "producer",
      "kafka": {
         "queue": "https://example.com/queue/1234"
      }
   }
}
```

No, I am not sure about that... I am thinking about two things: 

1. How would I model this as a protobuf?
2. How do we avoid the downsides of something like OSBAPI where there are generic overloaded fields that  based on some config in the response or context.

discovery/spec.md
```diff
@@ -365,775 +244,693 @@ The following sections define the attributes that appear in a Service entity.
...
+
+#### groups
```
n3wscott, line 301: I am still confused on why there are both `groups` and `channel`.

discovery/spec.md
```diff
@@ -365,775 +244,693 @@ The following sections define the attributes that appear in a Service entity.
...
-  Cloud Subscriptions subscribe() API call.
+  This property MAY be serialized as an empty array or MAY excluded from the
+  serialization entirely if there is no value.
 - Constraints:
   - OPTIONAL
+  - if present, the same Group, as identified by its `id` or `uri`, MUST NOT
+    appear more than once within this property.
 - Examples:
-  - `basic`
+  - `[ { ...group properties... }, { "uri": "https://example.com/discovery/group/987", "name", "John's blob store messages" } ]`
```
n3wscott, line 317: In this example, if the name defined for `https://example.com/discovery/group/987` differs from the name provided by the response from fetching `https://example.com/discovery/group/987`, which one wins? Is this a method for allowing me to overload properties defined in the group for a context of the endpoint?

discovery/spec.md
```diff
@@ -365,775 +244,693 @@ The following sections define the attributes that appear in a Service entity.
...
 - Examples:
-  - ??
+  - `{ "protocol": "kafka" , "endpoint": "https://example.com/queue/1234" }`
+
+#### groups
 
-##### subscriptiondialects
+- Type: Array
+- Description: A set of Definition Groups supported by this Endpoint. Each
+  Group MAY be a full representation of the Group or MAY be a
```
n3wscott, line 305: Personally, I worry having the either or here for full definition or ref (via which, `id` or `uri`??) makes the implementation complicated for both the consumer and the discovery server here.

I would prefer a `$ref` style pointer and have it be a URI-Ref

discovery/spec.md
```diff
@@ -365,775 +244,693 @@ The following sections define the attributes that appear in a Service entity.
...
 
-##### authscope
+#### definitions
 
-- Type: `String`
-- Description: Authorization scope needed for creating subscriptions. The actual
-  meaning of this field is determined on a per-service basis.
+- Type: Array
+- Description: A set of Message Definitions supported by this Endpoint. Each
+  Definition MAY be a full representation of the Definition or MAY be a
```
n3wscott, line 323: same comments on not wanting to mix full resources and pointers here, I prefer URI-Ref

discovery/spec.md
```diff
@@ -1,51 +1,41 @@
...
-to subscribe to those events.
+The Discovery Service specification allows for the discovery of messaging
+Endpoints and their related metadata such as Message Definitions and the
+mechanisms by which interactions with those Endpoints can be established.
 
 ## Table of Contents
 
 - [Overview](#overview)
 - [Notations and Terminology](#notations-and-terminology)
 - [Resource Model](#resource-model)
```
n3wscott, line 13: I would list out the resources here in an expanded table to be able to find the specific resource in question.

duglin: done

discovery/spec.md
```diff
@@ -1,51 +1,41 @@
...
-those events - in the example case, the blob storage service itself.
+Message Endpoints provide a location from which Message producers and consumers
+can share Messages. Endpoints could act as either consumers or producers
+of Messages, and the mechanisms by which those Messages are transferred
+might vary.
+
+This specification defines a set of APIs, and related documents, that allow
+for the discovery of these Endpoints and their related metadata. This
+information can then be used to perform actions such as:
+- subscribing for Messages
```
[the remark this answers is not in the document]
duglin: oddly enough that's the only occurrence of the word 'subscribe' or 'subscribing' in the spec. Changed it anyway

discovery/spec.md
```diff
@@ -1,51 +1,41 @@
...
+- pushing or pulling of Messages with the Endpoint
```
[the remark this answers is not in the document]
duglin: ok

discovery/spec.md
```diff
@@ -1,51 +1,41 @@
...
+- generating of Message validation code based on the schema associated with
```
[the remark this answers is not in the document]
duglin: Right, the text says "This info can then be used to perform actions such as..." not "the spec generates...". Have a suggested rewording?

discovery/spec.md
```diff
@@ -55,305 +45,194 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
-Service resource, the following non-normative pseudo json shows its basic
-structure:
+### Endpoint
 
-(`*` means zero or more, `+` means one or more, `?` means optional)
-
-Service:
+An Endpoint is ...
 
+The following pseudo JSON shows the basic defintion of an Endpoint:
```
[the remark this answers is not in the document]
duglin: Let's chat about this on the call today because I'm starting to wonder about how things like epoch and the timestamps will work in environments where the metadata isn't stored in a DB but instead is just a visualization of resources in other systems. In cases like that, it's not clear to me how easy it would be for there to be an "epoch" value without forcing some kind of DB to be created just to persist it or how we'd know when it needs to be updated based on changes in those other systems. Maybe it's ok to force that model on people, but maybe not...

discovery/spec.md
```diff
@@ -55,305 +45,194 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
-          "name": "[CE context attribute name]",
-          "type": "[CE type string]",
-          "specurl": "[URL to specification defining the extension]" ?
-        }
-      ]
-    }
-  ]
+    "id": "URI-reference",
+    "name": "STRING",
+    "self": "URI",
```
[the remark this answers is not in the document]
duglin: Yes I think the intent was to have `id` and `self` have the same endings. Would `self` always look the same for all clients or could the leading part of the self URI change based on how the client is accessing the system (e.g. public URL vs private URL)? I kind of assumed it might change at times, but `id` would always be static. WDYT?

discovery/spec.md
duglin: so glad someone else wants to do this as well :-)
Done

discovery/spec.md
```diff
@@ -55,305 +45,194 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
-  For example, if a Service's implementation is upgraded to a new version then
-  whether this would result in a new Service (and `id`) or is an update to the
-  existing Service's metadata, is an implementation choice. Likewise, this
-  specification makes no statement, or guarantees, as to the forwards or
-  backwards compatibility of a Service as it changes over time.
-
-  See the Primer for more information.
+#### id
 
+- Type: URI Reference
```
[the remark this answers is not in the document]
duglin: Not sure how we landed on this other than perhaps UUIDs are expensive or people wanted the flexibility to use IDs that are a bit more human friendly??? Let's ask on the call today.

discovery/spec.md
```diff
@@ -55,305 +45,194 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
+- Description: A unique identifier for this Endpoint.
 - Constraints:
-  - MUST be immutable
-  - REQUIRED in responses from the Discovery Endpoint.
-  - MUST be a non-empty string
-  - MUST conform with [RFC3986/3.3](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3) `segment-nz-nc` syntax.
+  - REQUIRED
+  - MUST be a non-empty URI reference
+  - MUST be unique across all Endpoints within the authorization scope
```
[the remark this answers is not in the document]
duglin: probably the string, but you raise a good point... maybe I goofed by dropping the segment-nz-nc stuff ?
I'll add that back in for now - we can always open it up more broadly later. Made ID a string too.

discovery/spec.md
```diff
@@ -55,305 +45,194 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
+- Description: The name of the Endpoint.
 - Constraints:
-  - OPTIONAL. If the attribute is absent or empty, its implied default value is the base
-    URI of the API endpoint.
-  - MUST be a valid URI.
-  - For services imported from other discovery endpoints in replication scenarios, the
-    attribute is REQUIRED to be not empty. If the value is empty or absent
-    during import, it MUST be explicitly set to its implied default value.
+  - REQUIRED
+  - MUST be a non-empty string
```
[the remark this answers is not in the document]
duglin: Since this is human readable and just FYI metadata I'm not sure if we should restrict it. As of now we don't even require it to be unique.

discovery/spec.md
```diff
@@ -55,305 +45,194 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
-  the value used. For example, implementations are free to use a value that
-  represents a timestamp or could choose to simply use a monotonically
-  increasing number. The only requirement is that the value MUST always
-  increase each time the Service Entry is updated. This allows for a quick
-  integer comparision to determine which version of this Service Entry is the
-  latest - meaning, the one with the largest integer value.
+#### self
 
+- Type: URI
+- Description: A unique URI for this Endpoint. If the URI is a URL then an
```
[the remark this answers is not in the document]
duglin: I'll drop the GET stuff.

discovery/spec.md
```diff
@@ -55,305 +45,194 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
 
-##### name
+#### epoch
 
-- Type: `String`
-- Description: A unique human readable identifier for this Service. This value
-  MUST be unique (case insensitive) within the scope of this Discovery Endpoint.
+- Type: Unsigned Integer
+- Description: A number representing the version number of this Endpoint. Each
+  time this Endpoint is modified this property MUST be updated with a new value
```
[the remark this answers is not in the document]
duglin: Let's chat about this - I have lots of concerns about some of our metadata, including epoch

discovery/spec.md
```diff
@@ -55,305 +45,194 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
+  - REQUIRED
+  - MUST be a non-empty URI
+
 - Examples:
-  - `42`
-  - `915148800`
+  - `https://example.com/discovery/endpoints/1234`
 
-##### name
+#### epoch
```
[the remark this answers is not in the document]
duglin: Let's chat on the call

discovery/spec.md
```diff
@@ -55,305 +45,194 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
-  entity.
-  Note: this value will most likely be Discovery Endpoint defined and therefore
-  will most likely be ignored when provided by clients.
+  - `0`
+  - `1001`
+
+#### description
+
+- Type: String
+- Description: A brief summary of the purpose of the Endpoint.
```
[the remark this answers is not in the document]
duglin: removed

discovery/spec.md
```diff
@@ -55,305 +45,194 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
-    the Service's `id` attribute.
+  - OPTIONAL
+  - if present, MUST be a non-empty string
 - Examples:
-  - `http://example.com/services/bf5ff5cc-d059-4c79-a89a-2513e45a1340`
-  - `http://discovery.github.com/services/892abefc-0293-9273-bead-92830efaefa0`
+  - "A queue of the sensor generated messages"
 
-##### description (Service)
+#### tags
```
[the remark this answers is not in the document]
duglin: For someone not familiar with K8s, would "tags" or "annotations" more likely imply name/value pairs? I'm wondering if "annotations" would be thought of as just "list of strings". But let's see what others think - I can technically go either way.

discovery/spec.md
```diff
@@ -55,305 +45,194 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
+  This property MAY be serialized as an empty map or MAY excluded from the
+  serialization entirely if there is no value.
 - Constraints:
   - OPTIONAL
-  - If present, MUST be a non-empty string
+  - if present, MUST be a map of name/value string pairs
+  - Each name MUST be a non-empty, unique within the scope of this map, string.
+    Values MAY be empty strings
+- Examples:
+  - `{ "owner": "John", "isVerified: "" }`
```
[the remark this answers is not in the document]
duglin: switched it to "verified".

discovery/spec.md
duglin: done

discovery/spec.md
duglin: done

discovery/spec.md
duglin: done

discovery/spec.md
duglin: Would it help if we made it clear that while we leave it pretty open, we expect it to be used to help identify Endpoints for a shared messaging bus instance?

discovery/spec.md
duglin: I decided to go with:
```
  When this property has no value it MUST either be serialized as an empty
  string or excluded from the serialization entirely.
```

discovery/spec.md
duglin: I'll wait for the original proposers of this one to speak up - I'm ok with the change though

discovery/spec.md
duglin: Let's ask on the call... I think `role` did pop up in our off-line chats but I can't remember why we didn't use it. It may have been because we wanted a word that implied it was the role as viewed by the "user of the endpoint" instead of it's "role as viewed from the endpoint's perspective".

discovery/spec.md
duglin: yes

discovery/spec.md
duglin: Do you mean a doc that shows all possible config fields/values?

discovery/spec.md
duglin: Let's chat on the call. I don't have a strong opinion on this one as it feels like just a syntax diff to me and not one that impacts the semantics.

discovery/spec.md
duglin: Groups are for correlating Message Definitions, while Channels are for grouping Endpoints.

discovery/spec.md
duglin: Do you mean to not allow in-lining at all? We wanted to be able to support the "get it all in one GET" scenario, which is what in-lining does. However, we could mandate that for the "single GET" scenario people need to always query from the top of the resource model since that allows for many different resource types to be returned - then a local ref could always be used. This would mean that a GET on an individual resource group would always have refs. Would this break anyone's model/usecases?

discovery/spec.md
duglin: I think that would be an error but in general I'm not sure how we can mandate that they always be the same since both could be correct at the point in time of each query. Meaning, at 12:01 name was "foo" but at 12:02 when someone did a GET to the group it was "bar" because someone modified it during that minute. This is basically true for just about any query, no?

PR #1045, opened 2022-08-12
fix: use SHOULD instead of needs

### LINKED ISSUE #1044: Ambiguous "needs" wording in spec
Under the [Notational Conventions](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#notational-conventions) section in the CloudEvents spec the wording is ambiguous under
> The producer **needs** to be prepared for the situation where a consumer ignores that feature

It is not clear if "needs" mean `MUST` or mean `SHOULD` under RFC 2119 .

My personal interpretation is SHOULD. 

The reason I is a problem is because while working on the hebrew tralsnation (#1042) the word "needs" is synonymous with "should" and this MAY cause confusion.

sasha-tkachev:
Fixes #1044

## Proposed Changes

- Change the word "needs" to "SHOULD" under the [Notational Conventions](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#notational-conventions) section in the CloudEvents spec to avoid ambiguity 

**Release Note**

```release-note
Change the word "needs" to "SHOULD" under the Notational Conventions section in the CloudEvents spec to avoid ambiguity 
```

### DIFF
--- a/cloudevents/spec.md
+++ b/cloudevents/spec.md
@@ -49,10 +49,16 @@ For clarity, when a feature is marked as "OPTIONAL" this means that it is
 OPTIONAL for both the [Producer](#producer) and [Consumer](#consumer) of a
 message to support that feature. In other words, a producer can choose to
 include that feature in a message if it wants, and a consumer can choose to
-support that feature if it wants. A consumer that does not support that feature
-will then silently ignore that part of the message. The producer SHOULD be
+support that feature if it wants. A consumer that does not support that feature 
+is free to take any action it wishes, including no action or generating an error, as
+long as doing so does not violate other requirements defined by this specification.
+However, the RECOMMENDED action is to ignore it. The producer SHOULD be
 prepared for the situation where a consumer ignores that feature. An
 [Intermediary](#intermediary) SHOULD forward OPTIONAL attributes.
+
+```
+
+```
 
 ### Terminology
 
--- a/cloudevents/spec.md
+++ b/cloudevents/spec.md
@@ -50,7 +50,7 @@ OPTIONAL for both the [Producer](#producer) and [Consumer](#consumer) of a
 message to support that feature. In other words, a producer can choose to
 include that feature in a message if it wants, and a consumer can choose to
 support that feature if it wants. A consumer that does not support that feature
-will then silently ignore that part of the message. The producer needs to be
+will then silently ignore that part of the message. The producer SHOULD be
 prepared for the situation where a consumer ignores that feature. An
 [Intermediary](#intermediary) SHOULD forward OPTIONAL attributes.
 
--- a/cloudevents/spec.md
+++ b/cloudevents/spec.md
@@ -56,10 +56,6 @@ However, the RECOMMENDED action is to ignore it. The producer SHOULD be
 prepared for the situation where a consumer ignores that feature. An
 [Intermediary](#intermediary) SHOULD forward OPTIONAL attributes.
 
-```
-
-```
-
 ### Terminology
 
 This specification defines the following terms:


duglin:
Discussed on the 8/18 call - will continue the discussions on the next call.

cloudevents/spec.md
```diff
@@ -50,7 +50,7 @@ OPTIONAL for both the [Producer](#producer) and [Consumer](#consumer) of a
 message to support that feature. In other words, a producer can choose to
 include that feature in a message if it wants, and a consumer can choose to
 support that feature if it wants. A consumer that does not support that feature
-will then silently ignore that part of the message. The producer needs to be
+will then silently ignore that part of the message. The producer SHOULD be
```
duglin, line 53: Been thinking about our discussions yesterday around this, and I'm now thinking that the proposal to just remove that sentence might be the best action. While I think the original goal of that sentence (trying to make it really really clear that producers can't assume anything about what consumers will do with optional features), it does feel a bit odd to place constraints on the sender of a fire-n-forget message.

The other option that came to mind would be to merge it with the previous sentence and add a bit more to it, such as:
```
A consumer that does not support that feature
will then silently ignore that part of the message and any semantics
associated with that feature would not be applied.
```

jskeet: I'm not even sure we need to say that the corresponding part of the message (event?) will be silently ignored - it might be logged, for example, or even *partially* implemented. This paragraph also seems to treat a "feature" as something that's included in a message; I can imagine some features that might not quite work like that.

I still think it would be useful to reserve the word OPTIONAL (in caps) for optional attributes, and use MAY for other occurrences.

duglin: We could make it something like:
```
A consumer that does not support that feature is free to
take any action it wishes, including no action, as long as
doing so does not violate other requirements defined by
this specification.
```

jskeet: Yup, that feels a bit better... but do we then need to make it clearer that it at least *shouldn't* reject the message due to not understanding things? (I'm not sure whether that would "violate other requirements defined by this specification".) It's possible that we can be somewhat loose here, but perhaps tighten up SDK requirements? (If someone writes a consumption service which rejects any CloudEvent that doesn't have a type beginning "xyz" then they're perfectly at liberty to do so, after all... all of this is effectively advisory, whereas we can be strict with what we allow CNCF-blessed SDKs to do.)

duglin: While writing-up my previous comment at one point I had text that specifically mentioned that it was ok to reject it due to its presence - but then I figured "any action" covered it. But I would be ok with making it clearer:
```
A consumer that does not support that feature is free to
take any action it wishes, including no action or 
generating an error, as long as doing so does not violate
other requirements defined by this specification.
However, the RECOMMENDED action is to ignore it.
```
I'm leaning towards allowing it to reject the message because I'm picturing a very strict consumer that wants to make sure all incoming CEs only have known CE attributes and it doesn't want to lie to the producer by returning a 200 in cases where it's just going to ignore unknown ce-* attributes. We could say that the error generated can't be something like "invalid CE" and would need to be something more like "sorry I don't support ce-xxx", but I'm not sure if that's needed or how verbose we want to be.  It wouldn't be my implementation choice, and I think most producers would be unhappy with the consumer, but I'm not sure we should all-out ban it though if they have a good reason to do so.

jskeet: +1 to that change. I like it.

sasha-tkachev: I like it as-well

cloudevents/spec.md
```diff
@@ -49,11 +49,17 @@ For clarity, when a feature is marked as "OPTIONAL" this means that it is
...
-support that feature if it wants. A consumer that does not support that feature
-will then silently ignore that part of the message. The producer needs to be
+support that feature if it wants. A consumer that does not support that feature 
+is free to take any action it wishes, including no action or generating an error, as
+long as doing so does not violate other requirements defined by this specification.
+However, the RECOMMENDED action is to ignore it. The producer SHOULD be
 prepared for the situation where a consumer ignores that feature. An
 [Intermediary](#intermediary) SHOULD forward OPTIONAL attributes.
 
+```
```
duglin, line 59: qu'est-ce que c'est? :-)

sasha-tkachev: 🤡

jskeet (approved):
Yup, this now looks good to me.

duglin:
approved on the 8/25 call

Issue #1018, opened 2022-05-20
Ambiguous requirements for delivery target executing validation callback


krancour:
[This section](https://github.com/cloudevents/spec/blob/main/cloudevents/http-webhook.md#413-webhook-request-callback) states:

> The delivery target grants permission by issuing an HTTPS GET or POST request against the given URL.

The "or" here throws a real monkey wrench into the works, because it's unclear which party decides which is the appropriate method.

Case in point, cloudevents/sdk-go unconditionally uses a POST for the callback [here](https://github.com/cloudevents/sdk-go/blob/main/v2/protocol/http/abuse_protection.go#L72), while Azure Event Grids expects the callback to be invoked as a GET.

duglin:
I think the intent is that the sender can decide to use either one and the receiver needs to be prepared for either. Can’t explain why we didn’t just pick one though. Might be because folks have seen both being used. But @clemensv will know for sure.

duglin:
Maybe there’s supposed to be a “method” parameter? https://progrium.github.io/blog/2012/11/26/x-callback-header-an-evented-web-building-block/

krancour:
Thanks @duglin. And I will now stand corrected... experimentation revealed that Azure Event Grid _does_ accept a POST, even though the documentations says a GET is expected.

I agree that the receiver should probably be prepared to receive either, but humbly suggest that perhaps that should be spelled out explicitly to avoid further confusion on this.

krancour:
Once again, I may have spoken too soon, because Event Grid provisioning did eventually fail with the POST callback. Webhook validation incomplete. ☹️ 

I guess I'm going back to a custom options handler that does a GET _and_ a POST because it's impossible to predict whether the receiver is prepared for either.

duglin:
We chatted about this on last week's call and @clemensv will be commenting soon....

krancour:
@duglin I should have circled back to provide some more info...

I do still think there's room for at least a mention that the receiver needs to be prepared for either GET or POST, but I discovered what might possibly be another gotcha...

After a lot more experimentation, my _real_ trouble with Event Grid turned out to be timing. If the async ack (whether GET or POST) was sent before the synchronous response to the original request was received, it didn't work...

Obviously, this is a thing that can happen. Should the spec possibly specify either that the sender must not do this or that the receiver should be prepared for that possibility?

clemensv:
[...]

Issue #617, opened 2020-05-05
Unclear how to answer Web Hoo Callback in browser case


n3wscott:
Ref: https://github.com/cloudevents/spec/blob/master/http-webhook.md#413-webhook-request-callback

The spec says this: 

```
If the delivery target chooses to grant permission by callback, it withholds the response headers.
```

And also: 

```
The HTTP GET request can be performed manually using a browser client.
```

I can't see how the manual case sends the correct rate limits and other option headers. Can we get a bit more detail in how this handshake is suppose to work? It feels like `it withholds the response headers.` is wrong and should be removed. The OPTIONS handler should answer the request, and the producer should block until the callback has been validated to start sending events. But it blocks with the config retrieved in the first OPTIONS call that passed the callback to begin with.

duglin:
ping @clemensv

n3wscott:
Ping @clemensv @duglin still not sure on this one.

duglin:
I wonder if the assumption here was that the header info would have been encoded into the URL, which is why a browser could do it. Although, note that it does say this:
```
The URL is not formally constrained, but it SHOULD contain an identifier for the delivery target along with a secret key that makes the URL difficult to guess so that 3rd parties cannot spoof the delivery target.
```
which should cover the `WebHook-Request-Origin` header, but for the rate-limit header that gets a bit tricky since it doesn't explicitly say how that should be encoded into the URL, though it should be obvious. I do wonder why it's a SHOULD and not a MUST though.

ping @clemensv

clemensv:
"If the delivery target chooses to grant permission by callback, it withholds the response headers." 

This means that it the target handles OPTIONS for all of its other functions that OPTIONS has, but not for granting this permission.

If you then answer the callback with a GET (you picked the URL from a log), you implicitly allow whatever rates the sender is asking for because you don't send headers.

if you want to limit rates, you respond more deliberately with a POST and include the rate limiting headers.

n3wscott:
> If you then answer the callback with a GET (you picked the URL from a log), you implicitly allow whatever rates the sender is asking for because you don't send headers.

What should reasonable default be in the sdk's?

duglin:
I believe we can close this one now due to #1049 - if I'm mistaken please ping me and I'll reopen it.

PR #1049, opened 2022-08-18
Clarification for the abuse protection callback

clemensv:
Addresses #617 and #1018


duglin:
To address @n3wscott's questions in #617, do we need to say something about how the sender of the callback would need to use headers if they wanted to modify some of the flags like rate-limit ? And while it can be done via a GET or POST, doing it just from a browser isn't quite really that easy.

clemensv:
GET accepts the proposed values by the caller. Makes sense to clarify.

duglin:
@n3wscott do these changes address your concerns?

duglin:
@krancour does this address your concerns from #1018 ?

n3wscott (approved):

duglin:
Approved on the 8/25 call

PR #1054, opened 2022-08-20
fix: world spelling mistakes in subscription spec

### LINKED ISSUE #1055: spelling mistakes world in subscription spec
Under the [Subscription Object config ](https://github.com/cloudevents/spec/blob/main/subscriptions/spec.md#config) section has a world of spelling mistakes
> When there is a Discovery `Enpoint` Service definition defined for the subscription manager, then the key MUST be one of the subscriptionconfig keys specified in the Discovery Endpoint Service definition

In the world of `Enpoint` spelling mistakes which lost a letter `d`, so the correct is `Endpoint`

xdlbdy:
Fixes  #1055


duglin:
executive decision - merging!
thanks @xdlbdy

PR #1067, opened 2022-08-26
chore: update Hebrew spec to reflect changes of #1045

sasha-tkachev:
Signed-off-by: Alexander Tkachev <sasha64sasha@gmail.com>


duglin:
Approved on the 9/1 call

PR #1069, opened 2022-09-01
Note around empty messages for Kafka

jskeet:
(As discussed in previous meetings.)

### DIFF
--- a/cloudevents/bindings/kafka-protocol-binding.md
+++ b/cloudevents/bindings/kafka-protocol-binding.md
@@ -182,6 +182,11 @@ directly to the CloudEvents `datacontenttype` attribute.
 The [`data`](#21-data) byte-sequence MUST be used as the value of the Kafka
 message.
 
+A CloudEvent with no `data` can be represented in a Kafka message in _binary_ mode,
+in which case the message has no value. In a topic with log compaction enabled,
+any such message will represent a _tombstone_ record, as described in the
+[Kafka compaction documentation](https://kafka.apache.org/documentation/#design_compactionbasics).
+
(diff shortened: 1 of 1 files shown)


cloudevents/bindings/kafka-protocol-binding.md
```diff
@@ -182,6 +182,11 @@ directly to the CloudEvents `datacontenttype` attribute.
 The [`data`](#21-data) byte-sequence MUST be used as the value of the Kafka
 message.
 
+A CloudEvent with no `data` can be represented in a Kafka message in _binary_ mode,
```
duglin, line 185: Is this saying that in order to send an event w/o any data it MUST be in binary mode, or are you just saying that it's legal to send a binary CE w/o a `data` value? It could be read both ways. I think you mean the latter, correct?

jskeet: Yes, the latter. Happy to clarify it. An event w/o data in structured mode is fine too, but *doesn't* represent a tombstone.

duglin: yup - that was my assumption - and it might be good to mention that tombstone doesn't apply for structured mode. Just to make it painfully clear

jskeet: Cool. Will amend the PR tomorrow; no need to talk about it today unless we think we'd get other and different feedback.

duglin:
Very clear now - thanks!
LGTM

duglin:
Approved on the 9/8 call

PR #1078, opened 2022-09-07
fix: force shell scripts to be LF

sasha-tkachev:
Force shell scripts to be LF on checkout to support executing on Windows WSL


.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Force shell scripts to be LF on checkout to support executing on Windows WSL.
+*.sh text eol=lf
```
duglin, line 2: minor thing but can you add a \n to the last line to remove the red icon?

duglin:
@sasha-tkachev can you sign your commits?

jskeet (approved):

duglin:
Approved on the 9/8 call

PR #1085, opened 2022-09-14
feat: verify languages

sasha-tkachev:


duglin:
Is this just making sure there's a non-english version of each doc, or will it tell us when a non-english version is out of date relative to the english one?

sasha-tkachev:
@duglin for now only verify that the file exists
I will add an outdating verification later

sasha-tkachev:
Merged with the refactor

PR #1086, opened 2022-09-14
perf: improve verification scripts speed x60

sasha-tkachev:
This PR improves the performance of the CI verification from 20 minutes to 20 seconds (60 times faster!)

In addition to that, it fixes some bugs which existed in the old scripts

### DIFF
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -7,7 +7,7 @@ verify:
 	@python tools/verify.py .
 
 test_tools:
-	@pytest -vvv tools/
+	@pytest tools/
 
 docker:
 	@docker run -ti -v $(PWD):/tmp/spec -w /tmp/spec python:latest \


duglin:
Per a slack chat with @sasha-tkachev - I had a few comments:
1 - let's add another Makefile target so people who don't have python installed (like me) can still test locally:
```
# Assumes we running from the root of the repo
docker:
    docker run -ti -v $(PWD):/tmp/spec python /bin/sh -c \
      "cd /tmp/spec ; \
       python -m pip install -qq -r tools/requirements.txt ; \
       make -f tools/Makefile"
```

2 - I will get this error about 1/2 the time when I run it:
```
            (   
                Path("fake-docs/text-verification.md"),
                "Translation file fake-docs/languages/my-lang/text-verification.md does not exist",
            ),
        }
E       assert {(PosixPath('...egment"), ...} == {(PosixPath('...egment"), ...}
E         Extra items in the right set:
E         (PosixPath('fake-docs/link-verification.md'), "'https://github.com/non-existing-page' was not found")
E         Use -v to get more diff

/tmp/spec/tools/test_verify.py:191: AssertionError
```

Otherwise it looks really good - and it's super fast!

I'd like to see if we can approve this on today's call with the assumption that we can resolve these 2 points I mentioned off-line and then merge. I don't think they're controversial or big changes.

sasha-tkachev:
@duglin I fail to re-create your symptoms
I think I maybe fixed it but not sure. 
Added your docker target as requested

tools/Makefile
```diff
@@ -1,10 +1,13 @@
...
-	@tools/verify-specs.sh -v .
-	@echo Running the doc phrase checker:
-	@tools/verify-docs.sh -v .
+	@python tools/verify.py .
+
+test_tools:
+	@pytest tools/
+
+docker:
+	@docker run -ti -v $(PWD):/tmp/spec -w /tmp/spec python:latest make -f tools/Makefile deps test_tools verify"
```
duglin, line 13: this is missing a quote. And it would be good to wrap it at 80 like my sample showed: https://github.com/cloudevents/spec/pull/1086#issuecomment-1255215354

duglin: Or just remove the trailing quote

duglin:
Have you tried the docker target?
Just tried it and I got:
```
...
Task was destroyed but it is pending!
task: <Task pending name='Task-2573' coro=<TCPConnector._resolve_host() running at /usr/local/lib/python3.10/site-packages/aiohttp/connector.py:874> wait_for=<Future pending cb=[_chain_future.<locals>._call_check_cancel() at /usr/local/lib/python3.10/asyncio/futures.py:385, Task.task_wakeup()]> cb=[TCPConnector._create_direct_connection.<locals>.drop_exception() at /usr/local/lib/python3.10/site-packages/aiohttp/connector.py:1155]>
```
Definitely random because it then ran successfully 3 times before I saw this:
```
...
                PurePosixPath("fake-docs/text-verification.md"),
                "Translation file fake-docs/languages/my-lang/text-verification.md does not"
                " exist",
            ),
        }
E       assert {(PurePosixPa...egment"), ...} == {(PurePosixPa...egment"), ...}
E         Extra items in the right set:
E         (PurePosixPath('fake-docs/link-verification.md'), "'https://github.com/non-existing-page' was not found")
E         Use -v to get more diff

/tmp/spec/tools/test_verify.py:185: AssertionError
```
Try running it in a loop until it fails

sasha-tkachev:
I ran it inside the docker container 10 times in a row and failed to re-create it. 
I don't think it is worth it to continue investigating it because it is only the verification tests and was never re-created in the CI.
What do you think?

duglin:
Approved on the 9/22 call

PR #1087, opened 2022-09-15
perf: refactor verify links

sasha-tkachev:
Improves performance significantly by using async operations (from 90 seconds to 20 seconds)

also fixed some bugs both in the script and in the specs


sasha-tkachev:
Merged with other PR

Issue #1083, opened 2022-09-09
dataschema for signed or encrypted payload


duglin:
Hi,

As mentioned in some issues (#373 and #379) a good option to sign or encrypt data in a CloudEvent is use of one of the JOSE format (JWS or JWE) my question here is, should then the `dataschema` describe the JWS/JWE structure or should focus on the schema of the decrypted information? i mean the payload of the JWT?

Use the schema to describe a JWS/JWE seems to not have any value for the consumer but the specs say 
> Identifies the schema that data adheres to.
so not sure what would be the correct approach here.

Best regards

_Originally posted by @jolivaSan in https://github.com/cloudevents/spec/discussions/1077_

duglin:
I believe it's meant to be the schema of the unencrypted payload.

@JemDay, since you did https://github.com/cloudevents/spec/pull/399/files, do you think we need to make this clear in the spec?

JemDay:
I would suggest that if the the _'data'_ represents an encoded and/or signed payload the the 'dataschema' isn't really appropriate in that situation.

The _datacontenttype_ would be something like _application/jose_ or _application/jwt_.

I believe there's a pattern whereby the schema of the business data can be declared in a JWS header - I'll try and find a reference.

JemDay:
The datacontenttype of the business data should be carried in the _cty_ JWS Header. [Ref](https://www.rfc-editor.org/rfc/rfc7515#section-4.1.10)

duglin:
Think we need an issue to clarify this in the spec?

jolivaSan:
Hi,

Just some opinions about that

> The _datacontenttype_ would be something like _application/jose_ or _application/jwt_.

I'm totally agree with that, the `datacontenttype` will indicate if the content is encrypted/signed or any other format. 

> I would suggest that if the the 'data' represents an encoded and/or signed payload then the 'dataschema' isn't really appropriate in that situation.

If the `dataschema` describe the JWS/JWE structure instead of the schema of the decrypted information then i see two main problems:

- Is not going to be useful for the consumer or any tool as JWS/JWE are standards that are already defined, not too much changes expected here...
- The considerations in the [CloudEvents Premier](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/primer.md#versioning-of-cloudevents) about version and `dataschema` are not going to be applicable in this case, because the model that is going to change/evolve is going to be the one of the decrypted information....

sasha-tkachev:
[...]

PR #1090, opened 2022-09-26
feat: cryptography extension 

sasha-tkachev:


duglin:
Couple of comments/questions help me better understand the goals here:
- How do people deal with this with normal HTTP messages, when CE isn't involved? I'm asking because I'm trying to understand what the ultimate goal here is. Is it to get CE involved in encryption so it becomes a formal part of our spec and CE event processing? Or is it simply provide a common location for data that already exists some place else in the message?
- Is this encrypted data needed to route the message properly or will it really (mainly) be used by the ultimate consumer who needed to examine the `data`?

I probably should have asked in: https://github.com/cloudevents/spec/issues/1083

sasha-tkachev:
@duglin 
* Q: How do people deal with this with normal HTTP messages, when CE isn't involved?
* A: As any other data blob, it is just encrypted. it has nothing to do with http messages. I just sent an encrypted file over HTTP. With or without CE it doesnt really matter. the content type will be probably `application/octet-stream` because it is an encrypted blob, but when decrypted it should be `application/xml` or something. If the encryption algorithm I chose was one time pad xored to the data, I had no place to put the data content type of the encrypted data. 
Also, propably over HTTPS the data is usually encrypted anyway so it does not really have a use case for it.
Encrypting data at-rest is a better use-case.

* Q: Is this encrypted data needed to route the message properly?
* A: yes, if you care about the encrypted data content type for example ( intermediary which filters only xml messages for example)

sasha-tkachev:
This extension tries to allow the explotation of the `datacontenttype` and `dataschema` for encrypted messages
Because otherwise the actual `datacontenttype` and `dataschema` are meaningless

duglin:
I probably wasn't clear...

> With or without CE it doesnt really matter. the content type will be probably application/octet-stream because it is an encrypted blob, but when decrypted it should be application/xml or something

> This extension tries to allow the explotation of the datacontenttype and dataschema for encrypted messages
Because otherwise the actual datacontenttype and dataschema are meaningless

This is why I'm asking about how things work w/o CE. It seems to me that "someone" should define a mechanism by which things can be encrypted and then decrypted and define the mechanism by which contentType will then be meaningful even when CE isn't involved since this isn't technically a CE issue. Then once that mechanism is defined, we can decide if CE should take it into account in some way with our context attributes (e.g. copy some of those "other spec defined fields" into CE attributes).

The alternative is that CE gets into the business of "manipulating the event", rather than just "defining where common metadata goes".

Or put another way... why does a CE enable event need these fields when a non-CE enabled event doesn't in order to be transferred and processed correctly?

sasha-tkachev:
@duglin JWE has a field for the encrypted content type for example. 
I dont know other formats which have this kind of metadata. 
This extension is a pragmatic approach on solving #1083

cloudevents/extensions/cryptography.md
```diff
@@ -0,0 +1,59 @@
...
+  changes to the schema SHOULD be reflected by a different URI. See
+  [Versioning of CloudEvents in the Primer](../primer.md#versioning-of-cloudevents)
+  for more information.
+- Constraints:
+  - OPTIONAL
+  - If present, MUST be a non-empty URI
+  
+  
+# Decryption and Signature Removal
+When the event `data` is decrypted or has it's signature removed.
```
jskeet, line 54: Nit: it's => its

cloudevents/extensions/cryptography.md
```diff
@@ -0,0 +1,59 @@
...
+Events with encrypted or signed `data` values create a few problems:
+ - [`datacontenttype`](../spec.md#datacontenttype) MUST hold the content type of 
+ the encrypted buffer, and not the plain data. 
+- [`dataschema`](../spec.md#dataschema) MUST hold the schema of the encrypted buffer, 
+ and not the the schema of the plain data
+
+This means consumers and intermediaries cannot route and validate events based on 
+the `datacontenttype` and `dataschema` attributes as they do with "normal" events.
+
+To solve this problem this extension proposes the attributes
```
jskeet, line 13: I think this would be clearer with a complete example.

jskeet (commented):
Initial very quick feedback

duglin:
Went back and looked at Jem's reference to https://www.rfc-editor.org/rfc/rfc7515#section-4.1.10 and I think an example that shows how the "cty" header is copied into this extensions ce-attribute might help.

For cryptdataschema, what value does it bring to have dataschema point to the encrypted schema? Wouldn't that be known based on the type of encryption? I'm guessing we're introducing "cryptdataschema" to make sure that ce-dataschema and ce-contenttype don't lie about 'data', is that correct?

sasha-tkachev:
@duglin yes, it is to remove ambiguity

duglin:
@sasha-tkachev were we going to close this one?

duglin:
Agreed to close this on the 9/29 call. @clemensv will write-up something (probably for the Primer) about how people should leverage existing security mechanisms and we might consider creating some "bindings" that show how to use those mechanisms with CE.

sasha-tkachev:
#1102 was proposed after work on this extension

PR #1097, opened 2022-10-07
refactor(discovery): change metadata to attrschema

sasha-tkachev:
In preparation for the `attrschema` extension I think these names are better for the field names in the discovery spec

### DIFF
--- a/discovery/spec.md
+++ b/discovery/spec.md
@@ -446,7 +446,7 @@ Talk about how each has metadata and data
     "docs": "URL", ?
 
     "format": "STRING",
-    "metadata": {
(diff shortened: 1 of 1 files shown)


duglin:
We chatted about this on the discovery sub-group call today and @clemensv reminded us that we called it "metadata" because the exact description of the metadata would vary based on the `format` property. While our examples focus on a format of CE, there could be others. And for those other formats it's possible the metadata doesn't describe "attributes" but something else. So, eventually we'll need to produce a set of "metadata binding" types of specs that describe what "metadata" looks like based on the "format" specified.

Make sense?

PR #1100, opened 2022-10-10
fix: rename `loglevel` to `severity`

sasha-tkachev:
this fixes this https://xkcd.com/927/


cloudevents/extensions/severity.md
```diff
@@ -0,0 +1,68 @@
+# Log Level Extension
```
duglin, line 1: Should this be "Severity Extension" or something like that?

duglin:
Presented on the 10/13 call. Folks asked for more time, so we'll vote next week

duglin:
All please review in prep for the call.
@jskeet @JemDay

jskeet (approved):
After a bit of reflection, I'm fine with this.

I do wonder whether we *might* want to restore suggested log bindings - just because it's not longer log-specific doesn't mean it won't often be *used* with logs...

JemDay:
I can't remember who proposed the original log-level extension but hopefully they are aware and onboard with this.

I would prefer that we enumerate the values for _severitynumber_ given it's a required attribute, or at least copy over the reference table from OT so readers don't have to jump around to much - it also raises the issue of what happens if the referenced OT mapping table changes.

duglin:
@sasha-tkachev was the original author - so hopefully he's ok with his own idea ;-)

sasha-tkachev:
@jskeet these log bindings already exist inside open telemetry

duglin:
Approved on the 10/20 call - minor edits coming in a follow-on PR.

PR #1101, opened 2022-10-11
next round of discovery edits

duglin:
Basically what we presented during last week's call.

### DIFF
--- a/discovery/spec.md
+++ b/discovery/spec.md
@@ -1,4 +1,4 @@
-# Discovery Service - Version 0.2-wip
+# Discovery Service - Version 0.3-wip
 
 ## Abstract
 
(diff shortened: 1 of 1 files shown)


discovery/spec.md
```diff
@@ -58,46 +57,23 @@ This specification defines the following terms:
...
-{ "uri": "https://example.com/discovery/endpoints/123", "name": "MyService" }
-```
+#### Group
 
-This allows for requestors to have some additional information about the
-target resource without the need to perform any additional queries. This
-is most commonly used for user-interfaces that wish to display some meaningful
-information about the resource instead of just its URI. It is RECOMMENDED that
-references include at least the `name` property.
+A set of Defintions that related in some way. This specification does not
```
deissnerk, line 70: Definitions

discovery/spec.md
```diff
@@ -114,6 +90,7 @@ defined:
 - Description: A unique identifier for the resource.
```
deissnerk, line 90: Unique within what scope?

discovery/spec.md
bigdavethewave: LGTM

duglin: fixed

discovery/spec.md
duglin: Is it ok that it says:
```
MUST be unique across all instances of the this resource type within the current Discovery Service.
```
down below?

sasha-tkachev: Do we have a definition for the identity of the Discovery Service?

duglin: @deissnerk or did you want me to repeat it in the Description line?

duglin: @sasha-tkachev what do you mean by "definition for the identity" ? Below we say it must conform to RFC3986/3.3. Or are you asking for the semantic meaning behind the string? If so, I'm not sure we can get into that.

duglin:
Approved on the 10/13 call - will submit a new PR if @deissnerk and/or @sasha-tkachev's comments need it.

Issue #1093, opened 2022-09-29
Conventional Encryption and Signing Headers


alexec:
To send or received encrypted to signed data, we need to pass two items:

* Encryption control fields, e.g algorithm, key ID.
* Encrypted data: a byte array.

S/MIME binding might be able to solve this.

duglin:
@sasha-tkachev @JemDay for your thoughts

duglin:
Also see: https://github.com/cloudevents/spec/pull/770/files

sasha-tkachev:
I see no reason why not use JWT as the de-facto standard for signing and encrypting CloudEvent `data` values. 
And to solve use case (1) we can use the russian-doll approach in which we sign/encrypt the whole event in any format of our liking and nesting it inside another event which has all of the required attributes duplicated or something along these lines

alexec:
I think this has been covered before, but JWT/E/S cannot support binary encoding without ballooning message  size by 30% due to the need to use base64 encoding.

Encryption needs to avoid copy operations.

sasha-tkachev:
@alexec  In thins binding we support binary encoding :)

We just don't use base64url encoding and we don't use the `.`  separation 

We relay on the `Binary` attribute and data types

alexec:
Also, JWT does not support XML etc.

PR #1102, opened 2022-10-14
feat: jwt extension

sasha-tkachev:
Solves #1093


sasha-tkachev:
@alexec  what do you think?

cloudevents/extensions/jwt.md
```diff
@@ -0,0 +1,147 @@
+# JWT Extension
+
+This extension provides a way to map JWT (JSON Web Token) values in the formats of JWE
```
duglin, line 3: can you turn these 3 into links to the appropriate specs?

cloudevents/extensions/jwt.md
```diff
@@ -0,0 +1,147 @@
...
+ (JSON Web Encryption) or JWS (JSON Web Signature).
+ 
+JWE MAY be used to encrypt a CloudEvent data and JWS MAY be used to sign CloudEvent data
```
duglin, line 6: Missing `.` at the end of most of the following paragraphs (and this one).

cloudevents/extensions/jwt.md
```diff
@@ -0,0 +1,147 @@
...
+    value is the empty octet sequence.
+
+    <!--Q: SHOULD we keep this?-->
+    If this attribute does not exist consumers MAY assume that the value is empty
+    an octet sequence
+- Constraints:
+  - OPTIONAL
+  - MAY be an empty octet sequence
+
+## Examples
```
duglin, line 85: Can you show a complete CE example? For example, show how each of the attributes defined in this spec would appear in a Binary HTTP CE.

cloudevents/extensions/jwt.md
```diff
@@ -0,0 +1,147 @@
...
+JWE MAY be used to encrypt a CloudEvent data and JWS MAY be used to sign CloudEvent data
+
+This extension does not support [JWE Json Serialization][jwe-json-serialization] 
+
+When using [JWE Cipher Text][jwe-ciphertext] MUST be mapped onto `data` and
+ `datacontenttype` MUST be set to `application/octet-stream`
+
+When using JWS the signed data MUST be decoded from base64url format and mapped onto
+`data` in `Binary` format where `datacontenttype` is taken from the `cty` property
+of the JOSE header, if no `cty` property exists `datacontenttype` MUST be set to `application/octet-stream`
```
duglin, line 15: While JOSE is a required property, do we need to say that at least one of the OPTIONAL properties is required? Or is it valid to have a JOSE attribute and nothing else? Or is it "valid" but just pointless/weird? :-)

sasha-tkachev:
1. Contextualize the use case
2. Clearify that the data is signed and not the attributes

alexec:
OK. So have (now) three problems to solve:

* Encryption.
* Signing.
* Compression (per-encryption).

JOSE solves all of these. So instead of signature extension + encryption extension + compression extension, we could have a single JOSE extension that supports all fields.

We still need an `rootcontenttype` to encode the true content type. 

Or? `{"datacontenttype": "application/jose+application/xml"?

alexec (commented):
This is going in the right direction. It is in line with my thinking of a JOSE extension which describes the mapping from JOSE to context attributes, which avoids needing to encoding the JWE or JWS into the body and get that 27% message size increase. My thoughts on this are:

* I want to use JWE and JWS at the same time. They’re structurally quiet different, you don’t (should not?) merge a JWE into a JWS or vice-versa. Instead, the plan_text of the JWE is the JWS. I’m not fan of that, and you are not either.
* I wonder if there is a more generic way to do this. 

I’ve created an issue #1114

cloudevents/extensions/jwt.md
```diff
@@ -0,0 +1,147 @@
+# JWT Extension
+
+This extension provides a way to map JWT (JSON Web Token) values in the formats of JWE
+ (JSON Web Encryption) or JWS (JSON Web Signature).
```
duglin, line 4: might want to add " to CloudEvents context attributes" to the end of the sentence. Otherwise we're missing what we're mapping the JWT values into.

cloudevents/extensions/jwt.md
duglin: I assume "data" here means the `data` property and not "CE metadata", right? If so, let's put back-ticks around `data` to make that clearer.

cloudevents/extensions/jwt.md
```diff
@@ -0,0 +1,147 @@
...
+ 
+JWE MAY be used to encrypt a CloudEvent data and JWS MAY be used to sign CloudEvent data
+
+This extension does not support [JWE Json Serialization][jwe-json-serialization]
```
duglin, line 8: s/Json/JSON/

cloudevents/extensions/jwt.md
```diff
@@ -0,0 +1,147 @@
+# JWT Extension
+
+This extension provides a way to map JWT (JSON Web Token) values in the formats of JWE
+ (JSON Web Encryption) or JWS (JSON Web Signature).
+ 
+JWE MAY be used to encrypt a CloudEvent data and JWS MAY be used to sign CloudEvent data
+
+This extension does not support [JWE Json Serialization][jwe-json-serialization] 
+
+When using [JWE Cipher Text][jwe-ciphertext] MUST be mapped onto `data` and
```
duglin, line 10: This sentence is a bit hard to parse, especially for a newbie, perhaps:
- - -
When using JWE, [Ciphertext][jew-ciphertext] MUST be mapped onto `data` and ...
- - -

cloudevents/extensions/jwt.md
```diff
@@ -0,0 +1,147 @@
...
+
+When using JWS the signed data MUST be decoded from base64url format and mapped onto
+`data` in `Binary` format where `datacontenttype` is taken from the `cty` property
+of the JOSE header, if no `cty` property exists `datacontenttype` MUST be set to `application/octet-stream`
+  
+## Attributes
+
+### jose
+- Type: `Binary`
+- Description: utf-8 encoded JOSE Header value.
```
duglin, line 21: utf-8 or base64-encoded?

cloudevents/extensions/jwt.md
```diff
@@ -0,0 +1,147 @@
...
+    Both JWS and JWE have JOSE Header definitions.
+- Constraints:
+  - REQUIRED
+  - MUST be a non-empty octet sequence
```
duglin, line 25: It may feel redundant since we say this is Binary, it might be more clear if we say:
```
- MUST be a base64-encoded non-empty octet sequence
```
We do this for `time` in the main spec too just to be extra clear that it must adhere to the time RFC

duglin: Same for the other Binary properties below

cloudevents/extensions/jwt.md
```diff
@@ -0,0 +1,147 @@
...
+- Constraints:
+  - OPTIONAL
+  - MUST be a non-empty octet sequence
+- Examples:
+  - `SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c` 
+
+### jweencryptedkey
+- Type: `Binary`
+- Description : [JWE Encrypted Key][jwe-encrypted-key].
+    Encrypted Content Encryption Key value. Note that for some algorithms,
```
duglin, line 42: Can we make the first two "sentences" into readable sentences? :-) Same for previous property. Even if we just add "The" to the start of it, at least then it's a complete sentence.

cloudevents/extensions/jwt.md
```diff
@@ -0,0 +1,147 @@
...
+  - `SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c` 
+
+### jweencryptedkey
+- Type: `Binary`
+- Description : [JWE Encrypted Key][jwe-encrypted-key].
+    Encrypted Content Encryption Key value. Note that for some algorithms, 
+    the JWE Encrypted Key value is specified as being the empty octet sequence
+    
+    <!--Q: SHOULD we keep this?-->
+    If this attribute does not exist consumers MAY assume that the value is empty
```
duglin, line 46: If they choose to NOT assume it's empty, what does that mean? I'm wondering if this should be a MUST instead.

duglin: This sentence isn't in the next attribute, should it be?

cloudevents/extensions/jwt.md
```diff
@@ -0,0 +1,147 @@
...
+    an octet sequence
+- Constraints:
+  - OPTIONAL
+  - MAY be an empty octet sequence
+
+### jweinitvector
+- Type: `Binary`
+- Description: [JWE Initialization Vector][jwe-initialization-vector].
+    Additional value to be integrity protected by the authenticated
```
duglin, line 55: s/value/values/

sasha-tkachev: I copy pasted the definition from the JWT spec. 

But yes, this vector is used to init the encryption operation.  Also known as IV

cloudevents/extensions/jwt.md
sasha-tkachev: No, it is confusing, because you will think you will need to encode it to base64 and then put the result as the binary content

sasha-tkachev:
@clemensv I don't understand your point at all. 
The only thing I'm proposing is how to map encrypted and signed `data` onto a cloudevent. the cloudevent itself is not encrypted or signed.

All I'm doing is taking parts of the JWT token  (without even parsing the JSON inside of it, because it is none of by concern) and defning how these token parts are mapped onto attributes.
Incidentally only the JOSE part of the token is encoded JSON object, but we are not parsing it.  We are just putting it as-is in the attribute `jose`

selfissued (changes requested):
When using the JOSE signing and encryption methods, JWS and JWE, the signed or encrypted payload can be any octet sequence; it can be JSON or anything else.  You can use JWS and/or JWE for securing anything.

Whereas JWT, which is built with them, requires that the payload be a set of claims represented as JSON.  So I think what you want for this use case is JOSE - not JWT.

jskeet:
Talking to folks internally at Google, it's unclear exactly what the purpose of this extension would be. The PR already gives examples of individual bits, but a motivating end-to-end example (with explanation as to the benefit of having this available vs not) would be really useful.

sasha-tkachev:
This extension will provide a standard way for encrypting cloudevent payloads at rest (for compliance reasons for example). Not only that, it allows also signing the payload for authentication and authorization purposes

jskeet:
> This extension will provide a standard way for encrypting cloudevent payloads at rest (for compliance reasons for example). Not only that, it allows also signing the payload for authentication and authorization purposes

IMO, that information, including complete examples of both use cases, should be clear in the PR itself.

sasha-tkachev:
I agree, I will complete this after we agree on the extension in todays call. 
It cannot be merged today anyway

duglin:
Just FYI - no call today. Next call is Jan 12

sasha-tkachev:
Waiting for security specialist opinion

duglin:
@sasha-tkachev is there a particular person you had in mind? There have been a couple of comments above from folks

duglin:
> > @sasha-tkachev is there a particular person you had in mind? There have been a couple of comments above from folks
> 
> No, I'm ready to merge it as-is.

Ok - can you address the open questions/comments and then we'll ask people to review?

sasha-tkachev:
After considiration I came to the conclusion this is not a good idea.

The reason for it is because the headers are not signed which is easy to miss, and can cause un-carefull implementors to introduce vulnerabilities

PR #1106, opened 2022-10-16
feat: attrschema extension

sasha-tkachev:
Technical definition is complete, only left to add an abstract


devinbost:
Looks good to me. Thanks for all the work on this @sasha-tkachev !
I'd like to hear from others like @clemensv and @deissnerk  since there may be edge cases I'm not considering.

cloudevents/extensions/attrschema-schema.yaml
```diff
@@ -0,0 +1,97 @@
+title: "Attrschema meta-schema"
+$schema: "http://json-schema.org/draft-07/schema#"
+$id: "https://cloudevents.io/attrschema/2022-10/schema#"
+allOf:
+  - { "$ref": "http://json-schema.org/draft-07/schema#" }
```
devinbost, line 5: Looks like this is using draft-07 of JSON Schema. The current stable version is 2020-12. Is there a particular reason for not updating to 2020-12?

sasha-tkachev: no, fixing now

cloudevents/extensions/attrschema-schema.yaml
```diff
@@ -0,0 +1,97 @@
...
+          - type: object
+            oneOf:
+              - properties:
+                  cetype:
+                    type: string
+                    enum:
+                      - boolean
+                      - integer
+                      - string
+                      - binary
```
devinbost, line 70: Is this saying that additional properties (are these for extensions?) can be either boolean, integer, string, or binary types?

sasha-tkachev: Yes,
Or URIs, references and timestamps

cloudevents/extensions/attrschema.md
```diff
@@ -0,0 +1,156 @@
...
+
+- Type: `URI`
+- Description: Identifies the [Attrschema](#attrschema-document) that attributes of
+  the current CloudEvent adhere to.
+ 
+  Incompatible changes to the schema SHOULD be reflected by a different URI. See
+  [Versioning of CloudEvents in the Primer](../primer.md#versioning-of-cloudevents)
+  for more information.
+- Constraints:
+  - OPTIONAL
```
duglin, line 15: REQUIRED

cloudevents/extensions/attrschema.md
```diff
@@ -0,0 +1,156 @@
...
+
+The Attrschema meta-schema is defined under [attrschema-schema.yaml](attrschema-schema.yaml)
+
+### Validation Algorithm
+To validate a given CloudEvent the attrschema implementation MUST:
+
+1. Create a map of all attributes of a given event, where the keys are the attribute
+    names and the values are the attribute values
+2. Strip all `null` OPTIONAL attributes from the event if present
+3. All assumed values MUST be evaluated as-if the event was translated to another
```
duglin, line 42: I'm not sure what "translated to another format" means. Can you elaborate?

cloudevents/extensions/attrschema.md
```diff
@@ -0,0 +1,156 @@
...
+    "myattr" : 5,
+    "yourattr" : true ,
+    "datacontenttype" : "text/xml",
+    "data" : "<much wow=\"xml\"/>"
+}
+```
+Notice that this event is adhering to the example attrschema even though the
+`myattr` is of type `integer` and not `string`. The reason for it is
+that the `myattr` value MUST be converted to the string representation
+before validation.
```
duglin, line 108: While I understand the idea of converting everything to a string for the purposes of the check, it feels odd that "myattr" isn't defined as an int. It seems like we should check its "type" as part of the original format (if the format support types), and then check to see if the value is syntactically correct per the pattern/enum/...

cloudevents/extensions/attrschema.md
```diff
@@ -0,0 +1,156 @@
...
+      "enum": [
+         "text/xml",
+         "application/json"
+      ],
+    }
+  },
+  "required": ["myattr"]
+}
+```
+##### Json Event
```
duglin, line 91: JSON

cloudevents/extensions/attrschema.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Attributes
+
+#### attrschema
+
+- Type: `URI`
+- Description: Identifies the [Attrschema](#attrschema-document) that attributes of
+  the current CloudEvent adhere to.
+ 
+  Incompatible changes to the schema SHOULD be reflected by a different URI. See
```
duglin, line 11: Not sure this first sentence is needed since it's not saying what it's incompatible with. Either this URI describes the current attributes or not.

cloudevents/extensions/attrschema.md
```diff
@@ -0,0 +1,156 @@
...
+Although JSON Schema is used to validate JSON Documents, Attrschema provides an
+algorithm to validate non-JSON formatted CloudEvents (XML, Protobuf, etc.) as well.
+
+The Attrschema meta-schema is defined under [attrschema-schema.yaml](attrschema-schema.yaml)
+
+### Validation Algorithm
+To validate a given CloudEvent the attrschema implementation MUST:
+
+1. Create a map of all attributes of a given event, where the keys are the attribute
+    names and the values are the attribute values
```
duglin, line 40: should "attribute values" be something more like "... are JSON objects that describe the value/schema of the attribute" ?

cloudevents/extensions/attrschema.md
```diff
@@ -0,0 +1,156 @@
...
+This keyword MUST be one of `boolean`, `integer`, `string`, `binary`, `uri`,
+`uri-reference`, or `timestamp`.
+
+When `uri` is used the `format` keyword MUST be set to `uri`
+When `uri-reference` is used the `format` keyword MUST be set to `uri-reference`
+When `timestamp` is used the `format` keyword MUST be set to `date-time`
+
+#### Examples
+Here is an example attrschema document
+```json
```
duglin, line 70: Can we add `cetype` and `format` to the example below?

clemensv:
There's a lot of work here, but this unfortunately overlaps quite a bit with the work we've done on Discovery, especially in the area of message definitions. 

The snippet below is code generated based out of Microsoft Azure Event Grid metadata for the Azure Storage service on the latest status of the discussion in the discovery group and uses the CloudEvents/1.0 format for a message description.  We obviously should avoid having competing models for this.

The formats have a slot in the discovery spec, but the prose hasn't been written, yet. I have formal schema definitions for the CloudEvents format, though: https://github.com/clemensv/cloudevents-discovery-interop/blob/master/src/Azure.CloudEvents.Discovery/specs/ce_discovery.yaml#L618

``` JSON
        "Microsoft.Storage.BlobCreated": {
            "schemaurl": "https://raw.githubusercontent.com/Azure/azure-rest-api-specs/master/specification/eventgrid/data-plane/Microsoft.Storage/stable/2018-01-01/Storage.json",
            "id": "Microsoft.Storage.BlobCreated",
            "epoch": 0,
            "self": "https://discovery.example.com/groups/Microsoft.storage/definitions/Microsoft.Storage.BlobCreated",
            "description": "Raised when a blob is created.",
            "origin": "https://azure.com",
            "format": "CloudEvents/1.0",
            "metadata": {
                "id": {
                    "required": true
                },
                "type": {
                    "value": "Microsoft.Storage.BlobCreated",
                    "required": true
                },
                "time": {
                    "required": true
                },
                "source": {
                    "value": "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Storage/storageAccounts/{storageAccountName}",
                    "required": false
                },
                "dataschema": {
                    "value": "https://raw.githubusercontent.com/Azure/azure-rest-api-specs/master/specification/eventgrid/data-plane/Microsoft.Storage/stable/2018-01-01/Storage.json",
                    "required": true
                },
                "datacontenttype": {
                    "value": "application/json",
                    "required": true
                }
            }
        },
```

devinbost:
@clemensv Thanks for sharing your thoughts on this! 
Does the discovery spec enable validation of the headers? 
I agree that we don't want competing work. 
I'm wondering where the line should be.

duglin:
Closing per @sasha-tkachev's comment. Please let me know if we need to reopen it

PR #1107, opened 2022-10-19
Discovery: More edits based on our 10/18 call

duglin:

### DIFF
--- a/discovery/spec.md
+++ b/discovery/spec.md
@@ -11,9 +11,9 @@ mechanisms by which interactions with those Endpoints can be established.
 - [Overview](#overview)
 - [Notations and Terminology](#notations-and-terminology)
 - [Resource Model](#resource-model)
-  - [Endpoint](#endpoint)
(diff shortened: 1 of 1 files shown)


discovery/spec.md
```diff
@@ -225,6 +225,7 @@ The following pseudo JSON shows the defintion of an Endpoint:
         "strict": true|false ?
     }, ?
 
+    "format": "STRING",
```
duglin, line 228: Since Endpoints are also Groups, we needed to add "format" since it's a field in Groups

discovery/spec.md
```diff
@@ -197,7 +197,7 @@ defined:
 - Examples:
   - `https://example.com/docs/myQueue`
 
-### Endpoint
+### Endpoint Resource
```
duglin, line 200: w/o "Resource" the TOC was messing up since we had 2 "Endpoint" titles.

PR #1108, opened 2022-10-19
More edits based on our 10/18 call

duglin:


discovery/spec.md
```diff
@@ -197,7 +197,7 @@ defined:
 - Examples:
   - `https://example.com/docs/myQueue`
 
-### Endpoint
+### Endpoint Resource
```
duglin, line 200: w/o "Resource" the TOC was messing up since we had 2 "Endpoint" titles.

duglin:
Approved on the 10/20 call

PR #1110, opened 2022-10-25
[Discovery] Fill in more info

duglin:


duglin:
Approved on the 11/3 call

PR #1120, opened 2022-11-08
Add security audit doc

duglin:


README.md
```diff
@@ -109,6 +109,10 @@ native ecosystem by making our systems interoperable with CloudEvents.
 If there is a security concern with one of the specifications in this
 repository please [open an issue](https://github.com/cloudevents/spec/issues).
 
+A security assessment was performed by
```
duglin, line 112: @JieDing any chance you could provide me the Chinese version of this text so I can update https://github.com/cloudevents/spec/blob/main/languages/zh-CN/README.md

JieDing: [Trail of Bits](https://www.trailofbits.com/) 于 2022 年 10 月对本项目进行了安全评估。详情参见[完整报告](docs/CE-SecurityAudit-2022-10.pdf)。
@duglin

JieDing: The link for the repot is a relative path, that might be a problem when you directly copy Chinese text into https://github.com/cloudevents/spec/blob/main/languages/zh-CN/README.md

duglin: would it be bad to make it `../../docs/CE-SecurityAudit-2022-10.pdf` for the Chinese version?

duglin: Thanks for the translation

JieDing: > would it be bad to make it `../../docs/CE-SecurityAudit-2022-10.pdf` for the Chinese version?

Or we can directly use `https://github.com/cloudevents/spec/tree/main/docs/CE-SecurityAudit-2022-10.pdf`. Both looks fine to me.

jskeet (approved):

duglin:
Approved on the Nov 10th call

PR #1123, opened 2022-11-17
Proposed "Registry" spec

duglin:
This PR shows an idea we've been thinking of where we abstract both the Discovery Service and the Schema Registry into one high-level Registry type of spec. We think this will work because both specs have similar concepts of "groups" and "resources" in those groups. This would then allow the generic spec to be customized for particular domains... endpoint discovery and schema registry in our case. This would also then potentially allow us to work with other projects like [Apicuri](https://www.apicur.io) to enhance community alignment.

This is a very rough draft but should show the direction so people can let us know if this seems like an ok way to go.

### DIFF
--- a/registry/README.md
+++ b/registry/README.md
@@ -1,0 +1,3 @@
+# Registry Service - Version 0.3-wip
+
+See the [Registry Service specfication](spec.md).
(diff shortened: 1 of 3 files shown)


duglin:
Approved on the 12/1 call

PR #1131, opened 2022-12-08
Clarity on who does event sampling (we don't mandate it)

### LINKED ISSUE #1091: Who does the sampling?
During our security audit they had a question about https://github.com/cloudevents/spec/blob/main/cloudevents/extensions/sampledrate.md, it's not clear which component should do the sampling... the source, producer, CE SDK, ....    Should the spec even say?

My recommendation is:
- add text to say that it is out of scope for the spec to mandate where the sampling is done, rather just that IF it is done then it will use the define CE attribute to convey the info.

Thoughts?

duglin:
Closes #1091


duglin:
Approved on the 1/12/23 call

Issue #1013, opened 2022-05-09
Support CDATA as body type for 'data' element


duglin:
### Discussed in https://github.com/cloudevents/spec/discussions/1011

<div type='discussions-op-text'>

<sup>Originally posted by **JimWeier-VeteransUnited** May  9, 2022</sup>
Data should be able to contain a CDATA string in the element named data </div>

duglin:
@JemDay

JemDay:
I think this should be doable ... this might mean we change such that "text" is always transferred as CDATA as opposed to an s:string, or we support both and leave it up to the SDKs to switch between the two styles - that has the potential for adding complexity into all of the SDKs though - @jskeet ??

jskeet:
Hmm.... will have to look into it in more detail. I'm not even sure whether we need to explicitly say anything here, or whether it just drops out. Fundamentally I need to get my head back into the XML CloudEvents space - including updating my C# SDK prototype...

duglin:
@JemDay @jskeet any update on this one?

duglin:
per 1/19 call - @JemDay to PR it

jskeet:
//remove-lifecycle stale

PR #1143, opened 2023-01-19
Support CDATA for textual data

JemDay:
Signed-off-by: Jem Day <Jem.Day@cliffhanger.com>

Fixes: #1013 

## Proposed Changes

- Update XML Format specification to include support for CDATA textual content.

**Release Note**

```release-note
Allows for textual content in an XML Format to be carried in a CDATA section if preferred.

```

### DIFF
--- a/cloudevents/working-drafts/xml-format.md
+++ b/cloudevents/working-drafts/xml-format.md
@@ -135,19 +135,35 @@ Example:
 
 ### 3.2 Text Data
 
-Text MUST be carried in an element with a defined type of `xs:string`.
-The element MUST NOT contain any child elements.
+Text MUST be carried in an element with a defined type of `xs:string`. The content
+of this element is either:
 
-Example:
+* Regular XML textual content, or
+* A CDATA section.
+
+Implementations MUST accept both textual representations.
+
+
+Examples:
 
 ``` xml
 <data xsi:type="xs:string">This is text</data>
 ```
 
+The following are equivalent representions of the string `if (X < Y) { .. }` :
+
+```xml
+<data xsi:type="xs:string">if X &lt; Y { .. }</data>
+```
+
+```xml
+<data xsi:type="xs:string"><![CDATA[if X < Y { .. }]]></data>
+```
+
 ### 3.3 XML Data
 
 XML data MUST be carried in an element with a defined type of `xs:any` with
-exactly one child XML element (with any mandatory namespace definitions).
+exactly one child XML element (including any mandatory namespace definitions).
 
 All XML nodes (including comments) within the child XML element MUST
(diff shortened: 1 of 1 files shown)


cloudevents/working-drafts/xml-format.md
```diff
@@ -135,19 +135,35 @@ Example:
 
 ### 3.2 Text Data
 
-Text MUST be carried in an element with a defined type of `xs:string`.
-The element MUST NOT contain any child elements.
+Text MUST be carried in an element with a defined type of `xs:string`. The content
```
duglin, line 138: Minor thing, the doc seems to switch between "xs" and "xsi" - we mean the same thing in both case right? Should we pick one for consistency?

duglin:
LGTM

cloudevents/working-drafts/xml-format.md
```diff
@@ -135,19 +135,35 @@ Example:
...
+of this element is either:
+
+* Regular XML textual content, or
+* A CDATA section.
+
+Implementations MUST accept both textual representations.
```
jskeet, line 144: This currently seems to restrict it to one or the other - I *suspect* a mixture makes sense, e.g.

```xml
<data xsi:type="xs:string">Text before <![CDATA[if X < Y { .. }]]> Text after</data>
```

Something like "The content of this element is text, which may be expressed as any combination of text nodes and CDATA nodes. Implementations MUST accept all combinations."

(Not sure about the "all combinations" it, but we can noodle on that further.)

jskeet (commented):
Definitely happy with the goal of this - just a nit on precisely what's allowed. It could be that I'm missing something about CDATA that makes combinations invalid in XML anyway. (It's definitely not an area I know much about.)

duglin:
@JemDay and @jskeet will think about this one off-line - per the 1/26 call

duglin:
@JemDay @jskeet what's the status of this one? Did you guys get a chance to chat?

jskeet:
Still pending, basically. I think the action is probably on me to come up with a proposal, then send it to Jem.

jskeet:
(On that front, if I could be given sufficient permissions to assign PR reviewers etc, that would be handy. Not to worry if you'd prefer not to though.)

duglin:
@jskeet give it a try now

jskeet:
> @jskeet give it a try now

Excellent, thanks :)

duglin:
just a reminder... rebase and action needed

jskeet:
Won't bother rebasing until we've worked out what we're going to do. No progress yet, I'm afraid.

JemDay:
Thanks @duglin - I think we're good.

PR #1145, opened 2023-01-23
Clarify XML format specification

### LINKED ISSUE #998: XML format "looseness" voting issue

### LINKED ISSUE #1005: XML format: is "XML data" a document or an element?
Currently we have three XML types for data:

- Text
- Binary (base64)
- "XML"

The XML form is required to have a single child element and no peer non-whitespace text nodes. I had interpreted that as "the data is an element" whereas Jem had written it intending "the data is a document". The two aren't the same - in particular, there could be other things we want for a document like a DTD. It also impacts how SDKs should deserialize the data - should it be as a Document object, or an Element object?

A few options:

- Have two separate types for this (at which point we *might* want to move away from `xsi:type` as the way of specifying the data type). We could even have a third type which is "list of XML nodes" to truly allow `xs:any` content. (That would open further questions such as whether the `data` element could contain additional attribute nodes...)
- Decide it's a document, and work out how to spec that in terms of other acceptable direct child nodes
- Decide it's an element, in which case the spec just needs to clarify it (there shouldn't be as much other work)

jskeet:
- Unknown attributes and elements in a different namespace should be ignored (other than within a data element)
- Make it clear that "XML data" is really a representation of a single XML element.

Closes #998.
Closes #1005.

### DIFF
--- a/cloudevents/working-drafts/xml-format.md
+++ b/cloudevents/working-drafts/xml-format.md
@@ -70,7 +70,7 @@ where all nodes MUST be preserved.
 XML elements in namespaces other than `http://cloudevents.io/xmlformat/V1`, and
 XML attributes other than those described in this specification MAY appear within an XML
 representation of a CloudEvent or CloudEvent batch. These SHOULD be ignored during
-processing, except for the child element of a `<data>` element containing
+CloudEvent processing, except for the child element of a `<data>` element containing
 [XML element data](#33-xml-element-data), where all nodes MUST be preserved.
 
 ## 2. Attributes
--- a/cloudevents/working-drafts/xml-format.md
+++ b/cloudevents/working-drafts/xml-format.md
@@ -64,8 +64,14 @@ an XML document preamble SHOULD be included to ensure deterministic processing.
 
 XML comments are permitted anywhere within an XML representation of a
 CloudEvent or CloudEvent batch. Comments MUST be ignored during processing, except
-for the child element of a `<data>` element containing [XML data](#33-xml-data),
+for the child element of a `<data>` element containing [XML element data](#33-xml-element-data),
 where all nodes MUST be preserved.
+
+XML elements in namespaces other than `http://cloudevents.io/xmlformat/V1`, and
+XML attributes other than those described in this specification MAY appear within an XML
+representation of a CloudEvent or CloudEvent batch. These SHOULD be ignored during
+processing, except for the child element of a `<data>` element containing
+[XML element data](#33-xml-element-data), where all nodes MUST be preserved.
 
 ## 2. Attributes
 
@@ -144,9 +150,9 @@ Example:
 <data xsi:type="xs:string">This is text</data>
 ```
 
-### 3.3 XML Data
+### 3.3 XML Element Data
 
-XML data MUST be carried in an element with a defined type of `xs:any` with
+XML element data MUST be carried in an element with a defined type of `xs:any` with
 exactly one child XML element (with any mandatory namespace definitions).
 
 All XML nodes (including comments) within the child XML element MUST


JemDay:
LGTM

cloudevents/working-drafts/xml-format.md
```diff
@@ -64,9 +64,15 @@ an XML document preamble SHOULD be included to ensure deterministic processing.
 
 XML comments are permitted anywhere within an XML representation of a
 CloudEvent or CloudEvent batch. Comments MUST be ignored during processing, except
-for the child element of a `<data>` element containing [XML data](#33-xml-data),
+for the child element of a `<data>` element containing [XML element data](#33-xml-element-data),
 where all nodes MUST be preserved.
 
+XML elements in namespaces other than `http://cloudevents.io/xmlformat/V1`, and
+XML attributes other than those described in this specification MAY appear within an XML
+representation of a CloudEvent or CloudEvent batch. These SHOULD be ignored during
```
duglin, line 72: I wonder if "SHOULD be ignored" is too strong since it doesn't limit the processing to just CE processing. Meaning, could it be interpreted as we're telling all aspects of the receiver to ignore them? Perhaps something more like: 
```
These SHOULD be ignored during the processing of CloudEvent attributes, except...
```
The other option is to say nothing at all since it could be considered out of scope for us to say anything at all about them, the same way we don't really say anything about non-CE HTTP headers (or at least I couldn't find anything).

JemDay: If I remember correctly this language was added to signal to people that they could include their own (non-CE) XML elements and attributes but that there shouldn't be any expectation that they would be considered as a part of the CE that is being marshaled.

Without such a statement an SDK writer might consider such additional and unexpected items to be a violation of the spec, and thus reject the whole thing. What we're saying here is "don't worry, just ignore what you're not expecting".

@jskeet Can probably give a more eloquent explanation.

jskeet: What Jem said, basically. I think this is somewhat different to HTTP headers, as they're naturally mixed in with non-CE headers. In this format, we're saying "This is CloudEvent content" - and as Jem says, I would personally *expect* to deem any element containing unexpected data to be invalid... so if we're saying it's valid, we need to be explicit about that.

I'm happy to reword the "should" to something else - probably worth discussion on Thursday. (I now will be there - I withdraw my previous apologies...)

duglin: Perhaps I'm not being clear...or I'm missing something :-)

> there shouldn't be any expectation that they would be considered as a part of the CE that is being marshaled.

Agreed, but that's not what the text says. The text says they "should be ignore during processing". Not "CE processing", but "processing". Implying all processing. W/o qualifying "processing" we're excluding the option of someone including an extension (in their own XML NS) in the XML doc that can be processed even IF the receiver understands it. We're, in essence, removing the "X" from XML.

I agree these non-CE namespaced things aren't part of the CE metadata. What's concerning me is that the text implies that a receiver should not process any extensions not in the CE namespace despite there being an out-of-band agreement between the sender and receiver to do so (the "X" in XML). And by "process", I don't mean "include them as part of the CE metadata", I mean "understand their semantics and do whatever the spec that defines those elements told them to do". Basically, people can't extend CEs (aka XML docs). I think we need to differentiate between "processing of CE metadata" and "processing of the XML in general beyond the CE semantics" - that's all I'm saying.

It seems to me we want to say:
- senders can include non-CE NS items in the XML (outside of "data") but they will not be included in the resulting CE
- senders can't assume they'll be interpreted/understood by the receivers
- receivers who don't understand them MUST NOT generate an error simply due to their presence - ie. they should ignore unknown items. That doesn't preclude them from generating an error due to those items not adhering to the spec defining those extensions though. I guess an open question is whether we want to mandate whether or not these error should be allowed to impact whether the CE itself is generated and passed-on. I'm leaning towards allowing the extension spec to make that choice where silence in that spec implies it doesn't stop the flow.

Or if we really want to lock things down we should say people MUST NOT include anything but CE namespaces items (outside of "data"), then there's only "CE processing" and no "X".

jskeet: I'd be happy with us going into more detail to be explicit about all of those bullet points.

The bottom paragraph was effectively what I was proposing originally - if you want to include additional data alongside your CloudEvent, it should be somewhere else in the XML document. As for whether doing so would remove the "X" - I see the extensibility as "XML itself allows all kinds of documents to be described" rather than saying that every document format must permit unrelated content within it.

duglin: > if you want to include additional data alongside your CloudEvent, it should be somewhere else in the XML document

Where else would that be ? I'm assuming we're talking about a case like this:
```
<event specversion="1.0" xmlns="http://cloudevents.io/xmlformat/V1">
    <foo:bar xmlns:foo="...">cool stuff</foo:bar>
    <time>2020-03-19T12:54:00-07:00</time>
    <datacontenttype>text/plain</datacontenttype>
    <id>000-1111-2222</id>
    <source>urn:uuid:123e4567-e89b-12d3-a456-426614174000</source>
    <type>SOME.EVENT.TYPE</type>
    <myboolean xsi:type="ce:boolean">false</myboolean>
    <data xsi:type="xs:string">Now is the winter of our discount tents...</data>
</event>
```
I'm not sure where else "foo:bar" could be where we can extend the CE w/o declaring new CE attributes. And telling people to ignore that extension even outside the scope of deserializing a CE pretty much bans extensions.

jskeet: An event (or a batch) is just an element. It can be anywhere within an XML document - so it would be reasonable to have something like:

```xml
<mycombinedevent>
  <!-- Actual CloudEvent -->
  <event specversion="1.0" xmlns="http://cloudevents.io/xmlformat/V1">
  ...
  </event>
  <other-data />
</mycombinedevent>
```

I would personally folks to extend CloudEvents using extension attributes, as that's what they're for. If the data is outside the domain of CloudEvents, I would personally prefer it to be outside the `<event/>` element itself. However, it looks like there at least wasn't significant support for my somewhat-conservative approach, hence the PR to explicitly *allow* this mixing.

Wording it appropriately is tricky though - partly because we usually try to avoid making SDK requirements within the format specs. Possibly that's a mistake.

duglin: interesting, I was assuming the `<event>` was the entire xml doc so extension elements have no choice but to be under `<event>`.

JemDay: I'd hope we wouldn't support the example above ... the cloudevent is the root element in the document.

Are we getting a bit carried away here? I only ask because I don't believe any of the other format specs cover this sort of stuff (i may be wrong). I'm not sure what would happen if somebody threw in a property with an object value in a JSON formatted message for example .. would that be silently ignored (as were suggesting for XML), or cause the whole deserialization process to fail.

duglin: > I only ask because I don't believe any of the other format specs cover this sort of stuff (i may be wrong).

When I looked, none of the others have the notion of namespaces in the body of the message that would show this same situation in structured mode. The closest to this _might_ be the binary cases (like http) where the http headers are (sort of) namespaced by the presence of the `ce-` prefix. And in those cases we only care about the `ce-` ones - we don't say anything about the others. I view this case as the structured version of that.

duglin: And to reiterate my proposal... I'm not suggesting anything more than to limit our "SHOULD" to just "ce processing":
```
These SHOULD be ignored during the processing of CloudEvent attributes
```
and we remain silent about any other receiver-side processing logic. I actually don't think this is really that different from what I think the PR is trying to do. It's just a bit too loose in the current wording.

jskeet:
>  I'm not sure what would happen if somebody threw in a property with an object value in a JSON formatted message for example .. would that be silently ignored (as we're suggesting for XML), or cause the whole deserialization process to fail.

In C# at least, it would absolutely fail ([source](https://github.com/cloudevents/sdk-csharp/blob/main/src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs#L257)) - and I'd *hope* it would in other SDKs too.

I do think the example Doug comes back to about HTTP is different, because an HTTP request is *expected* to have aspects unrelated to a CloudEvent - whereas we're saying "this is a CloudEvent element".

That said, if everyone's happy with the minor modification Doug is proposing (or a variant saying "CloudEvent processing", to avoid us having to worry about whether it's processing attributes or elements in a batch) then we don't need to get into any more detail :) I'll add that change in another commit (so it's easy to spot).

cloudevents/working-drafts/xml-format.md
jskeet: New commit added - PTAL.

duglin:
If we reject that sample how can someone do something like add an element representing a signature of "data" w/o also forcing it to be a CE attribute? They might be able to stick that info in other spots (eg http header) but that kind of feels like we're losing one of the values of xml.

jskeet:
> If we reject that sample how can someone do something like add an element representing a signature of "data" w/o also forcing it to be a CE attribute?

How would they do so in the JSON format? Is this a use case that we believe is actually wanted by anyone? Basically I don't see the point of the XML format as being "to allow side-channel information alongside a CloudEvent" so much as "to allow code that is already XML-oriented to represent CloudEvents that way".

(Note that I'm no longer proposing that we *do* reject it, although I still think that would be better, personally. I'm not a fan of Postel's Law.)

duglin:
> How would they do so in the JSON format?

that's JSON's problem 🤣

PR #1152, opened 2023-02-01
Registry Service clean-up

duglin:
Add some missing stuff to make it more real

### DIFF
--- a/registry/spec.md
+++ b/registry/spec.md
@@ -1,20 +1,275 @@
 # Registry Service - Version 0.3-wip
 
-## Registry Service
+## Abstract
 
-This specification defines the following APIs:
+A Registry Service exposes resources, and their metadata, for the purposes
+of enabling discovery of those resources for either end-user consumption or
+automation and tooling.
+
+## Table of Contents
+
+- [Overview](#overview)
+- [Notations and Terminology](#notations-and-terminology)
+  - [Notational Conventions](#notational-conventions)
+  - [Terminology](#terminology)
+- [Attributes and Extensions](#attributes-and-extensions)
+- [Registry APIs](#registry-apis)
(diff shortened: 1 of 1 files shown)


registry/spec.md
```diff
@@ -1,57 +1,289 @@
...
+  - `John Smith`
+  - `john.smith@example.com`
+
+### `modifiedOn`
+- Type: Timestamp
+- Description: The date/time of when the entity was last updated.
+- Constraints:
+  - MUST be an [RFC3339](https://tools.ietf.org/html/rfc3339) timestamp
+- Examples:
+  - `2030-12-19T06:00:00Z"
```
JemDay, line 243: nit: Inconsistent use of ' and "

registry/spec.md
```diff
@@ -1,57 +1,289 @@
...
+  - `John Smith`
+  - `john.smith@example.com`
+
+### `createdOn`
+- Type: Timestamp
+- Description: The date/time of when the entity was created.
+- Constraints:
+  - MUST be an [RFC3339](https://tools.ietf.org/html/rfc3339) timestamp
+- Examples:
+  - `2030-12-19T06:00:00Z"
```
JemDay, line 223: nit: Inconsistent use of ' and "

JemDay (changes requested):

registry/spec.md
duglin: LOL took me a while to realize you wrote "and" not "an". I just couldn't find "and" in this section!
Thanks for noticing my copy-n-paste error.

registry/spec.md
duglin: fixed

duglin:
Approved on the 2/16 call

PR #1155, opened 2023-02-16
Warn people about dataref

### LINKED ISSUE #1136: Event-data spec when data is too large, so truncated/pruned
Hi team,

I saw both #405, #451 , both kind of tackled this problem but in a different dimension:

What is the recommended way to communicate in the Cloud-event data/attributes spec from Producers point of view, when?
* the event data size is more than 64KB [limit](https://github.com/cloudevents/spec/pull/451)
* Subscribed consumers are expecting that `data` will always be available (or) there is no clear rationale established for missing `data`.
* producer wants to convey that it has either truncated / pruned the data to be in spec limits.

Few options:
1. Define `errors` as an additional data format, where either `data` or `errors` or `both` can be present in an event; While defining this `error` can you also define some form of format for it, i.e. a `http_status_code` or similar is present as `code`, an error message is present as `message` etc?
2. Define `error` as a new optional attribute; this has limitation when we want to provide more than one `error` or `error message` is required.
3. Make `data` also compatible with errors information; this also has limitations to provide 2 schemas for same event-types. It does not scale in my opinion.

duglin:
Fixes #1136 

Well, part of 1136's discussion. The originator of #1136 hasn't replied to our queries so until they do we'll assume using dataref + this fix should be good enough.

### DIFF
--- a/cloudevents/extensions/dataref.md
+++ b/cloudevents/extensions/dataref.md
@@ -10,7 +10,13 @@ MUST be identical.
 Both `data` and the `dataref` attribute MAY exist at the same time. A middleware
 MAY drop `data` when the `dataref` attribute exists, it MAY add
 the `dataref` attribute and drop the `data` attribute, or it MAY add the `data`
-attribute by using the `dataref` attribute.
+attribute by using the `dataref` attribute. Note that since the CloudEvents
+specification does not define a mechanism by which a sender can know if the
+receiver supports any particular CloudEvent extension, removing the `data`
+attribute in favor of just having the `dataref` attribute could yield
+unexpected results. As such, removing the `data` attribute SHOULD only be done
+when the sender is confident that all receivers support the `dataref`
+attribute - via some out-of-band agreement.
 
 ## Attributes
 


cloudevents/extensions/dataref.md
```diff
@@ -10,7 +10,13 @@ MUST be identical.
 Both `data` and the `dataref` attribute MAY exist at the same time. A middleware
 MAY drop `data` when the `dataref` attribute exists, it MAY add
 the `dataref` attribute and drop the `data` attribute, or it MAY add the `data`
-attribute by using the `dataref` attribute.
+attribute by using the `dataref` attribute. Note that since CloudEvent does
+not define a mechanism by which a sender can know if the receiver supports
+any particular CloudEvent extension, removing the `data` attribute in favor
+of just having the `dataref` attribute could yield unexpected results. As such,
+removing the `data` attribute SHOULD only be done when the sender is confident
+that all receivers support the `dataref` attribute - via some out-of-bands
```
duglin, line 18: s/bands/band/

cloudevents/extensions/dataref.md
```diff
@@ -10,7 +10,13 @@ MUST be identical.
 Both `data` and the `dataref` attribute MAY exist at the same time. A middleware
 MAY drop `data` when the `dataref` attribute exists, it MAY add
 the `dataref` attribute and drop the `data` attribute, or it MAY add the `data`
-attribute by using the `dataref` attribute.
+attribute by using the `dataref` attribute. Note that since CloudEvent does
```
jskeet, line 13:
```suggestion
attribute by using the `dataref` attribute. Note that since CloudEvents does
```

or

```suggestion
attribute by using the `dataref` attribute. Note that since the CloudEvents specification does
```

jskeet (approved):
LGTM

(I'd be generally nervous of using this extension, to be honest. If any receivers are themselves intermediaries, you'd need to know that all *their* receivers support dataref, etc. I'm not saying we should remove the extension, just that it feels quite limited in terms of usage scenarios.)

cloudevents/extensions/dataref.md
duglin: done!

deissnerk:
LGTM

duglin:
Approved on the 2/23 call

PR #1161, opened 2023-02-22
Add schemaformat

### LINKED ISSUE #1153: Discovery: "schemaformat" next to "schemaurl"
When `schemaurl` is specified any any place we use it, it SHOULD be accompanied by a "schemaformat" attribute that qualifies it. Reason for that is that the URL may be external and a code generator may not have the ability to resolve the URL in the stage where it is generating the event classes, but it may have to handle e.g. preoto differently than JSON. I have that exact problem at the moment. 

If the schema is inlined ("schema") property, that information is available via the "format" property of the embedded schema version object.

Example with `schemaformat` below:

```
                    "Contoso.ERP.Events.ReservationPlaced": {
                    "type": "cloudevent",
                    "id": "Contoso.ERP.Events.ReservationPlaced",
                    "description": "A reservation has been placed",
                    "format": "CloudEvents/1.0",
                    "metadata": {
                        "attributes": {
                            "id": {
                                "type": "string",
                                "required": true
                            },
                            "type": {
                                "type": "string",
                                "value": "Contoso.ERP.Events.ReservationPlaced",
                                "required": true
                            },
                            "time": {
                                "type": "datetime",
                                "required": true
                            },
                            "source": {
                                "type": "uritemplate",
                                "value": "/erp/orders",
                                "required": true
                            }
                        }
                    },
                    "schemaformat": "Protobuf/3.0",
                    "schemaurl": "https://example.com/registry/schemagroups/Contoso.ERP.Events/schemas/orderData:OrderData"
                }
```

duglin:
Closes #1153

/cc @clemensv

### DIFF
--- a/registry/spec.md
+++ b/registry/spec.md
@@ -1596,6 +1596,7 @@ The Resource entity is defined as:
       }
     } *
   }, ?
+  "schemaformat": "STRING", ?
   "schema": { ... }, ?
   "schemaurl": "URL" ?
 }


registry/spec.md
```diff
@@ -1596,6 +1596,7 @@ The Resource entity is defined as:
       }
     } *
   }, ?
+  "schemaformat": "STRING", ?
```
jskeet, line 1599: Is this defined anywhere else with more detail? I don't know what to expect as a value here.

duglin: Not yet, I was going to add some text to explain it better but then I realized I had a HUGE TODO to add descriptions for ALL of the other fields too, so I decided to wait and do them all at the same time... in a soon-to-come PR. In the meantime, see #1153

jskeet: In which case, LHTM.

duglin:
Approved on the 2/23 call

PR #1164, opened 2023-02-24
Group/Resource/Format Details for Registry

clemensv:
- Schema Registry section complete 
- Message Definitions section complete
- Endpoint Registry started

Covers CloudEvents, AMQP, MQTT (3.1.1/5.0) as transports
Covers JSON, XSD, Avro, Proto as schema formats

Plan to extend out to the remaining transports we supprt with CE after completing the initial draft.

### DIFF
--- a/registry/schemas/xregistry_openapi.json
+++ b/registry/schemas/xregistry_openapi.json
@@ -1,0 +1,1004 @@
+{
+  "openapi": "3.0.6",
+  "info": {
+    "title": "xRegistry API",
+    "description": "xRegistry API",
+    "version": "0.5-wip"
+  },
+  "servers": [],
+  "security": [
+    {
+      "api_key": []
+    }
+  ],
+  "paths": {
+    "/": {
+      "get": {
+        "operationId": "getAll",
+        "description": "Gets the root document",
+        "parameters": [
+          {
+            "in": "query",
+            "name": "inline",
+            "description": "Set if references shall be inlined",
+            "required": false,
+            "schema": {
+              "type": "boolean"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "The root document",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "xregistry_resources.json#/definitions/document"
+                }
+              }
+            }
+          }
+        }
+      },
+      "post": {
+        "operationId": "uploadDoc",
+        "description": "Uploads a registry document and upserts its contents into the registry",
+        "requestBody": {
+          "description": "A request to create or update the discovery endpoint's collection of endpoints with the given endpoints",
+          "content": {
+            "application/json": {
+              "schema": {
+                "$ref": "xregistry_resources.json#/definitions/document"
+              }
+            }
+          }
+        },
+        "responses": {
+          "200": {
+            "description": "The resulting document",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "xregistry_resources.json#/definitions/document"
+                }
+              }
+            }
+          }
+        }
+      }
+    },
+    "/{groupType}": {
+      "parameters": [
+        {
+          "$ref": "#/components/parameters/groupType"
+        }
+      ],
+      "get": {
+        "operationId": "getResourceGroupAll",
+        "description": "Gets all entries of the resource group",
+        "parameters": [
+          {
+            "in": "query",
+            "name": "inline",
+            "description": "Set if references shall be inlined",
+            "required": false,
+            "schema": {
+              "type": "boolean"
+            }
+          },
+          {
+            "in": "query",
+            "name": "name",
+            "description": "Query expression",
+            "required": false,
+            "schema": {
+              "type": "string",
+              "description": "Filter expression"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "Resource groups that match the query or all available resource groups if no query was specified",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "object",
+                  "additionalProperties": {
+                    "$ref": "xregistry_resources.json#/definitions/resource"
+                  }
+                }
+              }
+            }
+          },
+          "404": {
+            "description": "No groups found"
+          }
+        }
+      },
+      "post": {
+        "operationId": "postResourceGroup",
+        "requestBody": {
+          "description": "Create a new resource group",
+          "content": {
+            "application/json": {
+              "schema": {
+                "$ref": "xregistry_resources.json#/definitions/resource"
+              }
+            }
+          }
+        },
+        "responses": {
+          "201": {
+            "description": "The resulting resource group",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "xregistry_resources.json#/definitions/resource"
+                }
+              }
+            },
+            "headers": {
+              "Location": {
+                "description": "The location of the created resource",
+                "schema": {
+                  "type": "string"
+                }
+              }
+            }
+          },
+          "400": {
+            "description": "Bad Request - constraint failure"
+          },
+          "409": {
+            "description": "Conflict - epoch not greater"
+          }
+        }
+      }
+    },
+    "/{groupType}/{groupId}": {
+      "parameters": [
+        {
+          "$ref": "#/components/parameters/groupType"
+        },
+        {
+          "$ref": "#/components/parameters/groupId"
+        }
+      ],
+      "get": {
+        "operationId": "getResourceGroup",
+        "responses": {
+          "200": {
+            "description": "The schema group",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "xregistry_resources.json#/definitions/resource"
+                }
+              }
+            }
+          },
+          "400": {
+            "description": "Bad Request - constraint failure"
+          },
+          "404": {
+            "description": "Not Found"
+          }
+        }
+      },
+      "put": {
+        "operationId": "putResourceGroup",
+        "description": "creates or updates the resource group",
+        "requestBody": {
+          "description": "A request to create or update the discovery group's collection of groups with the given group",
+          "content": {
+            "application/json": {
+              "schema": {
+                "$ref": "xregistry_resources.json#/definitions/resource"
+              }
+            }
+          }
+        },
+        "responses": {
+          "200": {
+            "description": "",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "xregistry_resources.json#/definitions/resource"
+                }
+              }
+            }
+          },
+          "400": {
+            "description": "Bad Request - constraint failure"
+          },
+          "404": {
+            "description": "Not Found"
+          },
+          "409": {
+            "description": "Conflict - epoch not greater"
+          }
+        }
+      },
+      "delete": {
+        "operationId": "deleteResourceGroup",
+        "parameters": [
+          {
+            "in": "query",
+            "name": "epoch",
+            "description": "The epoch of the resource group to be deleted",
+            "required": false,
+            "schema": {
+              "type": "integer",
+              "description": "A number representing the version number of the resource.",
+              "format": "int64"
+            }
+          }
+        ],
+        "responses": {
+          "204": {
+            "description": "Ok"            
+          },
+          "400": {
+            "description": "Bad Request - constraint failure"
+          },
+          "409": {
+            "description": "Conflict - epoch not greater"
+          }
+        }
+      }
+    },
+    "/{groupType}/{groupId}/{resourceType}": {
+      "parameters": [
+        {
+          "$ref": "#/components/parameters/groupType"
+        },
+        {
+          "$ref": "#/components/parameters/groupId"
+        },
+        {
+          "$ref": "#/components/parameters/resourceType"
+        }
+      ],
+      "get": {
+        "operationId": "getResourcesAll",
+        "description": "Get an optionally filtered collection of resources",
+        "parameters": [
+          {
+            "in": "query",
+            "name": "query",
+            "description": "Query expression",
+            "required": false,
+            "schema": {
+              "type": "string",
+              "description": "Filter expression"
+            }
+          },
+          {
+            "in": "query",
+            "name": "inline",
+            "description": "Set if references shall be inlined",
+            "required": false,
+            "schema": {
+              "type": "boolean"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "A list of resources (optionally matching the query parameter)",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "object",
+                  "additionalProperties": {
+                    "$ref": "xregistry_resources.json#/definitions/resource"
+                  }
+                }
+              }
+            }
+          }
+        }
+      },
+      "put": {
+        "operationId": "putResources",
+        "requestBody": {
+          "description": "A request to create or update the discovery group's collection of groups with the given group",
+          "content": {
+            "application/json": {
+              "schema": {
+                "type": "object",
+                "additionalProperties": {
+                  "$ref": "xregistry_resources.json#/definitions/resource"
+                }
+              }
+            }
+          }
+        },
+        "responses": {
+          "200": {
+            "description": "Updated list of resources",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "object",
+                  "additionalProperties": {
+                     "$ref": "xregistry_resources.json#/definitions/resource"
+                  }
+                }
+              }
+            }
+          },
+          "400": {
+            "description": "Bad Request - constraint failure"
+          },
+          "409": {
+            "description": "Conflict - epoch not greater"
+          }
+        }
+      },
+      "delete": {
+        "operationId": "deleteResources",
+        "responses": {
+          "204": {
+            "description": "Ok"
+          },
+          "400": {
+            "description": "Bad Request - constraint failure"
+          },
+          "409": {
+            "description": "Conflict - epoch not greater"
+          }
+        }
+      }
+    },
+    "/{groupType}/{groupId}/{resourceType}/{resourceId}": {
+      "parameters": [
+        {
+          "$ref": "#/components/parameters/groupType"
+        },
+        {
+          "$ref": "#/components/parameters/groupId"
+        },
+        {
+          "$ref": "#/components/parameters/resourceType"
+        },
+        {
+          "$ref": "#/components/parameters/resourceId"
+        },
+        {
+          "in": "query",
+          "name": "meta",
+          "description": "Interact with the metadata",
+          "required": false,
+          "schema": {
+            "type": "boolean"
+          }
+        }
+      ],
+      "get": {
+        "operationId": "getResource",
+        "responses": {
+          "200": {
+            "description": "The corresponding resource",
+            "headers": {
+              "resource-id": {
+                "$ref": "#/components/headers/resource-id"
+              },
+              "resource-version": {
+                "$ref": "#/components/headers/resource-version"
+              },
+              "resource-name": {
+                "$ref": "#/components/headers/resource-name"
+              },
+              "resource-self": {
+                "$ref": "#/components/headers/resource-self"
+              },
+              "resource-description": {
+                "$ref": "#/components/headers/resource-description"
+              },
+              "resource-docs": {
+                "$ref": "#/components/headers/resource-docs"
+              },
+              "resource-origin": {
+                "$ref": "#/components/headers/resource-origin"
+              },
+              "resource-tags": {
+                "$ref": "#/components/headers/resource-tags"
+              },
+              "resource-createdby": {
+                "$ref": "#/components/headers/resource-createdby"
+              },
+              "resource-createdon": {
+                "$ref": "#/components/headers/resource-createdon"
+              },
+              "resource-modifiedby": {
+                "$ref": "#/components/headers/resource-modifiedby"
+              },
+              "resource-modifiedon": {
+                "$ref": "#/components/headers/resource-modifiedon"
+              },
+              "Content-Location": {
+                "description": "permalink location of the returned version",
+                "schema": {
+                  "type": "string",
+                  "format": "uri"
+                }
+              }
+            },
+            "content": {
+              "application/octet-stream": {
+                "schema": {
+                  "type": "string",
+                  "format": "binary"
+                }
+              },
+              "application/json": {
+                "schema": {
+                  "$ref": "xregistry_resources.json#/definitions/resource"
+                }
+              }
+            }
+          },
+          "400": {
+            "description": "Bad Request - constraint failure"
+          },
+          "404": {
+            "description": "Not Found"
+          }
+        }
+      },
+      "post": {
+        "summary": "Post new resource version",
+        "description": "Register resource version If resource of specified name does not exist in specified group, resource and resource version is created at version 1. If resource of specified name exists already in specified group, resource is created at latest version + 1. If resource with identical content already exists, existing resource's ID is returned. \n",
+        "operationId": "postResource",
+        "parameters": [
+          {
+            "$ref": "#/components/parameters/resource-description"
+          },
+          {
+            "$ref": "#/components/parameters/resource-docs"
+          },
+          {
+            "$ref": "#/components/parameters/resource-origin"
+          },
+          {
+            "$ref": "#/components/parameters/resource-tags"
+          },
+          {
+            "in": "header",
+            "description": "format",
+            "name": "format",
+            "schema": {
+              "type": "string"
+            }
+          }
+        ],
+        "requestBody": {
+          "description": "A request to add a new resource to the version collection",
+          "content": {
+            "application/octet-stream": {
+              "schema": {
+                "type": "string",
+                "format": "binary"
+              }
+            },
+            "application/json": {
+              "schema": {
+                "$ref": "xregistry_resources.json#/definitions/resource"
+              }
+            }
+          }
+        },
+        "responses": {
+          "201": {
+            "description": "The created resource",
+            "headers": {
+              "resource-id": {
+                "$ref": "#/components/headers/resource-id"
+              },
+              "resource-version": {
+                "$ref": "#/components/headers/resource-version"
+              },
+              "resource-name": {
+                "$ref": "#/components/headers/resource-name"
+              },
+              "resource-self": {
+                "$ref": "#/components/headers/resource-self"
+              },
+              "resource-description": {
+                "$ref": "#/components/headers/resource-description"
+              },
+              "resource-docs": {
+                "$ref": "#/components/headers/resource-docs"
+              },
+              "resource-origin": {
+                "$ref": "#/components/headers/resource-origin"
+              },
+              "resource-tags": {
+                "$ref": "#/components/headers/resource-tags"
+              },
+              "resource-createdby": {
+                "$ref": "#/components/headers/resource-createdby"
+              },
+              "resource-createdon": {
+                "$ref": "#/components/headers/resource-createdon"
+              },
+              "resource-modifiedby": {
+                "$ref": "#/components/headers/resource-modifiedby"
+              },
+              "resource-modifiedon": {
+                "$ref": "#/components/headers/resource-modifiedon"
+              }
+            },
+            "content": {
+              "application/octet-stream": {
+                "schema": {
+                  "type": "string",
+                  "format": "binary"
+                }
+              },
+              "application/json": {
+                "schema": {
+                  "$ref": "xregistry_resources.json#/definitions/resource"
+                }
+              }
+            }
+          },
+          "400": {
+            "description": "Bad Request - constraint failure"
+          },
+          "404": {
+            "description": "Not Found"
+          },
+          "409": {
+            "description": "Conflict - epoch not greater"
+          }
+        }
+      },
+      "delete": {
+        "summary": "Delete resource",
+        "description": "Delete resource",
+        "operationId": "deleteResource",
+        "parameters": [
+          {
+            "in": "query",
+            "name": "epoch",
+            "description": "The epoch of the resource to be deleted",
+            "required": false,
+            "schema": {
+              "type": "integer",
+              "description": "A number representing the version number of the resource.",
+              "format": "int64"
+            }
+          }
+        ],
+        "responses": {
+          "204": {
+            "description": "No Content"
+          },
+          "400": {
+            "description": "Bad Request - constraint failure"
+          },
+          "404": {
+            "description": "Not Found"
+          }
+        }
+      }
+    },
+    "/{groupType}/{groupId}/{resourceType}/{resourceId}/versions/{versionId}": {
+      "parameters": [
+        {
+          "$ref": "#/components/parameters/groupType"
+        },
+        {
+          "$ref": "#/components/parameters/groupId"
+        },
+        {
+          "$ref": "#/components/parameters/resourceType"
+        },
+        {
+          "$ref": "#/components/parameters/resourceId"
+        },
+        {
+          "$ref": "#/components/parameters/versionId"
+        },
+        {
+          "in": "query",
+          "name": "meta",
+          "description": "Interact with the metadata",
+          "required": false,
+          "schema": {
+            "type": "boolean"
+          }
+        }
+      ],
+      "get": {
+        "operationId": "getResourceVersion",
+        "description": "Gets the document stored for the schema version",
+        "responses": {
+          "200": {
+            "description": "The schema version document",
+            "headers": {
+              "resource-id": {
+                "$ref": "#/components/headers/resource-id"
+              },
+              "resource-version": {
+                "$ref": "#/components/headers/resource-version"
+              },
+              "resource-name": {
+                "$ref": "#/components/headers/resource-name"
+              },
+              "resource-self": {
+                "$ref": "#/components/headers/resource-self"
+              },
+              "resource-description": {
+                "$ref": "#/components/headers/resource-description"
+              },
+              "resource-docs": {
+                "$ref": "#/components/headers/resource-docs"
+              },
+              "resource-origin": {
+                "$ref": "#/components/headers/resource-origin"
+              },
+              "resource-tags": {
+                "$ref": "#/components/headers/resource-tags"
+              },
+              "resource-createdby": {
+                "$ref": "#/components/headers/resource-createdby"
+              },
+              "resource-createdon": {
+                "$ref": "#/components/headers/resource-createdon"
+              },
+              "resource-modifiedby": {
+                "$ref": "#/components/headers/resource-modifiedby"
+              },
+              "resource-modifiedon": {
+                "$ref": "#/components/headers/resource-modifiedon"
+              }
+            },
+            "content": {
+              "application/octet-stream": {
+                "schema": {
+                  "type": "string",
+                  "format": "binary"
+                }
+              }
+            }
+          },
+          "400": {
+            "description": "Bad Request - constraint failure"
+          },
+          "404": {
+            "description": "Not Found"
+          }
+        }
+      },
+      "delete": {
+        "operationId": "deleteResourceVersion",
+        "responses": {
+          "204": {
+            "description": "Ok"
+          },
+          "400": {
+            "description": "Bad Request - constraint failure"
+          },
+          "409": {
+            "description": "Conflict - epoch not greater"
+          }
+        }
+      }
+    }
+  },
+  "components": {
+    "securitySchemes": {
+      "api_key": {
+        "type": "apiKey",
+        "name": "code",
+        "in": "query"
+      }
+    },
+    "parameters": {
+      "groupType": {
+        "in": "path",
+        "name": "groupType",
+        "description": "The groupType (plural)",
+        "required": true,
+        "schema": {
+          "type": "string"
+        }
+      },
+      "resourceType": {
+        "in": "path",
+        "name": "id",
+        "description": "The resourceType (plural)",
+        "required": true,
+        "schema": {
+          "type": "string"
+        }
+      },
+      "id": {
+        "in": "path",
+        "name": "id",
+        "description": "The id of the endpoint",
+        "required": true,
+        "schema": {
+          "type": "string",
+          "description": "A unique identifier",
+          "format": "uri-reference"
+        }
+      },
+      "groupId": {
+        "in": "path",
+        "name": "groupId",
+        "description": "The id of the group",
+        "required": true,
+        "schema": {
+          "type": "string",
+          "description": "A unique identifier",
+          "format": "uri-reference"
+        }
+      },
+      "resourceId": {
+        "in": "path",
+        "name": "resourceId",
+        "description": "The id of the schema",
+        "required": true,
+        "schema": {
+          "type": "string",
+          "description": "A unique identifier",
+          "format": "uri-reference"
+        }
+      },
+      "versionId": {
+        "in": "path",
+        "name": "versionId",
+        "description": "The id of the schema",
+        "required": true,
+        "schema": {
+          "type": "string",
+          "description": "A unique identifier",
+          "format": "uri-reference"
+        }
+      },
+      "resource-type": {
+        "in": "header",
+        "name": "resource-type",
+        "required": false,
+        "schema": {
+          "type": "string"
+        }
+      },
+      "resource-id": {
+        "in": "header",
+        "name": "resource-id",
+        "required": false,
+        "description": "A unique identifier for this resource. This value MUST be globally unique",
+        "schema": {
+          "type": "string",
+          "format": "uri-reference"
+        }
+      },
+      "resource-version": {
+        "in": "header",
+        "name": "resource-version",
+        "required": false,
+        "description": "A number representing the version number of the resource.",
+        "schema": {
+          "type": "integer",
+          "format": "int64"
+        }
+      },
+      "resource-self": {
+        "in": "header",
+        "name": "resource-self",
+        "required": false,
+        "description": "A unique URI for the resource.",
+        "schema": {
+          "type": "string",
+          "format": "uri"
+        }
+      },
+      "resource-description": {
+        "in": "header",
+        "name": "resource-description",
+        "required": false,
+        "description": "A summary of the purpose of the resource.",
+        "schema": {
+          "type": "string"
+        }
+      },
+      "resource-name": {
+        "in": "header",
+        "name": "resource-name",
+        "required": false,
+        "description": "The name of the resource.",
+        "schema": {
+          "type": "string"
+        }
+      },
+      "resource-docs": {
+        "in": "header",
+        "required": false,
+        "name": "resource-docs",
+        "description": "Absolute URL that provides a link to additional documentation about the resource.",
+        "schema": {
+          "type": "string",
+          "format": "uri"
+        }
+      },
+      "resource-origin": {
+        "in": "header",
+        "required": false,
+        "name": "resource-origin",
+        "description": "A URI reference to the original source of this resource.",
+        "schema": {
+          "type": "string",
+          "format": "uri-reference"
+        }
+      },
+      "resource-tags": {
+        "in": "header",
+        "name": "resource-tags",
+        "required": false,
+        "schema": {
+          "type": "array",
+          "items": {
+            "$ref": "#/components/schemas/resource-tag"
+          }
+        }
+      },
+      "resource-createdby": {
+        "in": "header",
+        "name": "resource-createdby",
+        "required": false,
+        "description": "Identity of who created this entity",
+        "schema": {
+          "type": "string"
+        }
+      },
+      "resource-createdon": {
+        "in": "header",
+        "name": "resource-createdon",
+        "required": false,
+        "description": "Time when this entity was created",
+        "schema": {
+          "type": "string",
+          "format": "date-time"
+        }
+      },
+      "resource-modifiedby": {
+        "in": "header",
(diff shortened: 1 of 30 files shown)


registry/spec.md
```diff
@@ -1483,165 +1624,1172 @@ An attempt to delete all versions MUST generate an error.
...
+            "type": "schema",
+            "id": "com.example.telemetrydata",
+            "description": "device telemetry event data",
+            "format": "Protobuf/3.0",
+            "versions : {
+              "1.0" : {
+                "type" : "schemaversion",
+                "id" : "1.0",
+                "schema" : "syntax = \"proto3\"; message Metrics { float metric = 1; } }"
+            }
```
duglin, line 1815: missing a }

registry/spec.md
```diff
@@ -149,20 +187,22 @@ specification, MUST be adhere to.
...
 - Description: A mechanism in which additional metadata about the entity can
   be stored without changing the schema of the entity.
 - Constraints:
   - if present, MUST be a map of zero or more name/value string pairs
   - each name MUST be a non-empty string consisting of only alphanumeric
     characters, `-`, `_` or a `.`; be no longer than 63 characters;
-	start with an alphanumeric character and be unique within the scope of
-	this map. Values MAY be empty strings
+  start with an alphanumeric character and be unique within the scope of
+  this map. Values MAY be empty strings
```
duglin, line 200: I think these 2 lines need to be indented 2 more spaces

registry/spec.md
```diff
@@ -347,36 +397,40 @@ The following describes the attributes of Registry model:
   - The default value is zero (`0`), meaning no old versions will be stored
   - A value of negative one (`-1`) indicates there is no stated limit, and
     implementation MAY prune old versions at any time. Implementation MUST
-	NOT delete a version without also deleting all older versions.
+ NOT delete a version without also deleting all older versions.
```
duglin, line 400: needs more indentation

registry/spec.md
clemensv: Done

registry/spec.md
```diff
@@ -12,20 +12,23 @@ automation and tooling.
...
+- [Registry Formats and APIs](#registry-formats-and-apis)
+  - [Attributes and Extensions](#attributes-and-extensions)
+  - [Registry APIs](#registry-apis)
+    - [Retrieving the Registry Model](#retrieving-the-registry-model)
+    - [Retrieving the Registry](#retrieving-the-registry)
+    - [Managing Groups](#managing-groups)
+    - [Managing Resources](#managing-resources)
+    - [Managing versions of a Resource](#managing-versions-of-a-resource)
+- [CloudEvents Registry](#cloudevents-registry)
+  - [Schema Registry](#schema-registry)
```
duglin, line 24: If this is under a "CE Registry" will some folks think that it only applies to CEs and it's not a general purpose schema registry?

registry/spec.md
```diff
@@ -12,20 +12,23 @@ automation and tooling.
...
+    - [Schemagroups](#group-schemagroups)
```
duglin, line 25: s/Schemagroups/Schema Groups/   To be consistent with line 28 below.

registry/spec.md
```diff
@@ -83,13 +97,33 @@ include model specific Groups, Resources and extension attributes.
...
+way that is can be represented in a single file, including but not limited to
+JSON or YAML, or via the resource graph of a REST API.
+
+### Attributes and Extensions
 
 The follow attributes are used by one or more entities defined by this
 specification. They are defined here once rather than repeating them
 throughout the specification.
 
 List of attributes:
```
duglin, line 125: While you're in here, can you change this to just "Attributes:"?

registry/spec.md
```diff
@@ -254,12 +301,11 @@ specification, MUST be adhere to.
...
   - `https://example.com/docs/myQueue`
 
-
-## Registry APIs
+### Registry APIs
 
 This specification defines the following API patterns:
 
-```
+``` meta
```
duglin, line 308: I couldn't find "meta" as a defined language - where did you see that?

registry/spec.md
```diff
@@ -1483,165 +1624,1718 @@ An attempt to delete all versions MUST generate an error.
...
+  schema registry in file form in the parent directory, which formally organizes
+  and documents all versions of all Protobuf schemas that are used in the
+  directory.
+- The [Message Definitions Registry](#message-definitions-registry) section
+  describes the metadata description of events and messages. The payload schemas
+  for events and messages can be embedded in the definition, reference an
+  external schema document, or can be referenced into the schema registry. The
+  message definitions registry is universally applicable to any asynchronous
+  messaging and eventing scenario. You might define a group of definitions that
+  describe precisely which messageds with which metadata are permitted to flow
```
duglin, line 1650: s/messageds/messages/

duglin: Should it be `, with which metadata,` ? (added commas)

registry/spec.md
```diff
@@ -1483,165 +1624,1718 @@ An attempt to delete all versions MUST generate an error.
...
+  description of network endpoints that accept or emit events and messages. The
+  endpoint registry is a formal description of associations of message
+  definitions and network endpoints, which can be used to discover endpoints
+  that consume or emit particular messages or events via a central registry. The
+  message definitions can be embedded into the endpoint metadata or as
+  a reference into the message definitions registry.
+
+The metadata model is structured such that network endpoint information and
+message metadata and payload schemas can be described compactly in a single
+object in the simplest case or can be spread out and managed across separate
```
duglin, line 1666: I can't decide. Is "object" the best word here or would "document" make it easier to understand? Not a big deal either way - it just made me stop and wonder as I was reading it. It could just be me though. :-)

registry/spec.md
```diff
@@ -1483,165 +1624,1718 @@ An attempt to delete all versions MUST generate an error.
...
+registry products in a sophisticated large-enterprise scenario.
+
+The following is an exemplary, compact definition of an MQTT 5.0 consumer
+endpoint with a single, embedded message definition using an embedded Protobuf 3
+schema for its payload.
+
+``` JSON
+{
+    "$schema": "https://cloudevents.io/schemas/registry",
```
duglin, line 1675: We never define `$schema` - we should either do so or drop it. Do we need it?

registry/spec.md
```diff
@@ -1483,165 +1624,1718 @@ An attempt to delete all versions MUST generate an error.
...
+    "specversion": "0.3-wip",
```
duglin, line 1676: should be 0.4 since you bumped it.

registry/spec.md
```diff
@@ -1483,165 +1624,1718 @@ An attempt to delete all versions MUST generate an error.
...
+                                "type": "uritemplate",
+                                "value": "{deploymentid}/{deviceid}",
+                                "required": true
+                            }
+                        }
+                    },
+                    "schemaformat": "Protobuf/3.0",
+                    "schema" : "syntax = \"proto3\"; message Metrics { float metric = 1; } }"
+                }
+            },
```
duglin, line 1726: extra ,

registry/spec.md
```diff
@@ -1483,165 +1624,1718 @@ An attempt to delete all versions MUST generate an error.
...
+
+The following is an exemplary, compact definition of an MQTT 5.0 consumer
+endpoint with a single, embedded message definition using an embedded Protobuf 3
+schema for its payload.
+
+``` JSON
+{
+    "$schema": "https://cloudevents.io/schemas/registry",
+    "specversion": "0.3-wip",
+    "id": "urn:uuid:3978344f-8596-4c3a-a978-8fc9a6a469f7",
```
duglin, line 1677: We don't define `id` as a top-level attribute. Should we?

registry/spec.md
```diff
@@ -1483,165 +1624,1718 @@ An attempt to delete all versions MUST generate an error.
...
+    "endpoints" : 
+    {
+       "com.example.telemetry" : {
```
duglin, line 1680: while I know this is the ID and can be any string, as the first example people see should be make it something other than the URL of the endpoint so people don't assume the children of "endpoints" is a map where the key is a URL?

registry/spec.md
```diff
@@ -1483,165 +1624,1718 @@ An attempt to delete all versions MUST generate an error.
...
+The following is an exemplary, compact definition of an MQTT 5.0 consumer
+endpoint with a single, embedded message definition using an embedded Protobuf 3
+schema for its payload.
+
+``` JSON
+{
+    "$schema": "https://cloudevents.io/schemas/registry",
+    "specversion": "0.3-wip",
+    "id": "urn:uuid:3978344f-8596-4c3a-a978-8fc9a6a469f7",
+    "endpoints" :
```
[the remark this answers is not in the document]
duglin: this is missing these:
```
  "GROUPsURL": "URL",         # eg. "endpointsURL"
  "GROUPsCount": INT,         # eg. "endpointsCount"
```
In the `?inline` case, which is what you're showing here, we need to decide if these 2 attribute will still appear or if they're optional. It might be nice to always have them for consistency even if they're `{"endpointsURL": "endpoints", "endpointsCount":1}` in this case.

registry/spec.md
```diff
@@ -1483,165 +1624,1718 @@ An attempt to delete all versions MUST generate an error.
...
+{
+    "$schema": "https://cloudevents.io/schemas/registry",
+    "specversion": "0.3-wip",
+    "id": "urn:uuid:3978344f-8596-4c3a-a978-8fc9a6a469f7",
+    "endpoints" : 
+    {
+       "com.example.telemetry" : {
+           "type": "endpoint",
+           "id" : "com.example.telemetry",
+            "usage": "consumer",
```
duglin, line 1683: Missing "name" and "epoch".
The indent of "usage" (and following lines) are off by one char.

registry/spec.md
```diff
@@ -1483,165 +1624,1718 @@ An attempt to delete all versions MUST generate an error.
...
+                "endpoints": [
+                    "mqtt://mqtt.example.com:1883",
+                ],
+                "options" : {
+                    "topic": "{deviceid}/telemetry"
+                }
+            },
+            "format" : "CloudEvents/1.0"
+            "definitions": {
+                  "com.example.telemetry" : {
```
duglin, line 1696: As an intro example, it might be less confusing the make the endpoint's ID and this event ID be something different. Perhaps make this one something more eventing-ish, like "com.example.telemetry.activated" ??

registry/spec.md
```diff
@@ -1483,165 +1624,1718 @@ An attempt to delete all versions MUST generate an error.
...
+    {
+       "com.example.telemetry" : {
+           "type": "endpoint",
+           "id" : "com.example.telemetry",
+            "usage": "consumer",
+            "config": {
+                "protocol": "MQTT/5.0",
+                "strict": false,
+                "endpoints": [
+                    "mqtt://mqtt.example.com:1883",
```
duglin, line 1688: extra ,

registry/spec.md
```diff
@@ -1483,165 +1624,1718 @@ An attempt to delete all versions MUST generate an error.
...
+                    "topic": "{deviceid}/telemetry"
+                }
+            },
+            "format" : "CloudEvents/1.0"
+            "definitions": {
+                  "com.example.telemetry" : {
+                    "type": "cloudevent",
+                    "id": "com.example.telemetry",
+                    "description": "device telemetry event",
+                    "format": "CloudEvents/1.0",
```
duglin, line 1700: What is "type"? I don't see that defined, but maybe I'm missing it. If it remains, it might be hard for people to easily understand why we need it and "format" when they're both variants of "cloudevent"

registry/spec.md
```diff
@@ -1483,165 +1624,1718 @@ An attempt to delete all versions MUST generate an error.
...
+                "options" : {
+                    "topic": "{deviceid}/telemetry"
+                }
+            },
+            "format" : "CloudEvents/1.0"
+            "definitions": {
+                  "com.example.telemetry" : {
+                    "type": "cloudevent",
+                    "id": "com.example.telemetry",
+                    "description": "device telemetry event",
```
duglin, line 1699: Missing "name"

duglin: and "self"

duglin: and "ownerGroup"

registry/spec.md
clemensv: That is a very justified question. The "CloudEvents Registry" is probably just a profile of this that is scoped to the "CloudEvents" format. 

We basically have
* A generic registry model that would work for all sorts of schema- and metaschema-like documents
* A concrete schema registry
* A concrete registry for event and messaging envelope definitions
* A concrete registry for eventing and messaging endpoints

There's a "conformance trail" (a set of MUST rules) through those that would make up a "CloudEvents registry", but you can just as well make up a conformance trail that results in an "MQTT registry"

registry/spec.md
clemensv: Why not both? :)  "in a single metadata object (and therefore as a single document)"

registry/spec.md
clemensv: We need a formal identifier for "our" JSON schema. Yes, we need it.

registry/spec.md
clemensv: Yes. The doc itself should have an ID.

registry/spec.md
clemensv: OpenAPI has introduced its own sub-flavor of JSON schema for schemas that use inheritance as we do and that requires a "discriminator" 

https://spec.openapis.org/oas/latest.html#discriminator-object

The unfortunate effect of "discriminator" is that every instance of every derived type needs to labeled in some way that the type can be inferred. 

Since we need to use OpenAPI for the REST API, I cannot avoid buying into that scheme.

In our schema doc (schemas/ce_registry_doc.json in my PR), I have started to work around the "type" field and use "format" as the discriminator. That doesn't make me happy, but is an effect orf OpenAPI imposing that requirement and us needing OpenAPI:

```json
 "discriminator": {
          "propertyName": "format",
          "mapping": {
            "AMQP/1.0": "#/components/schemas/amqpDefinition10",
            "AMQP": "#/components/schemas/amqpDefinition",
            "MQTT/3.1.1": "#/components/schemas/mqttDefinition311",
            "MQTT/5.0": "#/components/schemas/mqttDefinition50",
            "MQTT": "#/components/schemas/mqttDefinition",
            "HTTP/1.1": "#/components/schemas/httpDefinition11",
            "HTTP": "#/components/schemas/httpDefinition",
            "HTTP/2": "#/components/schemas/httpDefinition2",
            "HTTP/3": "#/components/schemas/httpDefinition3",
            "KAFKA/0.11": "#/components/schemas/kafkaDefinition011",
            "KAFKA": "#/components/schemas/kafkaDefinition",
            "CloudEvents/1.0": "#/components/schemas/cloudEventDefinition10",
            "CloudEvents": "#/components/schemas/cloudEventDefinition",
            "Generic": "#/components/schemas/genericDefinition"
          }

```

`type` will have to remain for all derivatives of `resource` (including `definition`), but with leaning on `format` I will be able to avoid further subdivision of `definition`

``` json
 "resource": {
        "type": "object",
        "discriminator": {
          "propertyName": "type",
          "mapping": {
            "definitiongroup": "#/components/schemas/definitiongroup",
            "definition": "#/components/schemas/definition",
            "schemagroup": "#/components/schemas/schemaGroup",
            "endpoint": "#/components/schemas/endpoint",
            "schema": "#/components/schemas/schema",
            "schemaversion": "#/components/schemas/schemaVersion"
          }
        },
```

registry/spec.md
clemensv: YAML-trained eyes.

* We should discuss whether we really need `name`. It's at most an override for `id`.
* `epoch` makes sense to have in API-hosted docs, but I would like to make it optional (quasi: required, but default to zero) in files.

registry/spec.md
```diff
@@ -1483,165 +1624,1718 @@ An attempt to delete all versions MUST generate an error.
...
+                "strict": false,
+                "endpoints": [
+                    "mqtt://mqtt.example.com:1883",
+                ],
+                "options" : {
+                    "topic": "{deviceid}/telemetry"
+                }
+            },
+            "format" : "CloudEvents/1.0"
+            "definitions": {
```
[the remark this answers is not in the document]
clemensv: Since this is an example, it only shows one of the options. They are practically (even if not formally) mutually exclusive.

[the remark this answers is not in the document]
duglin: We should discuss this on the call tomorrow. I'm torn, while it does seem redundant to include the URL to the set when the set is inlined, it could be useful for someone to want both... the data itself now and the reference to it so they re-get it later. Especially if they know they'll want to paginate over it. W/o the URL in this response it'll require an extra GET.

registry/spec.md
```diff
@@ -1491,71 +1632,1350 @@ An attempt to delete all versions MUST generate an error.
 A `DELETE /GROUPs/ID/RESOURCEs/ID/versions` without a body MUST delete all
 versions (except the latest) of the Resource.
 
+## CloudEvents Registry
+
+The CloudEvents Registry is a universal catalog and discovery metadata format
+and metadata service API for messaging and eventing schemas and metaschemas and
```
duglin, line 1638: would love to know of anyone else had trouble parsing this due to all of the "and"s :-)

registry/spec.md
```diff
@@ -1491,71 +1632,1350 @@ An attempt to delete all versions MUST generate an error.
...
+
+``` JSON
+{
+  "$schema": "https://cloudevents.io/schemas/registry",
+  "specversion": "0.4-wip",
+  "id": "urn:uuid:3978344f-8596-4c3a-a978-8fc9a6a469f7",
+  "endpoints" : 
+  {
+    "com.example.telemetry" : {
+      "type": "endpoint",
```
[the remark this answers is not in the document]
duglin: Just as a popular example... github issues: https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28#get-an-issue

registry/spec.md
```diff
@@ -1491,71 +1632,1350 @@ An attempt to delete all versions MUST generate an error.
...
+        "protocol": "MQTT/5.0",
+        "strict": false,
+        "endpoints": [
+            "mqtt://mqtt.example.com:1883",
+        ],
+        "options" : {
+            "topic": "{deviceid}/telemetry"
+        }
+      },
+      "format" : "CloudEvents/1.0"
```
duglin, line 1703: missing  `,`

registry/spec.md
```diff
@@ -1491,71 +1632,1350 @@ An attempt to delete all versions MUST generate an error.
...
+        "protocol": "MQTT/5.0",
+        "strict": false,
+        "endpoints": [
+          "mqtt://mqtt.example.com:1883",
+        ],
+        "options" : {
+          "topic": "{deviceid}/telemetry"
         }
+      },
+      "format" : "CloudEvents/1.0"
```
duglin, line 1766: missing `,`

registry/spec.md
```diff
@@ -1491,71 +1632,1350 @@ An attempt to delete all versions MUST generate an error.
...
+              },
+              "source": {
+                "type": "uritemplate",
+                "value": "{deploymentid}/{deviceid}",
+                "required": true
+              }
+            }
+          },
+          "schemaformat": "Protobuf/3.0",
+          "schemaurl" : "#/schemagroups/com.example.telemetry/schema/com.example.telemetrydata/versions/1.0"
```
duglin, line 1805: s/schema/schemas/   - missing 's'

registry/spec.md
```diff
@@ -1491,71 +1632,1350 @@ An attempt to delete all versions MUST generate an error.
...
+      "schemas": {
+        "com.example.telemetrydata" : {
+          "type": "schema",
+          "id": "com.example.telemetrydata",
+          "description": "device telemetry event data",
+          "format": "Protobuf/3.0",
+          "versions" : {
+            "1.0" : {
+              "type" : "schemaversion",
+              "id" : "1.0",
```
duglin, line 1822: should this be "version" instead? There are other required fields so we need to decide if all of them should be required or optional.

registry/spec.md
```diff
@@ -1491,71 +1632,1350 @@ An attempt to delete all versions MUST generate an error.
...
+              "schema" : "syntax = \"proto3\"; message Metrics { float metric = 1; }"
+          }
```
duglin, line 1824: missing } for the "1.0" property

registry/spec.md
```diff
@@ -1491,71 +1632,1350 @@ An attempt to delete all versions MUST generate an error.
...
+      }
     }
-  ]
+  }
 }
 ```
 
+If we assume the message definitions and schemas to reside at an API endpoint,
```
duglin, line 1831: s/to reside/reside/

registry/spec.md
```diff
@@ -1491,71 +1632,1350 @@ An attempt to delete all versions MUST generate an error.
...
+  "endpoints" : 
+  {
+    "com.example.telemetry" : {
+      "type": "endpoint",
+      "id" : "com.example.telemetry",
+      "usage": "consumer",
+      "config": {
+        // ... details ...
+      },
+      "format" : "CloudEvents/1.0"
```
duglin, line 1849: comma

registry/spec.md
```diff
@@ -1491,71 +1632,1350 @@ An attempt to delete all versions MUST generate an error.
...
+  }
+}
+```
+
+All other references to other objects in the registry can be expressed in the
+same way.
+
+While the CloudEvents Registry is primarily motivated by enabling development of
+CloudEvents-based event flows, the registry is not limited to CloudEvents. It
+can be used to describe any asynchronous messaging or eventing endpoint and its
```
duglin, line 1891: s/endpoint/endpoints/   I think.

registry/spec.md
```diff
@@ -1491,71 +1632,1350 @@ An attempt to delete all versions MUST generate an error.
...
+messages, including such that do not use CloudEvents at all. The [Message
```
duglin, line 1892: does "such" refer to endpoints or messaging? Perhaps a rewording will clarify

registry/spec.md
```diff
@@ -1491,71 +1632,1350 @@ An attempt to delete all versions MUST generate an error.
...
+Formats](#message-formats) section therefore not only describes the attribute
+meta-schema for CloudEvents, but also meta-schemas for the native message
+envelopes of MQTT, AMQP, and other messaging protocols.
+
+The registry is designed to be extensible to support any structured data
+encoding and related schemas for message or event payloads. The [Schema
+Formats](#schema-formats) section describes the meta-schema for JSON Schema, XML
+Schema, Apache Avro schema, Protobuf schema.
```
duglin, line 1900: ...Apache schema and Profbuf schema.

added "and"

registry/spec.md
```diff
@@ -1491,71 +1632,1350 @@ An attempt to delete all versions MUST generate an error.
...
+
+### File format
+
+A CloudEvents Registry can be implemented using the Registry API or with plain
+text files.
+
+When using the file-based model, files with the extension `.cereg` use JSON
+encoding. Files with the extension `.cereg.yaml` or `.cereg.yml` use YAML
```
duglin, line 1908: .cereg.yaml is unfortunate, but I understand why you did it.
Not sure I prefer this idea but just to brainstorm... is the double extension idea better than:
- foo-cereg.json / foo-cereg.yaml
- cereg-foo.json / cereg-foo.yaml

Or some variant like that where the "reg" part is included in the base name rather than the extension.

registry/spec.md
```diff
@@ -1491,71 +1632,1350 @@ An attempt to delete all versions MUST generate an error.
...
+The object declares the roots of the three sub-registries, which are either
+embedded or referenced. Any of the three sub-registries MAY be omitted.
+
+``` meta
+{
+   "$schema": "https://cloudevents.io/schemas/registry",
+   "specversion": "0.4-wip",
+   "endpoints": { ... } | "endpointsUrl": "URL",
+   "definitiongroups": { ... } | "definitiongroupsUrl": "URL"
+   "schemagroups" : { ... } | "schemagroupsUrl": "URL",
```
duglin, line 1930: Including "schemagroups" here makes sense, but then in the registry metadata we'd need to include its definition in then too, alongside "endpoints" and "definitiongroups". Easy enough, however, does this imply that we should consider some kind of referencing mechanism in the registry metadata section so that people don't need to define/copy some other registry's metadata defs each time they want to include instances of those entities in their registry? Let's discuss

registry/spec.md
```diff
@@ -1491,71 +1632,1350 @@ An attempt to delete all versions MUST generate an error.
...
+A CloudEvents Registry file MUST contain a single JSON object or YAML document.
+The object declares the roots of the three sub-registries, which are either
+embedded or referenced. Any of the three sub-registries MAY be omitted.
+
+``` meta
+{
+   "$schema": "https://cloudevents.io/schemas/registry",
+   "specversion": "0.4-wip",
+   "endpoints": { ... } | "endpointsUrl": "URL",
+   "definitiongroups": { ... } | "definitiongroupsUrl": "URL"
```
duglin, line 1929: missing ,

registry/spec.md
duglin: remove trailing ,

registry/spec.md
```diff
@@ -1491,71 +1632,1350 @@ An attempt to delete all versions MUST generate an error.
...
+  "groups": [
+    {
+      "singular": "schemagroup",
+      "plural": "schemagroups",
+      "schema": "TBD",
+      "resources": [
+        {
+          "singular": "schema",
+          "plural": "schemas",
+          "versions": -1,
```
duglin, line 1954: extra ,

registry/spec.md
```diff
@@ -1491,71 +1632,1350 @@ An attempt to delete all versions MUST generate an error.
...
+```
+
+#### Resource: schemas
+
+The resources (RESOURCE) collection inside of schema groups is named `schemas`.
+The type of the resource is `schema`. Any single `schema` is a container for
+one or more `versions`, which hold the concrete schema documents or schema
+document references.
+
+Any new schema version that is added to a schema definition MUST be backwards
```
duglin, line 1995: I _believe_ the text here is accurate, however I think we may need to add some text about versioning of resources someplace because this text will confuse people into thinking that they can't have a v1 and v2 of a schema doc unless they're backwards compatible. Which isn't true - we just need to make it clear that there are two levels of versioning... similar to git commits vs git branches.  Part of me wonders if we should consider using a different term than "version" just to avoid the potential confusion. We'll also need to make it clear that branch-like versioning isn't really part of the spec as much as it's a naming convention of the ID of the resources.

Let's discuss.

registry/spec.md
```diff
@@ -1491,71 +1632,1350 @@ An attempt to delete all versions MUST generate an error.
...
+compatible with all previous versions of the schema, meaning that a consumer
+using the new schema MUST be able to understand data encoded using a prior
+version of the schema. If a new version introduces a breaking change, it MUST be
+regisetered as a new schema with a new name.
+
+When you retrieve a schema without qualifying the version, you will get the
+latest version of the schema, see [retrieving a
+resource](#retrieving-a-resource). The latest version is the lexically greatest
+version number, whereby all version ids MUST be left-padded with spaces to the
```
duglin, line 2004: s/ids/IDs/   but isn't it the `version` string?

registry/spec.md
```diff
@@ -1491,71 +1632,1350 @@ An attempt to delete all versions MUST generate an error.
...
+      "id": "com.example.schemas",
+      "schemas": {
+        ...
+      }
+    }
+  }
+}
+```
+
+#### Resource: schemas
```
duglin, line 1988: let's check that we're consistent on whether things are plural here or not. It feels like it should either be:
Resource: schema
or
Resources: schemas
I prefer the singular version because the 2nd on feels like "schemas" is defining a group not an singular resource type.

registry/spec.md
```diff
@@ -1491,71 +1632,1350 @@ An attempt to delete all versions MUST generate an error.
...
+  - MUST follow the naming convention `{NAME}/{VERSION}`, whereby `{NAME}` is
+    the name of the schema format and `{VERSION}` is the version of the schema
+    format in the format defined by the schema format itself.
+- Examples:
+  - `JsonSchema/draft-07`
+  - `Protobuf/3`
+  - `XSD/1.1`
+  - `Avro/1.9`  
+
+#### Resource Version: schemaversion
```
duglin, line 2031: Let's decide on the pattern here - some options:
#### `version` (schema)
#### `schema.version`
#### schema `version`
#### Schema `version`
#### Schema version
#### schema version
#### schema.version

I kind of like `schema.version`, but I could live with most of these.

registry/spec.md
```diff
@@ -1491,71 +1632,1350 @@ An attempt to delete all versions MUST generate an error.
...
+
+The `VERSION` object of the `schema` resource is of type `schemaversion`. The
```
duglin, line 2033: (if we keep `type`)
Looking at the text here and the example below, I wonder if this will confuse people because the `schema` type is actually a group, not a singular thing - but its name isn't `schemagroup`. That's already taken. My first reaction upon reading this section was that the "type" really should be `schema` for both but, of course, if we keep `type` we can't do that. Something feels off here and I want to blame the presence of `type` :-)

duglin: we need to discuss the relationship between ID, Name and Version. The sample uses the ID as the version string which doesn't seem consistent.

registry/spec.md
duglin: hmm maybe I'm wrong - maybe the text isn't right in that you should be allowed to have breaking changes across revisions, otherwise people would never be allowed to have major version number bumps. I may have been mixing up versions with the epoch stuff.....

registry/schemas/ce_registry_openapi_merged.json
```diff
@@ -0,0 +1,3191 @@
+{
```
matzew, line 1: whats `merged` ? merged with what?

registry/spec.md
```diff
@@ -1491,71 +1632,1350 @@ An attempt to delete all versions MUST generate an error.
...
+      { "my-application-property": {
+            "value": "my-application-property-value"
+          }
+      }
+    ]
+  }
+}
+```
+
+#### "Kafka/0.11" format
```
matzew, line 2831: Is this really how Kafka identifies the versions?

And why such an old version 0.11? 

Is it because the **KIP** for adding `headers` was added to 0.11.0? (see [KAFKA-4208](https://issues.apache.org/jira/browse/KAFKA-4208)) and since than there was no major update on the records API?

registry/spec.md
```diff
@@ -1483,165 +1624,1718 @@ An attempt to delete all versions MUST generate an error.
...
+  message definitions registry is universally applicable to any asynchronous
+  messaging and eventing scenario. You might define a group of definitions that
+  describe precisely which messageds with which metadata are permitted to flow
+  into a channel and can thus be expected by consumers of that channel and then
+  associate that definition group with a topic or queue in your eventing or
+  messaging infrastructure. That association might be a metadata attribute on
+  the topic or queue in the messaging infrastructure that embeds the metadata or
+  points to it.
+- The [Endpoint Registry](#endpoint-registry) section defines the metadata
+  description of network endpoints that accept or emit events and messages. The
```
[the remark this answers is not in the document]
clemensv: Yes

registry/spec.md
clemensv: It is a subordinate object of schema.

registry/spec.md
clemensv: No. It's the "id" of the schemaVersion object.

duglin: I probably wasn't clear. It seems like the key on this line should be "version" or the value should be "com.example.telemetrydata". I would have expected the fields of the latest version to match the fields of the resource (sibling fields to "versions") - including the "id". Meaning, the child of "1.0" is the same resource as the child of "com.example.telemetrydata"  - but w/o the "versions" field.

matzew:
Would it make sense to break the `spec.md` into smaller pieces?
* subscriton
* event groups / group definitions
* schema groups / schema definitions

registry/schemas/xregistry_openapi.json
```diff
@@ -0,0 +1,1004 @@
...
+            "name": "inline",
+            "description": "Set if references shall be inlined",
+            "required": false,
+            "schema": {
+              "type": "boolean"
+            }
+          },
+          {
+            "in": "query",
+            "name": "name",
```
duglin, line 91: s/name/filter/

registry/schemas/xregistry_openapi.json
```diff
@@ -0,0 +1,1004 @@
...
+            "description": "Query expression",
```
duglin, line 92: S/Query/Filter/

registry/schemas/xregistry_openapi.json
```diff
@@ -0,0 +1,1004 @@
...
+        },
+        {
+          "$ref": "#/components/parameters/groupId"
+        }
+      ],
+      "get": {
+        "operationId": "getResourceGroup",
+        "responses": {
+          "200": {
+            "description": "The schema group",
```
duglin, line 172: shouldn't be specific to schemas

registry/schemas/xregistry_openapi.json
```diff
@@ -0,0 +1,1004 @@
...
+            "required": false,
+            "schema": {
+              "type": "integer",
+              "description": "A number representing the version number of the resource.",
+              "format": "int64"
+            }
+          }
+        ],
+        "responses": {
+          "204": {
```
duglin, line 240: Should we add "200" too?

registry/schemas/xregistry_openapi.json
```diff
@@ -0,0 +1,1004 @@
...
+          "$ref": "#/components/parameters/resourceType"
+        }
+      ],
+      "get": {
+        "operationId": "getResourcesAll",
+        "description": "Get an optionally filtered collection of resources",
+        "parameters": [
+          {
+            "in": "query",
+            "name": "query",
```
duglin, line 270: filter

registry/schemas/xregistry_openapi.json
```diff
@@ -0,0 +1,1004 @@
...
+            "description": "Query expression",
```
duglin, line 271: s/Query/Filter/

registry/schemas/xregistry_openapi.json
```diff
@@ -0,0 +1,1004 @@
...
+            "description": "Set if references shall be inlined",
+            "required": false,
+            "schema": {
+              "type": "boolean"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "A list of resources (optionally matching the query parameter)",
```
duglin, line 290: s/query expression/filter expression/

registry/schemas/xregistry_openapi.json
```diff
@@ -0,0 +1,1004 @@
...
+          },
+          "409": {
+            "description": "Conflict - epoch not greater"
+          }
+        }
+      },
+      "delete": {
+        "operationId": "deleteResources",
+        "responses": {
+          "204": {
```
duglin, line 344: 200 & 202

registry/schemas/xregistry_openapi.json
```diff
@@ -0,0 +1,1004 @@
...
+                }
+              },
+              "application/json": {
+                "schema": {
+                  "$ref": "xregistry_resources.json#/definitions/resource"
+                }
+              }
+            }
+          },
+          "400": {
```
duglin, line 444: Should there be 307 or 308 in here?

registry/schemas/xregistry_openapi.json
```diff
@@ -0,0 +1,1004 @@
...
+            "$ref": "#/components/parameters/resource-tags"
+          },
+          {
+            "in": "header",
+            "description": "format",
+            "name": "format",
+            "schema": {
+              "type": "string"
+            }
+          }
```
duglin, line 476: In another PR we should add all of the other xreg headers (?meta values)

registry/schemas/xregistry_openapi.json
```diff
@@ -0,0 +1,1004 @@
...
+            "required": false,
+            "schema": {
+              "type": "integer",
+              "description": "A number representing the version number of the resource.",
+              "format": "int64"
+            }
+          }
+        ],
+        "responses": {
+          "204": {
```
duglin, line 578: 200/202

registry/schemas/xregistry_openapi.json
```diff
@@ -0,0 +1,1004 @@
...
+            "content": {
+              "application/octet-stream": {
+                "schema": {
+                  "type": "string",
+                  "format": "binary"
+                }
+              }
+            }
+          },
+          "400": {
```
duglin, line 670: we should allow for 3xx's

registry/schemas/xregistry_openapi_endpoint.json
```diff
@@ -0,0 +1,960 @@
+{
+  "openapi": "3.0.6",
+  "info": {
+    "title": "xRegistry API",
```
duglin, line 4: should probably make this specific to Endpoints

registry/schemas/xregistry_openapi_endpoint.json
```diff
@@ -0,0 +1,960 @@
...
+              }
+            }
+          }
+        }
+      }
+    },
+    "/endpoints": {
+      "get": {
+        "operationId": "getResourceGroupAll",
+        "description": "Gets all entries of the resource group",
```
duglin, line 73: Seems all references to generic xReg terms like "resource group" and "resource" should be "Endpoint" and "Definition", no?

registry/schemas/xregistry_openapi_messagedefinition.json
```diff
@@ -0,0 +1,960 @@
+{
+  "openapi": "3.0.6",
+  "info": {
+    "title": "xRegistry API",
```
duglin, line 4: same comments from xregistry_openapi_endpoints apply here (make it specific to message defs)

registry/schemas/xregistry_openapi_schema.json
```diff
@@ -0,0 +1,960 @@
+{
+  "openapi": "3.0.6",
+  "info": {
+    "title": "xRegistry API",
```
duglin, line 4: make this doc schema specific

registry/schemas/xregistry_openapi_template.json
```diff
@@ -0,0 +1,960 @@
+{
+  "openapi": "3.0.6",
```
duglin, line 2: is this doc an implementation specific doc?

registry/schemas/xregistry_resources.json
```diff
@@ -0,0 +1,106 @@
+{
+  "$schema": "https://json-schema.org/draft-07/schema",
+  "type": "object",
+  "title": "document",
+  "properties": {
+    "_ref": {      "$ref": "#/definitions/document"
```
duglin, line 6: extra char

registry/spec.md
duglin: I still think Schema Registry should not be under the CE TOC entry.

registry/spec.md
```diff
@@ -38,9 +41,20 @@ provide a common interaction pattern for these types of services with the goal
...
 
-This specification itself is meant to be a framework from which additional
-specifications will be defined that expose model specific resources and
-metadata.
+The first section of this specification, ["Registry Formats and
+APIs"](#registry-formats-and-apis), is meant to be a framework from which
+additional specifications can be defined that expose model specific resources
+and metadata.
+
+The second section of this specification ["CloudEvents
```
duglin, line 49: In prep for moving this to the xReg repo, it might be good to rename this something more generic

registry/spec.md
```diff
@@ -987,19 +1081,21 @@ Content-Length: nnnn
...
 This will update the latest version of a Resource. Missing Registry HTTP
 headers MUST NOT be interpreted as deleting the attribute. However, a Registry
 HTTP headers with an empty string for its value MUST be interpreted as a
 request to delete the attribute.
 
 The request MUST be of the form:
-```
+
+``` meta
 PUT /GROUPs/ID/RESOURCEs/ID[?epoch=EPOCH]
```
duglin, line 1099: For a follow-on PR... but I'm wondering if we should move `epoch` to be an HTTP like all other ?meta properties.

registry/spec.md
```diff
@@ -1498,71 +1630,1341 @@ An attempt to delete all versions MUST generate an error.
...
+be spread out and managed across separate registry products in a sophisticated
+large-enterprise scenario.
+
+The following is an exemplary, compact definition of an MQTT 5.0 consumer
+endpoint with a single, embedded message definition using an embedded Protobuf 3
+schema for its payload.
+
+``` JSON
+{
+  "$schema": "https://cloudevents.io/schemas/registry",
```
duglin, line 1682: we need to either define this or explain how this is different from the "schema" URL we're going to put into the `model` section

registry/spec.md
```diff
@@ -1498,71 +1630,1341 @@ An attempt to delete all versions MUST generate an error.
...
+  "specversion": "0.4-wip",
```
duglin, line 1683: 0.5-wip

registry/spec.md
```diff
@@ -1498,71 +1630,1341 @@ An attempt to delete all versions MUST generate an error.
...
+      "schemas": {
+        "com.example.telemetrydata" : {
+          "id": "com.example.telemetrydata",
+          "description": "device telemetry event data",
+          "format": "Protobuf/3.0",
+          "versions" : {
+            "1.0" : {
+              "type" : "schemaversion",
+              "id" : "1.0",
+              "schema" : "syntax = \"proto3\"; message Metrics { float metric = 1; }"
```
duglin, line 1816: Missing a `}` after this line

registry/spec.md
```diff
@@ -1498,71 +1630,1341 @@ An attempt to delete all versions MUST generate an error.
...
+  "$schema": "https://cloudevents.io/schemas/registry",
+  "specversion": "0.4-wip",
+  "id": "urn:uuid:3978344f-8596-4c3a-a978-8fc9a6a469f7",
+  "endpoints" : 
+  {
+    "com.example.telemetry" : {
+      "id" : "com.example.telemetry",
+      "usage": "consumer",
+      "config": {
+        // ... details ......
```
duglin, line 1866: if removing this line removes the red text below, let's do it

registry/spec.md
```diff
@@ -1498,71 +1630,1341 @@ An attempt to delete all versions MUST generate an error.
...
+```
+
+#### Group: schemaGroups
+
+The group (GROUP) name for the Schema Registry is `schemaGroups`. The group does
+not have any specific extension attributes.
+
+A schema group is a collection of schemas that are related to each other in some
+application-defined way. A schema group does not impose an restrictions on the
+contained schenmas, meaning that a schema group can contain schemas of different
```
duglin, line 1960: schenmas -> schemas

registry/spec.md
```diff
@@ -1498,71 +1630,1341 @@ An attempt to delete all versions MUST generate an error.
...
+The resources (RESOURCE) collection inside of schema groups is named `schemas`.
+The type of the resource is `schema`. Any single `schema` is a container for
+one or more `versions`, which hold the concrete schema documents or schema
+document references.
+
+Any new schema version that is added to a schema definition MUST be backwards
+compatible with all previous versions of the schema, meaning that a consumer
+using the new schema MUST be able to understand data encoded using a prior
+version of the schema. If a new version introduces a breaking change, it MUST be
+regisetered as a new schema with a new name.
```
duglin, line 1989: regisetered -> registered

registry/spec.md
```diff
@@ -1498,71 +1630,1341 @@ An attempt to delete all versions MUST generate an error.
...
+}
+```
+
+#### Group: schemaGroups
+
+The group (GROUP) name for the Schema Registry is `schemaGroups`. The group does
+not have any specific extension attributes.
+
+A schema group is a collection of schemas that are related to each other in some
+application-defined way. A schema group does not impose an restrictions on the
```
duglin, line 1959: s/an/any

registry/spec.md
```diff
@@ -1498,71 +1630,1341 @@ An attempt to delete all versions MUST generate an error.
...
+clarity, but MUST be identical. `schemaversion` has the following extension
+attributes.
+
+##### `schema`
+
+- Type: String | Object
+- Description: Embedded schema string or object. The format and encoding of the
+  schema is defined by the `format` attribute.
+- Constraints:
+  - Mutually exlusive with `schemaurl`. One of the two MUST be present.
```
duglin, line 2034: exlusive  -> exclusive

registry/spec.md
```diff
@@ -1498,71 +1630,1341 @@ An attempt to delete all versions MUST generate an error.
...
+
+##### `schemaurl`
+
+- Type: URI
+- Description: Reference to a schema document external to the registry.
+- Constraints:
+  - Mutually exlusive with `schemaurl`. One of the two MUST be present.
```
duglin, line 2041: exlusive  -> exclusive

registry/spec.md
```diff
@@ -1642,240 +3055,363 @@ extension attributes:
...
-- Description: Specifies the set of rules that govern how the messages for this
-  Endpoint MUST be formatted. This value can be used to definitively identify
-  the type of messages without checking the individual attributes defined by
-  the associated message definitions.
+- Description: The transport or application protocol used by the endpoint. This
+  specification defines a set of common protocol names that MUST be used for
+  respective protocol endpoints, but implementations MAY define and use
+  additional protocol names.
+
+  Predefined protocols are referred tp by name and version as
```
duglin, line 3124: tp -> to

deissnerk (commented):
I wasn't able to go through everything. Overall, this looks like a big step forward. Let's merge it and work on follow-ups.

duglin:
Approved on the 4/6 call

duglin:
@deissnerk picked up your edits in https://github.com/cloudevents/spec/pull/1184

PR #1175, opened 2023-03-13
Clarify CDATA handling in XML format

### LINKED ISSUE #1013: Support CDATA as body type for 'data' element

jskeet:
(We should definitely have a big conformance suite for this format at some point...)

The repetition of the language around XML element data is deliberate, but slightly clunky. Other suggestions welcome.

Also happy to add an example with both CDATA and a text node if that would be useful, but it feels like it's probably better in a conformance test.

This is an alternative to #1143, and I believe would fix #1013.

Initial review requested from Jem, and we can go on from there.

### DIFF
--- a/cloudevents/working-drafts/xml-format.md
+++ b/cloudevents/working-drafts/xml-format.md
@@ -67,6 +67,10 @@ CloudEvent or CloudEvent batch. Comments MUST be ignored during processing, exce
 for the child element of a `<data>` element containing [XML element data](#33-xml-element-data),
 where all nodes MUST be preserved.
 
+CDATA nodes and regular text nodes MUST be treated interchangably during processing,
+except for the child element `<data>` element containing [XML element data](#33-xml-element-data),
+where all nodes MUST be preserved.
+
 XML elements in namespaces other than `http://cloudevents.io/xmlformat/V1`, and
 XML attributes other than those described in this specification MAY appear within an XML
 representation of a CloudEvent or CloudEvent batch. These SHOULD be ignored during
@@ -155,7 +159,7 @@ Example:
 XML element data MUST be carried in an element with a defined type of `xs:any` with
 exactly one child XML element (with any mandatory namespace definitions).
 
-All XML nodes (including comments) within the child XML element MUST
+All XML nodes (including comments and CDATA nodes) within the child XML element MUST
 be preserved during processing.
 
 The `<data>` XML element MUST NOT contain any direct child text nodes with


JemDay (approved):
Thanks Jon, I think this makes sense ;-)

As you say, there's a growing need for conformance examples ..

cloudevents/working-drafts/xml-format.md
```diff
@@ -67,6 +67,10 @@ CloudEvent or CloudEvent batch. Comments MUST be ignored during processing, exce
 for the child element of a `<data>` element containing [XML element data](#33-xml-element-data),
 where all nodes MUST be preserved.
 
+CDATA nodes and regular text nodes MUST be treated interchangably during processing,
```
duglin, line 70: s/interchangably/interchangeably/ 

the word "interchangeably" is throwing me. Do you mean to say that they should be treated the same way in that they're just chunks of text? Or are you trying to say that someone could translate a CDATA node into a text node?

I guess I'm not clear on why we need to call out these two node types at all? What is the XML example that could cause confusion for a receiver?

jskeet: > Do you mean to say that they should be treated the same way in that they're just chunks of text? 

Yes.

> Or are you trying to say that someone could translate a CDATA node into a text node?

When processing the document, they should be treated in the same way, other than within the data element.

> I guess I'm not clear on why we need to call out these two node types at all? What is the XML example that could cause confusion for a receiver?

It means that this is valid, for example:

```xml
<event specversion="1.0">
    <id><![CDATA[abc]]>def</id>
    ... other CE attributes as normal ...
</event>
```

... and should be processed exactly as if it were written like this:

```xml
<event specversion="1.0">
    <id>abcdef</id>
    ... other CE attributes as normal ...
</event>
```

or

```xml
<event specversion="1.0">
    <id><![CDATA[abcdef]]></id>
    ... other CE attributes as normal ...
</event>
```


I want us to have conformance tests with examples like that, but I don't *think* that sort of example is worth including in the format spec itself.

duglin: Thanks for the clarification and that makes sense. Why did this come up as an issue rather than it just being obvious? Or, what other choice might someone make?

PR #1176, opened 2023-03-13
Modeling the registry API...

markweitzel:
Hi everyone...
I'm working on a project where, essentially, we need a registry of API specifications. Actually, there could be multiple API registries that are queried, browsed, etc. You can think of this kind of like what [apis.guru](https://apis.guru/) is trying to do, but in a more standard and programmatic way so that anyone can spin up their own API registry. My primary use case would be managing a collection of OpenAPI specifications. I'm still exploring if this spec would work for this use case and would appreciate an example of how we could use the the registry in this manner.

However, one thing I noticed is the API for the registry itself is still in the process of being defined. Already, there are some interesting choices being made, e.g. the use of `/` vs. `/registry` as the entry point, a query parameter `/?model` vs. treating the metadata as a distinct resource `/registry/metadata`. 

One thing that might be helpful is to model the API in [TypeSpec](https://microsoft.github.io/typespec/) and [typespec repo](https://github.com/Microsoft/typespec). Taking this approach would give us a clean model of the API and a way to represent this in OpenAPI. That is, OpenAPI could become the contract for the registry. This would greatly simplify the spec document and, at the same time, make the API definition very precise. 

This PR is a very quick, rough cut, of what the registry API would look like in TypeSpec.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ gen/
 gen/
 .DS_Store
 *.pyc
+/registry/typespec/node_modules
--- a/registry/typespec/main.tsp
+++ b/registry/typespec/main.tsp
@@ -1,0 +1,124 @@
+import "@typespec/http";
+import "@typespec/rest";
+import "@typespec/openapi3";
+
+using TypeSpec.Http;
+using TypeSpec.Rest;
+
+@service({
+    title: "xregistry",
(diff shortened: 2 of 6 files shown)


duglin:
Closing as it appears to have stalled. We can reopen if needed.

PR #1179, opened 2023-03-24
Specify the policy for whitespace in XML format attributes

### LINKED ISSUE #1004: XML format: do we have a whitespace policy?
For example, assuming appropriate namespaces, should we automatically preserve leading and trailing whitespace on context attributes?

```xml
<event>
  <!-- All the standard attributes etc -->
  <customattribute1 xsi:type="ce:string">    middle     </customattribute1>
  <customattribute2 xsi:type="ce:string">
    middle
  </customattribute2>
</event>
```

How does this apply to text and base64 data elements?

(Doug: please assign to me or Jem as a starting point.)

jskeet:
Fixes #1004.

I wasn't sure whether providing the parenthetical example was useful - I hope so. We could go further and explain *why* we prohibit line breaks (i.e. because no valid attribute value can ever contain a line break) but I wasn't sure whether that was worth it.

### DIFF
--- a/cloudevents/working-drafts/xml-format.md
+++ b/cloudevents/working-drafts/xml-format.md
@@ -105,7 +105,13 @@ An XML element representing a core context attribute MAY be decorated with
 an `xsi:type` XML attribute. If present, this designator MUST match that of the type specified
 by the [CloudEvent context attributes][ce-attrs].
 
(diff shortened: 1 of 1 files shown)


duglin:
Approved on the 3/30 call
@clemens to create a PR/issue to define a schema for the "ce" XSI prefix

Issue #1180, opened 2023-03-27
[xReg] add `versionsURL`  and `versionsCount` to Resource


duglin:
So people know where to find the versions.
Should also probably include `versionsCount` for consistency with `GROUPsURL` and `GROUPsCount` (and `RESOURCEsCount` and `RESOURCEsURL`).

duglin:
Discussed this on the off-line 3/28 call - and there was agreement to add them

PR #1189, opened 2023-04-10
Add versionsXXX fields and other misc tweaks

duglin:
- Added versionsUrl, versionsCount fields
- removed "type"
- s/URL/Url/ where appropriate
- "versions" in model MUST be >= 1 - meaning there will always be at least 1 version of a resource
- added missing fields in some samples
- wrap at 80 cols
- `id` must be case-insensitive unique since it can appear in case-insensitive locations, such as URLs or HTTP headers

Fixes: #1180


duglin:
Approved on the 4/13 call

PR #1191, opened 2023-04-12
Prevent CE-namespaced elements other than event in an XML batch.

jskeet:
Explanation: we view the CE namespace as "owned" by this spec, so while we allow for non-CE-namespaced elements in the batch to be present and ignored, there shouldn't be any elements in the CE namespace other than "event". (We don't permit nested batches, for example.)

Conformance tests in https://github.com/cloudevents/conformance/blob/format-tests/format/xml/invalid-batches.xml#L37

(Also a typo fix.)


duglin:
Approved on the 4/13 call

PR #1193, opened 2023-04-12
xReg - Lots of changes

duglin:
- Add support for ?model
- Add a Schema URL for the entire registry as part of the model
- Elaborate on ?inline=PATH
- s/version/versionID/, and remove "version" field from Version resource since
      `id` is now its version string
- fixed the spec verification tool so it doesn't flag optional query parmeters
      as markdown bookmarks (e.g. [?inline])
- define the 3 properties for each collection (xxxsUrl, xxxsCount, xxxs)

### DIFF
--- a/registry/spec.md
+++ b/registry/spec.md
@@ -15,7 +15,8 @@ automation and tooling.
 - [Registry Formats and APIs](#registry-formats-and-apis)
   - [Attributes and Extensions](#attributes-and-extensions)
   - [Registry APIs](#registry-apis)
-    - [Retrieving the Registry Model](#retrieving-the-registry-model)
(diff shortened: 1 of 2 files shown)


duglin:
PR #1189 should be merged first.

duglin:
Make sure a new PR lowercases "ID"->"Id"

duglin:
@jskeet the "ID" you were noticing should be "ID" not "Id" because it's not an attribute name, it's a variable name - even when it appears as an attribute name :-) Much like GROUP and RESOURCE in the spec. So think we're ok.

duglin:
approved on the 4/13 call

PR #1197, opened 2023-04-27
feat: add `avro-turbo` format

### LINKED ISSUE #1187: Avro
We want to use CloudEvents in high-performance settings. We do not want to use JSON or other verbose encoding. Instead we want to use compact encoding, e.g Protobuf or Avro.

What is the status of Avro? I see Protobuf support. It seems like it has not receive much TLC. That leaves me to believe that it is either unpopular, or stable.

We want to use Avro rather than Protobuf. We already use Avro, using Protobuf would complicate our tech stack.

This has already been implemented (in Java) by @fabiojose. 

https://github.com/kattlo/cloudevents-kafka-avro-serializer

alexec:
Fixes #1187

### DIFF
--- a/cloudevents/documented-extensions.md
+++ b/cloudevents/documented-extensions.md
@@ -42,6 +42,7 @@ for more information.
 
 - [Dataref (Claim Check Pattern)](extensions/dataref.md)
 - [Distributed Tracing](extensions/distributed-tracing.md)
+- [Expiry Time](extensions/expirytime.md)
 - [Partitioning](extensions/partitioning.md)
 - [Recorded Time](extensions/recordedtime.md)
 - [Sampling](extensions/sampledrate.md)
--- a/cloudevents/extensions/expirytime.md
+++ b/cloudevents/extensions/expirytime.md
@@ -1,0 +1,57 @@
+# Expiry Time Extension
+
+This extension provides a mechanism to hint to [consumers](../spec.md#consumer)
+or [intermediaries](../spec.md#intermediary) a timestamp after which an
+[event](../spec.md#event) can be ignored.
+
+In distributed systems with message delivery guarantees, events might be delivered
+to a consumer some significant amount of time after an event has been sent.
+In this situation, it might be desirable to ignore events that
+are no longer relevant. The [`time` attribute](../spec.md#time) could be used
+to handle this on the consumer side but can be tricky if the logic varies
+depending on the event type or producer.
+
+## Attributes
+
+### expirytime
+
+- Type: `Timestamp`
+- Description: Timestamp indicating an event is no longer useful after the
+  indicated time.
+- Constraints:
+  - REQUIRED
+  - SHOULD be equal to or later than the `time` attribute, if present
+
+## Usage
+
+When this extension is used, producers MUST set the value of the `expirytime`
+attribute.
+
+Intermediaries and consumers MAY ignore and discard an event that has an
+`expirytime` at or before the current timestamp at the time of any checks.
+Any system that directly or indirectly interacts with a consumer SHOULD NOT
+make any assumptions on whether a consumer will
+keep or discard an event based on this extension alone. The reasoning for this
+is that time-keeping can be inaccurate between any two given systems.
+
+Intermediaries MAY modify the `expirytime` attribute, however, they MUST NOT
+remove it.
+
+## Potential Scenarios
+
+### Web dashboard for sensors
+
+A series of sensors produce CloudEvents at regular intervals that vary per
+sensor. Each sensor can pick an `expirytime` that suits its configured sample
+rate. In the event that an intermediary delays delivery of events to a
+consumer, older events can be skipped to avoid excessive processing or UI
+updates upon resuming delivery.
+
+### Jobs triggered by Continuous Integration
+
+A Continuous Integration (CI) system uses CloudEvents to delegate a job to a
+runner machine. The job has a set deadline and needs to complete before that time
+has elapsed to be considered successful. The CI system can set the
+`expirytime` to match the deadline. The job runner would ignore/reject the job
+if the `expirytime` has elapsed since the CI might have likely already determined
+the job state.
--- a/cloudevents/languages/he/extensions/expirytime.md
+++ b/cloudevents/languages/he/extensions/expirytime.md
@@ -1,0 +1,2 @@
+# Expiry Time Extension
+מסמך זה טרם תורגם. בבקשה תשתמשו [בגרסה האנגלית של המסמך](../../../extensions/expirytime.md) לבינתיים.
--- a/cloudevents/languages/zh-CN/documented-extensions.md
+++ b/cloudevents/languages/zh-CN/documented-extensions.md
@@ -16,6 +16,8 @@
 
 - [Dataref (Claim Check Pattern)](../../extensions/dataref.md)
 - [Distributed Tracing](../../extensions/distributed-tracing.md)
+- [Expiry Time](../../extensions/expirytime.md)
 - [Partitioning](../../extensions/partitioning.md)
+- [Recorded Time](../../extensions/recordedtime.md)
 - [Sampling](../../extensions/sampledrate.md)
 - [Sequence](../../extensions/sequence.md)
--- a/cloudevents/languages/zh-CN/extensions/expirytime.md
+++ b/cloudevents/languages/zh-CN/extensions/expirytime.md
@@ -1,0 +1,6 @@
+# Expiry Time Extension
+
+本文档尚未被翻译，请先阅读英文[原版文档](../../../extensions/expirytime.md) 。
+
+如果您迫切地需要此文档的中文翻译，请[提交一个issue](https://github.com/cloudevents/spec/issues) ，
+我们会尽快安排专人进行翻译。


alexec:
I’ve raised this PR as a modification to the Avro format so it is clear how it is different. Not mergable in current state

alexec:
@JemDay @doug for review.

duglin:
Did we want to rename the existing one or create an alternative?

duglin:
Looks like the errors are real: https://github.com/cloudevents/spec/actions/runs/4824766723/jobs/8672389773?pr=1197

cloudevents/formats/avro-turbo-format.md
```diff
@@ -77,59 +83,80 @@ described by the [CloudEvent Avro Schema](cloudevents.avsc):
...
-      }
+        "values": [
+          "null",
+          "boolean",
+          "int",
+          "string",
+          "bytes"
+        ]
+      },
+        "default": {}
```
duglin, line 151: think the spacing is off here, needs to go left 2 spaces

duglin:
@alexec if you recreate the original file, add the new one to the README and address the comment above I think we should be able to get this one in on tomorrow's ca..

alexec:
> @alexec if you recreate the original file, add the new one to the README and address the comment above I think we should be able to get this one in on tomorrow's ca..

done

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,207 @@
...
+Example:
+
+```json
+["null", "string"]
+```
+
+### 2.4 Definition
+
+Users of Avro MUST use a message whose binary encoding is identical to the one
+described by the [CloudEvent Avro Schema](cloudevents.avsc):
```
duglin, line 85: Not knowing Avro I don't know if this is silly question or not, but.... in the other avro-format.md file it points to an avsc file looks to be a copy of the json that's in the avro-format.md file. So, for Turbo, do we need a Turbo version of the avsc file?

README.md
```diff
@@ -35,7 +35,8 @@ incubator project on [Oct 24, 2019](https://github.com/cncf/toc/pull/297).
 |                               |
 | **Optional Specifications:**  |
 | AMQP Protocol Binding         | [v1.0.2](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/bindings/amqp-protocol-binding.md)  | [WIP](cloudevents/bindings/amqp-protocol-binding.md)       |
-| AVRO Event Format             | [v1.0.2](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/formats/avro-format.md)             | [WIP](cloudevents/formats/avro-format.md)                  |
+| AVRO Event Format             | [v1.0.2](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/formats/avro-format.md)             | [WIP](cloudevents/formats/avro-format.md)     
+| AVRO Turbo Event Format       | [v1.0.2](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/formats/avro-turbo-format.md)             | [WIP](cloudevents/formats/avro-turbo-format.md)             |
```
duglin, line 39: I think we need to leave the first column blank until we release it - the link checker isn't happy.
I guess a question is whether we bump everything to 1.0.3 to just release it under 1.0.2. Let's discuss on this week's call.

cloudevents/SDK.md
```diff
@@ -173,6 +173,7 @@ Undo it when done:
 | [CloudEvents Core](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md)                                                       | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark:  | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark:    | :heavy_check_mark:  | :heavy_check_mark:  |
 | Event Formats |
 | [Avro](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/formats/avro-format.md)                                                    | :heavy_check_mark: |     | :x:  | :x: |     |     |        |      | :x:  |
+| [Avro Turbo](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/formats/avro-turbo-format.md)                                                    | :x: |     | :x:  | :x: |     |     |        |      | :x:  |
```
duglin, line 176: same here - we may need to point to the HEAD version for now - so a relative link would be ok for the PR.

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,207 @@
+# Avro Turbo Event Format for CloudEvents - Version 1.0.3-wip
+
+## Abstract
+
+The Avro Turbo Format for CloudEvents defines how events attributes are expressed in
+the [Avro 1.9.0 Specification][avro-spec].
+
+This differs from the [Avro format](avro-formad.md) in that:
```
duglin, line 8: s/formad/format/

duglin:
Link checker:
```
cloudevents/formats/avro-turbo-format.md: Translation file cloudevents/languages/zh-CN/formats/avro-turbo-format.md does not exist
cloudevents/formats/avro-turbo-format.md: Translation file cloudevents/languages/he/formats/avro-turbo-format.md does not exist
```

I think you can just create those 2 files and copy in the content from this file:
https://github.com/cloudevents/spec/blob/main/cloudevents/languages/zh-CN/formats/protobuf-format.md
and just change the English title. Then things should be good.

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,207 @@
...
+}
+```
+
+## 3 Data
+
+Before encoding, the AVRO serializer MUST first determine the runtime data type
+of the content. This can be determined by examining the data for invalid UTF-8
+sequences or by consulting the `datacontenttype` attribute.
+
+If the implementation determines that the type of the data is binary, the value
```
JemDay, line 174: Is this redundant since the definition only seems to allow 'bytes' (or null).

JemDay (changes requested):

JemDay:
One thing that's missing from this and the existing Avro format spec is the statement about the media-type to use for structured-mode.

My assumption is that that _application/cloudevents+avro_ is used for the existing format spec, do we need a different media-type for turbo mode ?

jskeet:
I haven't looked at this yet, and won't have time before the meeting, but will do so tomorrow or next week. Prepare for nitpicking :)

cloudevents/formats/avro-turbo-format.md
alexec: good point

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,207 @@
...
+    {
+      "name": "attribute",
+      "type": {
+        "type": "map",
+        "values": [
+          "null",
+          "boolean",
+          "int",
+          "string",
+          "bytes"
```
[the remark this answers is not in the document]
alexec: this did not exist in the non-turbo version, but yes

alexec:
`application/cloudevents+avro-tubro`?

duglin:
@alexec couple of more translation files needed:
cloudevents/formats/avro-turbo-format.md: Translation file cloudevents/languages/zh-CN/formats/avro-turbo-format.md does not exist
cloudevents/formats/avro-turbo-format.md: Translation file cloudevents/languages/he/formats/avro-turbo-format.md does not exist
cloudevents/languages/zh-CN/formats/avro-tubro-format.md: cloudevents/languages/zh-CN/formats/../../../formats/avro-tubro-format.md does not exist

cloudevents/languages/he/formats/avro-tubro-format.md
```diff
@@ -0,0 +1,2 @@
+# Avro Turbo Event Format for CloudEvents - Version 1.0.3-wip
```
duglin, line 1: filename type: tubro instead of turbo
same for the zh-CN one.

cloudevents/languages/he/formats/avro-tubro-format.md
```diff
@@ -0,0 +1,2 @@
...
+מסמך זה טרם תורגם. בבקשה תשתמשו [בגרסה האנגלית של המסמך](../../../formats/avro-turbo-format.md) לבינתיים.
```
duglin, line 2: I think all of the ..'s are off, but not sure. The checker says:
```
cloudevents/languages/zh-CN/formats/avro-tubro-format.md: cloudevents/languages/zh-CN/formats/../../../formats/avro-tubro-format.md does not exist
```
you can check it yourself by running `make -f tools/Makefile docker` from the root of the repo

cloudevents/languages/he/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,2 @@
+# Avro Turbo Event Format for CloudEvents - Version 1.0.3-wip
+מסמך זה טרם תורגם. בבקשה תשתמשו [בגרסה האנגלית של המסמך](../../../formats/avro-turbo-format.md) לבינתיים.
```
duglin, line 2: I think this has one too many ".."s

cloudevents/languages/zh-CN/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,6 @@
+# Avro Turbo Event Format for CloudEvents - Version 1.0.3-wip
+
+本文档尚未被翻译，请先阅读英文[原版文档](../../../formats/avro-tubro-format.md) 。
```
duglin, line 3: one to many ".."s  I think

duglin, line 3: tubro - typo

alexec: Maybe it should be called tubro?

duglin: your figures do seem to want to do that :-)

duglin:
ping @JemDay @jskeet for review

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,217 @@
...
+Before encoding, the AVRO serializer MUST first determine the runtime data type
+of the content. This can be determined by examining the data for invalid UTF-8
+sequences or by consulting the `datacontenttype` attribute.
+
+## 4 Transport
+
+Transports that support content identification MUST use the following designation:
+
+```text
+ application/cloudevents+avro-turbo
```
JemDay, line 184: I"m assuming we need to do this because avro doesn't (as far as I can tell) encode the namespace into the serialized representation ?

There may be some pushback on using a non-registered suffix, but given I don't think application/cloudevents is registered either it may not be an issue ...

JemDay (changes requested):

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,217 @@
...
+      ],
+      "default": "null"
+    }
+  ]
+}
+```
+
+## 3 Data
+
+Before encoding, the AVRO serializer MUST first determine the runtime data type
```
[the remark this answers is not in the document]
duglin: If @JemDay is correct, perhaps just say that the content MUST be serialized as a series of bytes ??
I could see an argument for removing the entire section, but "data" is kind of special so it might be nice to call it out.

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,217 @@
...
+
+| CloudEvents | Type   | Exemplary Avro Value                           |
+| ----------- | ------ | ---------------------------------------------- |
+| type        | string | `"com.example.someevent"`                      |
+| specversion | N/A    | Spec version is always `1.0`.                  |
+| source      | string | `"/mycontext"`                                 |
+| id          | string | `"7a0dc520-c870-4193c8"`                       |
+| time        | long   | `1234`                                         |
+| dataschema  | string | `"http://registry.com/schema/v1/much.json"`    |
+| contenttype | string | `"application/json"`                           |
```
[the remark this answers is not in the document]
alexec: Copy and paste!

jskeet:
I'm planning on reviewing this, but it's likely to be tomorrow - today I really want to create the issue around formats and data, and that'll take quite a lot of my time.

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,217 @@
+# Avro Turbo Event Format for CloudEvents - Version 1.0.3-wip
+
+## Abstract
+
+The Avro Turbo Format for CloudEvents defines how events attributes are expressed in
```
jskeet, line 5: I think this paragraph could do with rewriting - both here and in the original format - given that it's not just about attributes. It's how the whole event is represented in Avro, right? I'd also say it defines *one way* that CEs can be represented in Avro. Otherwise we've got two formats that say "This format defines how CEs are represented in Avro" and they can't both do that definitively.

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,217 @@
...
+the [Avro 1.9.0 Specification][avro-spec].
+
+This differs from the [Avro format](avro-format.md) in that:
+
+- It is optimized for performance, preferring a more compact representation.
+- Only supports spec version 1.0 (any changes to spec version, requires 
+  changes to Avro schema, which changes the finger-print, 
+  breaking compatibility).
```
jskeet, line 13:
```suggestion
- It only supports spec version 1.0 (any changes to spec version requires changes to the Avro schema,
  which changes the fingerprint, breaking compatibility).
```

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,217 @@
...
+- Does not natively support JSON (JSON can be straight-forwardly serialized 
+  to bytes and this was therefore not considered neccessary).
```
jskeet, line 15:
```suggestion
- It does not natively support JSON (JSON can be straight-forwardly serialized 
  to bytes and this was therefore not considered necessary).
```

Beyond the consistency/typo changes, it's not clear to me what "natively supporting JSON" means here. Perhaps describe what the original Avro format allows? (I assume this is related to `doc`, but I'm not sufficiently versed in Avro to really understand that in detail.)

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,217 @@
...
+1. [Introduction](#1-introduction)
+2. [Attributes](#2-attributes)
+3. [Data](#3-data)
+4. [Transport](#4-transport)
+5. [Examples](#5-examples)
+
+## 1. Introduction
+
+[CloudEvents][ce] is a standardized and protocol-agnostic definition of the
+structure and metadata description of events. This specification defines how the
```
jskeet, line 28: Again, this needs to be less definitive if there are two alternative formats.

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,217 @@
...
+| String        | [string][avro-primitives]                                              |
+| Binary        | [bytes][avro-primitives]                                               |
+| URI           | [string][avro-primitives] following [RFC 3986 §4.3][rfc3986-section43] |
+| URI-reference | [string][avro-primitives] following [RFC 3986 §4.1][rfc3986-section41] |
+| Timestamp     | [long][avro-primitives]                                                |
+
+Extension specifications MAY define secondary mapping rules for the values of
+attributes they define, but MUST also include the previously defined primary
+mapping.
+
```
jskeet, line 70: We appear to be missing 2.2...

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,217 @@
...
+### 2.3 OPTIONAL Attributes
+
+CloudEvents Spec defines OPTIONAL attributes. The Avro format defines that these
```
jskeet, line 73: Would someone who understands Avro know what this means? I don't understand this at all. That may be fine though...

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,217 @@
...
+fields MUST use the `null` type and the actual type through the
+[union][avro-unions].
+
+Example:
+
+```json
+["null", "string"]
+```
+
```
jskeet, line 82: What about extension attributes?

cloudevents/formats/avro-turbo-format.md
jskeet: And does this need to say "The Avro Turbo format"? (It's probably worth checking for "Avro format" throughout the doc.)

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,217 @@
...
+| CloudEvents   | Avro                                                                   |
+| ------------- | ---------------------------------------------------------------------- |
+| Boolean       | [boolean][avro-primitives]                                             |
+| Integer       | [int][avro-primitives]                                                 |
+| String        | [string][avro-primitives]                                              |
+| Binary        | [bytes][avro-primitives]                                               |
+| URI           | [string][avro-primitives] following [RFC 3986 §4.3][rfc3986-section43] |
+| URI-reference | [string][avro-primitives] following [RFC 3986 §4.1][rfc3986-section41] |
+| Timestamp     | [long][avro-primitives]                                                |
```
jskeet, line 65: I'd note here that this restricts the precision to milliseconds. (Assuming I've read things correctly later.) We also need to say somewhere that this is the number of milliseconds since 1970-01-01T00:00:00Z. (I'm okay with handwaving between UTC and TAI...)

cloudevents/formats/avro-turbo-format.md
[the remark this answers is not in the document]
jskeet: It looks like this is basically the first half of the equivalent section in the original Avro format.
Both documents probably need more work when we've reached some conclusions on #1204.

(I don't think we can remove the entire section - we've got to say *something* about data. It would be reasonable to say that the payload should be stored in the data field in the same way that it would be encoded to bytes in binary mode, perhaps.)

cloudevents/formats/avro-turbo-format.md
jskeet: I can't find details in RFC 2046 about what the use of "+" is intended to mean in a subtype...

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,217 @@
...
+```
+
+## 5 Examples
+
+The following table shows exemplary mappings:
+
+| CloudEvents | Type   | Exemplary Avro Value                           |
+| ----------- | ------ | ---------------------------------------------- |
+| type        | string | `"com.example.someevent"`                      |
+| specversion | N/A    | Spec version is always `1.0`.                  |
```
jskeet, line 194: If this isn't specified anywhere at all within the body, should it be in the content type, e.g. `application/cloudevents+avro-turbo-1.0`? If not, what do we expect a CE-2.0 version of this format to look like?

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,217 @@
...
+| source      | string | `"/mycontext"`                                 |
+| id          | string | `"7a0dc520-c870-4193c8"`                       |
+| time        | long   | `1234`                                         |
```
jskeet, line 197: An example value in the 21st century with the RFC3339 equivalent would probably be useful here.

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,217 @@
...
+| dataschema  | string | `"http://registry.com/schema/v1/much.json"`    |
+| contenttype | string | `"application/json"`                           |
+| data        | bytes  | `"{"much":{"wow":"json"}}"`                    |
```
jskeet, line 200: I suspect the use of a string here could cause confusion, particularly with JSON inside it. Possibly just say "Arbitrary bytes"?

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,217 @@
...
+
+Transports that support content identification MUST use the following designation:
+
+```text
+ application/cloudevents+avro-turbo
+```
+
+## 5 Examples
+
+The following table shows exemplary mappings:
```
jskeet, line 189: Having one table with multiple examples in is a bit confusing IMO. It gives the impression that you can include data twice, for example.

cloudevents/formats/avro-turbo-format.md
jskeet: Reading further, it feels like it's more than this - my possibly-hazy understanding is that the original format supports multiple data representations - *including* (but not limited to) JSON maps. (For example, I *think* you could just specify a string.) Whereas the Avro Turbo format requires that data is represented as a sequence of bytes - that's the only representation available. Is that a correct reading?

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,217 @@
...
+```json
+["null", "string"]
+```
+
+### 2.4 Definition
+
+Users of Avro MUST use a message whose binary encoding is identical to the one
+described by the [CloudEvent Avro Turbo Schema](cloudevents-turbo.avsc):
+
+```json
```
jskeet, line 88: Do we need to inline the file? That feels fragile to me in terms of the two schemas getting out of sync.

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,217 @@
...
+        "null",
+        {
+          "type": "long",
+          "logicalType": "timestamp-millis"
+        }
+      ],
+      "default": null
+    },
+    {
+      "name": "attribute",
```
jskeet, line 144: I assume this is provided for extension attributes. Does this mean we can't differentiate between string, URI and URI-ref attributes? This was unfortunate in JSON (and, I suspect, in the original Avro format) - if there's a way of including the type unambiguously when an extension attribute is specified, that would be great.

alexec:
@jskeet most of this doc is copied from the original and updated. Whenever you suggest a change, in many cases it must mean you want the original changed too. 

But that can’t be true, because if that was needed you would commented on the original. 

Please advise.

jskeet:
> But that can’t be true, because if that was needed you would commented on the original.

I've never gone through the original carefully, as it was already in place when I became interested in CloudEvents, and hasn't come up much in discussions. There may well be many things I'd have raised before it was approved, but which can't be changed now - but which *can* be fixed in this new format.

alexec:
> > But that can’t be true, because if that was needed you would commented on the original.
> 
> I've never gone through the original carefully, as it was already in place when I became interested in CloudEvents, and hasn't come up much in discussions. There may well be many things I'd have raised before it was approved, but which can't be changed now - but which _can_ be fixed in this new format.

Would you like to suggest changes by way of Github suggestions or a pull request? Do you want to do that to the original version?

jskeet:
> Would you like to suggest changes by way of Github suggestions or a pull request?

I don't have time to do that at the moment, I'm afraid.

> Do you want to do that to the original version?

No, because the original format has already been approved and implemented - a lot of the changes I'm suggesting here *can't* be retrofitted because they'd be breaking changes. (Some changes to wording could be, but I don't have time to go through it right now.)

> To be frank, I don’t really want to fix these issues because I want to move on so writing the Java SDK, and so I don’t want to spend another three weeks on rewriting this pull request.

I would hope that the issues I've raised wouldn't actually take that long - and I'd also hope that fixing them now would save time in the long run.

duglin:
@alexec let me know when you think this one is ready (all comments have been addressed) and then we can add it to the agenda for a vote.

duglin:
@alexec rebase needed. Did you want to get this one in soon?

cloudevents/formats/avro-turbo-format.md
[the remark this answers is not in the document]
alexec: deleting

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,217 @@
...
+| type        | string | `"com.example.someevent"`                      |
+| specversion | N/A    | Spec version is always `1.0`.                  |
+| source      | string | `"/mycontext"`                                 |
+| id          | string | `"7a0dc520-c870-4193c8"`                       |
+| time        | long   | `1234`                                         |
+| dataschema  | string | `"http://registry.com/schema/v1/much.json"`    |
+| contenttype | string | `"application/json"`                           |
+| data        | bytes  | `"{"much":{"wow":"json"}}"`                    |
+| dataschema  | string | `"http://registry.com/subjects/ce/versions/1"` |
+| contenttype | string | `"application/avro"`                           |
```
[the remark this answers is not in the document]
alexec: fixing

cloudevents/formats/avro-turbo-format.md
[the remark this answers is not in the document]
alexec: fixing

cloudevents/formats/avro-turbo-format.md
alexec: I'm assuming this is not an issue, because it is widely used in CE.

cloudevents/formats/avro-turbo-format.md
alexec: fixing

cloudevents/formats/avro-turbo-format.md
alexec: fixing

cloudevents/formats/avro-turbo-format.md
alexec: this ALWAYS happens with numbered heading is Markdown, annoying, would not number headings myself, fixing

cloudevents/formats/avro-turbo-format.md
alexec: fixing

cloudevents/formats/avro-turbo-format.md
alexec: I've unwrapped the string, but yes

cloudevents/formats/avro-turbo-format.md
alexec: fixing

cloudevents/formats/avro-turbo-format.md
alexec: I'm not a fan of this, I already make mistakes

alexec: prior-art

cloudevents/formats/avro-turbo-format.md
alexec: looking...

alexec: URI not supported

alexec: https://avro.apache.org/docs/1.8.1/spec.html#Logical+Types

cloudevents/formats/avro-turbo-format.md
alexec: I think this section is not need, this is just how Avro works.

cloudevents/formats/avro-turbo-format.md
alexec: > The smallest time unit that RFC3339 can support is one nanosecond (1 ns).

alexec: In practical cases, milli seconds will be enough. However, IHMO we should not truncate time to save space.

alexec: > a nanosecond is smaller than a microsecond.

> 1 microsecond (µs) = 1000 nanoseconds (ns)

alexec: > A long cannot represent a nanosecond.

alexec:
Had to create new PR due to Git-spolsion.

PR #1201, opened 2023-05-04
feat: add the `recordedtime` extension attribute

### LINKED ISSUE #1200: Add a known extension attribute to support bitemporal event data
In my current and previous role, where I have worked on analytical data platforms, I have had a need for bitemporal event data [[ref](https://en.wikipedia.org/wiki/Bitemporal_modeling)]. For every event, I want to know the best approximation of when the event actually occurred (_actual time_), together with when the event was recorded in a specific message or representation (_recorded time_).

## Example

For a payment event, for example, I want to know the best approximation of the actual time when a person made the payment, and the time that the CloudEvent representing the payment was created.

If there were errors in the initial CloudEvent representation of a specific payment, or the schema of the event data changed, or the event had to be redacted, a new amended CloudEvent describing the same payment would be created with the same ID and the same _actual time_ as the previous CloudEvent, but the amended event would have a different _recorded time_.

## Benefits

Bitemporal event data makes it possible to do get reproducible inputs for data science, which in turn makes it possible to then do reproducible training and make reproducible predictions.

Knowing when an event was recorded also makes it possible to capture events into time interval blobs and to measure latency. If the actual event time is the only time associated with events, then consumers can't differentiate between an event that was republished long after the actual event time, and an event which just took a long time to propagate.

The recorded time also makes it easier to differentiate republished messages in an analytical data warehouse.

## Proposal

In my previous role, I used something similar to `orgexamplerecordedtime` for the attribute with the recorded time, but I would prefer to just use `recordedtime`. However, I think for that, it would be appropriate to register the attribute name in the CloudEvents specification repository as a known extension.

I will make a pull request to add this attribute, but meanwhile I would like to get feedback on the proposal.

aucampia:
This extension defines an attribute that represents the time when an occurrence was recorded in a particular event, i.e. the time when the CloudEvent was created by a producer.

This enables bitemporal data to be represented with CloudEvents which enables several use cases described in the specification of the extension attribute.

Fixes #1200

## Proposed Changes

- Add the `recordedtime` extension attribute.

**Release Note**

Extension attributes are not tracked in the release notes as far as I know, so I'm eliding this. If it is needed let me know.

### DIFF
--- a/cloudevents/extensions/recordedtime.md
+++ b/cloudevents/extensions/recordedtime.md
@@ -1,0 +1,83 @@
+# Recorded Time Extension
+
+This extension defines an attribute that represents the time when an
+[_occurrence_](../spec.md#occurrence)
+was recorded in a particular
+[_event_](../spec.md#event),
+which is the time when the CloudEvent was created by a producer.
+
+This attribute is distinct from the [`time`
+attribute](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#time),
+which, according to the CloudEvents specification, SHOULD be the time when the
+occurrence happened, if it can be determined.
+
+This attribute makes it possible to represent
+[bitemporal](https://en.wikipedia.org/wiki/Bitemporal_modeling) data with
+CloudEvents so that, for every event, both of the following times can be known
+by consumers:
+
+- _Occurrence time_: timestamp of when the occurrence recorded in the event
+  happened, which corresponds to the `time` attribute.
+- _Recorded time_: the timestamp of when the occurrence was recorded in a
+  specific CloudEvent instance, which is represented by this extension.
+
+## Attributes
+
+### recordedtime
+
+- Type: `Timestamp`
+- Description: Timestamp of when the occurrence was recorded in this CloudEvent,
+  i.e. when the CloudEvent was created by a producer.
+- Constraints:
+  - OPTIONAL
+  - If present, MUST adhere to the format specified in
+    [RFC 3339](https://tools.ietf.org/html/rfc3339)
+
+## Usage
+
+When this extension is used, producers MUST set the value of the `recordedtime`
+attribute to the timestamp of when they create the owning CloudEvent.
+
+If the same occurrence MUST be recorded differently, or the event data or
+attributes of a previous record of the occurrence MUST be amended or redacted,
+then the new CloudEvent with the necessary changes SHOULD have a different
+`recordedtime` attribute value than the previous record of the occurrence.
+
+Intermediaries MUST NOT change the value of the `recordedtime` attribute.
+
+## Use cases
+
+Examples of when an occurrence might need to be recorded differently are:
+
+- When incompatible changes to the event data schema are made, and there are
+  systems that can only process the new schema.
+- When a previous record contains incorrect information.
+- When a previous record contains personal information that can no longer be
+  kept because of regulatory or statutory reasons and needs to be redacted.
+
+Having bitemporal data makes it easier to get reproducible datasets for
+analytics and data science, as the datasets can be created by placing
+constraints on both the `time` and `recordedtime` attributes of events.
+
+Knowing when an occurrence was recorded in a particular event was recorded also
+makes it possible to determine latency between event producers and consumers. It
+also makes it possible to do operations which are sensitive to the time when an
+event was recorded, such as capturing events into time-intervalled files.
+
+The recorded time also makes it easier to differentiate different records of the
+same occurrence in analytical data stores.
+
+## Terminology
+
+This specification uses the [terminology from the CloudEvents core specification](../spec.md#terminology).
+Additionally, this specification introduces the following terms:
+
+### Occurrence time
+
+The timestamp of when the occurrence happened.
+
+### Recorded time
+
+The timestamp of when the occurrence was recorded in a specific CloudEvent. A
+single occurrence might result in more than one event, and each event might have a
+different recorded time.
--- a/cloudevents/extensions/recordedtime.md
+++ b/cloudevents/extensions/recordedtime.md
@@ -29,7 +29,7 @@ by consumers:
 - Description: Timestamp of when the occurrence was recorded in this CloudEvent,
   i.e. when the CloudEvent was created by a producer.
 - Constraints:
-  - OPTIONAL
+  - REQUIRED
   - If present, MUST adhere to the format specified in
     [RFC 3339](https://tools.ietf.org/html/rfc3339)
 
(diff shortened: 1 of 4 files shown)


aucampia:
This is an early draft and I'm happy to make substantial changes to it.

cloudevents/extensions/recordedtime.md
```diff
@@ -0,0 +1,106 @@
...
+- _Recorded time_: the timestamp of when the occurrence was recorded in a
+  specific CloudEvent instance, which is represented by this extension.
+
+## Attributes
+
+### recordedtime
+
+- Type: `Timestamp`
+- Description: Timestamp of when the occurrence was recorded in this CloudEvent,
+  i.e. when the CloudEvent was created by a producer. If this attribute is used
```
duglin, line 30: I think the 2nd sentence is unneeded and could be misleading because it implies that if `recordedtime` isn't used then `time` might mean something different. Or are you just trying to say: "If this attribute is used then `time` SHOULD be present in the CloudEvent"  ?? I think that would be ok.

cloudevents/extensions/recordedtime.md
```diff
@@ -0,0 +1,106 @@
...
+  best approximation thereof.
+- Constraints:
+  - OPTIONAL
+  - If present, MUST adhere to the format specified in
+    [RFC 3339](https://tools.ietf.org/html/rfc3339)
+
+## Usage
+
+When this extension is used, producers MUST set the value of the `recordedtime`
+attribute to the timestamp of when they create a new CloudEvent.
```
duglin, line 41: s/create a new/created the owning/

Something about the phrase "create a new" that feels like it could be interpreted as "some other CE rather the one with the recordedtime".

cloudevents/extensions/recordedtime.md
```diff
@@ -0,0 +1,106 @@
...
+If the same occurrence MUST be recorded differently, or the event data or
+attributes of a previous record of the occurrence MUST be amended or redacted,
+then the new CloudEvent with the necessary changes SHOULD have a different
+`recordedtime` attribute value than the previous record of the occurrence.
+
+Intermediaries MUST NOT change the value of the `recordedtime` attribute.
+
+## Use cases
+
+Examples of when an occurrence may need to be recorded differently are:
```
duglin, line 52: s/may/might/ since the "may" here isn't meant to be normative.

cloudevents/extensions/recordedtime.md
```diff
@@ -0,0 +1,106 @@
...
+
+Knowing when an occurrence was recorded in a particular event was recorded also
+makes it possible to determine latency between event producers and consumers. It
+also makes it possible to do operations which are sensitive to the time when an
+event was recorded, such as capturing events into time-intervalled files.
+
+The recorded time also makes it easier to differentiate different records of the
+same occurrence in analytical data stores.
+
+## Terminology
```
duglin, line 72: I'd prefer if we didn't duplicate these terms here because then we need to worry about them getting out of sync.

duglin:
From the spec checker:
```
cloudevents/extensions/recordedtime.md: cloudevents/extensions/recordedtime.md does not contain '#event-data' segment
cloudevents/extensions/recordedtime.md: cloudevents/extensions/recordedtime.md does not contain '#context' segment
cloudevents/extensions/recordedtime.md: line 52: 'may' MUST be capitalized ('MAY')
cloudevents/extensions/recordedtime.md: line 57: 'may' MUST be capitalized ('MAY')
cloudevents/extensions/recordedtime.md: line 58: 'should' MUST be capitalized ('SHOULD')
cloudevents/extensions/recordedtime.md: Translation file cloudevents/languages/zh-CN/extensions/recordedtime.md does not exist
cloudevents/extensions/recordedtime.md: Translation file cloudevents/languages/he/extensions/recordedtime.md does not exist
```
For the translation files errors, just use https://github.com/cloudevents/spec/blob/main/cloudevents/languages/zh-CN/extensions/sequence.md and https://github.com/cloudevents/spec/blob/main/cloudevents/languages/he/extensions/sequence.md as guides to add placeholders.

duglin:
Overall it looks nice to me - thanks

cloudevents/extensions/recordedtime.md
aucampia: The CloudEvent spec says:

> If the time of the occurrence cannot be determined then this attribute MAY be set to some other time (such as the current time) by the CloudEvents producer

The second sentence here is an attempt to negate that. However, it is probably best to omit it, it really does not add much value and could add negative value.

cloudevents/extensions/recordedtime.md
aucampia: I removed this and just link to the core spec, also happy to remove that, not entirely sure if it is needed, it is somewhat implied already.

cloudevents/extensions/recordedtime.md
aucampia: I changed this and a couple of other places where I think there were similar problems.

cloudevents/extensions/recordedtime.md
aucampia: Changed.

cloudevents/extensions/recordedtime.md
aucampia: I removed this now.

aucampia:
@duglin thanks for the review. The CI passes now, the commit is signed and signed off, and I have addressed all the comments.

Not entirely sure what is up with GitHub, but to see diff from last review, check: https://github.com/cloudevents/spec/compare/07199e0..683f65f

cloudevents/extensions/recordedtime.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Attributes
+
+### recordedtime
+
+- Type: `Timestamp`
+- Description: Timestamp of when the occurrence was recorded in this CloudEvent,
+  i.e. when the CloudEvent was created by a producer.
+- Constraints:
+  - OPTIONAL
```
duglin, line 32: Mark this as REQUIRED. Extensions in general are optional, but once an extension is used this attribute is required.

cloudevents/extensions/recordedtime.md
duglin: I don't think this section is needed at all since you already define "recordedtime" above.

duglin:
Just a few minor things, still LGTM

cloudevents/extensions/recordedtime.md
aucampia: Removing it, it does not really add much value.

cloudevents/extensions/recordedtime.md
aucampia: Removed it now.

cloudevents/extensions/recordedtime.md
aucampia: I changed it to required.

aucampia:
> Just a few minor things, still LGTM

Thanks, changes are made.

cloudevents/extensions/recordedtime.md
```diff
@@ -0,0 +1,68 @@
...
+## Attributes
+
+### recordedtime
+
+- Type: `Timestamp`
+- Description: Timestamp of when the occurrence was recorded in this CloudEvent,
+  i.e. when the CloudEvent was created by a producer.
+- Constraints:
+  - REQUIRED
+  - If present, MUST adhere to the format specified in
```
jskeet, line 33: Is there a constraint that this must be "at or later than" the occurrence time? Logically that would make sense, but if there are different clocks involved, I could imagine it not actually being strictly true in all cases.

duglin:
@aucampia just one open comment for you to address and then I think we can vote on it

cloudevents/extensions/recordedtime.md
aucampia: I would say it should be later than the occurrence time, but as you say, it may be that there are different clocks involved, and the system creating CE may only be able to trust its own clock and may not want to assert this constraint, so I would not say it must be.

I will add this to the constraints:

```
- SHOULD be at or later than the _occurrence time_.
```

aucampia:
> @aucampia can you sign all of your commits? Or squash down to one and sign that one

I signed off on it now, and I added it to my repo config, so I won't forget again.

duglin:
thanks LGTM

jskeet (approved):

duglin:
Approved on the 5/18 call

PR #1208, opened 2023-05-23
Update SDK support status for RocketMQ binding

### LINKED ISSUE #1207: Update SDK support status for RocketMQ binding
Since https://github.com/cloudevents/sdk-java/pull/554 has been merged, we need to update the status of RocketMQ in SDK.md

aaron-ai:
Fixes #1207 

## Proposed Changes

- Update SDK support status for RocketMQ binding

**Release Note**

N/A

### DIFF
--- a/cloudevents/SDK.md
+++ b/cloudevents/SDK.md
@@ -189,6 +189,8 @@ Undo it when done:
 | [NATS Structured](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/bindings/nats-protocol-binding.md)                              | :x: |     | :x:  | :x: |     |     |        | :x: | :heavy_check_mark:  |
 | [WebSockets Binary](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/bindings/websockets-protocol-binding.md)                      | :x: |     | :x:  | :heavy_check_mark: |     |     |        | :x: | :x:  |
 | [WebSockets Structured](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/bindings/websockets-protocol-binding.md)                  | :x: |     | :x:  | :heavy_check_mark: |     |     |        | :x: | :x:  |
+| Proprietary Bindings |
+| [RocketMQ](https://github.com/apache/rocketmq-externals/blob/master/rocketmq-cloudevents-binding/rocketmq-transport-binding.md)               | :x: |     | :heavy_check_mark:  | :x: |     |     |        | :x: | :x:  |
 ||
 | **[v0.3](https://github.com/cloudevents/spec/tree/v0.3)** |
 | [CloudEvents Core](https://github.com/cloudevents/spec/blob/v0.3/spec.md)                                                                     | :x: | :heavy_check_mark: | :heavy_check_mark:  | :heavy_check_mark: | :x: | :x: | :heavy_check_mark:    | :heavy_check_mark:  | :heavy_check_mark:  |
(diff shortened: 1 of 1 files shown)


duglin:
Thanks!
LGTM

jskeet:
Is the RocketMQ binding documented in this repo? (I can't see it.) I'd personally prefer to keep the SDK table to "standards we've defined in this repo". Happy to be persuaded otherwise though - or just directed at the RocketMQ page in this repo if I've missed it :)

pierDipi:
What's the reason for not mentioning it here?

CloudEvents will win when users will have the broader ecosystem support and easy access for users to underlying technologies and integrations for such technologies (including documentation and visibility).

If someone is evaluating CloudEvents and is already using or has easy access to whatever technology we integrate with, giving visibility can only help CloudEvents and the broader ecosystem.

duglin:
We chatted about this on today's call and @jskeet will add a comment about some of the options we discussed.

jskeet:
We came up with two options which would be preferable to me personally:

Option 1: Create a new section within the existing table, to clearly separate out the "CE-working-group-owned" specs from the proprietary specs.

Option 2 (preferred by me): Just add links to the [proprietary bindings/formats](https://github.com/cloudevents/spec/blob/main/cloudevents/proprietary-specs.md) page - in this case, a single sub-bullet saying "This is implemented by the Java SDK" with a link to it would be fine, IMO.

Motivation: I think it's reasonable to encourage SDK owners to implement all of the bindings/formats in the core repo, as far as they have time. But I don't want owners to feel under any pressure to implement proprietary specs.

duglin:
I'm leaning towards option 1 because I fear that users of the SDK will rarely look at the [proprietary bindings/formats](https://github.com/cloudevents/spec/blob/main/cloudevents/proprietary-specs.md) page and not get the benefit of the status we're adding, or have the benefit of (by accident) noticing that one of the SDK supports one of those "other" bindings.

jskeet:
> I'm leaning towards option 1 because I fear that users of the SDK will rarely look at the [proprietary bindings/formats](https://github.com/cloudevents/spec/blob/main/cloudevents/proprietary-specs.md) page and not get the benefit of the status we're adding, or have the benefit of (by accident) noticing that one of the SDK supports one of those "other" bindings.

How about option 2 + a node on the SDK page saying "Some SDKs support proprietary bindings as well. For more details see (link to page)"?

pierDipi:
fwiw, I'm +1 on option 1.

> Motivation: I think it's reasonable to encourage SDK owners to implement all of the bindings/formats in the core repo, as far as they have time. But I don't want owners to feel under any pressure to implement proprietary specs.

I see and I agree, this should not be forced, however, I believe we can clarify this so that there is not confusion

duglin:
@aaron-ai we chatted about this on today's call. Would be willing to change the PR to do option 1? Create a new section for "Proprietary Bindings" after the "Bindings / Content Modes" section?

aaron-ai:
> @aaron-ai we chatted about this on today's call. Would be willing to change the PR to do option 1? Create a new section for "Proprietary Bindings" after the "Bindings / Content Modes" section?

Thanks! I will update this pull request later.

aaron-ai:
@duglin PTAL

duglin:
Looks like what we discussed on today's call - thanks a ton!
ping @jskeet

jskeet (approved):
Works for me, thanks!

PR #1210, opened 2023-05-26
Avro tubro

alexec:


alexec:
@duglin @jskeet @JemDay 

Types are a challenge with Avro. The Avro type system cannot support the CE types.

# Timestapm

CE Timestamp is RFC3339. Which is nanosecond accuracy + timezone.

The  Avro types options:

1. `long` with `timestamp-micros`. This is only microsecond accuracy, no timezone. 4 bytes.
2. `string`. This can fully represent the time. 20 bytes
3. `union` of those.
4. Special `Timestamp` type. Complex.

Give the goal is to make this fast, I would chose option 1 or 3.

# URI

This is a problem with customer attributes. If you have a custom attribute you can marshal to a string, but you lose that type information. When you unmarshal you therefore get a string.

This is a pre-existing issue with the `avro` and `proto` format.

1. `string` - simple.
2. Special `URI` type - complex.

I think I would choose option 1.

jskeet:
> CE Timestamp is RFC3339. Which is nanosecond accuracy + timezone

Well, sort of. A CE timestamp has a *string encoding* that is RFC3339, but we don't actually specify required precision or whether a time zone or UTC offset is supported. (RFC3339 doesn't support an actual time zone, only a UTC offset anyway.)

My expectation - which would be good to clarify at some point in the spec - is that SDKs/intermediaries should be able to *parse* an RFC3339 value that includes a UTC offset, but don't need to be able to *preserve* all the information.

Option 1 looks good to me.

For URIs, I don't know enough about Avro to make a recommendation in terms of how we achieve the goal, but I really think there's a lot of value in preserving the type of the attribute.

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,109 @@
+# Avro Turbo Event Format for CloudEvents - Version 1.0.3-wip
+
+## Abstract
+
+The Avro Turbo Format for CloudEvents defines how events are expressed in
```
jskeet, line 5: While I don't love how both Avro formats sound definitive as if they're the only format, we do the same for Protobuf so I'm not suggesting an actual change here.

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,109 @@
...
+the [Avro 1.9.0 Specification][avro-spec].
+
+This differs from the [Avro format](avro-format.md) in that:
+
+- It is optimized for performance, preferring a more compact representation.
```
jskeet, line 10: *If* we work out a way to differentiate between string/URI/URI-ref attribute values, we should include that here too.

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,109 @@
...
+|---------------|------------------------------------------------------------------------|
+| Boolean       | [boolean][avro-primitives]                                             |
+| Integer       | [int][avro-primitives]                                                 |
+| String        | [string][avro-primitives]                                              |
+| Binary        | [bytes][avro-primitives]                                               |
+| URI           | [string][avro-primitives] following [RFC 3986 §4.3][rfc3986-section43] |
+| URI-reference | [string][avro-primitives] following [RFC 3986 §4.1][rfc3986-section41] |
+| Timestamp     | [long][avro-primitives]  using `timestamp-micros` logical type         |
+
+Extension specifications MAY define secondary mapping rules for the values of
```
jskeet, line 64: I've never understood this in any of our formats. Maybe @JemDay or @duglin could clarify... otherwise I'm happy to take an action item to remove it everywhere possible :)

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,109 @@
...
+| CloudEvents     | Type   | Exemplary Avro Value                      |
+|-----------------|--------|-------------------------------------------|
+| id              | string | `7a0dc520-c870-4193c8`                    |
+| source          | string | `https://github.com/cloudevents`          |
+| specversion     | N/A    | Spec version is always `1.0`.             |
+| type            | string | `com.example.object.deleted.v2`           |
+| datacontenttype | string | `application/octet-stream`                |
+| dataschema      | string | `http://registry.com/schema/v1/much.json` |
+| subject         | string | `mynewfile.jpg`                           |
+| time            | long   | `1685121689691`                           |
```
jskeet, line 94: I suspect this should have 3 more digits, to reflect that it's micros rather than millis.

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,109 @@
...
+
+[CloudEvents][ce] is a standardized and protocol-agnostic definition of the
+structure and metadata description of events. This specification defines how the
+CloudEvents are to be represented as [Avro 1.9.0][avro-primitives].
+
+The [Attributes](#2-attributes) section describes the naming conventions and
+data type mappings for CloudEvents attributes for use as Avro message
+properties.
+
+This specification does not define an envelope format. The Avro type system's
```
jskeet, line 33: I'm unclear on exactly what this means. (Isn't it still a wrapper around the data, surfacing the metadata? Isn't that sort of an envelope?)

cloudevents/formats/cloudevents-turbo.avsc
```diff
@@ -0,0 +1,79 @@
...
+        "null",
+        {
+          "type": "long",
+          "logicalType": "timestamp-millis"
+        }
+      ],
+      "default": null
+    },
+    {
+      "name": "attribute",
```
jskeet, line 56: Is "attribute" rather than "attributes" appropriate here? (I have no idea what the conventions are in Avro.)

cloudevents/formats/cloudevents-turbo.avsc
```diff
@@ -0,0 +1,79 @@
...
+      ],
+      "default": null
+    },
+    {
+      "name": "time",
+      "type": [
+        "null",
+        {
+          "type": "long",
+          "logicalType": "timestamp-millis"
```
jskeet, line 50:
```suggestion
          "logicalType": "timestamp-micros"
```

cloudevents/formats/avro-turbo-format.md
duglin: it just means that CE attribute values could also appear in other places. I believe https://github.com/cloudevents/spec/blob/main/cloudevents/extensions/distributed-tracing.md is an example of that. The traceparent value appears as both as CE attribute and an HTTP header. We didn't want to imply that people must replace some other spec's mechanism with CE's, but rather they can co-exist.

JemDay:
w.r.t the URI and 'type preservation' comment earlier: the proto spec does retain the CloudEvent type information for context attributes ..

Ref: https://github.com/cloudevents/spec/blob/306daeb0c7c72bbffa9017882b69a6907bfc5dbe/cloudevents/formats/cloudevents.proto#L48

cloudevents/formats/avro-turbo-format.md
alexec: Should we delete?

cloudevents/formats/cloudevents-turbo.avsc
alexec: It is “attribute” in the old format. I think should “attributes” too.

alexec:
> w.r.t the URI and 'type preservation' comment earlier: the proto spec does retain the CloudEvent type information for context attributes ..
> 
> Ref:
> 
> https://github.com/cloudevents/spec/blob/306daeb0c7c72bbffa9017882b69a6907bfc5dbe/cloudevents/formats/cloudevents.proto#L48

Ah. Yes. Old Avro does not.

cloudevents/formats/avro-turbo-format.md
jskeet: I'd be happy with deleting it, if that's okay with others.

duglin: If we do remove it perhaps we should do it from both Avro docs for consistency.

alexec: action?

alexec:
really, screw merge conflicts, so painful

duglin:
it may have done it if you pushed an empty branch

duglin:
I don't see any commits for it

PR #1213, opened 2023-06-01
Avro Compact Format

alexec:


alexec:
@duglin I've had a negative contributor experience with this PR:

- Fighting the DCO check. There is no need to sign every commit, just one is enough to meet this check.
- Merge conflicts. I've been editing files that are also edited on master branch. 

This has resulted in a messy PR very unlike my usual work.

alexec:
I've traced the cause of the conflict. The instruction for fixing the DCO check are wrong. You cannot push using `--force-with-lease` once you `git merge master`.

alexec:
https://github.com/cloudevents/sdk-java/pull/578

alexec:
@duglin open questions:

Existing event formats loose type information (`time` is rounded to microseconds and offset lost, URIs are de. Both `avro` and `json` formats lose URI type information. Ee to limitations in the spec and limitation in JSON respectively. 

It is not clear if the HTTP formats support this, but I'm not sure how they would. Say you head custom context attribute that was `foo`, how can you turn this into a HTTP header and back and preserve type?

Given how common the `json` format is, it is the case that de-facto URIs are not supported in custom context attributes. 

Avro Turbo chooses performance over complexity. Supporting these types is `<= O(9)` cost (the number of CE SDKs is 9). Complex systems tend to perform poorly.

For `time` , the argument is not so strong. we could use an Avro, but it makes implementation complex.

My Java implementation  is 126 lines long. 

* `json` is 923 lines.
* `protobuf` is 700 lines.
* `xml` is 1017 lines long

Less code means less bugs. I don't want to make implementing it harder and more buggy than needed.

---

@pierDipi is v3x of the Java SDK backwards incompatible with v2.x? If that is the case, then I'll put this work on hold because we will not have resource to do a version upgrade anytime in the next 3-6 months.

pierDipi:
> @duglin open questions:
> 
> Existing event formats loose type information (`time` is rounded to microseconds and offset lost, URIs are de. Both `avro` and `json` formats lose URI type information. Ee to limitations in the spec and limitation in JSON respectively.
> 
> It is not clear if the HTTP formats support this, but I'm not sure how they would. Say you head custom context attribute that was `foo`, how can you turn this into a HTTP header and back and preserve type?
> 
> Given how common the `json` format is, it is now de-facto URIs are not supported in custom context attributes.
> 
> Avro Turbo chooses performance over complexity. Supporting these types is `<= O(9)` cost (the number of CE SDKs is 9). Complex systems tend to perform poorly.
> 
> For `time` , the argument is not so strong. we could use an union, but it makes implementation costly, complex and buggy.
> 
> My Java implementation is 126 lines long.
> 
> * `json` is 923 lines.
> * `protobuf` is 700 lines.
> * `xml` is 1017 lines long
> 
> It is a full _order of magnitude_ smaller. This is achieved by keenly eschewing complexity.
> 
> Less code means less bugs. I don't want to make implementing it harder and more buggy than needed.
> 
> @pierDipi is v3x of the Java SDK backwards incompatible with v2.x? If that is the case, then I'll put this work on hold because we will not have resource to do a version upgrade anytime in the next 3-6 months.

In v3, there will be a few breaking changes ([current list of breaking changes](https://github.com/cloudevents/sdk-java/milestone/13?closed=1), however, avro-turbo will be a new module, so we can backport it for a 2.x release

cloudevents/formats/avro-format.md
```diff
@@ -148,24 +148,29 @@ For other types (non-binary data without a `datacontenttype` attribute), the
...
+| CloudEvents     | Type   | Exemplary Avro Value                      |
+|-----------------|--------|-------------------------------------------|
+| id              | string | `7a0dc520-c870-4193c8`                    |
+| source          | string | `https://github.com/cloudevents`          |
+| specversion     | N/A    | Spec version is always `1.0`.             |
+| type            | string | `com.example.object.deleted.v2`           |
+| datacontenttype | string | `application/octet-stream`                |
+| dataschema      | string | `http://registry.com/schema/v1/much.json` |
+| subject         | string | `mynewfile.jpg`                           |
+| time            | long   | `1685121689691`                           |
```
duglin, line 172: Isn't this (and `specversion` change) breaking changes? Didn't we create the new format to avoid breaking current impls?

alexec: Mistake in docs

alexec:
Thank you. You answered my questions (even if not intentionally).

* URIs custom context attributes not being back to URIs is acceptable.
* Time being truncated to micros is acceptable.
* The main breaking change in the Java SDK appears to be the upgrade to Kafka 3 libraries.

cloudevents/formats/avro-format.md
alexec: The more changes you make, the more mistakes you make. Same as more code means more bugs.

cloudevents/formats/avro-format.md
```diff
@@ -148,24 +148,29 @@ For other types (non-binary data without a `datacontenttype` attribute), the
...
+| CloudEvents     | Type   | Exemplary Avro Value                      |
+|-----------------|--------|-------------------------------------------|
+| id              | string | `7a0dc520-c870-4193c8`                    |
+| source          | string | `https://github.com/cloudevents`          |
+| specversion     | string | `1.0`                                     |
+| type            | string | `com.example.object.deleted.v2`           |
+| datacontenttype | string | `application/octet-stream`                |
+| dataschema      | string | `http://registry.com/schema/v1/much.json` |
+| subject         | string | `mynewfile.jpg`                           |
+| time            | long   | `2019-06-05T23:45:00Z`                    |
```
duglin, line 172: Not being an Avro expert this looks odd to have this as a "long" but then show its value as a string.

cloudevents/formats/avro-format.md
```diff
@@ -148,24 +148,29 @@ For other types (non-binary data without a `datacontenttype` attribute), the
...
-| contenttype | string | `"application/avro"`                           |
-| data        | bytes  | `[avro-serialized-bytes]`                      |
+| CloudEvents     | Type   | Exemplary Avro Value                      |
+|-----------------|--------|-------------------------------------------|
+| id              | string | `7a0dc520-c870-4193c8`                    |
+| source          | string | `https://github.com/cloudevents`          |
+| specversion     | string | `1.0`                                     |
+| type            | string | `com.example.object.deleted.v2`           |
+| datacontenttype | string | `application/octet-stream`                |
+| dataschema      | string | `http://registry.com/schema/v1/much.json` |
```
duglin, line 170: This is kind of an odd example. Source is a git org, not repo, so it seems odd that an objDeleted event is coming from it. The dataschema is jsonSchema (I assume) but then the "subject" is a jpg - which I doubt would be defined by json schema. Am I missing something/

cloudevents/formats/avro-format.md
alexec: this is correct for avro

cloudevents/formats/avro-format.md
alexec: this is copied from the spec page

cloudevents/formats/avro-format.md
clemensv: it's correct but not precise. it's an "int" with the logicalType "date". While we're doing edits here we should probably correct that, maybe as "int (date)"

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,109 @@
+# Avro Turbo Event Format for CloudEvents - Version 1.0.3-wip
```
clemensv, line 1: I like the spec except for the "turbo" moniker and in spite of my affinity to the automaker that has "turbo" prominently in its model designations. I'd prefer the somewhat more boring "compact".

duglin:
@alexec is this one ready for final review?
Also, in the previous version you updated the SDK.md file - did you want to do that again for this one?

alexec:
ready for review

duglin:
Discussed on the 6/15 call and the general direction looks good.
Just a few outstanding comments that need to be dealt with, and some folks wanted a little more time to review.
Hopefully we'll vote next week.

cloudevents/formats/avro-turbo-format.md
jskeet: +1 for compact (I should have noted this earlier).

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Abstract
+
+The Avro Turbo Format for CloudEvents defines how events are expressed in
```
jskeet, line 5: This and line 25 still make each format sound definitive. This can probably be fixed later, but we should do so at some point.

cloudevents/formats/cloudevents-turbo.avsc
```diff
@@ -0,0 +1,79 @@
...
+        "null",
+        {
+          "type": "long",
+          "logicalType": "timestamp-micros"
+        }
+      ],
+      "default": null
+    },
+    {
+      "name": "attributes",
```
jskeet, line 56: Perhaps this should be extensionattributes or extensions? (The other values are all attributes too, apart from data.)

cloudevents/formats/avro-turbo-format.md
```diff
@@ -0,0 +1,109 @@
...
+The CloudEvents type system MUST be mapped to Avro types as follows.
+
+| CloudEvents   | Avro                                                                   |
+|---------------|------------------------------------------------------------------------|
+| Boolean       | [boolean][avro-primitives]                                             |
+| Integer       | [int][avro-primitives]                                                 |
+| String        | [string][avro-primitives]                                              |
+| Binary        | [bytes][avro-primitives]                                               |
+| URI           | [string][avro-primitives] following [RFC 3986 §4.3][rfc3986-section43] |
+| URI-reference | [string][avro-primitives] following [RFC 3986 §4.1][rfc3986-section41] |
```
jskeet, line 61: So just to check:

- In the current proposal, there's no way of knowing the desired attribute type for string-oriented extension attributes
- There's no clean way of adding that information?

jskeet (commented):
I agree with the suggestion to rename this to "compact" - I'm happy to do that in an extra commit in this PR if you'd like, so that we're not adding work for you.

Other than that:

- One wording nitpick which can be addressed later
- One question about type differentiation for String/URI/URI-Ref
- One suggested renaming

cloudevents/formats/avro-turbo-format.md
alexec: There are two intents:

- go fast
- be cheap

Being "compact" is the solution.

cloudevents/formats/cloudevents-turbo.avsc
alexec: I don't feel strongly - @clemensv _

alexec:
I've renamed this to "Avro Compact" because 3 of us preferred that.

duglin:
> Also, in the previous version you updated the SDK.md file - did you want to do that again for this one?

duglin:
can you rebase? There's a conflict on the SDK.md

alexec:
there is always a conflict with that file, I'll fix conflicts once approved

duglin:
I fixed the rebase for you so we can have the checker do it's job

duglin:
I also added a change to the primer since the Adobe docs moved. Might as well just reuse this PR. Now the only failing check is an ok one.

cloudevents/formats/cloudevents-compact.avsc
```diff
@@ -0,0 +1,79 @@
...
+    },
+    {
+      "name": "attributes",
+      "type": {
+        "type": "map",
+        "values": [
+          "null",
+          "boolean",
+          "int",
+          "long",
```
JemDay, line 63: I believe this should be represented as follows to clarify that we're talking about 'timestamp' types..

```
...
"boolean",
"int",
{ 
  "type": "long",
  "logicalType" : "timestamp-micros"
}
"string",
....
```

JemDay (changes requested):
Minor tweak ....

cloudevents/formats/cloudevents-compact.avsc
alexec: but then you could not have any `long`

JemDay: That's true .. but "long" is not a supported CloudEvent attribute primitive  type ...

alexec: ah - did not know that

alexec, line 63:
```suggestion
          { 
            "type": "long",
            "logicalType" : "timestamp-micros"
          },
```

cloudevents/formats/cloudevents-compact.avsc
```diff
@@ -0,0 +1,82 @@
...
+        }
+      ],
+      "default": null
+    },
+    {
+      "name": "attributes",
+      "type": {
+        "type": "map",
+        "values": [
+          "null",
```
jskeet, line 60: I hadn't thought about this before, but what's the purpose of having null as a valid extension attribute value here? It makes some sense for the optional core attributes, but I don't believe CloudEvents have the concept of "here's an extension attribute, but it has no value". (The JSON format explicitly states that null values are discarded.)

jskeet (commented):
Ping @clemensv for the naming of "attributes" vs "extensions" for the map of extension attributes. I'd personally like to see that resolved before merging. One new thing I've noticed as well around nullability... but we're getting there :)

cloudevents/formats/cloudevents-compact.avsc
JemDay: I *think* I agree with @jskeet here.. The map as-a-whole is optional, but if it's present then each entry must have a valid value.

cloudevents/formats/avro-turbo-format.md
alexec: Done

cloudevents/formats/cloudevents-turbo.avsc
alexec: Changed to “extensions”

cloudevents/formats/cloudevents-compact.avsc
alexec: Removed “null”

jskeet (approved):
Thanks for being patient with us :)

duglin:
@alexec can you please move the new spec into the cloudevents/working-drafts dir?

duglin:
@alexec couple of bad refs now:
```
cloudevents/SDK.md: 'https://github.com/cloudevents/spec/blob/main/cloudevents/formats/avro-compact-format.md' was not found
cloudevents/working-drafts/avro-compact-format.md: Translation file cloudevents/languages/he/working-drafts/avro-compact-format.md does not exist
cloudevents/working-drafts/avro-compact-format.md: Translation file cloudevents/languages/zh-CN/working-drafts/avro-compact-format.md does not exist
```
actually, I think the first one is supposed to fail but it needs to point to "working-drafts" instead of "formats". The other 2 probably need the files moved to the working-drafts dir

duglin:
@alexec thanks - merging as it was approved during last week's call.

PR #1214, opened 2023-06-01
Clarify RFC language

### LINKED ISSUE #1209: Singular attribute extensions w/REQUIRED seems odd
Raised by @jskeet on the 5/25 call : When we have an extension that only has one attribute, marking the attribute as REQUIRED feels a bit odd and potentially confusing

The reasoning for the current state is: IF the extension is being used THEN the attribute becomes mandatory, so the `REQUIRED` is scoped to when the extension is enabled

However, this nuance may not be obvious to people.

Some potential options:
- leave it as is
- add text along the lines of `REQUIRED if this extensions is being used` instead of just `REQUIRED`
- add text in a new cloudevents/extensions/README.md file to explain this (but will anyone notice?)
- add text to each spec explaining this outside of the "Constraints" section of each attribution (meaning not on the `REQUIRED` line per 2nd bullet, but rather just once near the top of each extension spec). E.g. add something that says "The RFC2119 keywords apply only when this extension is enabled" - or something like that
- find a new word/phrase

other options?

duglin:
Fixes #1209 

ping @jskeet 

```release-note
Clarify scope of RFC2119 keywords in CE extensions
```


duglin:
Also included removing some extra (trailing) whitespace.

jskeet (approved):

duglin:
Approved on the 6/15 call

PR #1227, opened 2023-09-20
Expand tck tests

### LINKED ISSUE #1225: Expand the CESQL tck tests
As we work in Knative Eventing to move the subscription API filters into beta, we would like expand the CESQL tck tests to cover more complex use cases. Specifically, expressing some of the other filter dialects (filter, suffix, exact) in CESQL with a combination of ANDs and ORs on required, optional, extension, and non-existing attributes.

cc @pierDipi

Cali0707:
Fixes #1225 

## Proposed Changes

- Add tck test cases which verify that the "LIKE" operator works when the operand is not a string and needs to be coerced to string
- Add tck test cases for suffix, prefix, and exact filter implementations
- Add tck test cases for ANDs and ORs of suffix, prefix, and exact filter implementations

### DIFF
--- a/cesql/cesql_tck/like_expression.yaml
+++ b/cesql/cesql_tck/like_expression.yaml
@@ -93,3 +93,26 @@ tests:
     eventOverrides:
       myext: "abc123123%456_dzf"
     result: false
+  
+  - name: With type coercion from int (1)
+    expression: "234 LIKE '23_'"
+    result: true
+  - name: With type coercion from int (2)
+    expression: "2344 LIKE '23%'"
+    result: true
+  - name: With type coercion from int (3)
+    expression: "2344 LIKE '23%'"
+    result: false 
+
+  - name: With type coercion from bool (1)
+    expression: "TRUE LIKE 'tr%'"
+    result: true
+  - name: With type coercion from bool (2)
+    expression: "TRUE LIKE '%ue'"
+    result: true
+  - name: With type coercion from bool (3)
+    expression: "FALSE LIKE 'tr%'"
+    result: false
+  - name: With type coercion from bool (4)
+    expression: "FALSE LIKE 'fal%'"
+    result: true 
--- a/cesql/cesql_tck/like_expression.yaml
+++ b/cesql/cesql_tck/like_expression.yaml
@@ -101,7 +101,7 @@ tests:
     expression: "2344 LIKE '23%'"
     result: true
   - name: With type coercion from int (3)
-    expression: "2344 LIKE '23%'"
+    expression: "2344 LIKE '23_'"
     result: false 
 
   - name: With type coercion from bool (1)
--- a/cesql/cesql_tck/subscriptions_api_recreations.yaml
+++ b/cesql/cesql_tck/subscriptions_api_recreations.yaml
@@ -105,12 +105,6 @@ tests:
     eventOverrides:
       id: "notmyId"
       source: "http://www.some-website.com"
-  - name: Prefix OR Suffix filter (5)
-    expression: "id LIKE 'my%' OR myext LIKE '%ext'"
-    error: missingAttribute
-    eventOverrides:
-      id: "notmyId"
-      source: "http://www.some-website.com"
 
   - name: Disjunctive Normal Form (1)
     expresion: "(id = 'myId' AND type LIKE '%.success') OR (id = 'notmyId' AND source LIKE 'http://%' AND type LIKE '%.warning')"
@@ -128,13 +122,6 @@ tests:
   - name: Disjunctive Normal Form (3)
     expresion: "(id = 'myId' AND type LIKE '%.success') OR (id = 'notmyId' AND source LIKE 'http://%' AND type LIKE '%.warning')"
     result: false 
-    eventOverrides:
-      id: "notmyId"
-      type: "example.event.warning"
-      source: "https://localhost.localdomain"
-  - name: Disjunctive Normal Form (4)
-    expresion: "(id = 'myId' AND type LIKE '%.success') OR (id = 'notmyId' AND source LIKE 'http://%' AND type LIKE '%.warning') OR (myext = 'cusomext')"
-    error: missingAttribute
     eventOverrides:
       id: "notmyId"
       type: "example.event.warning"
--- a/cesql/cesql_tck/subscriptions_api_recreations.yaml
+++ b/cesql/cesql_tck/subscriptions_api_recreations.yaml
@@ -1,0 +1,176 @@
+name: SubscriptionsAPI Recreations
+tests:
+  - name: Prefix filter (1)
+    expression: "source LIKE 'https://%'"
+    result: true
+    eventOverrides:
+      source: "https://example.com"
+  - name: Prefix filter (2)
+    expression: "source LIKE 'https://%'"
+    result: false
+    eventOverrides:
+      source: "http://example.com"
+  - name: Prefix filter on string extension
+    expression: "myext LIKE 'custom%'"
+    result: true
+    eventOverrides:
+      myext: "customext"
+  - name: Prefix filter on missing string extension
+    expression: "myext LIKE 'custom%'"
+    error: missingAttribute
+
+  - name: Suffix filter (1)
+    expression: "type like '%.error'"
+    result: true
+    eventOverrides:
+      type: "com.github.error"
+  - name: Suffix filter (2)
+    expression: "type like '%.error'"
+    result: false 
+    eventOverrides:
+      type: "com.github.success"
+  - name: Suffix filter on string extension
+    expression: "myext LIKE '%ext'"
+    result: true
+    eventOverrides:
+      myext: "customext"
+  - name: Suffix filter on missing string extension
+    expression: "myext LIKE '%ext'"
+    error: missingAttribute
+
+  - name: Exact filter (1)
+    expression: "id = 'myId'"
+    result: true
+    eventOverrides:
+      id: "myId"
+  - name: Exact filter  (2)
+    expression: "id = 'myId'"
+    result: false
+    eventOverrides:
+      id: "notmyId"
+  - name: Exact filter on string extension
+    expression: "myext = 'customext'"
+    result: true
+    eventOverrides:
+      myext: "customext"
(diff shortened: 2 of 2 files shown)


Cali0707:
cc @duglin @pierDipi

cesql/cesql_tck/like_expression.yaml
```diff
@@ -93,3 +93,26 @@ tests:
...
     result: false
+  
+  - name: With type coercion from int (1)
+    expression: "234 LIKE '23_'"
+    result: true
+  - name: With type coercion from int (2)
+    expression: "2344 LIKE '23%'"
+    result: true
+  - name: With type coercion from int (3)
+    expression: "2344 LIKE '23%'"
```
duglin, line 104: did you mean for this to be 23_ ?

cesql/cesql_tck/like_expression.yaml
```diff
@@ -93,3 +93,26 @@ tests:
...
+    result: false 
+
+  - name: With type coercion from bool (1)
+    expression: "TRUE LIKE 'tr%'"
```
duglin, line 108: where does the spec say "TRUE" is a bool? I only see "true"

cesql/cesql_tck/subscriptions_api_recreations.yaml
```diff
@@ -0,0 +1,176 @@
...
+      source: "http://www.some-website.ca"
+  - name: Prefix OR Suffix filter (4)
+    expression: "id LIKE 'my%' OR source LIKE '%.ca'"
+    result: false 
+    eventOverrides:
+      id: "notmyId"
+      source: "http://www.some-website.com"
+  - name: Prefix OR Suffix filter (5)
+    expression: "id LIKE 'my%' OR myext LIKE '%ext'"
+    error: missingAttribute
```
duglin, line 110: Just curious, does the spec say that the stuff after the OR must be evaluated if the first part evaluates to 'true'? We should add a test for that, unless I missed it. In particular, if the 2nd part would generate an error if evaluated... what should happen if someone chooses to NOT evaluate it ?

cesql/cesql_tck/subscriptions_api_recreations.yaml
```diff
@@ -0,0 +1,176 @@
...
+    eventOverrides:
+      id: "notmyId"
+      source: "http://www.some-website.com"
+
+  - name: Disjunctive Normal Form (1)
+    expresion: "(id = 'myId' AND type LIKE '%.success') OR (id = 'notmyId' AND source LIKE 'http://%' AND type LIKE '%.warning')"
```
duglin, line 116: This might answer my previous question :-) but where is that in the spec?

cesql/cesql_tck/subscriptions_api_recreations.yaml
```diff
@@ -0,0 +1,176 @@
...
+      type: "example.event.success"
+  - name: Conjunctive Normal Form (3)
+    expression: "(id = 'myId' OR type LIKE '%.success') AND (id = 'notmyId' OR source LIKE 'https://%' OR type LIKE '%.warning')"
+    result: false
+    eventOverrides:
+      id: "notmyId"
+      type: "example.event.warning"
+  - name: Conjunctive Normal Form (4)
+    expression: "(id = 'myId' OR type LIKE '%.success') AND (id = 'notmyId' OR source LIKE 'https://%' OR type LIKE '%.warning')"
+    result: true
```
duglin, line 163: shouldn't this generate an error due to missing 'source' ?

duglin: or is 'source' always there with a prefix of "https://" ?

cesql/cesql_tck/subscriptions_api_recreations.yaml
```diff
@@ -0,0 +1,176 @@
...
+    eventOverrides:
+      id: "myId" 
+      type: "example.event.success"
+  - name: Conjunctive Normal Form (5)
+    expression: "(id = 'myId' OR type LIKE '%.success') AND (id = 'notmyId' OR source LIKE 'https://%' OR type LIKE '%.warning') AND (myext = 'customext')"
+    error: missingAttribute
```
duglin, line 169: I agree with the error but I think it'll be because of 'source' not 'myext' as you're hinting at. If you add 'source' then I think you'll get what you intended.

cesql/cesql_tck/like_expression.yaml
Cali0707: Oh yeah, thanks for catching that!

cesql/cesql_tck/like_expression.yaml
Cali0707: Ah I wasn't referring to the spec here, rather the existing tck tests: https://github.com/cloudevents/spec/blob/755e45863e136ee35ac26941405c75666c091c9b/cesql/cesql_tck/casting_functions.yaml#L30-L35

Do you think I should switch this to `true` here?

cesql/cesql_tck/subscriptions_api_recreations.yaml
Cali0707: I'm not 100% sure, what I could find was:

> Because every operator and function is total, an expression evaluation flow is defined statically and cannot be modified by expected or unexpected errors. Nevertheless CESQL includes the concept of errors: when an expression is evaluated, in case an error arises, the evaluator collects a list of errors, referred in this spec as error list, which is then returned together with the evaluated value of the CESQL expression.

So, I guess the spec does not explicitly specify whether or not there should be an error if the `OR` has an early return. Do you think we should specify this?

cesql/cesql_tck/subscriptions_api_recreations.yaml
Cali0707: As above, don't think that this is in the spec - I guess we should clarify whether or not this is an error that would occur, or whether it would be implementation specific

cesql/cesql_tck/subscriptions_api_recreations.yaml
Cali0707: Yep my bad, I forgot that source was not set by default in these test - I'll fix it!

cesql/cesql_tck/like_expression.yaml
duglin: You might be in a better position than me since my only knowledge of the spec is based on this PR and my current reading of the spec to review your PR ;-) , but is the "expression" string meant to adhere to the syntax defined by the spec? If so, then it seems like it may need to be `true` or the spec needs to allow for both upper and lower case bools. Actually, the spec should probably say which things are case sensitive or not, in general.

cesql/cesql_tck/subscriptions_api_recreations.yaml
duglin: I sure does seem like the spec should be clear on what `true OR error` results in... either `true` or `error`. Otherwise we'll have an interop issue. I would think most people would expect `true` but I'm not an expert in what query languages normally do for this. @lionelvillard @embano1 any thoughts?

cesql/cesql_tck/subscriptions_api_recreations.yaml
duglin: thanks for clarifying my thought

cesql/cesql_tck/like_expression.yaml
Cali0707: @duglin from https://github.com/cloudevents/spec/blob/main/cesql/spec.md#2-language-syntax it looks like the grammar is case-insensitive so both `true` and `TRUE` should be valid. Maybe we can open a new issue to add clarification to the spec where we mention the boolean literals that this is the case?

cesql/cesql_tck/subscriptions_api_recreations.yaml
Cali0707: no problem!

cesql/cesql_tck/subscriptions_api_recreations.yaml
Cali0707: @duglin on the go sdk v2, when evaluating `(type = 'example.knative.dev') OR (missingattribute = 'test')` on an event with `type: 'example.knative.dev'` but no `missingattribute` I get: `missing attribute 'missingattribute'`

So, it seems like the current behaviour (at least for go) is that it reports the error, even if the first part evaluated to true

Cali0707: Should I open an issue to clarify this behaviour, or do we want to open an issue to change this behaviour and make it behave like you think most people would expect (`true` instead of `error`)?

Cali0707: It seems like from https://github.com/cloudevents/spec/blob/main/cesql/spec.md#41-error-handling that we would always return the error regardless of the result though (either fail fast, or continue and find the result and then return result and error)

cesql/cesql_tck/like_expression.yaml
duglin: I think that would be good since it may not be clear that "keywords" don't just mean function names, or operators, but also constants.  And thanks for the pointer - I missed that.

cesql/cesql_tck/subscriptions_api_recreations.yaml
duglin: I think we should open an issue to get feedback from the group and then make it clear in the spec... and then align the go-sdk if needed.

duglin:
@Cali0707 is this one ready to be merged? I assume so, and we have the follow-on issues to continue to clarifying questions. Right?

Cali0707:
> @Cali0707 is this one ready to be merged? I assume so, and we have the follow-on issues to continue to clarifying questions. Right?

@duglin I think if we want to merge this now, I should remove the tests which may or may not error on a  OR, otherwise we might break tests on some of the SDKs depending on their implementation since that isn't specified currently.

duglin:
ok sounds good. Let me know when done and then we can vote on it during the Thursday call

Cali0707:
@duglin I believe I have removed the two tests with issues in them. If you think this is ready, then I think it is ready to be voted on on Thursday. Thanks for all of your help on this!

PR #1229, opened 2023-09-23
more review changes

duglin:
- mention the SDKs are (sort of) independent of the spec


duglin:
Approved on the 9/28 call

PR #1233, opened 2023-10-05
Submission of draft Cloud Events DDS Protocol Binding and DDS Event Format

protimabanerjee:
Fixes #

This PR includes the following new files added to the cloudevents/working-drafts folder:
- dds-protocol-binding.md
- dds-format.xml
- dds-format.md

## Proposed Changes

- cloudevents/working-drafts/dds-protocol-binding.md (new contribution)
- cloudevents/working-drafts/dds-format.xml (new contribution)
- cloudevents/working-drafts/dds-format.md (new contribution)


cloudevents/working-drafts/dds-format.md
```diff
@@ -0,0 +1,243 @@
...
+| URI-reference | String (255) following [RFC 3986 §4.1][rfc3986-section41] |
+| Timestamp     | seconds: Int64 nanoseconds: UInt32     |
+
+Extension specifications MAY define secondary mapping rules for the values of
+attributes they define, but MUST also include the previously defined primary
+mapping.
+
+### 2.2 OPTIONAL Attributes
+
+CloudEvents Spec defines OPTIONAL attributes. The set of possible Attribute Types and Values
```
duglin, line 68: s/Spec/spec/

cloudevents/working-drafts/dds-format.md
```diff
@@ -0,0 +1,243 @@
...
+   </case>
+   <case>
+      <caseDiscriminator value="(io::cloudevents::TIMESTAMP)"/>
+      <member name="ce_timestamp" type="nonBasic" nonBasicTypeName="io::cloudevents::Timestamp"/>
+   </case>
+</union>
+
+```
+
+Based on the above, the type definition for optional Attributes is as follows:
```
duglin, line 116: I feel like I'm missing something... what in the example below indicates that the attribute being defined is optional vs mandatory?  Also is `Attribute` (in `<struct name="Attribute">`) the name of the attribute? If so, perhaps call it something less ambiguous, like `MyExtension` - otherwise it looks like a keyword and not a user defined name.

cloudevents/working-drafts/dds-format.md
```diff
@@ -0,0 +1,243 @@
...
+of the content. This can be determined by examining by consulting the `datacontenttype`
+and `datacontentencoding` attributes.
+
+If the implementation determines that the type of the data is binary, the value
+MUST be stored in the `body` field using the `bytes` type.
+
+For other types, the implementation MUST translate the data value into a text or JSON
+representation using the union type described for the message body.
+
+## 4 Examples
```
duglin, line 210: I wonder if it would be useful to DDS folks to see a full XML CE serialization, in addition to the table below?

duglin:
I'm curious about the relationship between this and the XML stuff we have... should there be a relationship? Should this be a specific usage of the XML format? Meaning, we just make sure that any of our XML format rules are not violated by this PR.

@JemDay @jskeet any thoughts?

jskeet:
It feels odd to me to have a format specific to one particular use case. I would prefer it if the XML format could be used by DDS as-is - and would like to know what benefit there is in having a whole separate format vs "DDS uses the XML format, and specifies that attributes X, Y and Z must be present."

duglin:
@protimabanerjee any update on this one coming?

protimabanerjee:
I'm going to be working on this today and likely to be reaching back with
some questions shortly. My apologies for the delay, I had a high priority
task pop up over the last few weeks

cloudevents/working-drafts/dds-format.md
protimabanerjee: updated and language cleaned up

cloudevents/working-drafts/dds-format.md
protimabanerjee: I changed "Attribute" -> "DDS-CE-Attribute" and "Attributes" -> "DDS-CE-Attributes". Does that read any better to you?

cloudevents/working-drafts/dds-format.md
protimabanerjee: The long story: The XML format here is what is used to define what is required/optional in the DDS CE message. This format information is then provided to a DDS code generator that defines the message classes (code in C, C++, Java, Javascript, other supported languages) with the required attributes. The data values are populated in code into the message classes and the data is actually sent over the wire in a CDR (Common Data Representation) format which is then decoded by the DDS message receiver on the other side. So the XML is not the serialization format itself, it is simply the information used to configure the CDR serialization/deserialization process.

This was obviously not clearly written down by me. I could use some advice though on how to best put state this information in the document. Could we have a discussion on this topic? I will see if I can get on the agenda at tomorrow's meeting.

cloudevents/working-drafts/dds-protocol-binding.md
```diff
@@ -0,0 +1,266 @@
...
+This specification does not further define any of the [CloudEvents][ce] event
+attributes.
+
+### 2.1. data
+
+`data` is assumed to contain opaque application data that is encoded as declared
+by the `datacontenttype` attribute. 
+
+There are two values of `datacontenttype` currently supported:
+- `cloudevent/json` for JSON data
```
JemDay, line 105: Unless I'm mis-understanding this statement it doesn't make sense to me ... iis the suggestion that I can have a receive a could-event over-the-wire that has a data payload which represents another cloud-event represented in JSON ?

I'm not clear why this needs to be called out explicitly as clouds-events can wrap *any* content .. what's special about json in this context.

cloudevents/working-drafts/dds-protocol-binding.md
```diff
@@ -0,0 +1,266 @@
...
+- `application/cloudevent+dds` for a wider range of data types (text, binary, JSON)
```
JemDay, line 106: I think we're muddling the content-type of the data inside the CE with the transport-level content-type of the enclosing event.

JemDay (commented):
I think I need to read up a bit more on DDS before I comment further

duglin:
@protimabanerjee some verification failures: https://github.com/cloudevents/spec/actions/runs/6730075140/job/18416748295?pr=1233

protimabanerjee:
Fixed now I think.

cloudevents/working-drafts/dds-format.md
```diff
@@ -0,0 +1,247 @@
...
+4. [Transport](#4-transport)
+5. [Batch Format](#5-batch-format)
+6. [Examples](#6-examples)
+
+## 1. Introduction
+
+[CloudEvents][ce] is a standardized and protocol-agnostic definition of the
+structure and metadata description of events. This specification defines how the
+elements defined in the CloudEvents specification are are represented using
+a protobuf schema.
```
duglin, line 29: "protobuf" ? :-)  and on line 32 below

cloudevents/working-drafts/dds-format.md
```diff
@@ -0,0 +1,247 @@
...
+     <member name="ce_uri_reference" type="nonBasic" nonBasicTypeName="io::cloudevents::URI_Ref"/>
+   </case>
+ </union>
+
+ <struct name="DDS-CE-Attribute">
+   <member name="key" type="string" stringMaxLength="255"/>
+   <member name="value" type="nonBasic" nonBasicTypeName="io::cloudevents::AttributeValue"/>
+ </struct>
+
+ <typedef name="DDS-CE-Attributes" type="nonBasic" nonBasicTypeName="io::cloudevents::DDS-CE-Attribute" sequenceMaxLength="100"/>
```
duglin, line 124: does this mean there can only be 100 CE attributes? Is that a limit per the DDS spec or just a "sounds about right" value? Does there have to be a max specified at all?

cloudevents/working-drafts/dds-format.md
```diff
@@ -0,0 +1,247 @@
...
+     <member name="ce_uri" type="nonBasic" nonBasicTypeName="io::cloudevents::URI"/>
+   </case>
+   <case>
+     <caseDiscriminator value="(io::cloudevents::URI_REF)"/>
+     <member name="ce_uri_reference" type="nonBasic" nonBasicTypeName="io::cloudevents::URI_Ref"/>
+   </case>
+ </union>
+
+ <struct name="DDS-CE-Attribute">
+   <member name="key" type="string" stringMaxLength="255"/>
```
duglin, line 120: The spec says attribute names SHOULD NOT exceed 20 chars.... notice it's "SHOULD NOT", not "MUST NOT", so the 255 is probably ok here, but I wonder if we should add some text to enforce the "SHOULD NOT" aspect... something like: the above definition mandates that attribute names are not longer than 255 characters, however, per the CE spec they SHOULD NOT be longer than 20 characters.  Maybe? BTW, where did "255" come from? Is that a DDS thing?

cloudevents/working-drafts/dds-format.md
```diff
@@ -0,0 +1,247 @@
...
+
+## 2.4 OPTIONAL Attributes & Extensions
+
+OPTIONAL and extension attributes are represented using a sequence of key-value
+constructs (DDS-CE-Attributes) enabling direct support of the CloudEvent
+[type system][ce-types]. An underlying union and enumeration data structure
+are needed to construct the key-value sequence to establish the association
+between the IDL types and the CloudEvent type system.
+
+```xml
```
duglin, line 79: Is this xml normative or just a sample and people can tweak as needed? If normative, we may want to say that DDS encoded messages MUST adhere to this definition.

cloudevents/working-drafts/dds-format.md
```diff
@@ -0,0 +1,247 @@
...
+ <enum name="DataKind">
+   <enumerator name="BINARY"/>
+   <enumerator name="TEXT"/>
+ </enum>
+	  
+ <union name="Data" extensibility="final">
+   <discriminator type="nonBasic" nonBasicTypeName="io::cloudevents::DataKind"/>
+   <case>
+     <caseDiscriminator value="(io::cloudevents::BINARY)"/>
+     <member name="binary_data" type="byte" sequenceMaxLength="100"/>
```
duglin, line 151: Does this mean DDS doesn't support binary data longer than 100 bytes?

cloudevents/working-drafts/dds-protocol-binding.md
```diff
@@ -0,0 +1,265 @@
...
+The CloudEvents specification defines three content modes for transferring
+events: _structured_, _binary_ and _batch_. 
+
+Every compliant implementation of the CloudEvent specification SHOULD support
+both structured and binary modes. The DDS protocol binding supports both these modes.
+
+The DDS protocol binding does not currently support the _batch_ content mode.
+
+In the _structured_ content mode, event metadata attributes and event data are
+placed into the DDS message using an [event format](#14-event-formats).
```
duglin, line 73: Just curious, the next paragraph mentions the `data` section but this paragraph does not, is it correct to assume that in structured mode the entire CE goes into the `data` section?

cloudevents/working-drafts/dds-protocol-binding.md
```diff
@@ -0,0 +1,265 @@
...
+
+`data` is assumed to contain opaque application data that is encoded as declared
+by the `datacontenttype` attribute. 
+
+There are two values of `datacontenttype` currently supported:
+- `cloudevent/json` for JSON data
+- `application/cloudevent+dds` for binary and plain text data types
+
+If the `datacontenttype` attribute is set to `application/cloudevent+dds`, the
+`datacontentencoding` attribute defines the encoding of the message body.
```
duglin, line 108: s/defines/MUST be present and MUST define/
I assume? Or is it s/defines/, if present, MUST define/ ?

cloudevents/working-drafts/dds-protocol-binding.md
```diff
@@ -0,0 +1,265 @@
...
+event can distinguish between content modes by inspecting the `content-type`
+header of the DDS message.
+
+If the header is present and its value is `cloudevent/json`, the receiver
+decodes the message as JSON data.
+
+If the header is present and its value is `application/cloudevent+dds`, the
+receiver decodes the message into the DDS message format, using the
+`datacontentencoding` attribute to determine the encoding of the message body.
+There are three valid values of the `datacontentencoding` attribute
```
duglin, line 125: 3 or 2?

cloudevents/working-drafts/dds-protocol-binding.md
```diff
@@ -0,0 +1,265 @@
...
+can not mandate that senders do not include any of the CloudEvents headers
+when the message is not a CloudEvent, it would be reasonable for a receiver
+to assume that if the message has all of the mandatory CloudEvents attributes
+as headers then it`s probably a CloudEvent. However, as with all CloudEvent
+messages, if it does not adhere to all of the normative language of this
+specification then it is not a valid CloudEvent.
+
+### 3.1. Keys
+
+The `datakey` of the DDS message MAY be populated. A _key field_ in DDS is way
```
duglin, line 148: s/is way/is a way/   ?

cloudevents/working-drafts/dds-protocol-binding.md
```diff
@@ -0,0 +1,265 @@
...
+to uniquely identify individual instances of data being published to a topic. For
+example, if you are publishing data to a Temperature Topic dealing with
+temperature readings from different sensors, the sensor ID could be a key field.
```
duglin, line 151: This example feels odd since "sensor ID" won't be unique per "instance of data", right?

cloudevents/working-drafts/dds-protocol-binding.md
```diff
@@ -0,0 +1,265 @@
...
+
+The chosen [event format](#14-event-formats) defines how all attributes, and
+`data`, are represented.
+
+The event metadata and data are then rendered in accordance with the
+[event format](#14-event-formats) specification and the resulting data becomes
+the DDS application [data](#21-data) section.
+
+Similar to binary mode, the DDS representation of a structured CloudEvent with
+no `data` is a DDS message with no body, and transmission of such a message
```
duglin, line 233: `is a DDS message with no body`... is this true? Wouldn't a structured CE still have a DDS message+body(`data`) but just missing the `data` attribute of the CE structure? Meaning, the DDS message's `data` would have the CE JSON but in that JSON the `data` attribute would be empty/missing. Or am I reading this incorrectly?

cloudevents/working-drafts/dds-protocol-binding.md
```diff
@@ -0,0 +1,265 @@
...
+is allowable.
+
+#### 3.3.3. Metadata Headers
+
+Implementations include the same DDS headers as defined for the
```
duglin, line 238: I'm reading this as the CE attributes are to appear as DDS Headers AND in the CE structure too. Is that the intent?

cloudevents/working-drafts/dds-format.md
```diff
@@ -0,0 +1,247 @@
...
+     <caseDiscriminator value="(io::cloudevents::INT32)"/>
+     <member name="ce_integer" type="int32"/>
+   </case>
+   <case>
+     <caseDiscriminator value="(io::cloudevents::STRING)"/>
+     <member name="ce_string" type="string" stringMaxLength="255"/>
+   </case>
+   <case>
+     <caseDiscriminator value="(io::cloudevents::BYTES)"/>
+     <member name="ce_bytes" type="byte" sequenceMaxLength="100"/>
```
JemDay, line 107: nit: The CloudEvent type system calls this 'binary' not 'bytes'.

Additionally I don't believe there's any mention of max-length in the spec, rightly or wrongly. So I'm not sure what to suggest as a length constraint if that is required.

cloudevents/working-drafts/dds-format.md
```diff
@@ -0,0 +1,247 @@
...
+   <enumerator name="BOOL"/>
+   <enumerator name="INT32"/>
+   <enumerator name="STRING"/>
+   <enumerator name="BYTES"/>
+   <enumerator name="URI"/>
+   <enumerator name="URI_REF"/>
+   <enumerator name="TIMESTAMP"/>
+ </enum>
+	  
+ <union name="AttributeValue" extensibility="final">
```
JemDay, line 91: Do you need to add a <case> to represent a Timestamp type ???

cloudevents/working-drafts/dds-format.md
```diff
@@ -0,0 +1,247 @@
...
+	  
+ <union name="Data" extensibility="final">
+   <discriminator type="nonBasic" nonBasicTypeName="io::cloudevents::DataKind"/>
+   <case>
+     <caseDiscriminator value="(io::cloudevents::BINARY)"/>
+     <member name="binary_data" type="byte" sequenceMaxLength="100"/>
+   </case>
+   <case>
+     <caseDiscriminator value="(io::cloudevents::TEXT)"/>
+     <member name="text_data" type="string" stringMaxLength="255"/>
```
JemDay, line 155: Similar to @duglin comment above ... I'm not sure we place any constraints (rightly or wrongly) on the size of event data.

cloudevents/working-drafts/dds-format.md
```diff
@@ -0,0 +1,247 @@
...
+  const source = "urn:event:from:myapi/resource/123";
+  const time = new Date().toISOString();
+  
+  // cloudevent+dds / binary
+  const ce_dds_binary_obj = new CloudEvent({
+    specversion: "1.0",
+    id: "b46cf653-d48a-4b90-8dfa-355c01061364",
+    type,
+    source,
+    datacontenttype: "application/cloudevent+dds",
```
JemDay, line 201: This doesn't seem to align, the data added in the code below is binary content (such as `image/jpg`)

JemDay (changes requested):
I'm still a little confused about whether I can use 'dds format' indepentantly of the 'dds protocol'.

Eg .. Can I format an event using DDS, send it over HTTP, and de-serialize it back into a CE

cloudevents/working-drafts/dds-cloudevent.xml
```diff
@@ -0,0 +1,105 @@
...
+              <member name="binary_data" type="byte" sequenceMaxLength="100"/>
+            </case>
+            <case>
+              <caseDiscriminator value="(io::cloudevents::TEXT)"/>
+              <member name="text_data" type="string" stringMaxLength="255"/>
+            </case>
+          </union>
+
+          <struct name="Event" extensibility="mutable">
+            <member name="headers" type="nonBasic" nonBasicTypeName="io::cloudevents::Headers"/>
```
JemDay, line 85: What does this represent ?

cloudevents/working-drafts/dds-format.md
protimabanerjee: Fixed :-)

cloudevents/working-drafts/dds-format.md
protimabanerjee: A max doesn't have to be specified because DDS does support unbounded strings and sequences. We actually ran some examples to verify the structure and setting a max makes the example code a bit simpler. There are a couple of extra memory management parameters you have to add with unbounded sequences. I just added those and verified that the examples work as expected without the maxlength needing to be specified.

cloudevents/working-drafts/dds-format.md
protimabanerjee: 255 is the default max string length for DDS. I will update max length to be 20 characters for attribute name - I don't see any good reason to extend a DDS attribute name longer than that.

cloudevents/working-drafts/dds-format.md
protimabanerjee: No, it supports binary data of any length (unbounded). (It just made my example code simpler when I was doing a test run - but there is no need for a max here.)

cloudevents/working-drafts/dds-format.md
```diff
@@ -0,0 +1,247 @@
...
+    subject: "SQUARE",
+    time,
+    datakey:"Somekey",
+    datacontentencoding: "binary",
+    data: Buffer.from("Some text string" as string) // send the binary representation of a string
+    // [ext1Name]: ext1Value,
```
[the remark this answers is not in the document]
protimabanerjee: We started digging a bit deeper into optional attributes and extensions and would like to defer support for these until a later revision of the protocol binding. 

The issue is that there is some overlap between optional attributes that can be specified at the DDS protocol level using DDS User Data, Topic Data and Group Data Quality of Service (QoS) settings. It is not exactly clear if these QoS (if they are set) should be propagated up to CE Optional Attributes or if CE Optional Attributes should be mapped down to DDS QoS. Similarly, for extension attributes there is some overlap between DDS QoS and CE Extension attributes. The clearest overlap is between the Expiry Time and DDS Lifespan - we would like to consider use cases to determine how best to map from one to the other. 

At first glance, it seemed like it would be possible to simply carry the OPTIONAL attributes through, but now we are not so sure that this will make sense for DDS->CloudEvents use cases.

Does this approach seem reasonable?

cloudevents/working-drafts/dds-format.md
protimabanerjee: Actually, I just removed the Attribute name altogether in light of the earlier comment.

cloudevents/working-drafts/dds-format.md
protimabanerjee: I think it is the right approach to make the XML normative. I updated the wording to say that DDS encoded messages must adhere to the schema.

cloudevents/working-drafts/dds-protocol-binding.md
protimabanerjee: That is correct. In structured mode, the entire CE message would go into the data section.

protimabanerjee: I updated the wording a bit between the two sections to make the difference between structured and binary modes a bit more clear.

cloudevents/working-drafts/dds-protocol-binding.md
protimabanerjee: it is the latter - if present, MUST define 
(thank you!)

cloudevents/working-drafts/dds-protocol-binding.md
protimabanerjee: It is 2! Another good catch - at first, I was thinking the JSON structured mode should go here too - but that is defined by its own format.

cloudevents/working-drafts/dds-protocol-binding.md
protimabanerjee: Fixed - thank you!

cloudevents/working-drafts/dds-protocol-binding.md
protimabanerjee: In DDS terminology, an instance is a set of samples that all have the same key value. A data sample is an individual message. So you can think of an instance as a group of samples that all have the same key. But this terminology is kind of confusing so let me see if I can make the language better.

cloudevents/working-drafts/dds-protocol-binding.md
protimabanerjee: No, you are reading it correctly, my wording is incorrect. I changed the wording to the following:

The DDS representation of a structured CloudEvent with no `data` is a message
that contains all of the CE attributes but has an empty body. Transmission of
such a message is allowable.

Does this seem reasonable?

cloudevents/working-drafts/dds-protocol-binding.md
protimabanerjee: Currently we're really only supporting (reading) the content-type header for both binary and structured data, so I updated the text to just say this.

cloudevents/working-drafts/dds-format.md
protimabanerjee: We thought about it and at this time we're not ready to support optional and extension attributes yet, since there might be potential conflicts with what is defined at the DDS protocol level in QoS settings. So this section is going away.

cloudevents/working-drafts/dds-format.md
protimabanerjee: Same as above. (But this was actually a copy/paste error from my sample code. I forgot the timestamp for some reason.)

cloudevents/working-drafts/dds-format.md
protimabanerjee: I removed all bounds on length/sizes, so we should be good here.

cloudevents/working-drafts/dds-format.md
protimabanerjee: The DDS serialization format (which is called the Common Data Representation or CDR) is actually the format that is used to send the data "over the wire". The receiver on the other side would need to know that binary data was being sent over the DDS wire protocol to know how to decode it.

Does this make sense?

cloudevents/working-drafts/dds-cloudevent.xml
protimabanerjee: The DDS protocol doesn't have any mechanism for sending headers as some other protocols do so we have to create a custom header structure within the DDS message for the users to add their headers.

cloudevents/working-drafts/dds-format.md
protimabanerjee: Responding to: Can I format an event using DDS, send it over HTTP, and de-serialize it back into a CE

The DDS format is tightly coupled to the DDS protocol. So when you send a message using the DDS protocol, you have to define the message format in DDS IDL, run the IDL through a code generator, and then get a set of classes that serialize/deserialize to the DDS wire protocol. 

I don't think there's really a way to use that serialization/deserialization code outside of DDS, the way you can with protobuf. So the end result is that the DDS format works only for the DDS protocol. I agree with you - this is an important point and I will add it to the doc.

cloudevents/working-drafts/dds-format.md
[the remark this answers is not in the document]
protimabanerjee: Btw, the working code (no extension attributes) is here: https://github.com/protimabanerjee/sdk-javascript/tree/main/examples/dds-ex-simple

duglin:
@protimabanerjee, from your perspective is this ready to merge?

All - if 'yes', I'd like to see if we can vote on this before the holidays, perhaps even on next week's call so please review it and add comments/concerns.

cloudevents/working-drafts/dds-format.md
JemDay: Apologies for the slow response...

If the formats & transports are tightly coupled I'm concerned it might not be appropriate to make a part of the CE specification.

If I'm not able to plug-and-play formats (wire-encoding) and different transport protocols this _may_ be better presented as a separate document that describes how CE's fit, or are used, in the context of DDS.

Ideally we should be able to realize a flow such as this (where Int == 'Intermediary'):

``` text
Source --[HTTP]--> Intermediary --[DDS]--> Intermediary --[AMQP]--> Intermediary --[KAFKA]--> SInk
```

Where the CE received by the Sink is equivalent to the CE emitted by the source, and the 'data'  is exactly what the source put into the original CE.

If we're not able to slot DDS into that type of pipeline I'm not sure this is the right place.

Apologies if I'm still not joining the (mental) dots.

protimabanerjee: This is a valid concern and, honestly, I was having this same doubt after watching Doug's demo showing how CloudEvents is able to pass messages across protocols. If a message goes to DDS that it is not able to deserialize, it does not look at any other information in the message to determine who to forward to. It simply drops the message. Now, there could be updates made at the protocol level to cause this kind of forwarding behavior but it does not exist in the standard today. There is a DDS Web Integration standard that defines how DDS traffic can be tunneled over HTTP, but nothing that does the reverse type of action.

So this might be the blocking point. :-(

cloudevents/working-drafts/dds-format.md
protimabanerjee: I think this is all part of the same blocker, unfortunately. :-(

cloudevents/working-drafts/dds-cloudevent.xml
duglin: if @JemDay is correct and "headers" are used for extensions, then we may need to rethink this because CE extensions are meant to be serialized just like spec-defined attributes, not within a separate "bucket"

cloudevents/working-drafts/dds-protocol-binding.md
```diff
@@ -0,0 +1,265 @@
...
+
+The CloudEvents specification defines three content modes for transferring
+events: _structured_, _binary_ and _batch_. 
+
+Every compliant implementation of the CloudEvent specification SHOULD support
+both structured and binary modes. The DDS protocol binding supports both these modes.
+
+The DDS protocol binding does not currently support the _batch_ content mode.
+
+In the _structured_ content mode, event metadata attributes and event data are
```
duglin, line 72: s/metadata attributes/context attributes/  I think

duglin: or just "event attributes" w/o "metadata"

cloudevents/working-drafts/dds-format.md
```diff
@@ -0,0 +1,209 @@
...
+| URI-reference | [string][dds-schema] following [RFC 3986 §4.1][rfc3986-section41] |
+| Timestamp     | [Timestamp][dds-schema]  |
+
+## 2.3 REQUIRED Attributes
+
+REQUIRED attributes are represented explicitly as fields in the [Event type definition][dds-schema].
+
+## 2.4 OPTIONAL Attributes & Extensions
+
+At this time OPTIONAL and extension attributes are not supported in the DDS format.
```
clemensv, line 80: That clause is in conflict with the base spec.

cloudevents/working-drafts/dds-protocol-binding.md
```diff
@@ -0,0 +1,265 @@
...
+This specification does not further define any of the [CloudEvents][ce] event
+attributes.
+
+### 2.1. data
+
+`data` is assumed to contain opaque application data that is encoded as declared
+by the `datacontenttype` attribute. 
+
+There are two values of `datacontenttype` currently supported:
+- `cloudevent/json` for JSON data
```
clemensv, line 104: That is not a valid content type for CloudEvents

cloudevents/working-drafts/dds-protocol-binding.md
```diff
@@ -0,0 +1,265 @@
...
+- `application/cloudevent+dds` for binary and plain text data types
+
+If the `datacontenttype` attribute is set to `application/cloudevent+dds`, the
+`datacontentencoding` attribute, if present, MUST define the encoding of the
```
clemensv, line 108: datacontentencoding only existed in a draft version of the spec and is not part of CloudEvents 1.0 and can therefore no longer be referred to and used.

cloudevents/working-drafts/dds-format.md
protimabanerjee: @clemensv as always, thank you for this thoughtful review.

@JemDay had also remarked earlier about the interoperability concerns with DDS.

I agree with both of you. I actually ran a few simple experiments to see what would happen if a message that DDS did not know how to interpret was sent to a DDS subscriber. The DDS subscriber simply drops the message with a serialization error. And I believe that this is dictated by the DDS specification. So I believe this submission is at its termination point.

I was not aware that RTI had a DDS patent though. We always have to ensure that DDS is marked as an OMG specification for any public distribution. @clemensv would you send me a link to the patent if you have it?

protimabanerjee: Please see response above

cloudevents/working-drafts/dds-protocol-binding.md
protimabanerjee: Please see response above

cloudevents/working-drafts/dds-protocol-binding.md
protimabanerjee: Please see response above

duglin:
@protimabanerjee to be sure... you want to close this PR right?

cloudevents/working-drafts/dds-format.md
JemDay: @protimabanerjee -  

> would happen if a message that DDS did not know how to interpret was sent to a DDS subscriber

I'm a little confused by this as I would expect from a DDS perspective you're exchanging CE's, so the applications would produce, and consume CE's as defined in your format specification.

protimabanerjee:
I think closing the PR is the right decision at this time. If in the future
DDS evolved to allow for more interoperability, we might re-submit. But it
will likely be a while.

PR #1243, opened 2023-10-23
SQL filter: Clarify that only boolean TRUE values pass filter

### LINKED ISSUE #1221: sql filter: clarification of non-boolean cesql expression result coercion to a boolean
From sql filter docs:

> Use of this MUST have a string value, representing a [CloudEvents SQL Expression](https://github.com/cloudevents/spec/blob/main/cesql/spec.md). The filter result MUST be true if the result value of the expression, coerced to boolean, equals to the TRUE boolean value, otherwise MUST be false if an error occurred while evaluating the expression or if the result value, coerced to boolean, equals to the FALSE boolean value.

If cesql expression evaluation result is not boolean type, it should be coerced to boolean. Different programming languages have different type coercion rules. So what about cloudevents? 

If expression result is of type int32, is non-zero integer coerced to true?
If expression result is of type string, is non-empty string coerced to true?

Cali0707:
Fixes #1221

## Proposed Changes

- Only CESQL results which are the `TRUE` boolean value should pass the filter.

**Release Note**

```release-note
Only CESQL results which are equal to the `TRUE` boolean value pass the CESQL subscriptions API filter.
```

### DIFF
--- a/subscriptions/spec.md
+++ b/subscriptions/spec.md
@@ -872,10 +872,10 @@ this specification:
 
 Use of this MUST have a string value, representing a [CloudEvents SQL 
 Expression](../cesql/spec.md).
-The filter result MUST be true if the result value of the expression, 
-coerced to boolean, equals to the `TRUE` boolean value,
-otherwise MUST be false if an error occurred while evaluating the expression or if the result value,
-coerced to boolean, equals to the `FALSE` boolean value.
+The filter result MUST be true if the result value of the expression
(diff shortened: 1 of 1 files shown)


duglin:
LGTM
@lionelvillard any thoughts?
@jskeet too

jskeet:
Yup, that works for me, personally - as someone who's never used CloudEvents SQL at all!

duglin:
Approved on the 10/26 call

PR #1247, opened 2023-11-15
Add missing ext & make at least one attribute REQUIRED for each ext

duglin:
ping @JU-2094


duglin:
Approved on the 11/16 call

PR #1248, opened 2023-11-25
Clarify boolean literals in specification

### LINKED ISSUE #1232: CESQL: clarify `TRUE` and `true`, `FALSE` and `false` are the same
In the current CESQL spec, it is stated that:

> Although in the EBNF keywords are defined using uppercase characters, they are case-insensitive.

Additionally, in the tck tests, we can see that `TRUE` evaluates to `true`, and the same is true for `FALSE` and `false`.

However, this may not be clear to users when reading the section on literals which states:

> CESQL defines 3 different literal kinds: integer numbers, `true` or `false` booleans, and '' or "" delimited strings.

We should clarify that `TRUE` and `true` are both valid, as well as `FALSE` and `false`

MaryamTaj:
Fixes #1232 

## Proposed Changes

Include "TRUE" | "FALSE" when you define boolean literals in the spec.md

**Release Note**

```release-note

```

### DIFF
--- a/cesql/spec.md
+++ b/cesql/spec.md
@@ -117,7 +117,7 @@ CESQL defines 3 different literal kinds: integer numbers, `true` or `false` bool
 digit ::= [0-9]
 number-literal ::= digit+
 
-boolean-literal ::= "true" | "false"
+boolean-literal ::= "true" | "false" (* Case insensitive *)
 
 string-literal ::= ( "'" ( [^'] | "\'" )* "'" ) | ( '"' ( [^"] | '\"' )* '"')
 


cesql/spec.md
```diff
@@ -117,7 +117,7 @@ CESQL defines 3 different literal kinds: integer numbers, `true` or `false` bool
 digit ::= [0-9]
 number-literal ::= digit+
 
-boolean-literal ::= "true" | "false"
+boolean-literal ::= "true" | "false" | "TRUE" | "FALSE"
```
duglin, line 120: Is there a format that means "any case", so that `TrUe` would be acceptable too?

duglin: @MaryamTaj I couldn't find a way... can you add a comment to indicate that it's case insensitive? 
e.g.
```
boolean-literal ::= "true" | "false"       (* Case insensitive *)
```

duglin: @MaryamTaj we chatted about this on today's call... are you ok with the above suggestion?

MaryamTaj: @duglin Thank you so much for letting me know! I'd be happy to make the changes:)

duglin:
LGTM

cesql/spec.md
jskeet: So just to check, this is intended to match True and False (or, less likely, tRUE and fALSE)? I'm pretty sure that's the aim, just checking.

duglin:
Approved on the 12/7 call

PR #1257, opened 2024-02-05
Clarify that missing attributes should not result in errors

### LINKED ISSUE #1230: Clarify whether missing attributes should return an `error` or `false`
Currently, when evaluating a CESQL expression with a missing attribute we return an error. However, the spec does not specify whether we should return an error or whether we should return `false` in this case. The convention of returning an error is coming from the tck tests, which have a `missingAttribute` error. 

From conversation [on slack](https://cloud-native.slack.com/archives/C9DB5ABAA/p1695654435459459?thread_ts=1695218436.855979&cid=C9DB5ABAA) there is a suggestion to return `false` instead of a `missingAttribute` error when there is a missing attribute in a CESQL expression.

Regardless of the decision to return an `error` or `false`, the behaviour should be clarified in the spec. Hence, the scope of this issue is to:
1. Decide whether `error` or `false` should be the result of a missing attribute in a CESQL expression
2. Clarify the spec to explain the result

Cali0707:
Fixes #1230

## Proposed Changes

- Clarify that missing attributes should not result in an error

Note: this is actually something that the SDKs currently implement incorrectly, so if this PR is merged I will go and fix them.

**Release Note**

```release-note
CESQL expressions should not return an error when there is a missing attribute
```

### DIFF
--- a/cesql/spec.md
+++ b/cesql/spec.md
@@ -197,7 +197,8 @@ Unless otherwise specified, every attribute and extension MUST be represented by
 Through implicit type casting, the user can convert the addressed value instances to _Integer_ and
 _Boolean_.
 
-When addressing an attribute not included in the input event, an empty _String_ MUST be assumed as its value.
+When addressing an attribute not included in the input event, the entire subexpression MUST evaluate to `false`.
+For example, `true AND (missingAttribute = "")` would evaluate to `false` as the subexpression `missingAttribute = ""` would be false.
 
 ### 3.3. Errors
 
@@ -205,9 +206,6 @@ Because every operator and function is total, an expression evaluation flow is d
 by expected or unexpected errors. Nevertheless CESQL includes the concept of errors: when an expression is evaluated, in
 case an error arises, the evaluator collects a list of errors, referred in this spec as _error list_, which is then
 returned together with the evaluated value of the CESQL expression.
-
-Addressing an attribute which is missing from the input event MUST NOT return an error due to the missing attribute. Instead, 
-an empty string MUST be assumed as its value, and execution MUST proceed as it would with any other string value.
 
 Whenever possible, some error checks SHOULD be done at compile time by the expression evaluator, in order to prevent
 runtime errors.
--- a/cesql/spec.md
+++ b/cesql/spec.md
@@ -206,7 +206,7 @@ by expected or unexpected errors. Nevertheless CESQL includes the concept of err
 case an error arises, the evaluator collects a list of errors, referred in this spec as _error list_, which is then
 returned together with the evaluated value of the CESQL expression.
 
-Addressing an attribute which is missing from the input event MUST not return an error due to the missing attribute. Instead, 
+Addressing an attribute which is missing from the input event MUST NOT return an error due to the missing attribute. Instead, 
 an empty string MUST be assumed as its value, and execution MUST proceed as it would with any other string value.
 
 Whenever possible, some error checks SHOULD be done at compile time by the expression evaluator, in order to prevent
--- a/cesql/spec.md
+++ b/cesql/spec.md
@@ -206,6 +206,9 @@ by expected or unexpected errors. Nevertheless CESQL includes the concept of err
 case an error arises, the evaluator collects a list of errors, referred in this spec as _error list_, which is then
 returned together with the evaluated value of the CESQL expression.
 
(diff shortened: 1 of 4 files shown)


duglin:
@lionelvillard any comments?

duglin:
@slinkydeveloper any comment?  Howdy! :-)

duglin:
Not deep into the cesql stuff, but it seems like there are a few options for `missingAttr=""` (aside from erroring):
1 - returns `true` because we treat `missingAttr` the same as `""`. Which means `missingAttr<>""` returns `false`
2 - returns `false` because we treat any SINGLE expression with an unknown attribute as `false` (so `missingAttr<>""` would also return `false`, but `NOT(missingAttr<>"")` returns `true`)
3 - returns `false` because we treat an ENTIRE complex expression with an unknown attribute as `false`, so `missingAttr="" or true` returns `false`

What does `missingAttr = 0` yield? I think the spec says to convert `missingAttr` to an int...but converting "" to an int generates an error, no? I can't seem to find it in the spec, but does an error get mapped to `false`? If so, is it for just that single expression or for the entire complex expression? (2 vs 3 in the list above)

Cali0707:
Hmm you bring up a good point @duglin - one other option would be to have the missing attribute cast to the "default" value for the SINGLE expression it is found within.

For example, if we had `missingAttr = ""` it would default to a string `""`, and if we had `missingAttr = 0`, it would default to `0`. I think this approach would allow people to still make comparisons ( and generally avoid having a special case when evalutating the expression as we already have type casting logic ).

However, I also like the idea of having the SINGLE expression with the unknown attribute evaluate to false, because then expressions like `(type = "my.example.type") OR (missintattr = 50` would evaluate to the result I expect (the first expression already returned true).

WDYT?

duglin:
I didn't offer a suggested answer in my previous comment because I just don't know yet :-)  Way too many options and they all seem to have pros/cons.

Let me walk thru some samples and see what I think might be the right answer for each (xxx=missingAttribute):
| expression | expected answer? |
| --- | --- |
| xxx=true | false |
| xxx=false | false. At first this makes sense, but down below "xxx" by itself is false, so is it odd that this isn't true? |
| xxx=5 | false |
| xxx=0 | false |
| (xxx=5 or true) | true due to the "or true" |
| xxx="abc" | false |
| xxx="" | false but could feel odd |
| not(xxx=5) | true due to xxx=5 returning false|
| not(xxx="") | true due to xxx="" returning false|
| xxx | false - alias for exists(xxx) ? |
| not(xxx) | true due to xxx returning false |
| xxx or false | false since both are false |

If the above makes sense, then I guess I'm leaning towards the current single operation (if there is one) returning `false`

Note that people can always use the `exists` function if they don't like whatever semantics we come up with.
e.g. xxx="" might need to be: `not(exists(xxx)) or xxx=""` if they want "missing" to be the same as "".

Cali0707:
@duglin I think I like the idea of the current single operation returning `false`. 

> e.g. xxx="" might need to be: `not(exists(xxx)) or xxx=""` if they want "missing" to be the same as "".

Maybe here we could follow normal SQL and provide a `COALESCE` function, which returns the value of the first espression that is non null (in our case, the first expression which is not a missing attribute).

So, someone could write `COALESCE(xxx, "")` to get the desired default behaviour

cesql/spec.md
```diff
@@ -197,7 +197,8 @@ Unless otherwise specified, every attribute and extension MUST be represented by
 Through implicit type casting, the user can convert the addressed value instances to _Integer_ and
 _Boolean_.
 
-When addressing an attribute not included in the input event, an empty _String_ MUST be assumed as its value.
+When addressing an attribute not included in the input event, the entire subexpression MUST evaluate to `false`.
+For example, `true AND (missingAttribute = "")` would evaluate to `false` as the subexpression `missingAttribute = ""` would be false.
```
duglin, line 201: It might be clearer to say "the subexpression referencing the missing attribute MUST evaluate to 'false'". Use of the word "entire" could be easily misread as the "entire thing".... like I did at first 🤣

duglin:
Approved (with edit) on the 2/22 call.

@Cali0707 let me know when to merge

Cali0707:
@duglin should be good now! I just updated the wording as you suggested. Thanks for the suggestion, I think it's much clearer now!

duglin:
@Cali0707 thanks

PR #1261, opened 2024-02-22
Suggestion about Structuring a Composite "subject" Field

evarga:
## Background

The core specification lacks any recommendation how to cope with a composite `subject` field. Leaving everything to a producer may create an undesirable situation of packing, even semantically similar, `subject` content in arcane manner. 

## Proposed Changes

This PR adds an additional example, suggesting a different way to structure the content of the `subject` field, while fully preserving compatibility with the existing standard.

### DIFF
--- a/cloudevents/spec.md
+++ b/cloudevents/spec.md
@@ -419,7 +419,7 @@ on the definition of OPTIONAL.
 #### subject
 
 - Type: `String`
-- Description: This describes the subject of the event in the context of the
+- Description: This identifies the subject of the event in the context of the
   event producer (identified by `source`). In publish-subscribe scenarios, a
   subscriber will typically subscribe to events emitted by a `source`, but the
   `source` identifier alone might not be sufficient as a qualifier for any
@@ -450,9 +450,9 @@ on the definition of OPTIONAL.
     subscription scope (CRM part of an eCommerce system), the `type` identifies
     the "client updated" event, and the `id` uniquely identifies the event
     instance to distinguish separate occurrences of a same client being
-    updated multiple times; the affiliate partner and client ids of the updated
+    updated multiple times; the affiliate partner id and client id of the updated
     client are carried in `subject`. Here, it is a composite string
-    (affiliate partner id:client id) that is made up of one or more identifying
+    (`affiliate partner id:client id`) that is made up of one or more identifying
     pieces of data separated by some symbol (like, colon), assuming that each
     affiliate partner has its own client base:
       - `source`: `https://example.com/eCommerce/crm`
--- a/cloudevents/spec.md
+++ b/cloudevents/spec.md
@@ -450,11 +450,9 @@ on the definition of OPTIONAL.
     subscription scope (CRM part of an eCommerce system), the `type` identifies
     the "client updated" event, and the `id` uniquely identifies the event
     instance to distinguish separate occurrences of a same client being
-    updated multiple times; the affiliate partner id and client id of the updated
-    client are carried in `subject`. Here, it is a composite string
-    (`affiliate partner id:client id`) that is made up of one or more identifying
-    pieces of data separated by some symbol (like, colon), assuming that each
-    affiliate partner has its own client base:
+    updated multiple times; the `subject` uniquely identifies the client within
+    the scope of the `source` by including a "partner id" and "client id"
+    (which is unique within the scope of the "partner id") separated by a colon:
       - `source`: `https://example.com/eCommerce/crm`
       - `subject`: `5:100`
 
--- a/cloudevents/spec.md
+++ b/cloudevents/spec.md
@@ -454,7 +454,7 @@ on the definition of OPTIONAL.
     the scope of the `source` by including a "partner id" and "client id"
     (which is unique within the scope of the "partner id") separated by a colon:
       - `source`: `https://example.com/eCommerce/crm`
-      - `subject`: `5:100`
+      - `subject`: `partnerid/5/clientid/100`
 
 #### time
 
--- a/cloudevents/spec.md
+++ b/cloudevents/spec.md
@@ -448,7 +448,7 @@ on the definition of OPTIONAL.
   - Building on the previous example, a subscriber might register interest for when new
      blobs are created inside a blob-storage container. However, this time both the name
      and/or size of the newly created blob are relevant for decision making purposes.
-     In this case, the `subject` may simply list meta-properties (including custom extensions),
+     In this case, the `subject` MAY simply list meta-properties (including custom extensions),
      separated with comma, containing additional data:
        - `source`: `https://example.com/storage/tenant/container`
        - `subject`: `myextensionfilename,myextensionfilesize`
--- a/cloudevents/spec.md
+++ b/cloudevents/spec.md
@@ -436,15 +436,24 @@ on the definition of OPTIONAL.
 - Constraints:
   - OPTIONAL
   - If present, MUST be a non-empty string
-- Example:
+- Examples:
   - A subscriber might register interest for when new blobs are created inside a
     blob-storage container. In this case, the event `source` identifies the
     subscription scope (storage container), the `type` identifies the "blob
     created" event, and the `id` uniquely identifies the event instance to
     distinguish separate occurrences of a same-named blob having been created;
     the name of the newly created blob is carried in `subject`:
-    - `source`: `https://example.com/storage/tenant/container`
-    - `subject`: `mynewfile.jpg`
+      - `source`: `https://example.com/storage/tenant/container`
+      - `subject`: `mynewfile.jpg`
+  - Building on the previous example, a subscriber might register interest for when new
+     blobs are created inside a blob-storage container. However, this time both the name
+     and/or size of the newly created blob are relevant for decision making purposes.
+     In this case, the `subject` may simply list meta-properties (including custom extensions),
+     separated with comma, containing additional data:
+       - `source`: `https://example.com/storage/tenant/container`
+       - `subject`: `myextensionfilename,myextensionfilesize`
+       - `myextensionfilename`: `mynewfile.jpg`
+       - `myextensionfilesize`: `100MB`
 
 #### time
 
--- a/cloudevents/spec.md
+++ b/cloudevents/spec.md
@@ -446,14 +446,14 @@ on the definition of OPTIONAL.
       - `source`: `https://example.com/storage/tenant/container`
       - `subject`: `mynewfile.jpg`
   - Building on the previous example, a subscriber might register interest for when new
-     blobs are created inside a blob-storage container. However, this time both the name
-     and/or size of the newly created blob are relevant for decision making purposes.
-     In this case, the `subject` MAY simply list meta-properties (including custom extensions),
-     separated with comma, containing additional data:
-       - `source`: `https://example.com/storage/tenant/container`
-       - `subject`: `myextensionfilename,myextensionfilesize`
-       - `myextensionfilename`: `mynewfile.jpg`
-       - `myextensionfilesize`: `100MB`
+    blobs are created inside a blob-storage container. However, this time both the name
+    and/or size of the newly created blob are relevant for decision making purposes.
+    In this case, the `subject` MAY simply list meta-properties (including custom extensions),
+    separated with comma, containing additional data:
+      - `source`: `https://example.com/storage/tenant/container`
+      - `subject`: `myextensionfilename,myextensionfilesize`
+      - `myextensionfilename`: `mynewfile.jpg`
+      - `myextensionfilesize`: `100MB`
 
 #### time
 
--- a/cloudevents/spec.md
+++ b/cloudevents/spec.md
@@ -446,14 +446,12 @@ on the definition of OPTIONAL.
       - `source`: `https://example.com/storage/tenant/container`
       - `subject`: `mynewfile.jpg`
   - Building on the previous example, a subscriber might register interest for when new
-    blobs are created inside a blob-storage container. However, this time both the name
+    blobs are created inside a blob-storage container. Here, both the name
     and/or size of the newly created blob are relevant for decision making purposes.
-    In this case, the `subject` MAY simply list custom extensions, separated with comma,
-    containing additional data:
+    In this case, the `subject` MAY simply list multiple values, separated with comma,
+    adorned with custom prefixes:
       - `source`: `https://example.com/storage/tenant/container`
-      - `subject`: `myextensionfilename,myextensionfilesize`
-      - `myextensionfilename`: `mynewfile.jpg`
-      - `myextensionfilesize`: `100MB`
+      - `subject`: `filename:mynewfile.jpg,filesize:100MB`
 
 #### time
 
--- a/cloudevents/spec.md
+++ b/cloudevents/spec.md
@@ -445,13 +445,18 @@ on the definition of OPTIONAL.
     the name of the newly created blob is carried in `subject`:
       - `source`: `https://example.com/storage/tenant/container`
       - `subject`: `mynewfile.jpg`
-  - Building on the previous example, a subscriber might register interest for when new
-    blobs are created inside a blob-storage container. Here, both the name
-    and/or size of the newly created blob are relevant for decision making purposes.
-    In this case, the `subject` MAY simply list multiple values, separated with comma,
-    adorned with custom prefixes:
-      - `source`: `https://example.com/storage/tenant/container`
-      - `subject`: `filename:mynewfile.jpg,filesize:100MB`
+  - A subscriber might register interest for when new updates are made to a client
+    in an eCommerce system. In this case, the event `source` identifies the
+    subscription scope (CRM part of an eCommerce system), the `type` identifies
+    the "client updated" event, and the `id` uniquely identifies the event
+    instance to distinguish separate occurrences of a same client being
+    updated multiple times; the affiliate partner and client ids of the updated
+    client are carried in `subject`. Here, it is a composite string
+    (affiliate partner id:client id) that is made up of one or more identifying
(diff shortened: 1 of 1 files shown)


cloudevents/spec.md
```diff
@@ -436,15 +436,24 @@ on the definition of OPTIONAL.
...
+      - `subject`: `mynewfile.jpg`
+  - Building on the previous example, a subscriber might register interest for when new
+    blobs are created inside a blob-storage container. However, this time both the name
+    and/or size of the newly created blob are relevant for decision making purposes.
+    In this case, the `subject` MAY simply list custom extensions, separated with comma,
+    containing additional data:
+      - `source`: `https://example.com/storage/tenant/container`
+      - `subject`: `myextensionfilename,myextensionfilesize`
+      - `myextensionfilename`: `mynewfile.jpg`
+      - `myextensionfilesize`: `100MB`
```
duglin, line 456: While subject can be just about any string, I can't help but wonder whether anyone who would know how to interpret `myextensionfilename` and `myextensionsize` attributes would need the extra hint of including them in the `subject`. Why not just look for them w/o examination of the `subject`, thus leaving `subject` to be what it was intended for... the identifier of the thing the event was about w/o extra metadata (or metadata pointers).
My concern with this indirection is that it's not clear how someone would know it's an indirection vs the actual subject? Especially if there's no `,` in there.

evarga: Thanks @duglin for your comment!

The semantics of the `subject` field would not change, it would still remain to be _some sort of an identifier of the thing the event was about_. The extra level of indirection could be relevant for couple of reasons as follows:

- Any subscriber would be able to discern what set of additional meta-data uniquely describe the source of an event, even those who are not familiar with a producer of a message.
- Having custom extensions specified in unified fashion brings extra clarity and order even into the `subject` content. For example, if `subject` is just 100 does it designate a client id or a file size? Stating that `subject` is `myextensionfilesize` and putting that 100 there it is clear what it is about. The whole point is making events capable of outliving producers and this is the idea behind such an extra indirection.

Parsing the `subject` property would be simple, since if it contains stuff that are custom extensions, then the meaning is clear.

I have only proposed an extra example about how changing perspective on the content of the `subject` may open up new possibilities. This is the best tesimony of CloudEvents flexibility and versatility. I hope this additional explanation brings more clarity about the intent.

evarga: @duglin Let me expand on the 3rd bullet point in my edited comment above. Leaving parts of the specification, like the `subject` field, as any string is a double-edged sword. Without any further steps from an enterprise, developers will sooner or later, intentionally, or not, start abusing it, since it may hold anything. Extreme schedule pressures, need for urgent production fixes would just rise the chance of hacking something into `subject` to make things work. 

I am not saying that a specification should give answers to all use cases nor be stringent about every aspect of it but providing an extra example, sparkling creative approaches of structuring and handling the `subject`, could help get the ball rolling.

jskeet: I'd argue that this really *is* changing the semantics of `subject`. It's described as:

> Description: This describes the subject of the event in the context of the event producer

In your proposal, the value `myextensionfilename,myextensionfilesize` *doesn't* describe the subject of the event. It describes other attributes. Two events for entirely different subjects within the same source could have the same value. (And indeed, for any given source I'd expect the `subject` value under your proposal to basically end up looking the same, which defeats the point of `subject` itself.)

Rather than try to use an existing attribute to mean something different, I would recommend proposing an additional extension attribute (e.g. `subjectattributes`) which would be specifically designed to say "Hey consumer, you should look at these other attributes because they provide extra metadata about the subject".

duglin: I would much prefer a secondary attribute with the list as @jskeet suggests - it keeps subject's value unambiguous. However, I would still question the need for that. Let's take the example of `myextensionfilesize`. Someone would need to define it and it would probably be defined as something like `the size of the file after the event`.  Any event that includes this attribute is providing the receiver with this info and the receiver can look for this attribute if they want to. Absence of the attribute means the info is not provided. Having a `subjectattributes` that lists them seems redundant. If the thought is that "the list" makes it clear that `myextensionfilesize` is related to the `subject` and not something else, then I would question the definition of that extension because if it could be used for more than one purpose it's probably not well defined enough - and it should be called `sourcefilesize`.

evarga: @jskeet I don't think that my proposal is changing anything in semantics of `subject`. According to the current CloudEvents specification the `subject` is anything that may adorn the `source` with extra contextual information about an event using meta-data as a carrier instead of placing the stuff into the `data` section. I have only proposed an additional example to be added to the specification hinting to users, that they are free to change perspectives in how they may control and leverage the `subject` property.

Would anything change in semantics if I organized a composite `subject` as follows:

- `subject`: `myextensionfilename:mynewfile.jpg,myextensionfilesize:100MB`

Whether someone would embed values into `subject` or simply use an additional indirection is more a matter of style rather than change of semantics. I don't quite understand why differently organizing the content of the `subject`, whilst still serving the same purpose of embellishing the `source`, would cause an issue.

Here is the crucial quite from the specification:

> ...the `source` identifier alone might not be sufficient as a qualifier for any specific event if the `source` context has **internal sub-structure**.

This advertised internal sub-structure entails that the `subject` must be capable of representing it in some fashion. Only a flat structure (pure value), as currently given in the example with `subject`: `mynewfile.jpg` is not enough. Therefore, my example was just one possible idea how `subject` could address the previously mentioned objective using indirect references. The exact form is not even that important. The point is just to remind users that the content of `subject` is not limited and mandated to be something bare simple.

jskeet: > According to the current CloudEvents specification the subject is anything that may adorn the source with extra contextual information about an event using meta-data as a carrier instead of placing the stuff into the data section.

No, that's not what the spec says. As I wrote before, the spec says: "This describes the subject of the event in the context of the event producer". That's not the same as "extra contextual information about an event" - because it's specifically meant to be information about the *subject* of the event.

> Would anything change in semantics if I organized a composite subject as follows: `subject: myextensionfilename:mynewfile.jpg,myextensionfilesize:100MB`

Yes, that would be *much* better, because then the attribute value contains information *about the subject* rather than about the rest of the event.

> This advertised internal sub-structure entails that the subject must be capable of representing it in some fashion. 

Right, but that's not what the original proposal suggests - it suggests encoding information about the rest of the attributes, *not* about the subject itself, within the `subject` attribute.

evarga: @jskeet OK, I perhaps wrongly phrased my sentence about extra contextual information, but I meant to say the same thing as you have cited.

> Yes, that would be much better, because then the attribute value contains information about the subject rather than about the rest of the event.

If I change my proposal to follow this approach would that be acceptable as an additional example to the specification?

jskeet: > If I change my proposal to follow this approach would that be acceptable as an additional example to the specification?

I'll need to think it a bit more, but I *think* so (personally - I wouldn't want to speak for anyone else).

evarga: @duglin I agree that such extra attribute to list additional ones would be redundant. Please, take a look at my answer to @jskeet above by embedding values into `subject` with extra prefix what that value is about.

evarga: @jskeet, @duglin Just for the sake of completeness, I've altered my proposal according to the results of our discussion up so far and made a new commit.

duglin: I'm willing to admit that I may need my head shaped a bit to align my thinking on this :-) but as of now I'm struggling with this from a few perspectives:

The spec says the source is used to `Identifying the subject of the event in context metadata`. While not explicit, to me this is akin to saying it should be something like the ID of the entity. I say this because it seems to me that someone who gets two events about "the same source" should be able to do a strcmp of the two events `source` attributes to know if it's about the same entity. Well, if the size of the file changed between the two events then the strcmp will fail, and that seems like it's hurting some of the value proposition of `source`. IOW, `source` is no longer unique per source, it's (maybe) unique per "source version" - and I don't think that was the intent.

Now, it's possible to get what I want but ONLY IF I do some extra parsing of the `source` field and break it down into sub-parts, AND know which sub-part is the `ID` (`filename` in this case). How as a receiver do I know that?

I think it would be good to back-up a bit and ask why this extra info is needed? I'm going to try to read @evarga's mind and say it's because in this example, the receiver wants to route the event based on the new file size, not just its name. if that's right, then between these two choices:
```
"source": "filename=myfile, size=100"
```
and
```
"source": "myfile",
"filesize": 100
```
I don't see why the 2nd one isn't clearer because each attribute serves a single purpose and the info of interest is available extra parsing. Additionally, it allows people who don't care about the size to not even know it is present in the event ... meaning they can ignore "filesize" and don't need to do extra parsing of "source".

I would assert that any receiver that knows about the special parsing of "source" AND knows about the "size" sub-part, would be just as happy (if not happier) to extract the "100" from a dedicate attribute than from "source" because it's less work.

evarga: @jskeet, @duglin
FYI: I have edited my previous post and reordered the options to summarize the current state of affairs.

evarga: @duglin Let me cite the current specification again, just to emphasize how the concept of a `subject` is pretty vague and why I have concerns and disagreement with your statements:

> In the above example, the subscriber might **only** be interested in blobs with names ending with '.jpg' or '.jpeg' and the subject attribute allows for constructing a simple and efficient string-suffix filter for that subset of events.

In this example, a consumer is not even interested in to look at the file name itself. Is this fact about image type _identifying_ or _describing_ a `subject`? Would a `subject` stop identify the subject of the event by simply being `jpg` in this case? What if a consumer is only interested in a file size?

The specification goes on and gives an example of:

`subject`: `mynewfile.jpg`

Does `mynewfile.jpg` instead of `jpg` a more simple and efficient string-suffix filter for a consumer that doesn't care about file name? Would it be a concern breaking up a file name and image type (jpg or whatever) into separate constituent parts of `subject` a semantic change or perhaps bad practice? Why is it a good practice to give people an example in the CE spec. having these things bundled together? 

> ...but I don't think choosing "filesize" is a valid choice at all - there's nothing unique or stable about that value w.r.t. "identifying" the subject. It's a poor choice. "filename" is the only valid choice in this example.

That same file name that is not even used by a consumer in the current CE example, right? Would letting people put `mynewfile.jpg` and `mynewfile.jpeg` help "stabilize" a `subject`? Are file extensions unique and stable _w.r.t. "identifying" the subject_, if nothing else matters to consumers?

> Plus the obvious use of the word subject in it :-)

I assume you meant to say _nonobvious_ instead of _obvious_. 😃 

> To be clear, people are free to stick whatever values they want in the these fields (as long as they're compliant with the spec). And while I'm still leaning towards this being a non-compliant use, my bigger concern with this PR is that I don't think we should be encouraging people to overload subject in this way via an example in the spec.

Based on all previous messages related to this PR, my impression is that currently no clear idea nor consensus exist what is a compliant and non-compliant usage of `subject`. Sorry.

> ...that this mechanism is being used is going to hurt interop.

Interoperability will be jeopardized the most by letting people cope with these aspects of `subject` on their own, as currently the specification gives little help here. Plus the obvious confusion around the word subject in it.

duglin: I think you're misinterpreting the `mynewfile.jpg` example. That value is used because it is the identity of the subject, and it is put there w/o the producer knowing which parts of that identity the consumer is interested in. So, even if the consumer is only interested in the suffix I would still expect the subject to be the full filename. Now, if you want to say that it should be ok for the subject to be serialized as `base:mynewfile, suffix:jpg` then I think that's ok - odd, but ok, because it is still conveying the identity of it. Filesize, in this example, has no bearing on it's identity. I could have an intermediary "zero out" the fileSize and the identity would not change - it might now be mis-routed in your system, but not due to a misaligned identity, but rather due to a misaligned "extra metadata checking". And that's a different situation.

> Are file extensions unique and stable w.r.t. "identifying" the subject, if nothing else matters to consumers?

Not the point. It doesn't matter what the consumer does. If they choose to just use the suffix or just the 3rd character of the fileName, that's up to them. What matters from a spec perspective is what the producer sends - and that needs to be the identity, not a mixture of identity and other non-identifying data about it.

You never did answer my questions...
- what breaks down if you put fileSize in a new extension attribute?
- how can a receiver know whether `filename=myfile,size=100` is a composite subject that needs to be broken-up, or whether that actually is the real subject's identity (e.g. filename) ? After all, that is a valid filename.

btw - I did mean `obvious` in the sense that an extension named `subjectfilesize` _obviously_ is the "filesize" of the "subject".

---

I reread your opening statement:
```
The core specification lacks any recommendation how to cope with a composite subject field
```
trying to better understand why you think there's an interop problem. Let me see if this gets closer...
- the spec might be implying that the subject is always just a simple (and single) ID-like field. E.g. a name, or UUID
- you see cases where what "identifies" the subject isn't that simple and it might be made up of more than one "thing". 
- so, you'd like for people to know that it's ok to include multiple "things" in the string that identifies the subject

If that's getting closer, then if what you're looking for text (or a sample) in the spec that makes it clear that the subject string: `could be a composite string that is made up of one or more identifying pieces of data about the subject` then I think that's ok. And how the producer chooses to serialize those pieces is up to it. For example, if we showed an example that shows a subject is made of up a two part string, DBname + `:` + the row's unique employee number, then I think that's ok.

What I'm less comfortable with is an example that includes one of those "parts" being data that is clearly not used for identification purposes. Other "interesting" data is not what subject is for - a new extension should be used instead. Even in your example you don't say "size" is used for identifying the source, you just say it's used for "decision making purposes" which I interpret as something like "routing". And any of the CE attributes could be used for that w/o being merged into "subject".

evarga: @doug In retrospect, I admit that file size was a misfortunate attribute to convey the idea of having a composite `subject` consisted from multiple identifying pieces. I should have simply used `affiliatepartnerid:clientid` as an example. All these pieces are related to identifying the subject, like, `filename:extension`.

> If that's getting closer, then if what you're looking for text (or a sample) in the spec that makes it clear that the subject string: could be a composite string that is made up of one or more identifying pieces of data about the subject then I think that's ok. And how the producer chooses to serialize those pieces is up to it. For example, if we showed an example that shows a subject is made of up a two part string, DBname + : + the row's unique employee number, then I think that's ok.

Exactly! This is what I would like to put as an additional example. I've altered my proposal accordingly and made a new commit.

evarga: @duglin, @jskeet I hope that my last comment (I've edited it to include one more topic) and the latest commit fixes all issues that we have found related to `subject`.

cloudevents/spec.md
```diff
@@ -436,15 +436,27 @@ on the definition of OPTIONAL.
...
+  - A subscriber might register interest for when new updates are made to a client
+    in an eCommerce system. In this case, the event `source` identifies the
+    subscription scope (CRM part of an eCommerce system), the `type` identifies
+    the "client updated" event, and the `id` uniquely identifies the event
+    instance to distinguish separate occurrences of a same client being
+    updated multiple times; the affiliate partner id and client id of the updated
+    client are carried in `subject`. Here, it is a composite string
+    (`affiliate partner id:client id`) that is made up of one or more identifying
+    pieces of data separated by some symbol (like, colon), assuming that each
+    affiliate partner has its own client base:
```
duglin, line 457: I'd prefer if it said something like (on lines 453-457):
- - -
the `subject` uniquely identifies the client within the scope of the `source` by including a "partner id" and "client id" (which is unique within the scope of the "partner id") separated by a colon.
- - -

This makes it clear that the value can be a composite value, but also makes it clear that each sub-part is there to help uniquely identify the subject and not just other interesting info.

TBH I'm still not sure all of this is needed since the same concerns could be said about any other attribute that identifies something (like "source" or even "id" itself) - those can be composites too.

evarga: > I'd prefer if it said something like (on lines 453-457):

Done.

> TBH I'm still not sure all of this is needed since the same concerns could be said about any other attribute that identifies something (like "source" or even "id" itself) - those can be composites too.

@duglin Don't forget that I've also fixed the word _describes_ into _identifies_ at line 422, an inconsistency in the specification that has generated a lots of confusion. Consequently, this extra example would help people understand the true intention behind `subject` and how to cope with a composite content.

The `source` is by its very nature following a composite (hierarchical) structure since it is an URI reference. An `id` is usually opaque and rarely, if ever, parsed, so its content doesn't really matter except for providing uniqueness.

duglin: yes I do agree with the s/describes/identifies/ change :-)

evarga: > yes I do agree with the s/describes/identifies/ change :-)

@duglin As the proverb goes: _"Little by little, a little becomes a lot.”_ At the end, I am glad that all these _little_ messages in this PR have revealed a _lot_, that assumptions about `subject` were not fully aligned with facts from the specification. :-)

cloudevents/spec.md
```diff
@@ -436,15 +436,25 @@ on the definition of OPTIONAL.
...
+  - A subscriber might register interest for when new updates are made to a client
+    in an eCommerce system. In this case, the event `source` identifies the
+    subscription scope (CRM part of an eCommerce system), the `type` identifies
+    the "client updated" event, and the `id` uniquely identifies the event
+    instance to distinguish separate occurrences of a same client being
+    updated multiple times; the `subject` uniquely identifies the client within
+    the scope of the `source` by including a "partner id" and "client id"
+    (which is unique within the scope of the "partner id") separated by a colon:
+      - `source`: `https://example.com/eCommerce/crm`
+      - `subject`: `5:100`
```
duglin, line 457: @evarga, @jskeet  had a minor suggested edit on today's call:
```
`subject`: `partnerid/5/clientid/100`
```
so it's a bit more descriptive. Would you be ok with this?

duglin:
approved on the 2/29 call. Just waiting for one possible last minute tweak

duglin:
thanks @evarga 
Merging

PR #1264, opened 2024-02-29
move documented-extensions to extensions/README.md

duglin:
we were missing a README in this dir so I think it makes more sense to move our documented-extensions file into the dir to become the README


duglin:
Approved on the 3/7 call

PR #1267, opened 2024-03-12
CESQL OR short-circuit evaluation

### LINKED ISSUE #1260: Clarify CESQL AND and OR evaluation
In the subscriptions spec when talking about the CESQL filter it states that the result `MUST be false if an error occurred while evaluating the expression`.

This can cause issues because while CESQL expressions are total (they always return a result, and optionally an error), if an error occurs during the evaluation of an expression using an AND or an OR, the result of the filter would depend not only on whether or not the expression evaluated to true but how quickly it evaluated to true and whether it evaluated far enough into the expression to encounter an error. For example:
```sql
(type = "my.event.type") OR (numericAttribute / 0)
```
If the implementation decides to return early when evaluating an OR it would never encounter an error and so the filter would pass. However, if the implementation evaluates the entire expression it would then get the arithmetic error when the division by 0 occurs.

To ensure consistency of CESQL behaviour when used as a filter, we should specify whether the implementations of CESQL must return early when possible, or must not return early.

Cali0707:
Fixes #1260 

## Proposed Changes

- OR expressions MUST be short-circuit evaluated

**Release Note**

```release-note
CESQL OR expressions are now required to be short-circuit evaluated
```

### DIFF
--- a/cesql/spec.md
+++ b/cesql/spec.md
@@ -258,6 +258,8 @@ The modulo and divisions MUST follow the [truncated divisions definition][modulo
 - The remainder of the modulo MUST have the same sign as the dividend.
 - The quotient MUST be rounded towards zero, _truncating_ the decimal part.
 
+The OR operator MUST be short-circuit evaluated. This means that evaluation of the operation MUST stop as soon as a `true` value is encountered.
+
 #### 3.4.3. Like operator
 
 | Definition                                       | Semantics                                           |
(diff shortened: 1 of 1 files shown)


duglin:
LGTM
do any testcases need to change because of this?

Cali0707:
> do any testcases need to change because of this?

Not off the top of my head, I will look closely at them tomorrow though.

duglin:
Approved on today's call HOWEVER @jskeet asked if we should short-circuit AND (meaning stop on first 'false'). @Cali0707  wanna take care of that in a new PR?

Cali0707:
@duglin I’d be happy to add that as a new PR 👍

duglin:
cool thanks - merging....

PR #1268, opened 2024-03-14
First pass at considering content-type in a case-insensitive way.

jskeet:
We don't note what "kind" of case insensitivity to consider though - should we only consider ASCII values as matching
"application/cloudevents" for example? (Case insensitivity is hard, basically...)

There may be other aspects I've missed, or I may have been too aggressive in imposing case-insensitivity in some bindings. Thoughts very welcome.

### DIFF
--- a/cloudevents/SDK.md
+++ b/cloudevents/SDK.md
@@ -128,6 +128,10 @@ and encoding:
   format, both in the [table below](#feature-support) and in any SDK-specific
   documentation provided.
 
+Note that when decoding an event, media types MUST be matched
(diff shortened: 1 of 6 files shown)


jskeet:
2045: https://www.rfc-editor.org/rfc/rfc2045.html#section-5.1

- Matching of media type and subtype is ALWAYS case-insensitive.
- Matching of attributes is ALWAYS case-insensitive.

jskeet:
@duglin I've made the change, and also renamed the commit to avoid the "first pass" aspect.

duglin:
Approved on the 3/21 call

PR #1271, opened 2024-03-20
Typo: filter->filters

### LINKED ISSUE #1219: Filters
Reviewing this [section](https://github.com/cloudevents/spec/blob/main/subscriptions/spec.md#324-filters)

> The `filter` property in a subscription is a set of filter expressions, where each expression evaluates to either true or false for each event generated.
> 
> If any of the filter expressions in the set evaluate to false, the event MUST NOT be sent to the sink. If all the filter expressions in the set evaluate to true, the event MUST be attempted to be delivered.

We are trying to get the meaning of _filter is a set of filter expressions_. From the snippets we understand that `filter` is an object and not an array  which implies I can write something like this
```
{
  "filter": {
    "exact": {},
    "not": {},
    "suffix": {}
  }
}
```
Each of those 3 expressions need to resolve to TRUE for the event to be sent to the sink. This overlaps with the ALL filter with the difference that keys might not be unique which might be a problem with some framework or json parsers. A better syntax would be to allow only one dialect at the root level this would require ALL dialect if for some reason a given dialect needs to appear more than once.

duglin:
Fixes #1219

### DIFF
--- a/subscriptions/spec.md
+++ b/subscriptions/spec.md
@@ -726,9 +726,9 @@ to enable delivery of events to a target.
 #### 3.2.4 Filters
 
 Filters allow for subscriptions to specify that only a subset of events are to
-be delivered to the sink based on a set of criteria. The `filter` property in a
-subscription is a set of filter expressions, where each expression evaluates to
-either true or false for each event generated.
+be delivered to the sink based on a set of criteria. The `filters` property in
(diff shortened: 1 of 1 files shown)


jskeet (approved):

duglin:
Approved on the 3/28 call

PR #1278, opened 2024-04-18
Adding extension for BAM

katallaxie:
Fixes #

BAM is a method to track the progress of your business processes via the occurrence of business events in your organization. Famously, many companies used BizTalk to implement BAM in their organization, but with open standards like CloudEvents this is changing. This extension is necessary, as the payload of an event is processed by primary systems, like payment or ERP and the payload of these events should not change; but the progress should be tracked in an BAM system for business users to observe the processing and to control these transactions.

## Proposed Changes

- Add extension with attributes that support the implementation or use of a BAM system in an enterprise.

**Release Note**

```release-note
Adds the business activity monitoring (BAM) known extension
```


cloudevents/extensions/bam.md
```diff
@@ -0,0 +1,96 @@
...
+described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+However, the scope of these key words is limited to when this extension is
+used. For example, an attribute being marked as "REQUIRED" does not mean
+it needs to be in all CloudEvents, rather it needs to be included only when
+this extension is being used.
+
+## Attributes
+
+### businessTransactionId
```
duglin, line 27: We require attribute names to be all lowercase since they can appear in case insensitive locations, like http headers

cloudevents/extensions/bam.md
```diff
@@ -0,0 +1,96 @@
...
+
+- Type: `String`
+- Description: A unique identifier for the transaction.
+- Constraints
+  - REQUIRED
+  - MUST be a non-empty string
+  - RECOMMENDED as monotonically increasing and contiguous identifier that is lexicographically-sortable.
```
duglin, line 34: Can you wrap all lines at 80 columns?

cloudevents/extensions/bam.md
```diff
@@ -0,0 +1,96 @@
...
+  - MUST be a non-empty string
+  - RECOMMENDED a alphanumeric string that contains non-whitespace characters and only hyphens, underscores, and periods.
+
+### businessProcessTransactionStepStatus
+
+- Type: `String`
+- Description: The status of the business process transaction step associated with the event.
+- Constraints
+  - OPTIONAL
+  - MUST be a non-empty string
```
duglin, line 69: minor thing, add "if present, " to the start of lines like this where the attribute is OPTIONAL

duglin:
All, please review before tomorrow's call

cloudevents/extensions/bam.md
```diff
@@ -0,0 +1,117 @@
...
+This extension can be used in cases in which a business activity monitoring
+system is used to monitor the progress of a business process, and the events
+generated by the process are used to track the progress of the process.
+Usually these systems have their own modelling language to describe the
+business process, and the events generated by the process
+are used to track the progress of the process.
+
+## References
+
+- [Gartner Business Activity Monitoring](https://www.gartner.com/en/information-technology/glossary/business-activity-monitoring-bam)
```
duglin, line 114: I get a 404 for this

cloudevents/extensions/bam.md
```diff
@@ -0,0 +1,117 @@
...
+
+- Type: `String`
+- Description: A unique identifier for the transaction.
+- Constraints
+  - REQUIRED
+  - MUST be a non-empty string
+  - RECOMMENDED as monotonically increasing and contiguous identifier
+    that is lexicographically-sortable.
+
+### businessprocessid
```
duglin, line 38: Is there a doc we can point to that explains the relationship of the various entities described here?
For example, do process-transactions live within processes which live within transactions? So it's a M:N:1 relationship ?

duglin:
We discussed this on yesterday's call and overall there's no concern with add it, however @JemDay was wondering about the length of some of those attributes. Would it be possible to use `bam` in there to shorten them?

Just tossing this out there:
```
bamtxid
bampid
bamptxid
bamptxpid
bamptxstatus
bamptxcompleted
```

Does there need to be a `bamtxcompleted` too?

cloudevents/extensions/bam.md
katallaxie: I will provide an brief explanation to those fields

cloudevents/extensions/bam.md
katallaxie: :) it so happend that they changed their content structure. They started to include the language in their urls.

katallaxie:
@duglin great suggestion.

- Updated the naming of the fields.
- Provided a more in-depth explanation of how this works (like a state machine of the business process that is controlled by the events).
- Fixed the link (they change the URL schema).

cloudevents/extensions/bam.md
```diff
@@ -0,0 +1,129 @@
...
+described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+However, the scope of these key words is limited to when this extension is
+used. For example, an attribute being marked as "REQUIRED" does not mean
+it needs to be in all CloudEvents, rather it needs to be included only when
+this extension is being used.
+
+## Attributes
+
+### bamtxid
```
markpeek, line 34: This is shortening the name further than I had thought on the call. As such, would it make sense to have it expanded for readability such as:
```
### bamtxid (Business Transaction ID)
```
Note: neither the spec of other extensions have this but those attributes are more readable. @duglin thoughts?

cloudevents/extensions/bam.md
```diff
@@ -0,0 +1,129 @@
...
+  actually been completed, or if the transaction has somehow failed.
+  This is a mechanism to indicate a final completion or failure that is
+  not captures by the model of the business process..
+- Constraints
+  - OPTIONAL
+  - if present, MUST be a boolean value
+
+## Usage
+
+When this extension is used, producers MUST set the value of
```
markpeek, line 105: This should be changed to:
```
When this extension is used, producers MUST set the value of
the `bamtxid`, `bampid`,
`bamptxid`, and `bamptxsid`
attributes to the unique identifiers of the business process, transaction,
and transaction step associated with the event.

Intermediaries MUST NOT change the value of the `bamtxid`,
`bampid`, `bamptxid`,
and `bamptxsid` attributes.
```
since those other attribute names are no longer specified above. Again, I could see expanding the attribute name in parens for readability.

cloudevents/extensions/bam.md
katallaxie: I think this makes sense.

```markdown
### `bamtxid` (Business Transaction ID)
```

Would strike a balance between efficiency of the fields and a good documentation. I think the convention on not using CamelCase, for the obvious reason of compatibility between platforms, this strikes the balance of actually make the long intention readable by shortening it. But on the other hand make it clear in the docs what is should be used for.

duglin: I agree that some of the names I came up with are a bit short and I'd be ok with making them a bit longer - I'm just not that creative :-). But if folks are ok with these short names then yes I agree with adding the descriptive long names in parens.

cloudevents/extensions/bam.md
```diff
@@ -0,0 +1,127 @@
...
+described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+However, the scope of these key words is limited to when this extension is
+used. For example, an attribute being marked as "REQUIRED" does not mean
+it needs to be in all CloudEvents, rather it needs to be included only when
+this extension is being used.
+
+## Attributes
+
+### bamtxid (Business Transaction ID)
```
duglin, line 34: minor thing, but I wonder if the stuff in parens should be `BAM Transaction ID`? Then all of the letters in the attribute name are correlated with the descriptive name - as of now we're missing the words for "am"

duglin:
@JemDay are you ok with the direction of the latest set of changes?

JemDay:
LGTM :-)

cloudevents/extensions/bam.md
duglin: ping @katallaxie, any thoughts on ^^ ?

katallaxie: @duglin I changed it to `BAM` as we are defining the `BAM` term in the header. I think then this is consistent.

duglin:
thanks @katallaxie - and thanks for your patience.
Approved on the 5/2 call

PR #1280, opened 2024-05-09
test: verify that invalid string literals in LIKE expression are a parse error

Cali0707:
Follow up on discussion in sdk-go: https://github.com/cloudevents/sdk-go/pull/1046#discussion_r1595307651

## Proposed Changes

- Add a test verifying that an invalid string literal in a LIKE expression is a parse error

### DIFF
--- a/cesql/cesql_tck/like_expression.yaml
+++ b/cesql/cesql_tck/like_expression.yaml
@@ -116,3 +116,10 @@ tests:
   - name: With type coercion from bool (4)
     expression: "FALSE LIKE 'fal%'"
     result: true 
+
+  - name: Invalid string literal in comparison causes parse error
+    expression: "x LIKE 123"
+    result: false
+    error: parse
+    eventOverrides:
+      x: "123"


Cali0707:
cc @duglin @pierDipi

cesql/cesql_tck/like_expression.yaml
```diff
@@ -116,3 +116,10 @@ tests:
   - name: With type coercion from bool (4)
     expression: "FALSE LIKE 'fal%'"
     result: true 
+
+  - name: Invalid string literal in comparison causes parse error
+    expression: "x LIKE 123"
```
duglin, line 121: The doesn't say that % is required, so why is this an error?

Cali0707: Because the only valid string literals in CESQL are either single quoted or double quoted. So, since 123 is not quoted it is not a string literal. See the definition for LIKE here:
```ebnf
string-literal ::= ( "'" ( [^'] | "\'" )* "'" ) | ( '"' ( [^"] | '\"' )* '"')
like-operator ::= "LIKE"
not-operator ::= "NOT"
like-operation ::= expression not-operator? like-operator string-literal
```

Basically, the `LIKE` operator is ONLY defined where the right hand operand is a string literal, and string literals require a single or double quote

duglin:
LGTM

duglin:
Approved on the 5/16 call

PR #1281, opened 2024-05-09
clarify type casting in CESQL spec

Cali0707:
Follow up on discussion in https://github.com/cloudevents/sdk-go/pull/1046#discussion_r1589256405

## Proposed Changes

- Clarify which type casts are required to be supported
- Clarify how boolean -> string type casting works
- Clarify how type casting is performed in the LIKE operator


Cali0707:
cc @pierDipi @duglin

cesql/spec.md
```diff
@@ -402,6 +409,14 @@ sequence = 10
 `Integer x Integer`. Because the right operand of the operator is an _Integer_ and there is only one `=` definition
 which uses the type _Integer_ as the right parameter, `sequence` is cast to _Integer_.
 
+A CESQL engine MUST support the following type casts:
+
+1. `Integer -> String`
+1. `String -> Integer`
+1. `String -> Boolean`
```
duglin, line 416: where are ones like this defined?

Cali0707: Ah I had assumed that they were defined elsewhere and I was just seeking to clarify which ones had to be supported. But, looking more closely now I don't see those defined anywhere...

Let me add those

Cali0707:
cc @duglin @jskeet 

Let me know if this makes sense given your feedback last week!

cesql/spec.md
```diff
@@ -353,6 +357,25 @@ left operand of the OR operation evalues to `true`, the right operand MUST NOT b
 
 #### 3.7. Type casting
 
+A CESQL engine MUST support the following type casts:
+
+| Definition           | Semantics                                                                                                                                                                          |
+| -------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `Integer -> String`  | Returns the string representation of the integer value in base 10. If the value is less than 0, the '-' character is prepended to the result.                                      |
+| `String -> Integer`  | Returns the result of interpreting the string as a 32 bit base 10 integer. The string may begin with a leading sign '+' or '-'. If the result will overflow, an error is returned. |
```
jskeet, line 365: We may not want to worry about it, but are leading zeroes permitted? "0001" for example? (And if so, does that mean the string can be really long, e.g. "00000000000000000000000000000000001"? Leading zeroes foil length checks sometimes...)

cesql/spec.md
```diff
@@ -353,6 +357,25 @@ left operand of the OR operation evalues to `true`, the right operand MUST NOT b
...
+| `String -> Boolean`  | Returns `true` or `false` if the lower case representation of the string is exatly "true" or "false, respectively. Otherwise returns an error.                                     |
```
jskeet, line 366: exatly => exactly

cesql/spec.md
```diff
@@ -353,6 +357,25 @@ left operand of the OR operation evalues to `true`, the right operand MUST NOT b
...
+
+An example of how _Boolean_ values cast to _String_ combines with the case insensitivity of CESQL keywords is that:
+```
+TRUE = "true" AND FALSE = "false"
+```
+will evaluate to `true`, while
+```
+TRUE = "true" OR FALSE = "false"
+```
+will evaluate to `false.
```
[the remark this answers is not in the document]
Cali0707: Ah whoops, copy paste mistake - I had intended for the string literals to be all caps :facepalm:

cesql/spec.md
```diff
@@ -353,6 +357,25 @@ left operand of the OR operation evalues to `true`, the right operand MUST NOT b
 
 #### 3.7. Type casting
 
+A CESQL engine MUST support the following type casts:
```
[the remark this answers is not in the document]
Cali0707: Yeah I see your point. Personally, I think "returning" makes more sense because it's not like an exception that is raised can be caught somewhere - it is just a second return value you get on every evaluation (which may be null/nil). WDYT? If you think "returning" makes more sense, I'll open a PR to change that everywhere, otherwise I'll switch these to be "raising" an error.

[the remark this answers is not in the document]
jskeet: Sorry, I wasn't clear - I'm fine with the "raising" terminology, but the problem is that these casts don't say what the non-error part of the return is when there's a problem. If I cast "BOGUS" to Boolean, an error will be raised... but what's the value of the expression?

cesql/spec.md
Cali0707: Personally, I think leading zeroes are fine - we aren't using any length checks on the input strings afaik, rather we validate the parsed integer value

cesql/spec.md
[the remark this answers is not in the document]
Cali0707: Oh I see your point - let me add some clarification here.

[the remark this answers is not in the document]
Cali0707: I added some clarification, could you recheck when you have time @jskeet ?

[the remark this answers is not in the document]
jskeet: Yup, that definitely works.

duglin:
Error looks real: `cesql/spec.md: line 365: 'may' MUST be capitalized ('MAY')`

cesql/spec.md
```diff
@@ -278,6 +278,10 @@ For example, the pattern `_b*` will accept values `ab`, `abc`, `abcd1` but won't
 Both `%` and `_` can be escaped with `\`, in order to be matched literally. For example, the pattern `abc\%` will match
 `abc%` but won't match `abcd`.
 
+In cases where the left operand is not a `String`, it MUST be cast to a `String` before the comparison is made. 
+The pattern of the `LIKE` operator (that is, the right operand of the operator) MUST be a valid string predicate, 
+otherwise the parse MUST return a parse error.
```
duglin, line 283: Just to be clear... `subject LIKE TRUE` isn't valid because `TRUE` isn't a valid string, however, we talk about how there's implicit casting all over the place... so someone may wonder why `TRUE` isn't implicitly converted into `"true"`. Am I correct in this thinking? Should we be explicit and say that casting isn't allowed in this case? Or perhaps should we allow it for consistency?  @jskeet thoughts?

jskeet: It would definitely be good to be clear. Even without the "LIKE" aspect, I can't tell offhand whether `"TRUE" = true` is casting the LHS to Boolean, or the RHS to String. I suspect it's the former, based on bullet 2 in the list in 3.7.

cesql/spec.md
```diff
@@ -353,6 +357,25 @@ left operand of the OR operation evalues to `true`, the right operand MUST NOT b
 
 #### 3.7. Type casting
 
+A CESQL engine MUST support the following type casts:
+
+| Definition           | Semantics                                                                                                                                                                                                                                        |
+| -------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `Integer -> String`  | Returns the string representation of the integer value in base 10. If the value is less than 0, the '-' character is prepended to the result.                                                                                                    |
```
duglin, line 364: To be very anal... w/o leading zeros?

cesql/spec.md
```diff
@@ -353,6 +357,25 @@ left operand of the OR operation evalues to `true`, the right operand MUST NOT b
...
+| `String -> Integer`  | Returns the result of interpreting the string as a 32 bit base 10 integer. The string MAY begin with a leading sign '+' or '-'. If the result will overflow or the string is not a valid integer an error is returned along with a value of `0`. |
+| `String -> Boolean`  | Returns `true` or `false` if the lower case representation of the string is exactly "true" or "false, respectively. Otherwise returns an error along with a value of `false`                                                                    |
+| `Boolean -> String`  | Returns `"true"` if the boolean is `true`, and `"false"` if the boolean is `false`.                                                                                                                                                              |
```
duglin, line 367: I think we should have a 2x2 table (or something) that shows all combinations in one spot. Then it's east to see what's missing, like bool->int, int->bool

cesql/spec.md
Cali0707: @duglin I think the main problem is that currently the `LIKE` expression isn't defined with an expression as the right operand. It is defined as:
```ebnf
expression NOT? LIKE stringLiteral
```

So, any value on the right operand which is not a string literal MUST be a parse error currently. We can change this, but I'm not sure if it makes sense. The whole point of a `LIKE` expression is to compare against a pattern, not a specific value. If someone were to compare against another string value wouldn't it just makes sense to use `=` or something like that?

duglin: I think it's ok to leave it as "must be a string literal w/o casting", but let's be sure clear in the text that implicit type casting isn't allowed in this one spot. I agree defining it as `stringLiteral` sort of implies it, but saying that things should be implicitly case in all other spaces (even when the spec says the arg is a `string`) could be unclear to people.

Cali0707:
@duglin how does this look now?

cesql/spec.md
```diff
@@ -353,6 +357,33 @@ left operand of the OR operation evalues to `true`, the right operand MUST NOT b
 
 #### 3.7. Type casting
 
+The following table indicates which type casts a CESQL engine MUST or MUST NOT support:
+
+| Type    | Integer  | String | Boolean  |
+| ------- | -------- | ------ | -------- |
+| Integer | N/A      | MUST   | MUST NOT |
+| String  | MUST     | N/A    | MUST     |
+| Boolean | MUST NOT | MUST   | N/A      |
```
duglin, line 366: This is a valid choice, but I'm curious why you didn't choose to map `1` with `true` and `0` with `false` ?

duglin:
Just one question, but either way that one lands.... LGTM -thanks!

duglin (approved):

cesql/spec.md
Cali0707: Tbh, I just based this off the current implementation in the SDKs...

The `1` mapping to `true` and `0` to `false` sounds reasonable though, especially because then the zero value for integers would be cast to `false`.

Which do you think is better?

duglin: I'm partial to allowing it, then everything can be implicitly cast to anything else - and I think you can then remove an error type.

@jskeet any concerns?

jskeet: I'm *okay* with it - personally I do like being clear about expressions in terms of types, but that's mostly in C# etc. I haven't done significant amounts of SQL in years, so I don't know how it would "feel". If others are happy, that's fine by me.

Cali0707: @duglin if we do this:
> map `1` with `true` and `0` with `false`

then how would we handle an integer value like `200`? Would that be an error? Or to avoid errors would it be better to say that when casting from integer to boolean, any non-zero integer becomes `true`, and `0` becomes `false`.

duglin: Yup- I like that. Int->bool is zero vs non-zero.  But bool->int is 0 vs 1

duglin:
Approved on the 5/30 call

Issue #1282, opened 2024-05-16
CESQL v1 review


Cali0707:
We are hoping to move the CESQL spec to v1 in the next few weeks. The [spec can be found here](https://github.com/cloudevents/spec/blob/main/cesql/spec.md). This issue is a tracker issue for all discussions related to any remaining changes that need to be made before we can move this spec to v1.

If you have any concerns/questions about the spec please leave them below and we can discuss them!

Once you have finished reviewing the spec, if you think it is ready to go please react to this issue with a :+1:

Cali0707:
@jskeet thanks so much for pointing these out! My thoughts on your points (in order) are:

2.1(Bug): This seems to be a bug in the spec - the actual Parser file has all three
2.2(Suggestion) This seems to be a reasonable change to me, let's do it!
2.2(Bug) Yeah I see your point, let me try and update the `number-literal` definition to allow either `+` or `-` in the front (and rename it to `integer-literal`)
2.2(Validation): I believe in the SDKs we do validate that the literals do not overflow the 32 bit allowed integer values. We should add some clarification to the spec that integer literals MUST be a valid 32 bit integer value, otherwise there will be a parse error
2.3 My vote is to try and keep the number of changes to a minimum for now, so since this isn't a breaking change we should leave it out for now
3.5 IMO, we shouldn't allow overloading of pre-defined functions (at least not for the same arity as the pre-defined function). We should add text indicating that would be an error.
3.5.2: I'm not sure what would make the most sense here, but my initial guess would be to return a math error? WDYT?
3.6: Yes you are correct this should be clarified. I agree that left to right with parenthesized expressions taking priority when encountered makes sense. So "1 + (5 + 2)" would be evaluated as 1, 5+2, 1+7.

I'll work on opening a PR for these changes by tomorrow. Thanks again for taking a look, and let me know if you find anything else!!

Cali0707:
@jskeet @duglin I've opened a PR to address these issues in the spec, if you could TAL when you have some time that would be great!

@duglin I haven't addressed your feedback here in that PR yet, I will try and push those changes soon...

Cali0707:
> What other uses of this spec do we expect other than "filter predicate" ? That might help with my final concern at the end of this comment.

@duglin one other use case I have been exploring in Knative is allowing users to use CESQL to add new attributes to events based on other attributes as they are processed (for example based on an original event when sending a reply event to a trigger). 

[...]

PR #1286, opened 2024-05-22
CESQL v1 review changes

Cali0707:
Addresses the comments on #1282 

## Proposed Changes

Look at the discussion on #1282

### DIFF
--- a/cesql/spec.md
+++ b/cesql/spec.md
@@ -41,11 +41,11 @@ CloudEvent instances.
 CloudEvents SQL expressions (also known as CESQL) allow computing values and matching of CloudEvent attributes against complex expressions
 that lean on the syntax of Structured Query Language (SQL) `WHERE` clauses. Using SQL-derived expressions for message
 filtering has widespread implementation usage because the Java Message Service (JMS) message selector syntax also leans
-on SQL. Note that neither the SQL standard (ISO 9075) nor the JMS standard nor any other SQL dialect is used as a
+on SQL. Note that neither the [SQL standard (ISO 9075)][iso-9075] nor the JMS standard nor any other SQL dialect is used as a
 normative foundation or to constrain the expression syntax defined in this specification, but the syntax is informed by
 them.
 
-CESQL is a _[Total pure functional programming language][total-programming-language-wiki]_ in order to guarantee the
+CESQL is a _[total pure functional programming language][total-programming-language-wiki]_ in order to guarantee the
 termination of the evaluation of the expression. It features a type system correlated to the [CloudEvents type
 system][ce-type-system], and it features boolean and arithmetic operations, as well as built-in functions for string
 manipulation.
@@ -56,7 +56,10 @@ producer, or in an intermediary, and it can be implemented using any technology 
 The CloudEvents Expression Language assumes the input always includes, but is not limited to, a single valid and
 type-checked CloudEvent instance. An expression MUST NOT mutate the value of the input CloudEvent instance, nor any of
 the other input values. The evaluation of an expression observes the concept of [referential
-transparency][referential-transparency-wiki]. The output of a CESQL expression evaluation is always a _boolean_, an _integer_ or a _string_, and it might include an error.
+transparency][referential-transparency-wiki]. The primary output of a CESQL expression evaluation is always a _boolean_, an _integer_ or a _string_.
+The secondary output of a CESQL expression evaluation is a set of errors which occurred during evaluation. This set MAY be empty, indicating that no
+error occurred during execution of the expression. The value used by CESQL engines to represent an empty set of errors is out of the scope of this
+specification.
 
 The CloudEvents Expression Language doesn't support the handling of the data field of the CloudEvent instances, due to
 its polymorphic nature and complexity. Users that need this functionality ought to use other more appropriate tools.
@@ -70,7 +73,8 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 The CESQL can be used as a [filter dialect][subscriptions-filter-dialect] to filter on the input values.
 
-When used as a filter predicate, the expression output value is always cast to a boolean value.
+When used as a filter predicate, the expression output value MUST be a _Boolean_. If the output value is not a _Boolean_ or any errors are returned,
+the event MUST NOT pass the filter.
 
 <!-- TODO -->
 
@@ -103,12 +107,12 @@ Nested expressions MUST be correctly parenthesized.
 ### 2.2. Value identifiers and literals
 
 Value identifiers in CESQL MUST follow the same restrictions of the [Attribute Naming
-Convention][ce-attribute-naming-convention] from the CloudEvents spec. A value identifier MUST NOT be greater than 20
+Convention][ce-attribute-naming-convention] from the CloudEvents spec. A value identifier SHOULD NOT be greater than 20
 characters in length.
 
 ```ebnf
 lowercase-char ::= [a-z]
-value-identifier ::= ( lowercase-char | digit ) ( lowercase-char | digit )*
+value-identifier ::= ( lowercase-char | digit )+
 ```
 
 CESQL defines 3 different literal kinds: integer numbers, `true` or `false` booleans, and `''` or `""` delimited strings. Integer literals MUST be valid 32 bit signed integer values.
@@ -124,8 +128,8 @@ string-literal ::= ( "'" ( [^'] | "\'" )* "'" ) | ( '"' ( [^"] | '\"' )* '"')
 literal ::= integer-literal | boolean-literal | string-literal
 ```
 
-Because string literals can be either `''` or `""` delimited, in the former case, the `'` has to be escaped, while in
-the latter the `"` has to be escaped.
+Because string literals can be either `''` or `""` delimited, in the former case, the `'` character has to be escaped if it is to be used in the string literal, while in
+the latter the `"` has to be escaped if it is to be used in the string literal.
 
 ### 2.3. Operators
 
@@ -200,8 +204,6 @@ _Boolean_.
 When addressing an attribute not included in the input event, the subexpression referencing the missing attribute MUST evaluate to `false`.
 For example, `true AND (missingAttribute = "")` would evaluate to `false` as the subexpression `missingAttribute = ""` would be false.
 
-When addressing an attribute represented by the _Integer_ type, if the value of that attribute does not fit into a 32 bit signed integer, an error MUST be returned.
-
 ### 3.3. Errors
 
 Because every operator and function is total, an expression evaluation flow is defined statically and cannot be modified
@@ -214,7 +216,8 @@ runtime errors.
 
 ### 3.4. Operators
 
-The following tables show the operators that MUST be supported by a CESQL evaluator.
+The following tables show the operators that MUST be supported by a CESQL evaluator. When evaluating an operator,
+a CESQL engine MUST attempt to cast the operands to the correct types. If this type casting fails, a _Cast_ error will be returned, along with the zero value for the return type of the expression.
 
 All the operators in the following tables are listed in precedence order.
 
@@ -471,3 +474,4 @@ hop < ttl
 [subscriptions-filter-dialect]: ../subscriptions/spec.md#3241-filter-dialects
 [ebnf-xml-spec]: https://www.w3.org/TR/REC-xml/#sec-notation
 [modulo-operation-wiki]: https://en.wikipedia.org/wiki/Modulo_operation
+[iso-9075]: https://en.wikipedia.org/wiki/ISO/IEC_9075
--- a/cesql/spec.md
+++ b/cesql/spec.md
@@ -234,34 +234,29 @@ Corresponds to the syntactic rule `unary-operation`:
 
 Corresponds to the syntactic rule `binary-operation`:
 
-| Definition                              | Semantics                                                                                                  |
-| --------------------------------------- | ---------------------------------------------------------------------------------------------------------- |
-| `x = y: Boolean x Boolean -> Boolean`   | Returns `true` if the values of `x` and `y` are equal                                                      |
-| `x != y: Boolean x Boolean -> Boolean`  | Same as `NOT (x = y)`                                                                                      |
-| `x <> y: Boolean x Boolean -> Boolean`  | Same as `NOT (x = y)`                                                                                      |
-| `x AND y: Boolean x Boolean -> Boolean` | Returns the logical and of `x` and `y`                                                                     |
-| `x OR y: Boolean x Boolean -> Boolean`  | Returns the logical or of `x` and `y`                                                                      |
-| `x XOR y: Boolean x Boolean -> Boolean` | Returns the logical xor of `x` and `y`                                                                     |
-| `x = y: Integer x Integer -> Boolean`   | Returns `true` if the values of `x` and `y` are equal                                                      |
-| `x != y: Integer x Integer -> Boolean`  | Same as `NOT (x = y)`                                                                                      |
-| `x <> y: Integer x Integer -> Boolean`  | Same as `NOT (x = y)`                                                                                      |
-| `x < y: Integer x Integer -> Boolean`   | Returns `true` if `x` is strictly less than `y`                                                            |
-| `x <= y: Integer x Integer -> Boolean`  | Returns `true` if `x` is less than or equal to `y`                                                         |
-| `x > y: Integer x Integer -> Boolean`   | Returns `true` if `x` is strictly greater than `y`                                                         |
-| `x >= y: Integer x Integer -> Boolean`  | Returns `true` if `x` is greater or equal to `y`                                                           |
-| `x * y: Integer x Integer -> Integer`   | Returns the product of `x` and `y`                                                                         |
-| `x / y: Integer x Integer -> Integer`   | Returns the truncated division of `x` and `y`. Returns `0` and raises an error if `y = 0`                  |
-| `x % y: Integer x Integer -> Integer`   | Returns the remainder of the truncated division of `x` and `y`. Returns `0` and raises an error if `y = 0` |
-| `x + y: Integer x Integer -> Integer`   | Returns the sum of `x` and `y`                                                                             |
-| `x - y: Integer x Integer -> Integer`   | Returns the difference of `x` and `y`                                                                      |
-| `x = y: String x String -> Boolean`     | Returns `true` if the values of `x` and `y` are equal                                                      |
-| `x != y: String x String -> Boolean`    | Same as `NOT (x = y)`                                                                                      |
-| `x <> y: String x String -> Boolean`    | Same as `NOT (x = y)`                                                                                      |
-
-The modulo and divisions MUST follow the [truncated divisions definition][modulo-operation-wiki], that is:
-
-- The remainder of the modulo MUST have the same sign as the dividend.
-- The quotient MUST be rounded towards zero, _truncating_ the decimal part.
+| Definition                              | Semantics                                                                                                                          |
+| --------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------- |
+| `x = y: Boolean x Boolean -> Boolean`   | Returns `true` if the values of `x` and `y` are equal                                                                              |
+| `x != y: Boolean x Boolean -> Boolean`  | Same as `NOT (x = y)`                                                                                                              |
+| `x <> y: Boolean x Boolean -> Boolean`  | Same as `NOT (x = y)`                                                                                                              |
+| `x AND y: Boolean x Boolean -> Boolean` | Returns the logical and of `x` and `y`                                                                                             |
+| `x OR y: Boolean x Boolean -> Boolean`  | Returns the logical or of `x` and `y`                                                                                              |
+| `x XOR y: Boolean x Boolean -> Boolean` | Returns the logical xor of `x` and `y`                                                                                             |
+| `x = y: Integer x Integer -> Boolean`   | Returns `true` if the values of `x` and `y` are equal                                                                              |
+| `x != y: Integer x Integer -> Boolean`  | Same as `NOT (x = y)`                                                                                                              |
+| `x <> y: Integer x Integer -> Boolean`  | Same as `NOT (x = y)`                                                                                                              |
+| `x < y: Integer x Integer -> Boolean`   | Returns `true` if `x` is less than `y`                                                                                             |
+| `x <= y: Integer x Integer -> Boolean`  | Returns `true` if `x` is less than or equal to `y`                                                                                 |
+| `x > y: Integer x Integer -> Boolean`   | Returns `true` if `x` is greater than `y`                                                                                          |
+| `x >= y: Integer x Integer -> Boolean`  | Returns `true` if `x` is greater than or equal to `y`                                                                              |
+| `x * y: Integer x Integer -> Integer`   | Returns the product of `x` and `y`                                                                                                 |
+| `x / y: Integer x Integer -> Integer`   | Returns the result of dividing `x` by `y`, rounded towards `0` to obtain an integer. Returns `0` and raises an error if `y = 0`    |
+| `x % y: Integer x Integer -> Integer`   | Returns the remainder of `x` divided by `y`, where the result has the same sign as `x`. Returns `0` and raises an error if `y = 0` |
+| `x + y: Integer x Integer -> Integer`   | Returns the sum of `x` and `y`                                                                                                     |
+| `x - y: Integer x Integer -> Integer`   | Returns the difference of `x` and `y`                                                                                              |
+| `x = y: String x String -> Boolean`     | Returns `true` if the values of `x` and `y` are equal                                                                              |
+| `x != y: String x String -> Boolean`    | Same as `NOT (x = y)`                                                                                                              |
+| `x <> y: String x String -> Boolean`    | Same as `NOT (x = y)`                                                                                                              |
 
 The AND and OR operators MUST be short-circuit evaluated. This means that whenever the left operand of the AND operation evaluates to `false`, the right operand MUST NOT be evaluated.
 Similarly, whenever the left operand of the OR operation evaluates to `true`, the right operand MUST NOT be evaluated.
@@ -303,7 +298,8 @@ The matching is done using the same semantics of the equal `=` operator, but usi
 
 ### 3.5. Functions
 
-CESQL provides the concept of function, and defines some built-in that every engine MUST implement. An engine SHOULD also allow users to define their custom functions.
+CESQL provides the concept of function, and defines some built-in functions that every engine MUST implement. An engine SHOULD also allow users to define their custom functions, however, the mechanism
+by which this is done is out of scope of this specification.
 
 A function is identified by its name, its parameters and the return value. A function can be variadic, that is the arity is not fixed.
 
@@ -312,18 +308,23 @@ Because of implicit casting, no functions with the same name and same arity but 
 
 An overload on a variadic function is allowed only if the number of initial fixed arguments is greater than the maximum arity for that particular function name. Only one variadic overload is allowed.
 
-For example, the following definitions are valid:
+For example, the following set of definitions are valid and will all be allowed by the rules:
 
 * `ABC(x): String -> Integer`: Unary function (arity 1). 
 * `ABC(x, y): String x String -> Integer`: Binary function (arity 2).
 * `ABC(x, y, z, ...): String x String x String x String^n -> Integer`: n-ary function (variable arity), but the initial fixed arguments are at least 3.
 
-But the following are invalid, so the engine MUST reject them:
+But the following set is invalid, so the engine MUST reject them:
 
 * `ABC(x...): String^n -> Integer`: n-ary function (variable arity), but there are no initial fixed arguments.
 * `ABC(x, y, z): String x String x String -> Integer`: Ternary function (arity 3).
 
-When a function invocation cannot be dispatched, the return value is undefined.
+These two are incompatible because the n-ary function `ABC(x...)` can not be distinguished in any way
+from the ternary function `ABC(x, y, z)` if the n-ary function were called with three arguments.
+In order for these definitions to be valid, the n-ary function would need to have at least 4 fixed
+arguments.
+
+When a function invocation cannot be dispatched, the return value is `false`, and a `_FunctionError` is also returned.
 
 The following tables show the built-in functions that MUST be supported by a CESQL evaluator.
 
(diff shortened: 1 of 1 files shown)


cesql/spec.md
```diff
@@ -200,6 +200,8 @@ _Boolean_.
 When addressing an attribute not included in the input event, the subexpression referencing the missing attribute MUST evaluate to `false`.
 For example, `true AND (missingAttribute = "")` would evaluate to `false` as the subexpression `missingAttribute = ""` would be false.
 
+When addressing an attribute represented by the _Integer_ type, if the value of that attribute does not fit into a 32 bit signed integer, an error MUST be returned.
```
jskeet, line 203: A CloudEvent Integer attribute can't have such a value though... from the CE Spec:

> A whole number in the range -2,147,483,648 to +2,147,483,647 inclusive. This is the range of a signed, 32-bit, twos-complement encoding. Event formats do not have to use this encoding, but they MUST only use Integer values in this range.

To put it another way: if we've got this for integers, do we need something similar for attributes represented by the _Boolean_ type but with a value which isn't true or false?

Cali0707: Ah, if it isn't possible for this value to be in an event then I think we can remove this - I hadn't been aware that that was the case

jskeet: Yes, it's the CE type system rather than about specific attributes.

cesql/spec.md
```diff
@@ -56,7 +56,10 @@ producer, or in an intermediary, and it can be implemented using any technology
 The CloudEvents Expression Language assumes the input always includes, but is not limited to, a single valid and
 type-checked CloudEvent instance. An expression MUST NOT mutate the value of the input CloudEvent instance, nor any of
 the other input values. The evaluation of an expression observes the concept of [referential
-transparency][referential-transparency-wiki]. The output of a CESQL expression evaluation is always a _boolean_, an _integer_ or a _string_, and it might include an error.
+transparency][referential-transparency-wiki]. The primary output of a CESQL expression evaluation is always a _boolean_, an _integer_ or a _string_.
+The secondary output of a CESQL expression evaluation is a set of errors which occurred during evaluation. This set MAY be empty, indicating that no
+error occurred during execution of the expression. The value used by CESQL engines to represent an empty set of errors is out of the scope of this
+specification.
```
duglin, line 62: I like this better. Do you think we need to add something like:
> Generation of an error does not halt processing. This specification does not mandate how these errors are exposed to a user.

? Is the first sentence true?

cesql/spec.md
```diff
@@ -70,7 +73,8 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 The CESQL can be used as a [filter dialect][subscriptions-filter-dialect] to filter on the input values.
 
-When used as a filter predicate, the expression output value is always cast to a boolean value.
+When used as a filter predicate, the expression output value MUST be a _Boolean_. If the output value is not a _Boolean_ or any errors are returned,
+the event MUST NOT pass the filter.
 
 <!-- TODO -->
```
duglin, line 79: do we still need this TODO ?

cesql/spec.md
```diff
@@ -212,7 +216,8 @@ runtime errors.
 
 ### 3.4. Operators
 
-The following tables show the operators that MUST be supported by a CESQL evaluator.
+The following tables show the operators that MUST be supported by a CESQL evaluator. When evaluating an operator,
+a CESQL engine MUST attempt to cast the operands to the correct types. If this type casting fails, a _Cast_ error will be returned, along with the zero value for the return type of the expression.
```
duglin, line 220: do we define the zero value for each type? While probably obvious we should say it's: zero, "" and `false`.

cesql/spec.md
```diff
@@ -341,12 +346,11 @@ The following tables show the built-in functions that MUST be supported by a CES
...
 | ---------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `ABS(x): Integer -> Integer` | Returns the absolute value of `x` |
+| `ABS(x): Integer -> Integer` | Returns the absolute value of `x`. If the value of `x` is `-2147483648` (the most negative 32 bit integer value possible), then this returns `2147483647` as well as a math error. |
 
 ### 3.6. Evaluation of the expression
 
-Operators MUST be evaluated in order, where the parenthesized expressions have the highest priority over all the other
-operators.
+Operators MUST be evaluated in left to right order, where all operators within a parenthized expression MUST be evaluated before continuing to the right of the parenthized expression.
```
duglin, line 353: parenthized -> parenthesized

duglin:
Before we ship this - can we wrap it at 80 columns to match the other specs?

cesql/spec.md
Cali0707: I'm not 100% sure here, if you look at section 4.1 it states that:

> When evaluating an expression, the evaluator can operate in two _modes_, in relation to error handling:
> * Fail fast mode: When an error is triggered, the evaluation is interrupted and returns the error, without any result.
> * Complete evaluation mode: When an error is triggered, the evaluation is continued, and the evaluation of the expression returns both the result and the error(s).

But, as far as I can tell the SDKs implement the fail fast mode instead of the complete evaluation mode.

duglin: w/o any result... is that true even in the SDKs and Knative? I was assuming the was result 'false'.

I'm having a memory of us discussing what happens when a nested expression errors when that expression is part of an AND - doesn't it result in a false for the nested expression and then false again due to that side of the AND being false? Or am I thinking about some other situation?

Cali0707: In terms of w/o any result, I think the problem (from a spec standpoint at least) is what type should the return value have? For example, if my query is `LEFT(missingattribute, 10)`, then should I get `(false, missing attribute error)` as my return value? Does it make more sense to return `("", missing attribute error)`?

But, similarly if my expression is `LEFT(missingattribute, 10) LIKE prefix%`, then having a return value of `(false, missing attribute error)` would be what I would expect while a return value of `("", missing attribute error)` would not be what I expect.

Perhaps it makes sense to say something along the lines of "Fail fast mode: When an error is triggered, the evaluation is interrupted and returns the zero value for the return type of the root operation, along with the error". WDYT @duglin ?

Cali0707:
@duglin @jskeet would you be able to take another look when you have time? I _think_ I should have addressed all your feedback (minus wrapping at 80 characters), but I'm happy to make further fixes or address anything I missed.

Thanks again for all of your feedback so far!!

jskeet:
I'm unlikely to get time to review this again I'm afraid.

cesql/spec.md
Cali0707: yep I did :facepalm:

Cali0707:
> I'm unlikely to get time to review this again I'm afraid.

No worries, thank you so much for taking the time the last few weeks to review this!

cesql/spec.md
Cali0707: Fixed!

cesql/spec.md
Cali0707: We do now :)

cesql/spec.md
Cali0707: Removed

cesql/spec.md
```diff
@@ -56,7 +56,10 @@ producer, or in an intermediary, and it can be implemented using any technology
 The CloudEvents Expression Language assumes the input always includes, but is not limited to, a single valid and
 type-checked CloudEvent instance. An expression MUST NOT mutate the value of the input CloudEvent instance, nor any of
 the other input values. The evaluation of an expression observes the concept of [referential
-transparency][referential-transparency-wiki]. The output of a CESQL expression evaluation is always a _boolean_, an _integer_ or a _string_, and it might include an error.
+transparency][referential-transparency-wiki]. The primary output of a CESQL expression evaluation is always a _boolean_, an _integer_ or a _string_.
+The secondary output of a CESQL expression evaluation is a set of errors which occurred during evaluation. This set MAY be empty, indicating that no
+error occurred during execution of the expression. The value used by CESQL engines to represent an empty set of errors is out of the scope of this
```
duglin, line 61: Isn't the entire representation of, or even access to, errors out of scope of the spec, not just the empty set of errors?

cesql/spec.md
```diff
@@ -197,8 +207,13 @@ Unless otherwise specified, every attribute and extension MUST be represented by
...
+When addressing an attribute not included in the input event, the subexpression referencing the missing attribute MUST evaluate to the zero value for the return type of the subexpression.
+For example, `true AND (missingAttribute = "")` would evaluate to `false` as the subexpression `missingAttribute = ""` would be false, given that the return type for the `=` operator is _Boolean_.
+However, the expression `missingattribute * 5` would evaluate to `0` because the return type for the `*` operator is _Integer_. Note that this does not mean that the _value_ of the missing attribute
+is set to be the zero value for the type of the missing attribute. Rather, the subexpression with the missingAttribute returns the zero value of the return type of the subexpression.
+As an example, `1 / missingAttribute` does not raise a _MathError_ due to the division by zero, instead it returns `0` as that is the zero value for the return type of the subexpression.
+In cases where the return type of the subexpression cannot be determined by the CESQL engine, the CESQL engine MUST assume a return type of _Boolean_. In such cases, the return value would
+therefore be `false`.
```
duglin, line 216: geez I've been staring at this one for way too long....

duglin: the wording is fine... it's the semantics that I keep thinking about.... :-)

cesql/spec.md
```diff
@@ -210,9 +225,24 @@ returned together with the evaluated value of the CESQL expression.
...
 
+Every CESQL engine MUST support the following error types:
+
+- _ParseError_: An error that occurs during parseing
+- _MathError_: An error that occurs during the evaluation of a mathematical operation
+- _CastError_: An error that occurs during an implicit or explicit type cast
+- _MissingFunctionError_: An error that occurs due to a call to a function that has not been registered with the CESQL engine
+- _FunctionEvaluationError_: An error that occurs during the evaluation of a function
+
+Whenever an operator or function encounters an error, it MUST return a return value as well as an error value or values. In cases where there is not an obvious return value for the expression,
```
duglin, line 236: See if this is easier to read:
...it MUST result in a "return value" as well as one or more "error values".

Up to you

cesql/spec.md
```diff
@@ -210,9 +225,24 @@ returned together with the evaluated value of the CESQL expression.
...
+the operator or function SHOULD return the zero value for the return type of the operator or function.
+
 ### 3.4. Operators
 
-The following tables show the operators that MUST be supported by a CESQL evaluator.
+The following tables show the operators that MUST be supported by a CESQL evaluator. When evaluating an operator,
+a CESQL engine MUST attempt to cast the operands to the correct types. A CESQL evaluator MUST stop evaluating an operator
```
duglin, line 242: s/correct/specified/

cesql/spec.md
duglin: Is a missing attribute an error? Kind of feel like it should be and therefore should have an error type, no? I could be swayed into believing that it's special, but if so we may want to say that a missing attribute isn't an error in the spec so people don't think we overlooked it.

Also, I can't help but wonder if there should be some kind of "generic error" define for cases that you didn't think of above.

cesql/spec.md
```diff
@@ -328,25 +363,31 @@ The following tables show the built-in functions that MUST be supported by a CES
...
-| `SUBSTRING(x, pos, len): String x Integer x Integer -> String` | Returns the substring of `x` starting from index `pos` (included) of length `len`. Characters' index starts from `1`. If `pos` is negative, the beginning of the substring is `pos` characters from the end of the string. If `pos` is 0, then returns the empty string. If `len` is greater than the maximum substring starting at `pos`, then return the maximum substring. Returns the empty string and raises an error if `pos > LENGTH(x) OR pos < -LENGTH(x)` or if `len` is negative. |
+| `TRIM(x): String -> String`                                | Returns `x` with leading and trailing whitespaces trimmed. This does not remove any non-whitespace characters, such as control characters.                                                                                 |
```
duglin, line 369: I think we still need to define "whitespace characters" by pointing to some spec (or defining it here). Saying it's not control characters is misleading because `\t` is both a control character and whitespace

cesql/spec.md
```diff
@@ -70,9 +73,8 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 The CESQL can be used as a [filter dialect][subscriptions-filter-dialect] to filter on the input values.
 
-When used as a filter predicate, the expression output value is always cast to a boolean value.
-
-<!-- TODO -->
+When used as a filter predicate, the expression output value MUST be a _Boolean_. If the output value is not a _Boolean_ or any errors are returned,
```
duglin, line 76: add a comma after _Boolean_

cesql/spec.md
```diff
@@ -56,7 +56,10 @@ producer, or in an intermediary, and it can be implemented using any technology
 The CloudEvents Expression Language assumes the input always includes, but is not limited to, a single valid and
 type-checked CloudEvent instance. An expression MUST NOT mutate the value of the input CloudEvent instance, nor any of
 the other input values. The evaluation of an expression observes the concept of [referential
-transparency][referential-transparency-wiki]. The output of a CESQL expression evaluation is always a _boolean_, an _integer_ or a _string_, and it might include an error.
+transparency][referential-transparency-wiki]. The primary output of a CESQL expression evaluation is always a _boolean_, an _integer_ or a _string_.
```
duglin, line 59: Just curious, do you know why the reference to "referential transparency" is here? I'm just wondering what people should do with that info? If the sentence were removed would it hurt the spec or change the impls?

cesql/spec.md
```diff
@@ -197,8 +207,13 @@ Unless otherwise specified, every attribute and extension MUST be represented by
 Through implicit type casting, the user can convert the addressed value instances to _Integer_ and
```
duglin, line 207: "... MUST be represented by the _String_ type as its initial type"...  I'm curious what this really means and why it's here. Does the user who is creating the expression need to do something with this info? Or is this for an implementation/SDK ? And if so, what do they do with it? If someone defines an extension called myint (an int), why would it matter if it starts out (internally) as an int or a string if it will be type cast as needed?

The spec says: The primary output of a CESQL expression evaluation is always a boolean, an integer or a string.

Does this mean that an expression of just `myint` MUST result in a string? And if so, it's kind of interesting that the spec doesn't define an int() function to allow people to covert it back to an int if that's what they want to result to be.

cesql/spec.md
```diff
@@ -210,9 +225,24 @@ returned together with the evaluated value of the CESQL expression.
 Whenever possible, some error checks SHOULD be done at compile time by the expression evaluator, in order to prevent
 runtime errors.
 
+Every CESQL engine MUST support the following error types:
+
+- _ParseError_: An error that occurs during parseing
```
duglin, line 230: s/parseing/parsing/

cesql/spec.md
```diff
@@ -210,9 +225,24 @@ returned together with the evaluated value of the CESQL expression.
...
 ### 3.4. Operators
 
-The following tables show the operators that MUST be supported by a CESQL evaluator.
+The following tables show the operators that MUST be supported by a CESQL evaluator. When evaluating an operator,
+a CESQL engine MUST attempt to cast the operands to the correct types. A CESQL evaluator MUST stop evaluating an operator
+when it encounters an error value, and it MUST return both the zero value for the return type of the operator as well
+as the error. In cases where the return type for the operator is ambiguous, the return type MUST be assumed to be
+_Boolean_ and the return value MUST be `false`.
 
 All the operators in the following tables are listed in precedence order.
```
duglin, line 247: Below, it defines "x = y" before "x != y". Given the sentence above, does this mean that:
```
a != b = c
```
must have "b = c" evaluated before the "a != ..." part?

The spec also says "Operators MUST be evaluated in left to right order", which I'd interpret as "a != b" would be evaluated first.

And I'm reading this wrong or are things a bit ambiguous?

cesql/spec.md
```diff
@@ -229,38 +259,35 @@ Corresponds to the syntactic rule `unary-operation`:
...
+| `x >= y: Integer x Integer -> Boolean`  | Returns `true` if `x` is greater than or equal to `y`                                                                              |
+| `x * y: Integer x Integer -> Integer`   | Returns the product of `x` and `y`                                                                                                 |
+| `x / y: Integer x Integer -> Integer`   | Returns the result of dividing `x` by `y`, rounded towards `0` to obtain an integer. Returns `0` and a _MathError_ if `y = 0`      |
+| `x % y: Integer x Integer -> Integer`   | Returns the remainder of `x` divided by `y`, where the result has the same sign as `x`. Returns `0` and a _MathError_ if `y = 0`   |
+| `x + y: Integer x Integer -> Integer`   | Returns the sum of `x` and `y`                                                                                                     |
+| `x - y: Integer x Integer -> Integer`   | Returns the difference of `x` and `y`                                                                                              |
```
duglin, line 281: I'd prefer: Returns the value of `y` subtracted from `x`.   Or something like that.

"Returns the difference of x and y" is ambiguous as to which is subtracted from which. Technically, y-x is the diff too.

cesql/spec.md
```diff
@@ -229,38 +259,35 @@ Corresponds to the syntactic rule `unary-operation`:
...
+| `x = y: String x String -> Boolean`     | Returns `true` if the values of `x` and `y` are equal                                                                              |
+| `x != y: String x String -> Boolean`    | Same as `NOT (x = y)`                                                                                                              |
+| `x <> y: String x String -> Boolean`    | Same as `NOT (x = y)`                                                                                                              |
```
duglin, line 284: Something for future consideration:  String > String, etc...

cesql/spec.md
```diff
@@ -229,38 +259,35 @@ Corresponds to the syntactic rule `unary-operation`:
...
+| `x * y: Integer x Integer -> Integer`   | Returns the product of `x` and `y`                                                                                                 |
+| `x / y: Integer x Integer -> Integer`   | Returns the result of dividing `x` by `y`, rounded towards `0` to obtain an integer. Returns `0` and a _MathError_ if `y = 0`      |
+| `x % y: Integer x Integer -> Integer`   | Returns the remainder of `x` divided by `y`, where the result has the same sign as `x`. Returns `0` and a _MathError_ if `y = 0`   |
+| `x + y: Integer x Integer -> Integer`   | Returns the sum of `x` and `y`                                                                                                     |
+| `x - y: Integer x Integer -> Integer`   | Returns the difference of `x` and `y`                                                                                              |
+| `x = y: String x String -> Boolean`     | Returns `true` if the values of `x` and `y` are equal                                                                              |
```
duglin, line 282: For easier reading, I'd put something like `(case sensitive)` on each of these lines so that people don't need to look outside of each row to get the full definition of the operator. Then you can delete the sentence you added on line 289.

cesql/spec.md
```diff
@@ -278,6 +305,8 @@ For example, the pattern `_b*` will accept values `ab`, `abc`, `abcd1` but won't
 Both `%` and `_` can be escaped with `\`, in order to be matched literally. For example, the pattern `abc\%` will match
 `abc%` but won't match `abcd`.
 
+The pattern matching MUST be case sensitive.
```
duglin, line 308: As above, it might be nice to say it's "case sensitive" in each row rather than splitting the full definition between the table and a sentence 5 paragraphs away.

cesql/spec.md
```diff
@@ -307,18 +337,23 @@ Because of implicit casting, no functions with the same name and same arity but
 
 An overload on a variadic function is allowed only if the number of initial fixed arguments is greater than the maximum arity for that particular function name. Only one variadic overload is allowed.
```
duglin, line 338: I think the wording here might need tweaking:
- "An overload on a variadic function"... reads to me as "You can create an overload of a function that's already defined with variadic args", which I don't think is the intent
- "Only one variadic overload is allowed" need to be scoped to a particular function name

Perhaps something like:
A function name MAY have at most one variadic overload definition and only if the number of initial fixed arguments is greater than the maximum arity of all other function definitions for that function name.

cesql/spec.md
```diff
@@ -291,14 +320,15 @@ assumed valid, e.g. `EXISTS id` MUST always return `true`.
 
 | Definition                                             | Semantics                                                                  |
 | ------------------------------------------------------ | -------------------------------------------------------------------------- |
-| `x IN (y1, y2, ...): Any x Any^n -> Boolean`     | Returns `true` if `x` is equal to an element in the _Set_ of `yN` elements |
-| `x NOT IN (y1, y2, ...): Any x Any^n -> Boolean` | Same as `NOT (x IN set)`                                                   |
+| `x IN (y1, y2, ...): Any x Any^n -> Boolean`           | Returns `true` if `x` is equal to an element in the _Set_ of `yN` elements |
+| `x NOT IN (y1, y2, ...): Any x Any^n -> Boolean`       | Same as `NOT (x IN set)`                                                   |
 
 The matching is done using the same semantics of the equal `=` operator, but using `x` type as the target type for the implicit type casting.
```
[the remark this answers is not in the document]
duglin: If "n" can be zero, what does that mean for:
```
CONCAT(x1, x2, ...): String^n -> String
```
Is `CONCAT()` valid?  We may need to define the range of "n" on each use

cesql/spec.md
```diff
@@ -328,25 +363,31 @@ The following tables show the built-in functions that MUST be supported by a CES
...
 | ---------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `ABS(x): Integer -> Integer` | Returns the absolute value of `x` |
+| `ABS(x): Integer -> Integer` | Returns the absolute value of `x`. If the value of `x` is `-2147483648` (the most negative 32 bit integer value possible), then this returns `2147483647` as well as a _MathError_. |
+
+#### 3.5.3 Function Errors
+
+As specified in 3.3, in the event of an error a function MUST still return a valid return value for it's defined return type.
+A CESQL engine MUST guarantee that all built-in functions comply with this. For user defined functions, if they return one or more errors
+and fail to providea valid return value for their return type the CESQL engine MUST return the zero value for the return type of the
```
duglin, line 385: s/providea/provide a/

cesql/spec.md
```diff
@@ -355,26 +396,26 @@ left operand of the OR operation evalues to `true`, the right operand MUST NOT b
...
    1. If it's ambiguous, use the `y` type to search, in the set of ambiguous operators, every definition of the operator
       using the `y` type as the right parameter type:
       1. If such operator definition exists and is unique, cast `x` to the type of the left parameter
-      1. Otherwise, raise an error and the cast results are undefined
+      1. Otherwise, raise a _CastError_ and the result is `false`
 1. If the function is n-ary with `n > 1`:
-   1. If it's not ambiguous, cast all the arguments to the corresponding parameter types.
-   1. If it's ambiguous, raise an error and the cast results are undefined.
+   1. Cast all the arguments to the corresponding parameter types.
 1. If the operator is n-ary with `n > 2`:
```
duglin, line 416: I'm sure there a reason, but why "n>2" and not "n>0" ?

And I guess on the previous one, why "n > 1" and not "n > 0" ?

cesql/spec.md
Cali0707: @duglin I think that what actually happens is that subexpressions are evaluated left to right but that within that there is a precedence (coming from the antlr4 lexer for the language). Technically, what was originally written here is correct about the precedence of operations. Maybe we need to clarify that evaluation is left to right, but that the precedence of operators/functions takes priority over that L to R evaluation?

cesql/spec.md
Cali0707: Oh as far as I know the point of this is that there are no side effects in a CESQL statement - every function/operator/subexpression can be replaced with the output of that function/operator/subexpression and you would get the same result

cesql/spec.md
Cali0707: I keep going back to whether or not a missing attribute should be an error. I think with the introduction of zero values for each type along with the short circuiting of AND/OR, it makes sense to have missing attributes as an error again. A big part of the argument for not having an error was to try and understand an expression like:
```sql
missingAttribute AND true
```
as well as
```
true OR missingAttribute
```

But, with these new changes we have made I think that these can be reasoned about so including the error (with the zero type) makes more sense

cesql/spec.md
Cali0707: Yeah, having thought about it more (and based on my comment in the other thread about missing attributes) I think moving back to missing attributes resulting in an error makes sense.

cesql/spec.md
Cali0707: Thinking about this, I don't think it necessarily hurts the spec, but I find that it does give some clarity about the independence of parts of an expression. I'll try adding another sentence explaining the referential transparency

cesql/spec.md
Cali0707: Yeah, we could add support for that in the future. For now, I'd prefer to keep it out of the spec and expand if there is a request

cesql/spec.md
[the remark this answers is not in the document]
Cali0707: For `CONCAT`, I think having no parameters is allowed, it would just return an empty string

cesql/spec.md
Cali0707: For "n>2" instead of "n>0" this is because we already handled the "n=1" and "n=2" earlie with the unary and binary operator/function sections.

As for why we only handle binary operator resolution and not binary function resolution (the "n>1" vs "n>2"), I am not sure, this comes from before my time working on the spec. I'm inclined to believe there was a reason for it though

cesql/spec.md
Cali0707: I've removed the text about operators being listed in precedence order, since it did not seem to match the actual precedence from the parser. Instead, I have added a table with the updated precedence.

cesql/spec.md
duglin: I'm still not following. If I define an extension as an int, why must it be converted to a string ?
Especially if the cesql expression if `myint = 5` ?

duglin: Let me ask this... what breaks if you just remove these sentences and this notion of "initial type"? Things will get converted automatically when needed. I'm just not seeing the point of (or the meaning of) the phrase "initial type".

Thinking about this more... is this exposing some implementation detail because the SDKs convert all attributes to strings initially? If so, then it really shouldn't be in the spec since the spec should focus on the cesql language syntax and resulting semantics, not how the code goes from "syntax" to "result".

cesql/spec.md
```diff
@@ -193,12 +205,20 @@ be used in the `IN` operator.
...
+along with a _MissingAttributeError_.
+For example, `true AND (missingAttribute = "")` would evaluate to `false, (missingAttributeError)` as the subexpression `missingAttribute = ""` would be false, given that the return type for the `=` operator is _Boolean_.
+However, the expression `missingattribute * 5` would evaluate to `0, (missingAttributeError)` because the return type for the `*` operator is _Integer_. Note that this does not mean that the _value_ of the missing attribute
+is set to be the zero value for the type of the missing attribute. Rather, the subexpression with the missingAttribute returns the zero value of the return type of the subexpression.
+As an example, `1 / missingAttribute` does not raise a _MathError_ due to the division by zero, instead it returns `0, (missingAttributeError)` as that is the zero value for the return type of the subexpression.
+In cases where the return type of the subexpression cannot be determined by the CESQL engine, the CESQL engine MUST assume a return type of _Boolean_. In such cases, the return value would
+therefore be `false, (missingAttributeError)`.
```
duglin, line 221: These examples really help clarify things, even if I have to read it a few times to really let it sink in :-)

markpeek:
I appreciate the review and refinement that @jskeet @duglin @Cali0707 have been doing in this PR. Have the review and changes converged yet? If so, it might make sense for another week for the rest of the team to review before voting on the v1. Thoughts?

cesql/spec.md
```diff
@@ -327,26 +362,47 @@ The following tables show the built-in functions that MUST be supported by a CES
...
 ### 3.6. Evaluation of the expression
 
-Operators MUST be evaluated in order, where the parenthesized expressions have the highest priority over all the other
-operators.
+Operators and functions MUST be evaluated in order of precedence, and MUST be evaluated left to right when the precedence is equal. The order of precedence is as follows:
+
+1. Function invocations
+1. Unary operators
+  1. NOT unary operator
+  1. `-` unary operator
```
duglin, line 395: I don't think you got the indenting you were hoping for, I see this:
image

and this new direction is much clearer than what you had before.

cesql/spec.md
Cali0707: How about "If the value of the attribute or extension is not one of the primitive CESQL types, it MUST be represented by the _String_ type"

Cali0707:
> Have the review and changes converged yet? If so, it might make sense for another week for the rest of the team to review before voting on the v1. Thoughts?

To me it seems like we are converging, not sure if @duglin or @jskeet have other opinions. My thoughts would be that we should:
1. If everyone is okay with it in the meeting, merge these changes this week
2. Have another week for everyone to review the updated spec after these changes
3. Have the vote next Thursday

cesql/spec.md
```diff
@@ -56,7 +56,11 @@ producer, or in an intermediary, and it can be implemented using any technology
 The CloudEvents Expression Language assumes the input always includes, but is not limited to, a single valid and
 type-checked CloudEvent instance. An expression MUST NOT mutate the value of the input CloudEvent instance, nor any of
 the other input values. The evaluation of an expression observes the concept of [referential
-transparency][referential-transparency-wiki]. The output of a CESQL expression evaluation is always a _boolean_, an _integer_ or a _string_, and it might include an error.
+transparency][referential-transparency-wiki]. This means that any part of an expression can be replaced with it's output
```
jskeet, line 59: it's => its

cesql/spec.md
```diff
@@ -327,26 +359,47 @@ The following tables show the built-in functions that MUST be supported by a CES
...
 #### 3.5.2. Built-in Math functions
 
 | Definition                                                 | Semantics                                                                                                                                                                                                                  |
 | ---------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `ABS(x): Integer -> Integer` | Returns the absolute value of `x` |
+| `ABS(x): Integer -> Integer` | Returns the absolute value of `x`. If the value of `x` is `-2147483648` (the most negative 32 bit integer value possible), then this returns `2147483647` as well as a _MathError_. |
+
+#### 3.5.3 Function Errors
+
+As specified in 3.3, in the event of an error a function MUST still return a valid return value for it's defined return type.
```
jskeet, line 380: it's => its

duglin:
Approved on the 5/30 call.
Despite there being late changes to this PR (meaning, after Tuesday), we decided to merge it anyway so that people only need to look at the latest version of the spec to do their "v1.0" review. So, please look over the "main" branch version of the spec and add any comments to the "review issue": https://github.com/cloudevents/spec/issues/1282 

The current plans are to complete the final review by next week's call, and then (if the group approves), to start the official 1-week voting process.

PR #1287, opened 2024-05-23
attr names SHOULD start with a letter

### LINKED ISSUE #1285: Should attribute names be allowed to start with a digit?
Today the spec says:

```
In order to maximize the likelihood of interoperability and portability across transport protocols and
messaging formats, CloudEvents attribute names MUST consist of lower-case letters ('a' to 'z') or
digits ('0' to '9') from the ASCII character set. Attribute names SHOULD be descriptive and terse
and SHOULD NOT exceed 20 characters in length.
```

Normally digits aren't great as first characters of identifiers. 

If we want to restrict it w/o breaking the spec we could say they "SHOULD start with a-z" instead of "MUST"

duglin:
Some folks might try to map attribute names to variables or struct properties, and those often don't support starting with a digit, and while some kind of escaping or prefixing (e.g. with `_`) is possible, it's kind of annoying.

Fixed #1285


duglin:
Approved on the 5/30 call

Issue #1288, opened 2024-05-30
Improve Privacy and Security section in spec


AxelNennker:
The example provided in https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#privacy-and-security does not follow the recommendation of the privacy and security section to encrypt the data.

```JSON
{
    "specversion" : "1.0",
    "type" : "com.github.pull_request.opened",
    "source" : "https://github.com/cloudevents/spec/pull",
    "subject" : "123",
    "id" : "A234-1234-1234",
    "time" : "2018-04-05T17:31:00Z",
    "comexampleextension1" : "value",
    "comexampleothervalue" : 5,
    "datacontenttype" : "text/xml",
    "data" : "<much wow=\"xml\"/>"
}
```

Rana-KV:
Hi @AxelNennker, I'm interested in this topic. Can I work this issue if it is not assigned?

AxelNennker:
> Hi @AxelNennker, I'm interested in this topic. Can I work this issue if it is not assigned?

I am not a core member of cloudevents.
Also there is https://github.com/cloudevents/spec/pull/1290

So, read https://github.com/cloudevents/spec/blob/main/docs/GOVERNANCE.md and https://github.com/cloudevents/spec/blob/main/docs/CONTRIBUTING.md follow those guidelines

PR #1289, opened 2024-05-30
Update example in Privacy and Security Section

AxelNennker:
Fixes # https://github.com/cloudevents/spec/issues/1288

The example in the Privacy and Security section should follow its own recommendation and encrypt the data.

## Proposed Changes

- change content-type
- use jwe (value taken from the [jwe spec](https://www.rfc-editor.org/rfc/rfc7516#section-3.3))


AxelNennker:
oops, did not mean to have such a big PR. Not sure what happened.

PR #1291, opened 2024-06-04
Update CESQL tck tests to match spec changes

Cali0707:
With all of the CESQL spec changes recently, the spec needed some updates to reflect what had changed

## Proposed Changes

- Add results to every expression that has an error, as CESQL spec has been clarified on what should happen in that case
- Added missing attribute errors back, as well as tests to ensure that operators handle it correctly
- Add test for the `ABS` function overflow error
- Fix boolean casting now that it is defined for all types

### DIFF
--- a/cesql/cesql_tck/README.md
+++ b/cesql/cesql_tck/README.md
@@ -25,3 +25,4 @@ The `error` values could be any of the following:
 * `cast`: Casting error
 * `missingFunction`: Addressed a missing function
 * `functionEvaluation`: Error while evaluating a function
+* `missingAttribute`: Error due to a missing attribute
--- a/cesql/cesql_tck/README.md
+++ b/cesql/cesql_tck/README.md
@@ -26,3 +26,4 @@ The `error` values could be any of the following:
 * `missingFunction`: Addressed a missing function
 * `functionEvaluation`: Error while evaluating a function
 * `missingAttribute`: Error due to a missing attribute
+* `generic`: A generic error
--- a/cesql/cesql_tck/negate_operator.yaml
+++ b/cesql/cesql_tck/negate_operator.yaml
@@ -14,7 +14,11 @@ tests:
     expression: --'10'
     result: 10
 
-  - name: Invalid boolean cast
+  - name: Minus with boolean cast
     expression: -TRUE
+    result: -1
+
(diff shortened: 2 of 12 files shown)


cesql/cesql_tck/README.md
```diff
@@ -25,3 +25,4 @@ The `error` values could be any of the following:
 * `cast`: Casting error
 * `missingFunction`: Addressed a missing function
 * `functionEvaluation`: Error while evaluating a function
+* `missingAttribute`: Error due to a missing attribute
```
duglin, line 28: I think you added a generic error message to the spec - should we test for that too? It might end up being more of an impl-specific error case. If it's too hard to put into this generic tck stuff then I'm ok with skipping it. Was just curious.

cesql/cesql_tck/negate_operator.yaml
```diff
@@ -14,7 +14,11 @@ tests:
     expression: --'10'
     result: 10
 
-  - name: Invalid boolean cast
+  - name: Minus with boolean cast
     expression: -TRUE
+    result: -1
```
duglin, line 19: Correct result, but funny

cesql/cesql_tck/negate_operator.yaml
```diff
@@ -14,7 +14,11 @@ tests:
...
+
+  - name: Minus with missing attribute
+    expression: -missing
     result: 0
-    error: cast
+    error: match
```
duglin, line 24: What does "match' mean here - I don't see it listed as a valid error? I expected "missingAttribute"

duglin:
Just a few minor questions... otherwise LGTM

cesql/cesql_tck/negate_operator.yaml
Cali0707: Tbh I'm not sure how I ended up with `match` here...

You are totally correct, it should be `missingAttribute`

cesql/cesql_tck/README.md
Cali0707: I think we could mention it in the `README` for completeness, but I can't think of any use cases where we would actually use it

duglin:
Approved on the 6/6 call

PR #1292, opened 2024-06-06
fix: add trailing commas where appropriate for 'that is' in CESQL spec

Cali0707:
Addresses https://github.com/cloudevents/spec/issues/1282#issuecomment-2152782563

## Proposed Changes

- Add trailing commas where appropriate after "that is" in the spec (I audited all of them and only found the two mentioned in the original comment as being wrong

### DIFF
--- a/cesql/spec.md
+++ b/cesql/spec.md
@@ -2,8 +2,8 @@
 
 ## Abstract
 
-The goal of this specification is to define a SQL-like expression language which can be used to express predicates on
-CloudEvent instances.
+The goal of this specification is to define a SQL-like expression language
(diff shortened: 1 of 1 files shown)


Cali0707:
cc @markpeek @duglin

duglin:
@Cali0707 LGTM however, can you use this PR to wrap the spec at 80 columns? I just remember that it was the last outstanding change from the other PR. I think we've used: https://prettier.io/ in the past, if that helps

duglin:
@Cali0707 thanks!
Approved on the 7/6 call... merging

PR #1296, opened 2024-06-11
fix: CESQL conformance tests are correct with spec v1

Cali0707:
While implementing the v1 spec changes in the go sdk, I found that there were some inconsistencies with the conformance tests and the v1 spec. This PR fixes the conformance tests

## Proposed Changes

- Remove `IS_(BOOL|INT|STRING)` tests since the functions are not required by spec
- Fix cast of boolean no longer causing error
- Remove test that requires complete expression evaluation, as the spec leaves this to implementations
- Fix remaining tests to return missing attribute errors where necessary


duglin:
LGTM

duglin:
Approved on the 6/13 call

PR #1301, opened 2024-06-24
Add zero-trust-preproposal

xibz:
This commit adds the zero-trust-preproposal.md file which explores the various protocols to find their limitations around metadata which is a great place for signatures or any other security related information to live.

This is 1 of 2 proposals for zero trust. This initial proposal is to investigate each protocol and look into their limitations around metadata support which we will use to pass any necessary information to help verify any incoming requests.

## Proposed Changes

- Adds a pre-proposal for zero trust support

### DIFF
--- a/cloudevents/extensions/zero-trust-preproposal.md
+++ b/cloudevents/extensions/zero-trust-preproposal.md
@@ -1,0 +1,186 @@
+# CloudEvents Zero Trust (Pre-proposal)
+
+## Abstract
+
+The ultimate aim is to add zero trust capabilities to cloudevents. To achieve this, we will create two proposals. The first proposal, which we will call the pre-proposal, will explore the different protocols to understand their limitations and how we can implement zero trust. The second proposal will design the implementation of how we can achieve zero trust in cloudevents..
+
+It is important to note that cloudevents’ protocols operate at the application level, encompassing both encodings and wire protocols. This analysis will separate encodings from wire protocols, as encoding refers to the format in which data is sent and does not restrict the security libraries or tools we can use, unlike wire protocols.
+
+The primary focus will be on where metadata can be stored, such as headers in the HTTP protocol, in each protocol, rather than an in-depth examination of each protocol. Having a place to store metadata per request will allow us to configure and send any necessary security-related information, facilitating the implementation of zero trust. This proposal will not delve into various security algorithms or tools to be described in the metadata.
+
+## Wire Protocols
+
+### AMQP
+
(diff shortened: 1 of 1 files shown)


cloudevents/extensions/zero-trust-preproposal.md
```diff
@@ -0,0 +1,186 @@
...
+
+These are the encodings that cloudevents supports: AVRO, JSON, protobuf, and XML. Encodings do not affect whether we can rely on metadata, and will have no hindrance on the implementation of zero trust.
+
+### Webhook
+
+Webhooks are a high-level abstraction built on top of the HTTP protocol. They allow clients to send arbitrary HTTP requests to a specified endpoint in response to events. This flexibility means that webhooks do not limit the underlying wire protocol used for transmission. Consequently, webhooks enable the use of the protocol’s metadata format and other mechanisms to ensure the trustworthiness of requests.
+
+## Summary
+
+We have explored each wire protocol, and while some have more limitations than
```
clemensv, line 182: Any extension must indeed ignore all of these details and stay within the framework set by CloudEvents. If you want to provide end-to-end security for CloudEvents, the focus must be on the CloudEvents attributes. 

It appears feasible to apply S/MIME (https://datatracker.ietf.org/doc/html/rfc8551) to CloudEvents in the following (theoretical) model:

1. Map the CloudEvent to the "Internet Message" using the CloudEvents HTTP binding. The HTTP 1.1 message is a superset of the "Internet Message" model that is also used for SMTP.
2. Process the message using S/MIME for signature and/or encryption.
3. The result is an Internet Message that has now the desired security attributes applied.
4. You will then stash the entire resulting message into the "data" section of a wrapping CloudEvent, whose datacontenttype is "multipart/signed" or another applicable S/MIME format. The CloudEvents attributes that are okay to travel in the clear are copied onto this wrapper CloudEvent. You MUST copy all required CloudEvent attributes onto the wrapper: id, source, type, specversion. If that leaks too much data, you need to obscure the fact that you are sending CloudEvents at all.

Now you have a signed or encrypted CloudEvent.

Any receiver reverses the process, whereby it MUST check that the attributes that have been provided on the wrapper CloudEvent are exactly the same as in the contained, signed CloudEvent and that all the signatures are correct.

clemensv (changes requested):
The proposal must stay within the boundaries set by the CloudEvents framework.

xibz:
Discussed in SIG. Will move this to an issue where zero trust in general can be talked about

PR #1304, opened 2024-07-16
docs: link distributed tracing extension to otel semconv

joaopgrassi:
Replacement for #912

## Proposed Changes
This follows the same idea/goal as #912 - Point to the OpenTelemetry CloudEvents semantic conventions to help people to understand how to use the extension and how a trace structure should look like in systems using CloudEvents.

I took a slight different approach with this PR where I left most of the existing text. We have now finalized the CloudEvents semconv in OTel. The group there decided to only mention the distributed tracing extension as _an additional_ mechanism to propagate context, but the definition of it should stay here. OTel also doesn't document/specify other means of propagation in messaging systems for example, we rather only mention it. That's why we followed the same for CloudEvents.

The document now points to the CloudEvents OTel semconv which basically tells to follow the messaging semantic conventions. 

**Release Note**

```release-note
Clarify usage of Distributed Tracing Extension by OpenTelemetry
```

**Note:** The PR #912 is very old and I had many conflicts trying to rebase it. So I'm opening this new one. CC @duglin 

**Note 2:** The link to the OTel website doesn't contain the last version of the CloudEvents semantic conventions. The version that will go live soon is here: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/cloudevents/cloudevents-spans.md

### DIFF
--- a/cloudevents/extensions/distributed-tracing.md
+++ b/cloudevents/extensions/distributed-tracing.md
@@ -1,11 +1,28 @@
 # Distributed Tracing extension
 
 This extension embeds context from
-[Distributed Tracing](https://w3c.github.io/trace-context/) so that distributed
-systems can include traces that span an event-driven system. This extension is
-meant to contain historical data of the parent trace, in order to diagnose
-eventual failures of the system through tracing platforms like Jaeger, Zipkin,
-etc.
+[W3C TraceContext](https://www.w3.org/TR/trace-context/) into a CloudEvent.
+The goal of this extension is to offer means to carry context when instrumenting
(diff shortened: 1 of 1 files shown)


jskeet (approved):
I don't know much about the accuracy of the statements here, but they certainly seem useful.

manuelottlik:
I just read through the website, the changes and the discussion in the preceding PR – this looks like a very useful extension. Just one question (that should not block this PR): Is this actually implemented in the most common SDKs already or is it too early for such widespread adoption?

duglin:
Approved on the 7/25 call
thanks @joaopgrassi for all your work on this!

PR #1307, opened 2024-08-14
Add CloudEvents extension for deprecated events

vandewillysilva:
Introduced the `deprecation` attribute to indicate when an event type is deprecated, and added the `sunset` attribute to specify the date and time when the event will become unsupported. This extension provides clear guidelines and examples for implementing these attributes, aiming to improve lifecycle management and ensure better communication with consumers.

### DIFF
--- a/cloudevents/extensions/deprecation.md
+++ b/cloudevents/extensions/deprecation.md
@@ -27,7 +27,7 @@ extension is being used.
   `false` if not specified.
 - Constraints
   - OPTIONAL
-- Example: `"deprecated": "true"`
+- Example: `"deprecated": true`
 
 ### deprecationfrom
 
@@ -61,8 +61,8 @@ the migration path from the deprecated event to an alternative. This helps
 consumers transition away from the deprecated event.
 - Constraints
   - OPTIONAL
-  - The URI SHOULD point to valid and accessible resources that help the
-    consumer understand what SHOULD replace the deprecated event
+  - The URI SHOULD point to a valid and accessible resource that helps the
+    consumer understand what SHOULD replace the deprecated event.
 - Example: `"deprecationmigration": "https://example.com/migrate-to-new-evt"`
 
 ## Usage
--- a/cloudevents/extensions/deprecation.md
+++ b/cloudevents/extensions/deprecation.md
@@ -2,8 +2,8 @@
 
 This specification defines attributes that can be included in CloudEvents to
 indicate to [consumers](../spec.md#consumer) or
-[intermediaries](../spec.md#intermediary) the deprecation of the events. These
-attributes inform consumers of CloudEvents about upcoming changes or removals,
+[intermediaries](../spec.md#intermediary) the deprecation of events. These
+attributes inform CloudEvents consumers about upcoming changes or removals,
 facilitating smoother transitions and proactive adjustments.
 
 ## Notational Conventions
@@ -37,7 +37,7 @@ extension is being used.
 - Constraints
   - OPTIONAL
   - The `deprecationfrom` timestamp SHOULD remain stable once set and SHOULD
-    reflect a point in the past or the present. Pre-announcing deprecation by
+    reflect a point in the past or present. Pre-announcing deprecation by
     setting a future date is not encouraged.
 - Example: `"deprecationfrom": "2024-10-11T00:00:00Z"`
 
@@ -61,8 +61,8 @@ the migration path from the deprecated event to an alternative. This helps
 consumers transition away from the deprecated event.
 - Constraints
   - OPTIONAL
-  - The URI SHOULD point to a valid and accessible resource that helps the
-    consumer understand what SHOULD replace the deprecated event.
+  - The URI SHOULD point to a valid and accessible resource that helps
+    consumers understand what SHOULD replace the deprecated event.
 - Example: `"deprecationmigration": "https://example.com/migrate-to-new-evt"`
 
 ## Usage
--- a/cloudevents/extensions/deprecation.md
+++ b/cloudevents/extensions/deprecation.md
@@ -27,7 +27,6 @@ extension is being used.
   `false` if not specified.
 - Constraints
   - OPTIONAL
-  - The value MUST be a boolean (`true` or `false`).
 - Example: `"deprecated": "true"`
 
 ### deprecationfrom
--- a/cloudevents/extensions/deprecation.md
+++ b/cloudevents/extensions/deprecation.md
@@ -37,8 +37,6 @@ extension is being used.
   officially marked as deprecated.
 - Constraints
   - OPTIONAL
-  - If present, MUST adhere to the format specified in
-    [RFC 3339](https://tools.ietf.org/html/rfc3339)
   - The `deprecationfrom` timestamp SHOULD remain stable once set and SHOULD
     reflect a point in the past or the present. Pre-announcing deprecation by
     setting a future date is not encouraged.
@@ -54,8 +52,6 @@ extension is being used.
   - The timestamp MUST be later than or the same as the one given in the
     `deprecationfrom` field, if present. It MAY be extended to a later date but
     MUST NOT be shortened once set.
-  - If present, MUST adhere to the format specified in
-    [RFC 3339](https://tools.ietf.org/html/rfc3339)
 - Example: `"deprecationsunset": "2024-11-12T00:00:00Z"`
 
 ### deprecationmigration
--- a/cloudevents/extensions/deprecation.md
+++ b/cloudevents/extensions/deprecation.md
@@ -1,11 +1,10 @@
-# Deprecation and Sunset extension
+# Deprecation extension
 
 This specification defines attributes that can be included in CloudEvents to
 indicate to [consumers](../spec.md#consumer) or
-[intermediaries](../spec.md#intermediary) the deprecation and sunset status of
-the event type. These attributes inform consumers of CloudEvents about upcoming
-changes or removals, facilitating smoother transitions and proactive
-adjustments.
+[intermediaries](../spec.md#intermediary) the deprecation of the events. These
+attributes inform consumers of CloudEvents about upcoming changes or removals,
+facilitating smoother transitions and proactive adjustments.
 
 ## Notational Conventions
 
@@ -23,36 +22,65 @@ extension is being used.
 
 ### deprecation
 
-- Type: `String`
-- Description: Indicates that the event type is deprecated. This can be a
-  boolean `true` or a date in [RFC 5322](https://tools.ietf.org/html/rfc5322)
-  format indicating when the deprecation was announced.
+- Type: `Boolean`
+- Description: Indicates whether the event type is deprecated.
 - Constraints
   - REQUIRED
-  - If the value is a date, it MUST be formatted according to RFC 5322.
-- Examples:
-  - A boolean deprecation: `"deprecation": "true"`
-  - A date-based deprecation: `"deprecation": "Sun, 11 Aug 2024 23:59:59 GMT"`
+  - The value MUST be a boolean (`true` or `false`).
+- Example: `"deprecation": "true"`
 
-### sunset
+### deprecationfrom
 
-- Type: `String`
+- Type: `Timestamp`
+- Description: Specifies the date and time when the event type was
+  officially marked as deprecated.
+- Constraints
+  - OPTIONAL
+  - If present, MUST adhere to the format specified in
+    [RFC 3339](https://tools.ietf.org/html/rfc3339)
+- Example: `"deprecationfrom": "2024-10-11T00:00:00Z"`
+
+### deprecationsunset
+
+- Type: `Timestamp`
 - Description: Specifies the future date and time when the event type will
-  become unsupported, formatted according to
-  [RFC 8594](https://tools.ietf.org/html/rfc8594).
+  become unsupported.
 - Constraints
   - OPTIONAL
   - The timestamp MUST be later or the same as the one given in the
-    `deprecation` field.
-- Example: `"sunset": "Wed, 14 Aug 2024 23:59:59 GMT"`
+    `deprecationfrom` field.
+  - If present, MUST adhere to the format specified in
+    [RFC 3339](https://tools.ietf.org/html/rfc3339)
+- Example: `"deprecationsunset": "2024-11-12T00:00:00Z"`
+
+### deprecationmigration
+
+- Type: `URI`
+- Description: Provides a link to documentation or resources that describe
+the migration path from the deprecated event to an alternative. This helps
+consumers transition away from the deprecated event.
+- Constraints
+  - OPTIONAL
+  - The URI SHOULD point to valid and accessible resources that help the
+    consumer understand what SHOULD replace the deprecated event
+- Example: `"deprecationmigration": "https://example.com/migrate-to-new-evt"`
 
 ## Usage
 
 When this extension is used, producers MUST set the value of the `deprecation`
-attribute to `true` or the `timestamp` of when an event is being phased out but
-still supported. This gives consumers a heads-up that they SHOULD begin
-migrating to a new event type or version.
+attribute to `true`. This gives consumers a heads-up that they SHOULD begin
+migrating to a new event or version.
 
-The `sunset` attribute SHOULD specify the exact date and time when the event
-will no longer be supported. This is the final cutoff date after which the
-event will no longer function as expected.
+Consumers SHOULD use the `deprecation` and `deprecationfrom` attributes to
+monitor the lifecycle of events they rely on. If `deprecationsunset` has been
+reached, consumers SHOULD consider switching to the recommended replacement, as
+events may no longer be supported.
+
+If an event is received after the `deprecationsunset` timestamp, consumers
+SHOULD choose to stop processing such events, especially if unsupported events
+can cause downstream issues.
+
+Producers SHOULD stop emitting deprecated events after the `deprecationsunset`
+timestamp. They SHOULD also provide detailed documentation via the
+`deprecationmigration` attribute to guide consumers toward the correct replacement
+event.
--- a/cloudevents/extensions/deprecation.md
+++ b/cloudevents/extensions/deprecation.md
@@ -20,14 +20,15 @@ extension is being used.
 
 ## Attributes
 
-### deprecation
+### deprecated
 
 - Type: `Boolean`
-- Description: Indicates whether the event type is deprecated.
+- Description: Indicates whether the event type is deprecated. Defaults to
+  `false` if not specified.
 - Constraints
-  - REQUIRED
+  - OPTIONAL
   - The value MUST be a boolean (`true` or `false`).
-- Example: `"deprecation": "true"`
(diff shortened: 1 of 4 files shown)


cloudevents/extensions/deprecation.md
```diff
@@ -0,0 +1,58 @@
...
+- Examples:
+  - A boolean deprecation: `"deprecation": "true"`
+  - A date-based deprecation: `"deprecation": "Sun, 11 Aug 2024 23:59:59 GMT"`
+
+### sunset
+
+- Type: `String`
+- Description: Specifies the future date and time when the event type will
+  become unsupported, formatted according to
+  [RFC 8594](https://tools.ietf.org/html/rfc8594).
```
jskeet, line 42: While I can see the purpose of using RFC 8594 here, CE isn't tightly bound to HTTP, and as it already has a Timestamp format. I'd definitely prefer this to be a Timestamp-typed attribute, whatever we call it.

cloudevents/extensions/deprecation.md
```diff
@@ -0,0 +1,58 @@
...
+needs to be in all CloudEvents, rather it needs to be included only when this
+extension is being used.
+
+## Attributes
+
+### deprecation
+
+- Type: `String`
+- Description: Indicates that the event type is deprecated. This can be a
+  boolean `true` or a date in [RFC 5322](https://tools.ietf.org/html/rfc5322)
```
jskeet, line 28: Rather than having a String that can be one of two formats, I'd prefer to have two attributes, each of which uses a single existing CE type - one `Boolean`, one `Timestamp`. (e.g. `deprecated` and `deprecatedfrom`)

cloudevents/extensions/deprecation.md
```diff
@@ -0,0 +1,58 @@
...
+  format indicating when the deprecation was announced.
+- Constraints
+  - REQUIRED
+  - If the value is a date, it MUST be formatted according to RFC 5322.
```
jskeet, line 32: Why introduce a different (and relatively hard to parse, culture-sensitive) format when we already use RFC3339 for the CloudEvent Timestamp format?

cloudevents/extensions/deprecation.md
```diff
@@ -0,0 +1,58 @@
...
+- Examples:
+  - A boolean deprecation: `"deprecation": "true"`
+  - A date-based deprecation: `"deprecation": "Sun, 11 Aug 2024 23:59:59 GMT"`
+
+### sunset
```
jskeet, line 37: If we do end up with two attributes above, I'd prefer this to be something starting with "deprecated" or "deprecation" - perhaps "deprecationsunset" is best, but it keeps them all together.

cloudevents/extensions/deprecation.md
```diff
@@ -0,0 +1,58 @@
...
+## Usage
+
+When this extension is used, producers MUST set the value of the `deprecation`
+attribute to `true` or the `timestamp` of when an event is being phased out but
+still supported. This gives consumers a heads-up that they SHOULD begin
+migrating to a new event type or version.
+
+The `sunset` attribute SHOULD specify the exact date and time when the event
+will no longer be supported. This is the final cutoff date after which the
+event will no longer function as expected.
```
jskeet, line 58: We should probably have some guidance around what different actors might do with an event that they see after this time. For example, if you're an "event forwarder" should you just drop such events?

cloudevents/extensions/deprecation.md
```diff
@@ -0,0 +1,58 @@
...
+"MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT",
+"RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as
+described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+However, the scope of these key words is limited to when this extension is
+used. For example, an attribute being marked as "REQUIRED" does not mean it
+needs to be in all CloudEvents, rather it needs to be included only when this
+extension is being used.
+
+## Attributes
```
jskeet, line 22: I'd like some sort of attribute describing a migration path - "stop using this, start using X". We could make that a URI instead of text, for people to link to docs. Not sure of the details, but something would be good.

jskeet (commented):
Definitely happy with the idea of being able to signal deprecation - definitely don't want to introduce more timestamp formats.

cloudevents/extensions/deprecation.md
ttripp: Although a different purpose, the `Deprecation` name is aligned with [this IETF draft standard](https://datatracker.ietf.org/doc/html/draft-ietf-httpapi-deprecation-header-04).

jskeet: Again, that's HTTP-centric though. While I don't mind if attribute names happen to coincide with HTTP ones, I don't think that should be a driving factor.

cloudevents/extensions/deprecation.md
vandewillysilva: Changed type to `Timestamp`

cloudevents/extensions/deprecation.md
vandewillysilva: separated out the attributes:
- deprecation
- deprecationfrom

cloudevents/extensions/deprecation.md
vandewillysilva: fixed

cloudevents/extensions/deprecation.md
vandewillysilva: Renamed to deprecationsunset

cloudevents/extensions/deprecation.md
vandewillysilva: Added a new (optional) attribute: `deprecationmigration`

cloudevents/extensions/deprecation.md
vandewillysilva: Added a new paragraph in the `Usage` section. Please take a look at it.

cloudevents/extensions/deprecation.md
```diff
@@ -0,0 +1,86 @@
...
+
+### deprecation
+
+- Type: `Boolean`
+- Description: Indicates whether the event type is deprecated.
+- Constraints
+  - REQUIRED
+  - The value MUST be a boolean (`true` or `false`).
+- Example: `"deprecation": "true"`
```
jskeet, line 30: I'm torn between the consistency of having a common prefix, and the oddity of it sounding like it's not Boolean (whereas "deprecated: true" sounds fine).

Thinking about it more, when would this ever actually have a value of false? That would be unusual at least...

cloudevents/extensions/deprecation.md
```diff
@@ -0,0 +1,86 @@
...
+
+### deprecationfrom
+
+- Type: `Timestamp`
+- Description: Specifies the date and time when the event type was
+  officially marked as deprecated.
+- Constraints
+  - OPTIONAL
+  - If present, MUST adhere to the format specified in
```
jskeet, line 39: I suspect we don't need this, because it's inherent in the type. (And only the string representation needs to adhere to a format anyway - if this is formatted with a Protobuf formatter for example, it'll end up as a protobuf timestamp.)

Ditto deprecationsunset

cloudevents/extensions/deprecation.md
```diff
@@ -0,0 +1,86 @@
...
+- Example: `"deprecationmigration": "https://example.com/migrate-to-new-evt"`
+
+## Usage
+
+When this extension is used, producers MUST set the value of the `deprecation`
+attribute to `true`. This gives consumers a heads-up that they SHOULD begin
+migrating to a new event or version.
+
+Consumers SHOULD use the `deprecation` and `deprecationfrom` attributes to
+monitor the lifecycle of events they rely on. If `deprecationsunset` has been
```
jskeet, line 75: Presumably consumers should consider switching to the recommended replacement as soon as they see that the event is deprecated, really? (You don't want to wait until the sunset.) Perhaps this should be more "consumers should make efforts to switch to the recommended replacement before the specified sunset"?

cloudevents/extensions/deprecation.md
```diff
@@ -0,0 +1,86 @@
...
+reached, consumers SHOULD consider switching to the RECOMMENDED replacement, as
```
jskeet, line 76: I don't think RECOMMENDED needs to be in capitals here.

cloudevents/extensions/deprecation.md
```diff
@@ -0,0 +1,86 @@
...
+  - If present, MUST adhere to the format specified in
+    [RFC 3339](https://tools.ietf.org/html/rfc3339)
+- Example: `"deprecationfrom": "2024-10-11T00:00:00Z"`
+
+### deprecationsunset
+
+- Type: `Timestamp`
+- Description: Specifies the future date and time when the event type will
+  become unsupported.
+- Constraints
```
jskeet, line 48: I wonder whether it's worth constraining this to say that the sunset for a type should never be observed to become earlier, but may be observed to become later. (An event producer may choose to support an event for longer than initially expected, but shouldn't shorten that support window.)

cloudevents/extensions/deprecation.md
```diff
@@ -0,0 +1,86 @@
...
+  - REQUIRED
+  - The value MUST be a boolean (`true` or `false`).
+- Example: `"deprecation": "true"`
+
+### deprecationfrom
+
+- Type: `Timestamp`
+- Description: Specifies the date and time when the event type was
+  officially marked as deprecated.
+- Constraints
```
jskeet, line 37: Should we have a constraint around the stability of this? (It would be odd for it to vary over time.)

Should this always be in the (potentially very near) past or does it make sense to pre-announce deprecation?

jskeet (commented):
This looks a *lot* better to me, thanks so much. Various additional questions...

Fannon:
@vandewillysilva : Sorry, I didn't have time to look deeper into it, but just wanted to drop some links:

Here is a RFC for sunset as a header: https://datatracker.ietf.org/doc/html/rfc8594

At my company, SAP we have also standardized the same attributes as you propose here. In case you want to have a look, they're documented in the Open Resource Discovery specification, here: https://sap.github.io/open-resource-discovery/spec-v1/interfaces/document#event-resource_releasestatus

In particular, check `releaseStatus` (has deprecated as one value), `deprecationDate`, `sunsetDate`, `successors`. 

Also, we noticed that deprecating a resource does not necessarily have to imply a sunset. Of course you have to deprecate first, and sometimes you already plan the sunset when you do. But sometimes, you just created a successor resource and want just to indicate that the old resource is deprecated and the newer should be used.

cloudevents/extensions/deprecation.md
Fannon: Does this have to be REQUIRED? 

If you just have a `deprecated`: true and imply a default of false, there is no need to add a mandatory attribute.

cloudevents/extensions/deprecation.md
vandewillysilva: we have that in the recordedtime extension: https://github.com/cloudevents/spec/blob/main/cloudevents/extensions/recordedtime.md?plain=1#L45-L46

I was just following the same convention

cloudevents/extensions/deprecation.md
vandewillysilva: there is a lint script that complains about that. I'll reword it to `suggested`.

cloudevents/extensions/deprecation.md
jskeet: I'll happily create a PR to get rid of it from there :)

cloudevents/extensions/deprecation.md
vandewillysilva: added a new constraint.

cloudevents/extensions/deprecation.md
vandewillysilva: added a constraint.

cloudevents/extensions/deprecation.md
vandewillysilva: Rephrased this paragraph. Please take a look at it.

cloudevents/extensions/deprecation.md
vandewillysilva: renamed to deprecated

cloudevents/extensions/deprecation.md
vandewillysilva: Remove reference to RFC 3339

vandewillysilva:
> @vandewillysilva : Sorry, I didn't have time to look deeper into it, but just wanted to drop some links:
> 
> Here is a RFC for sunset as a header: https://datatracker.ietf.org/doc/html/rfc8594
> 
> At my company, SAP we have also standardized the same attributes as you propose here. In case you want to have a look, they're documented in the Open Resource Discovery specification, here: https://sap.github.io/open-resource-discovery/spec-v1/interfaces/document#event-resource_releasestatus
> 
> In particular, check `releaseStatus` (has deprecated as one value), `deprecationDate`, `sunsetDate`, `successors`.
> 
> Also, we noticed that deprecating a resource does not necessarily have to imply a sunset. Of course you have to deprecate first, and sometimes you already plan the sunset when you do. But sometimes, you just created a successor resource and want just to indicate that the old resource is deprecated and the newer should be used.

@Fannon thanks for the links and insights. It's great to see that we're aligned on similar attributes. :)

cloudevents/extensions/deprecation.md
```diff
@@ -0,0 +1,87 @@
...
+## Attributes
+
+### deprecated
+
+- Type: `Boolean`
+- Description: Indicates whether the event type is deprecated. Defaults to
+  `false` if not specified.
+- Constraints
+  - OPTIONAL
+  - The value MUST be a boolean (`true` or `false`).
```
jskeet, line 30: Just like the timestamp part, I don't think we need to say this - it's implicit in the type being Boolean.

jskeet (approved):
One more nit - but basically I'm otherwise happy :)

cloudevents/extensions/deprecation.md
vandewillysilva: agreed. Removed the constraint

santhoshhpe:
A must have feature

santhoshhpe (approved):

vandewillysilva:
@duglin Do you think it’s a good idea to proceed with this deprecation extension proposal? Let me know if there’s any other work needed or if you have any additional feedback.

cloudevents/extensions/deprecation.md
```diff
@@ -0,0 +1,86 @@
...
+## Attributes
+
+### deprecated
+
+- Type: `Boolean`
+- Description: Indicates whether the event type is deprecated. Defaults to
+  `false` if not specified.
+- Constraints
+  - OPTIONAL
+- Example: `"deprecated": "true"`
```
duglin, line 30: `"deprecated": true`
no quotes

cloudevents/extensions/deprecation.md
```diff
@@ -0,0 +1,86 @@
...
+
+### deprecationfrom
+
+- Type: `Timestamp`
+- Description: Specifies the date and time when the event type was
+  officially marked as deprecated.
+- Constraints
+  - OPTIONAL
+  - The `deprecationfrom` timestamp SHOULD remain stable once set and SHOULD
```
duglin, line 39: Just curious, aside from this being a possible "FYI" bit of info, do people use this timestamp for some purpose in real life? Doesn't seem like it would influence people's action.... the "sunset" one is the important one.

cloudevents/extensions/deprecation.md
```diff
@@ -0,0 +1,86 @@
...
+
+### deprecationmigration
+
+- Type: `URI`
+- Description: Provides a link to documentation or resources that describe
+the migration path from the deprecated event to an alternative. This helps
+consumers transition away from the deprecated event.
+- Constraints
+  - OPTIONAL
+  - The URI SHOULD point to valid and accessible resources that help the
```
duglin, line 64: s/help/helps/

cloudevents/extensions/deprecation.md
vandewillysilva: the `deprecationsunset` dictates the hard deadline, but `deprecationfrom` can prompt consumers to start planning their migration earlier.
IMO, it adds valuable context and encourages proactive behavior from consumers.

cloudevents/extensions/deprecation.md
vandewillysilva: fixed

cloudevents/extensions/deprecation.md
vandewillysilva: fixed

cloudevents/extensions/deprecation.md
```diff
@@ -0,0 +1,86 @@
...
+described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+However, the scope of these key words is limited to when this extension is
+used. For example, an attribute being marked as "REQUIRED" does not mean it
+needs to be in all CloudEvents, rather it needs to be included only when this
+extension is being used.
+
+## Attributes
+
+### deprecated
```
duglin, line 23: Since we basically say that even REQUIRED fields are really optional if the extension isn't used, I'm wondering if this one should be REQUIRED. It would be kind of weird to have this one "false" but any of the other attributes present, right?

cloudevents/extensions/deprecation.md
duglin: In theory, but I would think the presence of any of these would indicate that the event is deprecated regardless of when the deprecation period stated, right?

duglin:
> @duglin Do you think it’s a good idea to proceed with this deprecation extension proposal? Let me know if there’s any other work needed or if you have any additional feedback.

Yup - it's on the agenda for tomorrow's call. Will you be on the call? Not a requirement but it would be nice :-)
https://docs.google.com/document/d/1OVF68rpuPK5shIHILK9JOqlZBbfe91RNzQ7u_P7YCDE/edit#heading=h.6xidlvhv1goi

cloudevents/extensions/deprecation.md
vandewillysilva: that's correct.
`deprecationfrom` adds an extra layer of clarity. This can be useful for teams who need to plan their migrations earlier rather than waiting until the sunset approaches.

vandewillysilva:
PR is failing due to broken links in avro spec. This PR #1309 fixes the issue.

vandewillysilva:
@duglin I went over the recording and made the changes related to the deprecated field. Please take a look at it.

cloudevents/extensions/deprecation.md
```diff
@@ -0,0 +1,86 @@
...
+
+- Type: `Boolean`
+- Description: Indicates whether the event type is deprecated.
+- Constraints
+  - MUST be true
```
duglin, line 28: s/true/`true`/g

duglin:
@vandewillysilva just one minor tweak and then we can merge

vandewillysilva:
> @vandewillysilva just one minor tweak and then we can merge

Good catch! Fixed.

duglin:
Approved on the 8/22 call.
Thanks @vandewillysilva

PR #1311, opened 2024-09-26
Add RabbitMQ to proprietary specs and SDK support matrix

jskeet:
(This is no doubt too late for today's call, but I don't think it's urgent anyway.)


duglin:
LGTM thanks!

duglin:
Approved on the 9/26 call

PR #1320, opened 2024-11-20
add another trusted host

duglin:


duglin:
executive decision...merging so the other PR stops failing

PR #1328, opened 2025-02-12
lowercase all attributes in subscription spec

### LINKED ISSUE #824: [Registry] Align naming conventions for Registry and subscription API
The Open API specs for [discovery ](https://github.com/cloudevents/spec/blob/master/ce_discovery.yaml) and [subscription](https://github.com/cloudevents/spec/blob/master/subscriptions-api-openapi.yaml) do not follow the same conventions. The spec of the subscription API uses camelCase and has `operationIds`. While #821 adds `operationId`, there is still the difference regarding camelCase. 
If we agree on aligning this, I would be willing to come up with a PR for the discovery API.

duglin:
Fixes #824

### DIFF
--- a/subscriptions/spec.md
+++ b/subscriptions/spec.md
@@ -388,7 +388,7 @@ Each subscription is represented by an object that has the following properties:
 - Examples:
   - `https://example.com/event-processor`
 
-##### sinkCredential
+##### sinkcredential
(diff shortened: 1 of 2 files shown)


duglin:
Approved on the 2/13 call

PR #1332, opened 2025-03-20
Remove the "stale" action/bot per 3/20 call

duglin:
It's become just noise.
Will delete the label once the PR is merged


duglin:
Approved on the 3/27 call

PR #1333, opened 2025-03-27
fix ci issue

duglin:
x.com returns really really long headers now


duglin:
executive decision... merging

PR #1347, opened 2025-05-22
#1339 split opcua doc

### LINKED ISSUE #1339: Improve OPCUA documentation

thompson-tomo:
Fixes #1339

## Proposed Changes

- Splits the OPCUA document in 2

**Release Note**

```
This change splits the OPCUA into 2 distinct documents, with 1 describing the extensions necessary for the messages to be mapped to cloud events and the other document being an adapter document describing how the OPCUA message properties are to be mapped to either the core cloud event message or to generic extensions.
```

### DIFF
--- a/cloudevents/extensions/opcua.md
+++ b/cloudevents/extensions/opcua.md
@@ -25,7 +25,6 @@ is being used.
 
 - Type: `Integer`
 - Description: Links dataset message to the current version of the metadata.
-Contains value from `MajorVersion` of [Data Set Message Header](https://reference.opcfoundation.org/Core/Part14/v105/docs/7.2.5.4#Table164) field `MetaDataVersion`.
 - Constraints
   - OPTIONAL but MUST NOT be present if `dataschema` is used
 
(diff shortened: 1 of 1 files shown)


duglin:
ping @koepalex for your thoughts

duglin:
@thompson-tomo sorry for the delay, but would you be able to make a call at 9pm your time (I think it's 1pm CET and 7am ET) any time soon? @clemensv and I would like to discuss your issues/PRs and it would be easier via voice than via github.

duglin:
ping @thompson-tomo - see previous comment

duglin:
@thompson-tomo are you available for a voice chat one day?

duglin:
@thompson-tomo if we don't hear back from you soon we'll be forced to close your issues/PRs. Would you like to chat about them?

duglin:
Closing due to lack of response from author. Can reopen when they reengage

PR #1359, opened 2025-08-19
Fixup ce@v1.0.2 for new release process

duglin:


duglin:
Merging per agreement on new process

PR #1360, opened 2025-08-19
fix link for new release process

duglin:


duglin:
Merging per agreement in group to new release structure

PR #1361, opened 2025-08-19
fixup for new release process

duglin:


duglin:
Merging per agreement on new process

PR #1362, opened 2025-08-19
Fixup for new process & support shortened release names

duglin:
- typos
- use new tags names in URLs
- support shortened tag names - e.g. ce@v1.0 points to HEAD of ce@v1.0.2-branch (which is ce@v1.0.2 tag)


duglin:
If people review this,don't look at each individual commit - just look at the final results. It'll be easier

duglin:
Approved on the 8/21  call

PR #1371, opened 2025-10-06
feat: DSSE signing extension

### LINKED ISSUE #1302: Adding zero trust to cloudevents

xibz:
## Proposed Changes

feat: Add DSSE-based cryptographic signing for CloudEvents verification

Implements verifiable CloudEvents using DSSE (Dead Simple Signing Envelope) to ensure event authenticity and integrity across untrusted transport layers.

Key features:
- Sign CloudEvents with DSSE v1.0.2 protocol using SHA256 digests
- Transport verification material in 'dssematerial' extension attribute
- Support for binary, structured, and batch CloudEvent modes
- Backward compatible - unsigned events still work, consumers can ignore signatures (but highly inadvisable)

Technical approach:
- Creates SHA256 digest chain of all context attributes and event data
- Wraps digest in DSSE envelope with Base64 encoding
- Verifies by recomputing digests and comparing against signed payload
- Returns only verified data to consumers (strips unverified extensions)

This enables cryptographic proof that events:
1. Were produced by the claimed source (authenticity)
2. Were not modified in transit (integrity)

Does NOT address: event ordering, completeness, replay attacks, or confidentiality

Fixes #1302 

**Release Note**

```release-note

```


cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,443 @@
...
+## Non-goals
+
+This proposal only applies to individual events.
+It does not give consumers any guarantees about the completeness of the event stream or the order that events are delivered in.
+
+The threats of a malicious actor preventing events from being delivered or swapping their order are not addressed by this proposal.
+Neither are the possibilities of events being accidentally dropped, delivered in the wrong order or the same event being delivered multiple times.
+These challenges can be addressed through means of adding the necessary information inside the event payloads.
+
+Because the CloudEvents specification [requires](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#id) the combination of event `source` and `id` to be unique per event, signature replays for identical events are not considered.
```
duglin, line 22: is "signature replays" a "security" phrase? I wonder if everyone will know what's being said here. I'm not sure I do :-)     Is it saying "we don't guarantee that replays will look the same", or is it saying "different events with the same id/souce might look the same even if other data is different" ? Or something else?

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,443 @@
...
+
+While the design in this proposal *can* be used by authorized intermediaries to modify and re-sign events, it explicitly does not aim to provide a cryptographic audit trail of event modifications.
+
+As a general principle, this proposal aims to avoid cryptographic agility in favor of simplicity.
+
+## Constraints
+
+We have set the following constraints for the proposed design:
+
+**Verifiability MUST be OPTIONAL:** This ensures that the additional burden of producing verification material and performing verification only applies when verifiability is desired, which is not always the case.
```
duglin, line 36: I wonder if it's clear that just because a producer adds the extra verifiable stuff to the CE, the consumer is not required to do anything with it or even know what those extensions mean? Could just be me.

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,443 @@
...
+
+## Overview
+
+The producer passes a CloudEvent to the SDK, which creates the verification material and adds it to the CloudEvent. When the consumer’s CloudEvents SDK receives a message with event and verification material, it performs a verification of the signature against the key and passes on a verified event to the consumer:
+
+![An illustration showing how the producer's SDK signs an event, and the consumer's SDK verifies it](verifiability1.png)
+
+The verification material is transported in an [Extension Context Attribute](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#extension-context-attributes) called `dssematerial`:
+
+* Type: `String`
```
duglin, line 52: You may want to follow the pattern in other extensions of having an "Attributes" section - see: https://github.com/cloudevents/spec/blob/52387e31cc41688ba1ce56ec8f040554d3517592/cloudevents/extensions/authcontext.md#attributes

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,443 @@
...
+* Description: The [DSSE JSON Envelope](https://github.com/secure-systems-lab/dsse/blob/master/envelope.md) that can be used to verify the authenticity and integrity of CloudEvent.
+* Constraints:
+    * OPTIONAL
+    * If present, MUST be Base64 encoded
+
+The verification material, once Base64 decoded, looks something like this:
```
duglin, line 58: "looks something like" isn't precise enough. Are all of those 5 fields required? If so, then say "MUST adhere to the following form". If the fields can change based on the mechanism used to do the verifying then we'll need to be more creative, so let's discuss because across all mechanisms/formats there MUST be at least one consistent field (payloadType I assume) that people can rely upon to disambiguating them.

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,443 @@
...
+6. read the envelope’s `payloadType` value
+    1. if it is not equal to `https://cloudevents.io/verifiability/dsse/v0.1`, the verification payload is unknown and the event MUST be discarded
+7. read the envelope’s `payload` field and Base64 decode it into a byte sequence
+    1. if the result is not an sequence of the length 32, then the verification payload is corrupted and the event MUST be discarded
+8. create verification material from the event according to the signing [Protocol](#protocol)
+9. compare the newly created verification material to that in the envelope
+    1. if the values are not equal, the event has been modified in transit and MUST be discarded
+10. the event is returned as verified successfully.
+
+Upon verification of an CloudEvent, implementations MUST return a new event containing only verified data: the Context Attributes (REQUIRED and OPTIONAL) plus the event data. Extension Context Attributes MUST NOT be included in the verified event. This ensures clear separation between verified and unverified data. Users handle either a complete unverified event or a verified event with only verified values—never a mixture of both.
```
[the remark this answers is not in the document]
duglin: No, this isn't about audits or chains... it's about situations where there are environments where messages are sent thru middleware that will sign/verify messages with out the apps on either side knowing about it. Which means if there are multiple layers of that middleware then the design of this needs to support the idea of signing a message that has already been signed. And, of course, this will then also deal with cases where the middleware adds new attributes to the messages that need to be signed

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,443 @@
...
+    5. compute the SHA256 digest of the event's [`datacontenttype`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#datacontenttype) Context Attribute in UTF8 and append it to the byte sequence *(if the attribute is not set, use the digest of the empty byte sequence)*
+    6. compute the SHA256 digest of the event's [`dataschema`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#dataschema) Context Attribute in UTF8 and append it to the byte sequence *(if the attribute is not set, use the digest of the empty byte sequence)*
+    7. compute the SHA256 digest of the event's [`subject`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#subject) Context Attribute in UTF8 and append it to the byte sequence *(if the attribute is not set, use the digest of the empty byte sequence)*
+    8. compute the SHA256 digest of the event's [`time`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#time) in RFC 3339 Zulu format and append it to the byte sequence *(if the attribute is not set, use the digest of the empty byte sequence)*
```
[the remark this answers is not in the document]
duglin: Meaning, add the note text to this bullet with a MUST  ??

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,545 @@
...
+
+It is the digest of the concatenated digest list of the mandatory Context
+Attributes, the OPTIONAL Context Attributes as well as the event data itself.
+
+#### Protocol
+
+This is how to sign a CloudEvent using DSSE:
+
+1. choose a signing key
+2. create an empty byte sequence
```
jskeet: I like the fact that each SHA256 is created separately here, then they're all concatenated together. The benefit of that is that when verifying, if I get an event with data before the id, I can still handle it and remember the hash instead of having to retain the complete byte sequence in memory. It may be worth noting this as a design decision.

duglin:
@xibz I see it now supports extensions, nice!

Am I correct that it still doesn't support nested signing? E.g. app signs using this mechanism and then some middleware (blindly) also signs the event because that's it job regardless of what some other logic may have done beforehand.

xibz:
@duglin - Correct, this mechanism does not support nested signing in the sense of one signer blindly wrapping or re-signing an already signed event.

What you’re describing is more accurately chained signatures. That’s a distinct use case and is typically addressed by higher-level constructions (e.g., Biscuit-style signatures or other signature chaining schemes) that compose multiple signatures using an explicit and well-defined structure.

This proposal intentionally focuses on a single-signature model and does not attempt to define or standardize signature chaining. The goal is to keep the scope as small as possible as signature chaining would be expected to be handled by a separate proposal layered on top of this mechanism.

duglin:
Please see the error in the checker: https://github.com/cloudevents/spec/actions/runs/21483812306/job/61887191695?pr=1371

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,545 @@
...
+  5. compute the SHA256 digest of the event's [`datacontenttype`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#datacontenttype) Context Attribute in UTF8 and append it to the byte sequence *(if the attribute is not set, use the digest of the empty byte sequence)*
+  6. compute the SHA256 digest of the event's [`dataschema`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#dataschema) Context Attribute in UTF8 and append it to the byte sequence *(if the attribute is not set, use the digest of the empty byte sequence)*
+  7. compute the SHA256 digest of the event's [`subject`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#subject) Context Attribute in UTF8 and append it to the byte sequence *(if the attribute is not set, use the digest of the empty byte sequence)*
+  8. compute the SHA256 digest of the event's [`time`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#time) Context Attribute normalized to RFC 3339 Zulu format and append it to the byte sequence (if the attribute is not set, use the digest of the empty byte sequence)
```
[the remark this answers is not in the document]
jskeet: This is still an issue, I believe.

jskeet (commented):
Some more challenges, I'm afraid.

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,691 @@
...
+    8. compute the SHA256 digest of the event's [`time`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#time) Context Attribute normalized to RFC 3339 Zulu format and append it to the byte sequence (if the attribute is not set, use the digest of the empty byte sequence)
+       Note: Time normalization to Zulu format ensures signature verification remains valid even when intermediaries deserialize and reserialize events with different timezone representations of the same timestamp.
+    9. compute the SHA256 digest of the event's [`data`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#event-data) and append it to the byte sequence
+    10. for each extension attribute in the list from step 2 (in the given order, from lowest to highest index)
+        1. compute the SHA256 digest of the extension attribute's name and append it to the byte sequence
+        2. compute the SHA256 digest of the extension attribute's value and append it to the byte sequence
```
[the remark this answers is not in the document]
xibz: I disagree that this is inherently a protocol brittleness issue, after some thought. The protocol defines a deterministic canonicalization + hashing procedure, and the signature binds that canonical representation.

If an intermediary or SDK heuristically coerces values (e.g., parses a string into a timestamp) and then re-serializes it differently, that is a modification of signed content. In that case, signature verification should fail, by design, because the payload is no longer bitwise/canonically equivalent to what was signed.

Put differently: signature semantics are not “sign the meaning,” they’re “sign the canonical form.” Intermediaries that wish to forward signed envelopes must treat signed fields as immutable with respect to canonicalization. If they want to change representations (even to a semantically-equivalent timestamp), they must resign or persist the original value somewhere.

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,691 @@
...
+    5. compute the SHA256 digest of the event's [`datacontenttype`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#datacontenttype) Context Attribute in UTF8 and append it to the byte sequence *(if the attribute is not set, use the digest of the empty byte sequence)*
+    6. compute the SHA256 digest of the event's [`dataschema`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#dataschema) Context Attribute in UTF8 and append it to the byte sequence *(if the attribute is not set, use the digest of the empty byte sequence)*
+    7. compute the SHA256 digest of the event's [`subject`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#subject) Context Attribute in UTF8 and append it to the byte sequence *(if the attribute is not set, use the digest of the empty byte sequence)*
+    8. compute the SHA256 digest of the event's [`time`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#time) Context Attribute normalized to RFC 3339 Zulu format and append it to the byte sequence (if the attribute is not set, use the digest of the empty byte sequence)
```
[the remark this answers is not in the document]
duglin: > For other attributes, the canonical serialized representation defined by the event format applies.

is it inconsistent, or weird, to have some attributes signed pre-serialization and some signed post-serialization? I guess I'm also (indirectly) asking why timestamps are special? And if there's a good reason for them to be special, how do we know there won't be others as people add extensions?

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,674 @@
...
+  that can be used to verify the authenticity and integrity of the CloudEvent.
+- Constraints:
+  - REQUIRED
+  - If present, MUST be Base64 encoded
+  - When decoded, MUST be a valid DSSE JSON Envelope with:
+    - `payloadType` of `"https://cloudevents.io/verifiability/dsse/v0.1"`
+    - `payload` containing Base64-encoded verification material (32-byte SHA256 digest)
+    - `signatures` array with at least one signature object containing `keyid` and `sig` fields
+
+### signedextattrs
```
duglin: Does the signature include this attribute? I don't think it does, but shouldn't it? Granted if they twiddle the list then I think the verification should fail anyway, but for completeness I wonder if we should include it?

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,674 @@
...
+Infrastructure (PKI), as these decisions are highly dependent on organizational
+context, existing security infrastructure, and compliance requirements.
+Numerous technical solutions exist to make private and public key material
+available to producers and consumers.
+
+### Signature
+
+The `VERIFICATION_MATERIAL` of the type
+`https://cloudevents.io/verifiability/dsse/v0.1` in the DSSE envelope described
+in the Overview section is created as follows:
```
duglin:
```suggestion
in the Overview section MUST be created as follows:
```

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,674 @@
...
+    1. if it is not present or empty, the event was not signed and MUST be discarded
+3. decode the `dssematerial` as Base64 per [RFC4648](https://tools.ietf.org/html/rfc4648)
+    1. if decoding failed, the verification material is corrupted and the event MUST be discarded
+4. parse the result into a [JSON DSSE Envelope](https://github.com/secure-systems-lab/dsse/blob/master/envelope.md)
+    1. if parsing fails, the verification material is corrupted and the event MUST be discarded
+5. follow the [DSSE verification protocol](https://github.com/secure-systems-lab/dsse/blob/master/protocol.md#protocol)
+    1. filter signatures by `keyid` from list of acceptable keys from step 1
+    2. if verification fails, the event MUST be discarded
+6. read the envelope’s `payloadType` value
+    1. if it is not equal to `https://cloudevents.io/verifiability/dsse/v0.1`, the verification payload is unknown and the event MUST be discarded
```
duglin: ...verification fails

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,674 @@
...
+7. read the envelope’s `payload` field and Base64 decode it into a byte sequence
+    1. if the result is not an sequence of the length 32, then the verification payload is corrupted and the event MUST be discarded
```
duglin: ...verification fails

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,674 @@
...
+8. read the event's `signedextattrs` [Extension Context Attribute](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#extension-context-attributes) (see the [`signedextattrs`](#signedextattrs) attribute definition for constraints)
+    1. if it is not present or empty, the list of "signed extension attributes" is considered empty
+    2. if it is not empty, treat it as a comma separated string of "signed extension attribute names"
+    3. validate that the `signedextattrs` value adheres to the constraints defined in the [`signedextattrs`](#signedextattrs) attribute definition - if any constraint is violated, the event MUST be discarded
```
duglin: merge this into step 2, so:
... treat it as a comma separated string of "signed extension attribute names" and validate it adhere to the constraints defined in ....

.github/workflows/release-notes.yaml
```diff
@@ -27,6 +27,9 @@ on:
       end-sha:
         description: 'Ending SHA? (latest HEAD)'
 
+permissions:
```
duglin: Can you rebase? These are not part of your PR

cloudevents/extensions/verifiability.md
```diff
@@ -63,17 +64,19 @@ unsigned.
...
+
 
 ## Overview
 
 This extension enables event producers to sign CloudEvents and consumers to
-verify those signatures. The verification material is transported as extension
-attributes alongside the event data.
+verify those signatures. The verification material is transported in an
+extension attribute alongside the event data.
+
```
duglin, line 75: extra line

cloudevents/extensions/verifiability.md
```diff
@@ -221,52 +234,257 @@ material delivered with the event. If it is valid, the client creates the
...
-    1. if it is not present or empty, the list of "signed extension attributes" is considered empty
-    2. if it is not empty, treat it as a comma separated string of "signed extension attribute names"
-    3. validate that the `signedextattrs` value adheres to the constraints defined in the [`signedextattrs`](#signedextattrs) attribute definition - if any constraint is violated, the event MUST be discarded
-9. create verification material from the event according to the signing [Protocol](#protocol)
-10. compare the newly created verification material to that in the envelope
+8. recompute `core_digest` from the event according to steps 3-4 of the signing [Protocol](#protocol) and compare it to the `core` byte sequence from step 6.2
     1. if the values are not equal, the event has been modified in transit and MUST be discarded
-11. the event is returned as verified successfully.
+9. if the `ext` field is present (step 6.3) and the consumer wishes to verify extension attributes, recompute `ext_digest` from the event using the `signedextattrs` list from step 6.4 according to step 5 of the signing [Protocol](#protocol) and compare it to the `ext` byte sequence from step 6.3
```
duglin: Please wrap entire doc at 80 chars

cloudevents/extensions/verifiability.md
```diff
@@ -221,52 +234,257 @@ material delivered with the event. If it is valid, the client creates the
...
+10. the event is returned as verified successfully.
 
-Upon verification of an CloudEvent, implementations MUST return a new event
-containing only verified data: the Context Attributes (REQUIRED and OPTIONAL),
-the event data plus exactly those Extension Context Attributes that were signed
-by the producer. This ensures clear separation between verified and unverified
-data. Users handle either a complete unverified event or a verified event with
-only verified values—never a mixture of both.
+Upon verification of a CloudEvent, implementations MUST return a new event
+containing only verified data. If only core verification was performed (step 8),
```
duglin: So, if there are CE extensions but they're not verified, they're silently dropped?
Is this what we want instead of rejecting the event? I'm wondering if the presence of extra stuff is sign that someone is trying to mess with things and we need to play it safe.

duglin:
Some errors from the tooling:
```
cloudevents/extensions/verifiability.md: cloudevents/extensions/verifiability.md does not contain '#protocol-1' segment
cloudevents/extensions/verifiability.md: line 141: 'required' MUST be capitalized ('REQUIRED')
cloudevents/extensions/verifiability.md: line 141: 'optional' MUST be capitalized ('OPTIONAL')
cloudevents/extensions/verifiability.md: line 281: 'Optional' MUST be capitalized ('OPTIONAL')
cloudevents/extensions/verifiability.md: line 416: 'should' MUST be capitalized ('SHOULD')
cloudevents/extensions/verifiability.md: line 440: 'should' MUST be capitalized ('SHOULD')
cloudevents/extensions/verifiability.md: line 473: 'must' MUST be capitalized ('MUST')
cloudevents/extensions/verifiability.md: line 482: 'must' MUST be capitalized ('MUST')
cloudevents/extensions/verifiability.md: title ('# extension: Verifiable CloudEvents with DSSE') does not match the title of cloudevents/languages/he/extensions/verifiability.md ('# Verifiable CloudEvents with DSSE')
cloudevents/extensions/verifiability.md: title ('# extension: Verifiable CloudEvents with DSSE') does not match the title of cloudevents/languages/zh-CN/extensions/verifiability.md ('# Verifiable CloudEvents with DSSE')

```

You may want to squash and rebase

duglin:
We agreed to try to vote on this next week so:
- @xibz please try to address all open comments and CI issues
- everyone else PLEASE review and comment ASAP so @xibz can respond with enough time for folks to review his responses before next week's call

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,965 @@
...
+any guarantees about the completeness of the event stream or the order in
+which events are delivered. Solutions for these other issues are outside the
+scope of this proposal.
+
+Because the CloudEvents specification [requires](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#id)
+the combination of event `source` and `id` to be unique per event, signature
+replay attacks (where an attacker resubmits a legitimately signed event) are
+out of scope for this proposal - the existing source+id uniqueness requirement
+provides sufficient protection.
+
```
jskeet, line 33: Nit: remove redundant blank line.

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,965 @@
...
+enable *confidentiality*. Consequently, it does not address the threat of
+unauthorized parties being able to read CloudEvents that were not meant for
+them (see [Privacy & Security](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#privacy-and-security)
+in the CloudEvents spec).
+
+While the design in this extension *can* be used by authorized intermediaries
+to modify and re-sign events, it explicitly does not aim to provide a
+cryptographic audit trail of event modifications.
+
+As a general principle, this extension aims to avoid cryptographic agility in
```
jskeet: I'm not sure what "cryptographic agility" is - is that a term most folks in the relevant audience would understand? (It's entirely possible.)

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,965 @@
...
+producers can produce verifiable events without any knowledge about whether the
+consumers have been configured to and are able to verify events. Consumers
+that do not support verification can consume signed events as if they were
+unsigned.
+
+**The verification material MUST be contained in the same message as the
+event:** The design aims to be simple and robust, and so the verification
+material MUST be transported and delivered along with the event that it
+describes and not in separate events or even through different channels.
+
```
jskeet, line 69: Nit: extra blank line.

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,965 @@
...
+  "sig": "<Base64(SIGNATURE)>"
+ }]
+}
+```
+
+The `VERIFICATION_MATERIAL`, once Base64 decoded, MUST be a JSON object:
+
+```
+{
+ "core": "<Base64(core_digest)>",
```
jskeet: Can we capitalize these for consistency with VERIFICATION_MATERIAL, KEY_ID and SIGNATURE? Aside from anything else, that makes it easier to tell that they're not expected to be attribute names.

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,965 @@
...
+The `payloadType` links to the version of this spec that was used to create the
+verification material. The spec defines the version of the [DSSE Protocol](https://github.com/secure-systems-lab/dsse/blob/master/protocol.md)
+that is to be used.
+
+## Attributes
+
+This extension defines the following attribute:
+
+### dssematerial
+- Type: `String`
```
jskeet: It's not clear to me why this is a string *and* specified as base64. If the intention is to transport arbitrary bytes, I think it should be of type `Binary`, which would happen to be base64-encoded in the JSON CloudEvents format, but not necessarily in other formats.

I think it would be better to make it Binary, and explicitly state that logically it's a string, but binary is used to avoid the restrictions on the String type. Alternatively, would it be reasonable to put the JSON on a single line and have it as a string *without* any additional encoding? I suspect the only character disallowed in https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#type-system that causes a problem at the moment is for the line break.

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,965 @@
...
+- Description: The [DSSE JSON Envelope](https://github.com/secure-systems-lab/dsse/blob/master/envelope.md)
+  that can be used to verify the authenticity and integrity of the CloudEvent.
+- Constraints:
+  - OPTIONAL
+  - If present, MUST be Base64 encoded
+  - When decoded, MUST be a valid DSSE JSON Envelope with:
```
jskeet: Is it permitted for other fields to be present in the JSON object?

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,965 @@
...
+    - `payloadType` of `"https://cloudevents.io/verifiability/dsse/v0.1"`
+    - `payload` containing a Base64-encoded JSON object with:
+      - `core`: Base64-encoded 32-byte SHA256 digest of the core CloudEvent fields (REQUIRED)
+      - `ext`: Base64-encoded 32-byte SHA256 digest of the signed extension attribute values (OPTIONAL)
+      - `signedextattrs`: JSON array of extension attribute name strings that were signed (OPTIONAL)
+      - `ext` and `signedextattrs` MUST both be present or both be absent
+      - The `signedextattrs` array MUST adhere to the following rules. If any of these rules are violated, the signing function MUST refuse to produce material and the verifier MUST discard the event:
+        - Attribute names MUST NOT contain repetitions
+        - Attribute names MUST NOT include any required or optional [Context Attribute](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#context-attributes) name (`id`, `source`, `specversion`, `type`, `datacontenttype`, `dataschema`, `subject`, `time`)
```
jskeet:
```suggestion
        - Attribute names MUST NOT include any [core context attribute](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#context-attributes) name (`id`, `source`, `specversion`, `type`, `datacontenttype`, `dataschema`, `subject`, `time`)
```

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,965 @@
...
+  - OPTIONAL
+  - If present, MUST be Base64 encoded
+  - When decoded, MUST be a valid DSSE JSON Envelope with:
+    - `payloadType` of `"https://cloudevents.io/verifiability/dsse/v0.1"`
+    - `payload` containing a Base64-encoded JSON object with:
+      - `core`: Base64-encoded 32-byte SHA256 digest of the core CloudEvent fields (REQUIRED)
+      - `ext`: Base64-encoded 32-byte SHA256 digest of the signed extension attribute values (OPTIONAL)
+      - `signedextattrs`: JSON array of extension attribute name strings that were signed (OPTIONAL)
+      - `ext` and `signedextattrs` MUST both be present or both be absent
+      - The `signedextattrs` array MUST adhere to the following rules. If any of these rules are violated, the signing function MUST refuse to produce material and the verifier MUST discard the event:
```
jskeet: What about including the name of an attribute which isn't actually present in the event?

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,965 @@
...
+    5. compute the SHA256 digest of the event's [`datacontenttype`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#datacontenttype) Context Attribute in UTF8 and append it to the byte sequence *(if the attribute is not set, use the digest of the empty byte sequence)*
+    6. compute the SHA256 digest of the event's [`dataschema`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#dataschema) Context Attribute in UTF8 and append it to the byte sequence *(if the attribute is not set, use the digest of the empty byte sequence)*
+    7. compute the SHA256 digest of the event's [`subject`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#subject) Context Attribute in UTF8 and append it to the byte sequence *(if the attribute is not set, use the digest of the empty byte sequence)*
+    8. compute the SHA256 digest of the event's [`time`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#time) Context Attribute normalized to RFC 3339 Zulu format and append it to the byte sequence (if the attribute is not set, use the digest of the empty byte sequence)
```
jskeet: Starting a new comment on this as the others are outdated... I think there are a few aspects here:

- Canonicalization to UTC
- Normalization of precision (potentially, anyway)
- The *precise* format used (not just "what's valid in RFC3339" but a canonical "this is the only permissable result for the input value")

There are valid (according to the CloudEvents spec) "timestamp" values that don't have UTC offsets, so don't actually represent instants in time. I'd propose that trying to sign an event with such a timestamp should fail.

Different SDKs may well have different levels of precision for date/time values. I think we should settle on one, truncate to that, and be very specific about it. If trunction to the second would be acceptable, I think that would be simplest: state that there is *no* subsecond value in the signed material.

So for example, an input timestamp of "2026-05-13T10:49:12.123456+01:00" would *have* to be converted to "2026-05-13T09:49:12Z" and that's what gets signed.

We should specify that this is how time-based extension attributes are signed too.

cloudevents/extensions/verifiability.md
jskeet: From later on, it looks like that's okay - and could be used to assert that the given attribute is absent/empty - but I think it's worth calling out here.

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,965 @@
...
+       Note: Time normalization to Zulu format ensures signature verification remains valid even when intermediaries deserialize and reserialize events with different timezone representations of the same timestamp.
+    9. compute the SHA256 digest of the event's [`data`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#event-data) and append it to the byte sequence
+4. compute `core_digest` as the SHA256 digest of the byte sequence from step 3
+5. if the list from step 2 is not empty:
+    1. create an empty byte sequence for the ext digest
+    2. for each extension attribute in the list from step 2 (in the given order, from lowest to highest index)
+        1. compute the SHA256 digest of the extension attribute's value and append it to the byte sequence. Non-string attribute values MUST be serialized to their [CloudEvents string representation](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#type-system) before hashing. *(if the attribute is not present on the event, use the digest of the empty byte sequence)*
```
jskeet: I think we need more clarity here. Effectively I'd suggest a whole separate section about how each type is hashed, and then use that for both the core and extension attributes.

(I'd also suggest that when hashing a binary value, there's no point in converting it to base64 first - likewise we could use simple encodings for Integer and Boolean values.)

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,965 @@
...
+        1. validate that the `signedextattrs` value adheres to the constraints defined in the [`dssematerial`](#dssematerial) attribute definition - if any constraint is violated, the event MUST be discarded
+    5. if `signedextattrs` is present but the `ext` field is absent from the payload, or vice versa, the event MUST be discarded
+7. follow the [DSSE verification protocol](https://github.com/secure-systems-lab/dsse/blob/master/protocol.md#protocol)
+    1. filter signatures by `keyid` from list of acceptable keys from step 1
+    2. if verification fails, the event MUST be discarded
+8. recompute `core_digest` from the event according to steps 3-4 of the signing [Protocol](#protocol) and compare it to the `core` byte sequence from step 6.2
+    1. if the values are not equal, the event has been modified in transit and MUST be discarded
+9. if the `ext` field is present (step 6.3) and the consumer wishes to verify extension attributes, recompute `ext_digest` from the event using the `signedextattrs` list from step 6.4 according to step 5 of the signing [Protocol](#protocol) and compare it to the `ext` byte sequence from step 6.3
```
jskeet: This is where I think we need to be clear that the verifier *must* know the types of the extension attributes in order to verify them. The overall result of verification could be:

- Verified both core and extensions
- Verified core, there were no signed extensions
- Verified code, can't verify extensions due to at least one extension attribute's type not being known
- Verification failed (one of the signatures doesn't match)

cloudevents/extensions/verifiability.md
jskeet: I'm concerned about this too. I'd far rather be able to keep everything, but make it easy to determine which extensions were signed and which weren't.

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,965 @@
...
+
+*Notes:*
+
+* *In [CloudEvent's type system](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#type-system) a `Timestamp`'s string encoding is [RFC 3339](https://tools.ietf.org/html/rfc3339). This means that verification of the `time` Context Attribute can only be done with second precision, even though an SDK might allow passing in a timestamp with nanosecond precision.*
+* *In [CloudEvent's official Protocol Buffers format](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/formats/cloudevents.proto#L57), the `time` Context Attribute is encoded as a `google.protobuf.Timestamp` and hence does not include time zone information (which RFC 3339 would allow). For interoperability with CloudEvent setups using the Protocol Buffers format, time zone information is ignored in the signing and verification process.*
+* *The signature covers the exact byte representation of the event data. Intermediaries that deserialize and reserialize event data (e.g., reformatting JSON whitespace or reordering keys) will invalidate the signature. Implementations that route signed events MUST preserve the original byte sequence of the event data.*
+
+## Verification Walkthrough
```
jskeet, line 465: I've stopped reviewing at this point as I suspect changes above will affect the text below.

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,965 @@
...
+by being included in the `signedextattrs` array inside the signed container.
+The `ext` and `signedextattrs` fields are only present when extension
+attributes are signed.
+
+#### Protocol
+
+This is how to sign a CloudEvent using DSSE:
+
+1. choose a signing key
+2. choose the list of "signed extension attributes" (the list MUST adhere to the constraints defined in the [`dssematerial`](#dssematerial) attribute definition: no repetitions, no Context Attribute names, and no `dssematerial`)
```
jskeet: Note either here or elsewhere that extension attributes with a known value but not a known type (which could happen in some SDK implementations if this is an event which has been received and is being propagated) cannot be signed.

xibz:
@duglin @jskeet - I forgot an important commit T-T. i was so confused why you kept mentioning types because I was like "I thought that was in the proposal". Turns out I forgot to push that commit. Sorry about that :(

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,1139 @@
...
+      - `ext`: Base64-encoded 32-byte SHA256 digest of the signed extension attribute values (OPTIONAL)
+      - `signedextattrs`: JSON array of extension attribute name strings that were signed (OPTIONAL)
+      - `ext` and `signedextattrs` MUST both be present or both be absent
+      - The `signedextattrs` array MUST adhere to the following rules. If any
+        of these rules are violated, the signing function MUST refuse to
+        produce material and the verifier MUST discard the event:
+        - Attribute names MUST NOT contain repetitions
+        - Attribute names MUST NOT include any [core context attribute](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#context-attributes) name (`id`, `source`, `specversion`, `type`, `datacontenttype`, `dataschema`, `subject`, `time`)
+        - Attribute names MUST NOT include `dssematerial` (the verification material attribute)
+        - Extension attribute values MUST be of a supported CloudEvents type (`Boolean`, `Integer`, `String`, `Binary`, `URI`, `URI-reference`, or `Timestamp`)
```
jskeet: I'm not sure what this means - there's no concept of an attribute which exists but isn't of a supported type.

If this is intended to mean that the signer *knows* the type of the attribute, that's slightly different. I think I'd probably talk about that separate from this bullet list though.

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,1139 @@
...
+```
+
+Notation used above:
+- `+` denotes byte-sequence concatenation.
+- `UTF8(s)` is the UTF-8 encoding of string `s`.
+- `RFC3339(UTC(t))` is the RFC 3339 string representation of timestamp `t`
+  normalized to UTC ("Zulu") time with second precision.
+- `CEString(v)` is the CloudEvents canonical string encoding of an extension
+  attribute value, as defined in step 5.2.1 of the [Signing Protocol](#signing-protocol)
+  (`Boolean`, `Integer`, `String`, `Binary`, `URI`, `URI-reference`, `Timestamp`).
```
jskeet: Not sure what this list is intended for, but I think we can just remove the line (and move the period to the line above)

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,1139 @@
...
+- `event.data` is hashed as its raw byte representation — no UTF-8 wrapping
```
jskeet: Hmm... this could be tricky. I've been focused on the representations of the attributes, but data is a whole other tricky story. What *is* the "raw byte representation" of an in-memory CloudEvent which has a JSON object as its data? If this is intended to be "the bytes that appear in the serialized form" then that gets us back to implementation difficulties.

(Historically, attributes have been *relatively* straightforward - we've had many, many discussions about the nature of data, particularly for event formats which can represents certain data types in a different way, e.g. where the data for a JSON-formatted event is itself a JSON object...)

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,1139 @@
...
+by being included in the `signedextattrs` array inside the signed container.
+The `ext` and `signedextattrs` fields are only present when extension
+attributes are signed.
+
+#### Signing Protocol
+
+This is how to sign a CloudEvent using DSSE:
+
+1. choose a signing key
+2. choose the list of "signed extension attributes" (the list MUST adhere to the constraints defined in the [`dssematerial`](#dssematerial) attribute definition: no repetitions, no Context Attribute names, and no `dssematerial`)
```
jskeet:
```suggestion
2. choose the list of "signed extension attributes" (the list MUST adhere to the constraints defined in the [`dssematerial`](#dssematerial) attribute definition: no repetitions, no Core Context Attribute names, and no `dssematerial`)
```

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,1139 @@
...
+    1. create an empty byte sequence for the ext digest
+    2. for each extension attribute in the list from step 2 (in the given order, from lowest to highest index)
+        1. compute the SHA256 digest of the extension attribute's value and append it to the byte sequence. Extension attribute values MUST be serialized using their CloudEvents canonical string encoding before hashing:
+            - `Boolean`: "true" or "false" (case-sensitive)
+            - `Integer`: decimal representation without leading zeros, fraction, or exponent (per RFC 7159 Section 6)
+            - `String`: the string value as-is
+            - `Binary`: Base64 encoding per RFC 4648
+            - `URI`: the absolute URI string per RFC 3986 Section 4.3
+            - `URI-reference`: the URI-reference string per RFC 3986 Section 4.1
+            - `Timestamp`: RFC 3339 Zulu format with second precision (normalized to UTC as described in step 8)
```
jskeet: Again, I'd separate out "here's how you hash a value" (on a per type basis) from the uses of it. That way you don't need any repetition, and key implementation details (e.g. how to hash a timestamp value) aren't buried in the list.

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,1139 @@
...
+       Note: Time normalization to Zulu format with second precision ensures signature verification remains valid even when intermediaries deserialize and reserialize events with different timezone representations or subsecond precisions of the same timestamp.
+    9. compute the SHA256 digest of the event's [`data`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#event-data) and append it to the byte sequence
+4. compute `CORE_DIGEST` as the SHA256 digest of the byte sequence from step 3
+5. if the list from step 2 is not empty:
+    1. create an empty byte sequence for the ext digest
+    2. for each extension attribute in the list from step 2 (in the given order, from lowest to highest index)
+        1. compute the SHA256 digest of the extension attribute's value and append it to the byte sequence. Extension attribute values MUST be serialized using their CloudEvents canonical string encoding before hashing:
+            - `Boolean`: "true" or "false" (case-sensitive)
+            - `Integer`: decimal representation without leading zeros, fraction, or exponent (per RFC 7159 Section 6)
+            - `String`: the string value as-is
```
jskeet: A string value can't be hashed, as hashing is an operation on binary data. I suspect we want to say you hash the UTF-8-encoded form of the string.

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,1139 @@
...
+|---	|---	|---	|---	|---	|
+|Event Data (payload)	|✅	|✅	|✅	|	|
+|REQUIRED Context Attributes	|✅	|✅	|✅	|	|
+|OPTIONAL Context Attributes	|✅	|✅	|✅	|See notes below for time attribute	|
+|Extension Context Attributes	|✅	|✅	|✅	|Optional (per attribute)	|
+|Metadata added by transports	|❌	|❌	|❌	|	|
+
+*Notes:*
+
+* *In [CloudEvent's type system](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#type-system) a `Timestamp`'s string encoding is [RFC 3339](https://tools.ietf.org/html/rfc3339). This means that verification of the `time` Context Attribute can only be done with second precision, even though an SDK might allow passing in a timestamp with nanosecond precision.*
```
jskeet: It's not RFC3339 which limits us to second precision - it's interoperability. We don't know what degree of precision any given consumer will use, so it's simplest to truncate to the second.

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,1139 @@
...
+* *In [CloudEvent's official Protocol Buffers format](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/formats/cloudevents.proto#L57), the `time` Context Attribute is encoded as a `google.protobuf.Timestamp` and hence does not include time zone information (which RFC 3339 would allow). For interoperability with CloudEvent setups using the Protocol Buffers format, time zone information is ignored in the signing and verification process.*
+* *The signature covers the exact byte representation of the event data. Intermediaries that deserialize and reserialize event data (e.g., reformatting JSON whitespace or reordering keys) will invalidate the signature. Deployments requiring end-to-end verifiability SHOULD avoid such intermediaries or accept re-signing at trust boundaries.*
```
jskeet: This comes back to what "the exact byte representation of the event data" is, and how tricky that could be for implementations...

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,691 @@
...
+avoids prescribing specific approaches for key management and Public Key
+Infrastructure (PKI), as these decisions are highly dependent on organizational
+context, existing security infrastructure, and compliance requirements.
+Numerous technical solutions exist to make private and public key material
+available to producers and consumers.
+
+### Signature
+
+The `VERIFICATION_MATERIAL` of the type
+`https://cloudevents.io/verifiability/dsse/v0.1` in the envelope above is
```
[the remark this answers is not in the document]
xibz: It's the reference to the above example on line 89

cloudevents/extensions/verifiability.md
xibz: It's mostly cause security typically wants things to be VERY clear. No way to interpret anything.

cloudevents/extensions/verifiability.md
xibz: This only applies to event.data, not to context attributes. Context attributes use semantic canonicalization (Option 3 from our email thread), which you described as having “really significant benefits: Signature doesn’t change based on format, Much simpler implementation than either 1 or 2”, and that’s what the spec implements for all typed attributes (UTF8(s) for strings, RFC3339(UTC(t)) for timestamps, etc.).

For event.data, raw bytes are unavoidable: there is no CE type system for event data, so we cannot apply semantic canonicalization. The note at lines 327–328 calls this out explicitly, any intermediary that reserializes event data (whitespace changes, key reordering) must re-sign. This is the same trade-off you acknowledged in the email when discussing Option 2: “That’s ‘only’ inefficient rather than particularly difficult, but it does mean that support for this extension needs to be built into every event formatter.” We avoided that for attributes via Option 3; for data there is no equivalent.

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,674 @@
...
+1. obtain list of acceptable key ids for verification
+2. read the event’s `dssematerial` [Extension Context Attribute](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#extension-context-attributes)
+    1. if it is not present or empty, the event was not signed and MUST be discarded
+3. decode the `dssematerial` as Base64 per [RFC4648](https://tools.ietf.org/html/rfc4648)
+    1. if decoding failed, the verification material is corrupted and the event MUST be discarded
+4. parse the result into a [JSON DSSE Envelope](https://github.com/secure-systems-lab/dsse/blob/master/envelope.md)
+    1. if parsing fails, the verification material is corrupted and the event MUST be discarded
+5. follow the [DSSE verification protocol](https://github.com/secure-systems-lab/dsse/blob/master/protocol.md#protocol)
+    1. filter signatures by `keyid` from list of acceptable keys from step 1
+    2. if verification fails, the event MUST be discarded
```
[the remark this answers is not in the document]
xibz: So security really wanted this. I think a consumer may not know what is valid or not if they are reading from some system.

Maybe we change MUST to SHOULD? Thoughts?

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,674 @@
...
+7. read the envelope’s `payload` field and Base64 decode it into a byte sequence
+    1. if the result is not an sequence of the length 32, then the verification payload is corrupted and the event MUST be discarded
+8. read the event's `signedextattrs` [Extension Context Attribute](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#extension-context-attributes) (see the [`signedextattrs`](#signedextattrs) attribute definition for constraints)
+    1. if it is not present or empty, the list of "signed extension attributes" is considered empty
+    2. if it is not empty, treat it as a comma separated string of "signed extension attribute names"
+    3. validate that the `signedextattrs` value adheres to the constraints defined in the [`signedextattrs`](#signedextattrs) attribute definition - if any constraint is violated, the event MUST be discarded
+9. create verification material from the event according to the signing [Protocol](#protocol)
+10. compare the newly created verification material to that in the envelope
+    1. if the values are not equal, the event has been modified in transit and MUST be discarded
```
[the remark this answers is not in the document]
xibz: Same here

cloudevents/extensions/verifiability.md
xibz: Good call out. I remember there's a timestamp RFC that is specifically second precisions, but I may be misremembering or blurring RFCs.

But good callout

cloudevents/extensions/verifiability.md
xibz: Same answer as above

This applies only to event.data, not to context attributes. For context attributes the spec uses semantic canonicalization (Option 3 from our March 11 discussion), which you preferred precisely because it avoids the wire-bytes implementation pain you described: “often the bytes that were on the wire aren’t known by the time you’re dealing with the parsed representation... the JSON event formatter reads the whole body as a document. At that point, the original bytes on the wire have gone.”

For event.data we cannot apply the same trick, there is no CE type for data. Signing requires stable bytes, and the spec documents the consequence clearly (lines 327–328): intermediaries that reserialize data must resign. SDK authors should sign as close to the wire as possible. Happy to add a note in the spec making this explicit for SDK implementors.

jskeet: Right - sorry for missing this before. I think in many cases this may make implementation close to infeasible, unfortunately.

For example in https://github.com/cloudevents/sdk-csharp/blob/main/src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs#L138 (and similar event formatters) the first thing that's done is to read the event as a JSON document. At that point, we simply don't have the original bytes. So to read the wire data, we'd probably have to maintain our own *very* low-level JSON parser in order to get the raw bytes.

I hate to be a downer, but I think for at least a lot of languages, this will simply break any layered architecture of the SDK. It may be feasible in some other languages of course.

xibz: For event.data I want to acknowledge you raised this in our March email too, and at the time you said "That feels somewhat reasonable as a spec, but a real pain to implement in an SDK." We proceeded on that basis, and the spec documents the consequence clearly. So I dont think it was missed, per se.

On the C# SDK example: the formatter reading the whole JSON document before the raw bytes are available is a real constraint, but the solution is to capture the bytes during parsing rather than after. System.Text.Json exposes raw UTF-8 bytes for a value node via GetRawText() or an equivalent reader-level API, no custom low-level parser needed. The SDK would need to thread this through the signing path, which is additional work, but it's the same class of work any checksum/integrity system requires.

The alternative, defining a canonical serialization for event.data, would require specifying rules for every possible content type (JSON, XML, binary, protobuf, etc.), which is far out of scope. Raw bytes is the only content-type-agnostic option.

Producers typically sign before serialization where raw bytes are always available. For SDKs that receive-and-propagate, capturing bytes at the formatter layer is the right integration point. We can capture that in the spec if needed

jskeet: > but the solution is to capture the bytes during parsing rather than after. System.Text.Json exposes raw UTF-8 bytes for a value node via GetRawText() or an equivalent reader-level API, no custom low-level parser needed.

I may have chosen a bad example - it's *possible* that we can do this for System.Text.Json without very many changes. I think I'd need to check it really carefully for every format - for example, in protobuf it may be feasible due to parsing effectively preserving everything we need.

At the very least, I think it'll mean work on a "per language, per format" basis, and in a way which makes the implementation significantly more complex. That may well be unavoidable, and I acknowledge that there's probably no better approach - but with our current state of SDKs being "it's hard to find maintainers at all" that may well mean we end up with an extension that is *very* spottily implemented :(

xibz: As someone who used to maintain SDKs, I understand the pain :(.

I will have to check with legal, but we may be able to help with some implementations, e.g., golang, java, python.

C# is a little out of skillset though ><. At this point it seems like figuring out the best path to get this implemented.

I will have a commit that updates the other points, e.g., the duplicated canonicalization stuff, type lists, etc, soon

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,1139 @@
...
+verification of the signature against the key and passes on a verified event to
+the consumer:
+
+![An illustration showing how the producer's SDK signs an event, and the consumer's SDK verifies it](verifiability1.png)
+
+The verification material is transported in an [Extension Context Attribute](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#extension-context-attributes)
+called `dssematerial` (see [Attributes](#attributes) section below).
+
+The `dssematerial` binary value MUST be the UTF-8 encoding of a proper DSSE
+envelope:
```
clemensv: DSSE JSON envelope (like below)

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,1139 @@
...
+    5. compute the SHA256 digest of the event's [`datacontenttype`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#datacontenttype) Context Attribute in UTF8 and append it to the byte sequence *(if the attribute is not set, use the digest of the empty byte sequence)*
+    6. compute the SHA256 digest of the event's [`dataschema`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#dataschema) Context Attribute in UTF8 and append it to the byte sequence *(if the attribute is not set, use the digest of the empty byte sequence)*
+    7. compute the SHA256 digest of the event's [`subject`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#subject) Context Attribute in UTF8 and append it to the byte sequence *(if the attribute is not set, use the digest of the empty byte sequence)*
+    8. compute the SHA256 digest of the event's [`time`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#time) Context Attribute normalized to RFC 3339 Zulu format with second precision (no subsecond component) and append it to the byte sequence (if the attribute is not set, use the digest of the empty byte sequence)
```
clemensv: "Zulu" is mentioned once in RFC3339, but it will be helpful to show a formatted datetime example of what you mean here and also link to the RFC. That's a good potential source of hours of debugging fun if one misses this detail.

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,1139 @@
...
+|Event Data (payload)	|✅	|✅	|✅	|	|
+|REQUIRED Context Attributes	|✅	|✅	|✅	|	|
+|OPTIONAL Context Attributes	|✅	|✅	|✅	|See notes below for time attribute	|
+|Extension Context Attributes	|✅	|✅	|✅	|Optional (per attribute)	|
+|Metadata added by transports	|❌	|❌	|❌	|	|
+
+*Notes:*
+
+* *In [CloudEvent's type system](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#type-system) a `Timestamp`'s string encoding is [RFC 3339](https://tools.ietf.org/html/rfc3339). This means that verification of the `time` Context Attribute can only be done with second precision, even though an SDK might allow passing in a timestamp with nanosecond precision.*
+* *In [CloudEvent's official Protocol Buffers format](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/formats/cloudevents.proto#L57), the `time` Context Attribute is encoded as a `google.protobuf.Timestamp` and hence does not include time zone information (which RFC 3339 would allow). For interoperability with CloudEvent setups using the Protocol Buffers format, time zone information is ignored in the signing and verification process.*
```
clemensv: We should assume that all timezone-less datetime values are UTC.

clemensv (commented):
I am in favor of merging this once the remaining raised issues are addressed, which all seem editorial. The mechanism is clearly explained.

PR #1379, opened 2026-03-17
Add App Context extension

inlined:
Fixes (N/A)

Adds an extension to document future features Google is exploring so that there can be industry standardization 

## Proposed Changes

Document an extension to indicate an app that triggered an event. Allows for a BaaS-opque identifier as well as app store specific identifiers for Android and iOS.

**Release Note**

```release-note
(feat) extensions: add appcontext extension
```

### DIFF
--- a/cloudevents/extensions/appcontext.md
+++ b/cloudevents/extensions/appcontext.md
@@ -1,0 +1,50 @@
+# App Context
+
+This extension gives information about a mobile app that triggered an event.
+
+## Notational Conventions
+
+As with the main [CloudEvents specification](../spec.md), the key words "MUST",
+"MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT",
+"RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as
+described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+However, the scope of these key words is limited to when this extension is
+used. For example, an attribute being marked as "REQUIRED" does not mean
+it needs to be in all CloudEvents, rather it needs to be included only when
+this extension is being used.
+
+## Attributes
+
+### appid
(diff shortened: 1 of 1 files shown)


duglin:
Hey @inlined ! Been a while!

Could you elaborate a bit on how you see this being used w/o conflicting with `source` and/or `subject`?  The latest spec talks about how either of those could be compound strings made up of more than one "thing", so it might include appid and bundleid

inlined:
Hey Dug, always a pleasure!

Think of this as another form of AuthN rather than a true source. So a write was made against the Firebase Realtime Database. That is the source of the event. And we tell you the human who made the change to that database with the authcontext extension. This would allow you to also track the app that human used.

Can be used for a variety of purposes from tracking sources of engagement with particular features to applying bugfixes surgically to known bad versions of an app.

Has the spec changed so much since I was last looking at it that source/subject should be the app and not the database in this example?

duglin (changes requested):

duglin:
Your example and AuthN commentary helps a lot - thanks!

duglin:
@inlined we briefly discussed this on the weekly call and there were no major objections. @jskeet has a few questions that he'll ask as PR comments, and I have one more: any reason not to generalize the naming a bit more? E.g. "appid" seem pretty specific - would it lose value if we picked a phase like "initiatorid" or "clientid", or something that could then make sense in a non-mobile-app case but still falls into a similar category of usecases/workflows?

duglin:
@inlined did you want to address the open comments?

duglin:
Also,the DCO check is failing

duglin:
@inlined any update on this one?

duglin:
@inlined you still interested in this one getting in?

duglin:
per the May 7th call - closing this PR for now since it appears to be stale. Please let us know if we should reopen it.

PR #1385, opened 2026-05-07
Remove non-English docs

duglin:
Per the May 7th call, we agreed to remove all non-Engish docs from the repo ; tag the current commit so people can reference the old translated docs ; add a comment to the README telling people to use tooling/AI to do the translation and let us know if that's problematic for them.


duglin:
Approved on the 5/21 call

PR #1386, opened 2026-05-30
feat(extensions): add verifiable CloudEvents with DSSE

### LINKED ISSUE #1302: Adding zero trust to cloudevents

xibz:
## Proposed Changes

feat: Add DSSE-based cryptographic signing for CloudEvents verification

Implements verifiable CloudEvents using DSSE (Dead Simple Signing Envelope) to ensure event authenticity and integrity across untrusted transport layers.

Key features:
- Sign CloudEvents with DSSE v1.0.2 protocol using SHA256 digests
- Transport verification material in 'dssematerial' extension attribute
- Support for binary, structured, and batch CloudEvent modes
- Backward compatible - unsigned events still work, consumers can ignore signatures (but highly inadvisable)

Technical approach:
- Creates SHA256 digest chain of all context attributes and event data
- Wraps digest in DSSE envelope with Base64 encoding
- Verifies by recomputing digests and comparing against signed payload
- Returns only verified data to consumers (strips unverified extensions)

This enables cryptographic proof that events:
1. Were produced by the claimed source (authenticity)
2. Were not modified in transit (integrity)

Does NOT address: event ordering, completeness, replay attacks, or confidentiality

Fixes #1302 

**Release Note**

```release-note

```


xibz:
@duglin - I decided to recreate the PR. There were too many merge conflicts, 800+ T-T, because of how poorly it diverged.

I tried to force push by hard resetting then cherry picking, but it autoclosed the PR :(.

Anywho updated! But the verify tools look broken since the removal the locale files

references #1371

duglin:
@xibz no worries.  I’ll fix the CI issues tomorrow. They’re not related to your changes

duglin:
@xibz can you rebase using the latest 'main' branch? I think that should fix your CI issues.

xibz:
@duglin - Done!

duglin:
@xibz are all changed made and people should do a final review?

/cc @clemensv @jskeet

xibz:
Yep, it's good for a final review

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,1302 @@
...
+   [Signing Protocol](#signing-protocol) and compare it to the `ext` byte
+   sequence from step 6.3
+    1. if the values are not equal, the extension attributes have been modified
+       in transit and MUST be discarded
+10. the event is returned as verified successfully.
+
+Upon successful verification, implementations MUST return a result that
+identifies what was verified: core only, core and extensions, or core with
+extensions skipped due to unknown types. The verified event returned to
+application code MUST contain only verified data, what exactly is included
```
jskeet: This seems to contradict the `passthrough` mode below. I think we could just remove the first half of the sentence here.

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,1302 @@
...
+#### Proxy Guidance
+
+A proxy or intermediary that forwards signed CloudEvents without modifying
+the signed fields MAY forward the event unchanged and its `dssematerial`
+remains valid for downstream consumers.
+
+A proxy that modifies any signed field, including event data, core context
+attributes, or any attribute named in `signedextattrs`, SHOULD re-sign the
+event using a key that downstream consumers are configured to trust. A proxy
+that cannot re-sign MUST NOT silently forward the modified event as if the
```
jskeet: This requires that every proxy is aware of this extension, which I don't think is reasonable. (It would mean that a proxy which is valid today becomes invalid when this PR is submitted.)

I think it would be reasonable to say "a proxy that conforms with this extension" or something similar.

cloudevents/extensions/verifiability.md
```diff
@@ -0,0 +1,1302 @@
...
+
+When hashing a CloudEvents attribute value, the value MUST first be
+converted to a byte sequence using its CloudEvents type. The rules are:
+
+| Type | Byte sequence |
+|---|---|
+| `Boolean` | UTF-8 encoding of `"true"` or `"false"` (case-sensitive) |
+| `Integer` | UTF-8 encoding of the decimal representation without leading zeros, fraction, or exponent (per RFC 7159 Section 6) |
+| `String` | UTF-8 encoding of the string value |
+| `Binary` | Base64 encoding per RFC 4648, then UTF-8 encoded |
```
jskeet: This feels unnecessary to me - why not just hash the bytes themselves? (The point is to get a byte sequence, and this *is* a byte sequence already.)

jskeet (commented):
I won't claim to have exhaustively re-reviewed everything, but I've looked for the bits that I know I'm interested in :) Just a couple more comments.

duglin:
Approved on the 6/11 call

Thanks @xibz for all your hard work and time!!!

PR #1389, opened 2026-06-10
add another potential v2 work item

### LINKED ISSUE #1388: Suggestion: Same Header-Variables through different specs (example: Kafka and HTTP
Currently Kafka and the HTTP Specs uses differenct prefixes for their header information.
kafka uses "ce_" and http uses "ce-".
There is no technical explanation for that, so why it is different?

duglin:
Consider more consistent prefixes across bindings, see #1388 

Fixes #1388


jskeet (approved):

duglin:
Approved on the 6/11 call

PR #1394, opened 2026-08-22
feat(sdk-kotlin): mark JSON event format support in Kotlin column

octaviospain:
The CloudEvents Kotlin SDK now implements the JSON event format — structured-mode
`application/cloudevents+json` and batch-mode `application/cloudevents-batch+json` —
for both CloudEvents v1.0 and v0.3, on every Kotlin Multiplatform target.

This updates the `Kotlin` column of the feature-support table to mark the JSON
Event Format row as supported for both versions (the Kotlin SDK already claimed
CloudEvents Core for both).

No conformance-shared-suite claim is made — there is no shared conformance suite
upstream; the claim reflects the SDK's own vendored fixture suite.


duglin:
LGTM

duglin:
Approved on the 8/27 call

